PR #3, opened 2017-02-05
Upgrade event-loop-lag dependency due to security issue

jamessharp:
Running the nsp tool over a codebase with this sensor provides the following security warning:

https://nodesecurity.io/advisories/46

I tracked it down to the `event-loop-lag` package (which had an older version of the `debug` package which depended on the offending `ms` package)


bripkens:
Thanks for identifying and fixing the problem! Merged via f0f770c8bbcfba3d72359e768cea52475ce43bc5.

bripkens:
Also released as `1.20.2`.

PR #6, opened 2017-05-12
add option field for looking up the agent at a custom ip

lowsky:
add an additional option parameter for custom agentHost to override default '127.0.0.1'


bripkens:
Thanks for the PR! I'll release this later today.

PR #12, opened 2017-10-25
Parse and report HTTP query params.

pglombardo:
This adds URL query reporting for the HTTP server instrumentation.

### DIFF
--- a/src/tracing/instrumentation/httpServer.js
+++ b/src/tracing/instrumentation/httpServer.js
@@ -7,6 +7,7 @@ var tracingConstants = require('../constants');
 var transmission = require('../transmission');
 var tracingUtil = require('../tracingUtil');
 var hook = require('../hook');
+var url = require('url');
 
 var originalCreateServer = coreHttpModule.createServer;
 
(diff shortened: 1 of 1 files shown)


pglombardo:
screen shot 2017-10-25 at 15 20 44

src/tracing/instrumentation/httpServer.js
```diff
@@ -65,10 +66,12 @@ function requestListener(req, res) {
   }
 
   res.on('finish', function() {
+    var urlParts = url.parse(req.url);
```
bripkens: I noticed before that this is a really expensive operation and removed it in some location within this module (also within other Instana components). Alternatively, could we look for the first `?` and grab everything after that till the end of the string?

bripkens:
We should extend the test cases to test for params.

bripkens:
Maybe I am wrong, but I believe to remember that we don't include the leading `?` in the other sensors, no?

src/tracing/instrumentation/httpServer.js
pglombardo: I was wondering why you didn't use it...

pglombardo:
All good points.  I'll redo.  And add tests.

pglombardo:
BTW coding this was a total 5 minutes.  Fighting the ecosystem that I'm not familiar with...1 hour+ ¯\_(ツ)_/¯

PR #13, opened 2017-10-26
Add Request query capture & reporting

pglombardo:
This PR adds logging of HTTP request queries for HTTPServer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.27.1
+ - Add Request query capture & reporting
+
 ## 1.27.0
  - HTTP client spans do not have the error count field `ec` set.
  - It must be possible to run in manual tracing mode only.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.27.1
+## Unreleased
  - Add Request query capture & reporting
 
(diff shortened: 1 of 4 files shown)


src/tracing/instrumentation/httpServer.js
```diff
@@ -65,10 +64,12 @@ function requestListener(req, res) {
   }
 
   res.on('finish', function() {
+    var urlParts = req.url.split('?');
     span.data = {
       http: {
         method: req.method,
-        url: discardUrlParameters(req.url),
+        url: urlParts.shift(),
```
bripkens: Could we continue to use `discardUrlParameters`? It also discarded matrix parameters and the hash sequence for safety.

test/tracing_test.js
```diff
@@ -76,6 +76,30 @@ describe('tracing', function() {
         });
       });
     });
+    it.only('must properly capture request params', function() {
```
bripkens: Also, you checked in the `.only` restriction ;).

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.27.1
```
bripkens: We should start of by marking this as unreleased. Selecting a version is a release process decision that we need to make (which might even be a feature?).

bripkens:
LGTM 👍

PR #14, opened 2017-11-02
Migrate to universal context local storage

pglombardo:
This PR updates the sensor to use the `cls-hooked` package which supports both older and the most recent versions of Node.

- [x] Use cls-hooked
- [x] Add tests to validate CLS
- [x] Remove/update boot time checks
- [x] Update instrumentation
  - [x] http.Server
  - [x] http.Client
  - [x] Elastic
  - [x] Kafka
  - [x] MySQL
  - [x] Mongo
- [x] Update/add to tests
- [x] ES6 Lint decision on `function() {}` vs `() => {}`
- [x] My local Kafka is broken; need to fix and finish Kafka instrumentation validation


pglombardo:
PR summary in 6caa0ca but I'll re-post here:

Cleaner, leaner & meaner CLS implementation
- Unified context & span concepts under 'span' only
  - Span creation now done in cls module; less/more centralized/consistent code
  - Instrumentation code is now more concise
  - No longer need to track the destroy of old uid/contexts
  - `cls.startSpan(spanName, traceId, spanId)` - start new span in this context
  - `cls.getCurrentSpan()` - get current in this context
  - `cls.isTracing()` - returns boolean if we have a current span
- Update instrumentation to use new exports
- Revamp cls tests
- Store per request tracing level in request context (with cls exports to access)
  - `cls.setTracingLevel(level)` - to set the level
  - `cls.tracingLevel()` - to access the level
  - `cls.tracingSuppressed()` - boolean return; false if 0; true otherwise
- Keep a registry of exit span types and offer function to query spans
  - `cls.isExitSpan(span)` - boolean return if this span is an exit span

.eslintrc
```diff
@@ -1,16 +1,29 @@
 ---
   extends: airbnb/legacy
+  env:
+    node: true
+    mocha: true
+    es6: true
```
bripkens: We aren't using any es6 features besides lambdas and retaining some basic Node.js 0.12 support is also good. Hence I would say let's please disable es6

.eslintrc
```diff
@@ -1,16 +1,29 @@
 ---
   extends: airbnb/legacy
+  env:
+    node: true
+    mocha: true
```
bripkens: Enabling mocha globally has unintended side effects, e.g. it registering some globals that we do not want. Please disable it.

.eslintrc
```diff
@@ -1,16 +1,29 @@
...
+    es6: true
+  parserOptions:
+    sourceType: strict
   rules:
+    no-shadow: 0
+    require-yield: 0
```
bripkens: Why did you disable all these rules?

src/tracing/cls.js
```diff
@@ -0,0 +1,110 @@
...
+  return false;
+};
+
+/*
+ * Determine if <span> is an exit span
+ *
+ */
+var exitSpans = ['node.http.client', 'elasticsearch', 'mongo', 'mysql'];
+exports.isExitSpan = function isExitSpan(span) {
+  return (exitSpans.indexOf(span.n) > -1);
```
bripkens: How about we start reporting the `k` field as part of the span? Values would be:

- 1: Entry
- 2: Exit
- 3: Intermediate, local

This check would then be really simple :).

src/tracing/index.js
```diff
@@ -21,7 +21,7 @@ exports.init = function(_config) {
     require('./opentracing').init(config, automaticTracingEnabled);
 
     if (automaticTracingEnabled) {
-      require('./hook').init(config);
+      require('./cls');
```
bripkens: Why is it required if it doesn't need any initialization? As far as I can tell, requiring the module has no side effects, so this line can be removed.

src/tracing/instrumentation/elasticsearch.js
```diff
@@ -86,7 +71,7 @@ function instrumentApi(client, action, info) {
           onSuccess(response);
         }
 
-        return cb.apply(this, arguments);
+        return cb.apply(client, arguments);
```
bripkens: Why do you change the call context?

bripkens: Context: Passing `this` ensures that we do not modify user code.

src/tracing/instrumentation/httpClient.js
```diff
@@ -6,116 +6,110 @@ var discardUrlParameters = require('../../util/url').discardUrlParameters;
...
+            method: clientRequest.method,
+            url: completeCallUrl,
+            status: res.statusCode
+          }
+        };
+        span.d = Date.now() - span.ts;
+        span.error = res.statusCode >= 500;
+        span.ec = span.error ? 1 : 0;
+        transmission.addSpan(span);
+        logger.info('adding client Span is %s tid: %s sid: %s', span.n, span.t, span.s);
```
bripkens: This will result in a ton of log spam. Could we remove this?

src/tracing/instrumentation/httpServer.js
```diff
@@ -97,7 +83,7 @@ function getExistingSpanId(req, fallback) {
   fallback = arguments.length > 1 ? fallback : null;
 
   var spanId = req.headers[tracingConstants.spanIdHeaderNameLowerCase];
-  if (spanId == null) {
+  if (spanId === null) {
```
bripkens: Checking with eqeqeq will not respect `undefined`. Was this intentional?

src/tracing/instrumentation/httpServer.js
```diff
@@ -109,7 +95,7 @@ function getExistingTraceId(req, fallback) {
   fallback = arguments.length > 1 ? fallback : null;
 
   var traceId = req.headers[tracingConstants.traceIdHeaderNameLowerCase];
-  if (traceId == null) {
+  if (traceId === null) {
```
bripkens: Checking with eqeqeq will not respect `undefined`. Was this intentional?

src/tracing/instrumentation/httpClient.js
```diff
@@ -6,116 +6,110 @@ var discardUrlParameters = require('../../util/url').discardUrlParameters;
...
-      // listeners. Cleanup immediately.
-      hook.postAndDestroySimulated(uid);
-      throw e;
+      return clientRequest;
     }
 
-    clientRequest.setHeader(tracingConstants.spanIdHeaderName, span.s);
-    clientRequest.setHeader(tracingConstants.traceIdHeaderName, span.t);
-    clientRequest.setHeader(tracingConstants.traceLevelHeaderName, '1');
+    cls.ns.run(() => {
```
bripkens: We should avoid ES6 features for the sake of supporting Node.js 0.12 as mentioned above

src/tracing/instrumentation/httpServer.js
```diff
@@ -5,83 +5,69 @@ var coreHttpModule = require('http');
...
-
-    return server;
-  };
+  var proto = coreHttpModule.Server && coreHttpModule.Server.prototype;
+  shimmer.wrap(proto, 'emit', shimEmit);
 };
 
+function shimEmit(realEmit) {
+  return function(type, req, res) {
+    return cls.ns.runAndReturn(() => {
```
bripkens: We should avoid ES6 features for the sake of supporting Node.js 0.12 as mentioned above

src/tracing/instrumentation/kafka.js
```diff
@@ -108,37 +80,23 @@ function shimEmit(original) {
...
-      ts: Date.now(),
-      d: 0,
-      n: 'kafka',
-      stack: [],
-      data: {
-        kafka: {
-          access: 'consume',
-          service: message.topic
-        }
+    cls.ns.runAndReturn(() => {
```
bripkens: We should avoid ES6 features for the sake of supporting Node.js 0.12 as mentioned above

src/tracing/cls.js
pglombardo: I'm not sure if I understand this @bripkens.  We have a `type` field for SDK spans but a `k` field?

pglombardo: Added in 08e90c5

src/tracing/cls.js
```diff
@@ -0,0 +1,138 @@
+'use strict';
+
+var tracingUtil = require('./tracingUtil');
+var hooked = require('cls-hooked');
+var currentSpanKey = 'csKey';
+
+var exitSpans = ['node.http.client', 'elasticsearch', 'mongo', 'mysql'];
```
bripkens: My intention was that instrumentations will set it themselves and that we therefore do not need such a list. Is this not possible?

pglombardo: I did it this way in Ruby which is why I made an assumption.  I suppose we could instead add the `span.k` assignment in each instrumentation code directly but the  intention was to simplify instrumentation by moving all the span creation details into CLS.

If you have a strong preference I can change it.  Otherwise let's leave it and I'll revisit it on my next run through the code.  lmk

PR #16, opened 2017-11-17
Update supported versions; checks & tests.

pglombardo:
This updates the internal list of supported versions, adds a check before initializing tracing and updates tests to follow these changes.

### DIFF
--- a/src/tracing/index.js
+++ b/src/tracing/index.js
@@ -39,7 +39,7 @@ function setDefaults() {
 
 
 function shouldEnableTracing() {
-  if (config.tracing && config.tracing.enabled === false && exports.supportedVersion(process.versions.node)) {
+  if (config.tracing && config.tracing.enabled === false) {
     logger.info('Not enabling manual tracing as tracing is not enabled via config.');
     return false;
   }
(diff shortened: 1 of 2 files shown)


src/tracing/index.js
```diff
@@ -39,7 +39,7 @@ function setDefaults() {
 
 
 function shouldEnableTracing() {
-  if (config.tracing && config.tracing.enabled === false) {
+  if (config.tracing && config.tracing.enabled === false && exports.supportedVersion(process.versions.node)) {
     logger.info('Not enabling manual tracing as tracing is not enabled via config.');
```
bripkens: We previously supported manual tracing for all Node.js versions. Why is this now unsupported?

src/tracing/index.js
```diff
@@ -49,7 +49,7 @@ function shouldEnableTracing() {
 
 
 function shouldEnableAutomaticTracing() {
-  if (config.tracing && config.tracing.enabled === false) {
+  if (config.tracing && config.tracing.enabled === false && exports.supportedVersion(process.versions.node)) {
```
bripkens: In this case we should log a different message.

src/tracing/index.js
pglombardo: With the new CLS, we are now limited by the support provided by cls-hooked.  This is why I proposed also the `engine` entry as a warning on installation.

src/tracing/index.js
pglombardo: We can log a message but it would be better if we warned on install via `engine`: https://docs.npmjs.com/files/package.json#engines

src/tracing/index.js
pglombardo: Additionally, falling back to continuation-local-storage as an alternative for 8.0.0 & 8.1.4 won't work as it's broken on those versions.

bripkens: This restriction shouldn't exist. Manual tracing shouldn't have a dependency on automatic tracing.

src/tracing/index.js
bripkens: Nope, we shouldn't warn on install. The sensor even works on Node.js 0.12. IMHO we should consider automatic tracing an optional feature. This is how we treated it before. Also, correctly interpreting the engine field is harder than interpreting a good log message.

src/tracing/index.js
pglombardo: I forgot that this was manual tracing.  I'll remove.

src/tracing/index.js
pglombardo: Ok I'll update the log message.

bripkens:
👍

PR #17, opened 2017-11-17
Replaced link to non-public internal library

jhorowitz:
The README contains a link to what I assume is an internal and non-public library. This PR fixes the broken link and directs it to the external library.


bripkens:
Nicely spotted @jhorowitz. Thank you 👍

PR #15, opened 2017-11-17
process cmdline can be changed

pglombardo:
Changing the `process.title` ([1]) results in failing agent announce because the process title is cached in [2]. We should read the file on every agent announce to support this.

[1] https://nodejs.org/api/process.html#process_process_title
[2] https://github.com/instana/nodejs-sensor/blob/master/src/cmdline.js

### DIFF
--- a/src/agentConnection.js
+++ b/src/agentConnection.js
@@ -8,6 +8,7 @@ var atMostOnce = require('./util/atMostOnce');
 var agentOpts = require('./agent/opts');
 var buffer = require('./util/buffer');
 var pidStore = require('./pidStore');
+var cmdline = require('./cmdline');
 var http = require('./http');
 
 // How many extra characters are to be reserved for the inode and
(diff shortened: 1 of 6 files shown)


src/agentConnection.js
```diff
@@ -8,14 +8,14 @@ var atMostOnce = require('./util/atMostOnce');
...
 var pidStore = require('./pidStore');
-var cmdline = require('./cmdline');
 var http = require('./http');
 
 // How many extra characters are to be reserved for the inode and
 // file descriptor fields in the sensor announce cycle.
 var paddingForInodeAndFileDescriptor = 200;
 
 exports.announceNodeSensor = function announceNodeSensor(cb) {
+  var cmdline = require('./cmdline');
```
bripkens: Requiring something doesn't always reevaluate the the command line module. In fact, it only does it once (for the initial require). We probably need to properly handle this, i.e. add a `getCmdline()` function or similar to the `cmdline` module. WDYT?

pglombardo: Ok makes sense.  Will do.

bripkens:
Maybe we should have squashed this? 😏

pglombardo:
Bad habits are hard to break...  Gonna do this in all sensor repos going forward.

PR #31, opened 2018-01-24
feat(tracing): Add Bluebird tracing support

pglombardo:


schmod:
Any reason why you wouldn't want to just drop in [cls-bluebird](https://www.npmjs.com/package/cls-bluebird)?  

We've had reasonably good luck using `cls-bluebird` with the Instana CLS namespace.

bripkens:
I think this is a fair suggestion @schmod. Right now one of the first objectives is to clearly isolate what kind of bluebird usage results in the async context issues. Curiously, all tests are passing right now. Once identified, cls-bluebird would be a great alternative to a reimplementation (potentially even with the ability to let the user decide whether they want this).

schmod:
Bluebird's `Promise.all()` and `Promise.map()` methods consistently cause CLS issues for us if we aren't using cls-bluebird.

schmod:
Here's a simple test-case showing where bluebird loses the CLS context:
https://runkit.com/schmod/cls-bluebird-test

Interestingly, it works if you require Bluebird _after_ `continuation-local-storage`...

pglombardo:
Looks great @bripkens.  Thanks for the help/test case @schmod.

bripkens:
Thank you @schmod and @pglombardo!

PR #32, opened 2018-01-25
add support for ioredis

### LINKED ISSUE #30: ioredis support
Would it be possible to add support for [ioredis](https://github.com/luin/ioredis) instrumentation in addition to the recently-added support for [node_redis](https://github.com/NodeRedis/node_redis)?

schmod:
Adds instrumentation for the [ioredis](https://github.com/luin/ioredis) Redis client.

Closes #30.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "eslint": "1.7.2",
     "eslint-config-airbnb": "0.1.0",
     "express": "4.14.0",
+    "ioredis": "^3.2.2",
     "kafka-node": "^1.0.7",
     "lodash": "^4.15.0",
     "mocha": "2.3.3",
--- a/src/tracing/instrumentation/redis.js
+++ b/src/tracing/instrumentation/redis.js
@@ -19,18 +19,31 @@ exports.deactivate = function() {
 
 exports.init = function() {
   requireHook.on('redis', instrument);
+  requireHook.on('ioredis', instrumentIoredis);
 };
 
 
 function instrument(redis) {
-  shimmer.wrap(redis.RedisClient.prototype, 'internal_send_command', instrumentInternalSendCommand);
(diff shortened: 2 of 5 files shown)


bripkens:
Thank you for this great PR @schmod ❤️ ! I manually merged it with some changes to adapt it to our guidelines (admittedly not documented so you couldn't have known). Commit: https://github.com/instana/nodejs-sensor/commit/327750e01308b83c67afe57195e1056c8df18d26

Thanks again!

bripkens:
Once the tests are executed & green on travis, I will release your changes.

schmod:
@bripkens I think you conflated `multi` and `pipeline` in your commit -- ioredis has both.

bripkens:
Ah, I was wondering why you added this and why the tests were still green after I removed this condition. We will need additional tests for pipeline and then also adapt the implementation accordingly.

schmod:
You can probably just reuse the `multi` test -- the semantics are exactly the same.

bripkens:
The semantics are slightly different as I understand it. `multi` defines a Redis [transaction](https://redis.io/topics/transactions). So one failed command will fail the entire transaction. Because of this, it makes sense to model a transaction as a single span.

I understand `pipeline` as a why to optimize predominantly for [network latency](https://redis.io/topics/pipelining). Because of this, I think that commands should be handled differently when pipelined. More specifically, I do not think that they should be aggregated into a single span (as each command has its own [error state](https://github.com/luin/ioredis#pipelining)).

bripkens:
Alternatively, pipeline commands could be modeled as a single span, but the batch and error count semantics (`ec` field in the span) must be respected.

bripkens:
I will refrain from releasing this for now. We need at least a test to verify that pipelining still works.

bripkens:
Taking of this now.

bripkens:
Follow up in: https://github.com/instana/nodejs-sensor/pull/34

PR #34, opened 2018-01-26
fix(tracing): Support ioredis pipeline calls

bripkens:
This is a follow up to: https://github.com/instana/nodejs-sensor/pull/32

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
 ## Unreleased
- - - Add ioredis instrumentation.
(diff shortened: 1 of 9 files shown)


schmod:
Does it make sense to set the "start" timestamp for a pipeline/multi span as the time when `exec` was called, rather than the time when `multi()` or `pipeline()` were called?  The redis client doesn't do anything meaningful until `exec()` is called.

bripkens:
Great point @schmod. Should improve cases where the construction of the multi / pipeline query takes some time. I changed my commit to include this logic.

PR #37, opened 2018-02-14
Prepare a debug action

bripkens:


basti1302:
@bripkens do we still need this or can it be closed?

bripkens:
@basti1302: I didn't get to continue this. The idea here was to support backchannel messages which would gather all the typically requested debugging data. In this case specifically around loaded modules, e.g. so that these may be requested via the backchannel. There is currently no need for it, which is why I didn't continue with it. This being said, I still think that this will be valuable in the future.

test/actions/source_test.js
```diff
@@ -9,13 +9,13 @@ var config = require('../config');
 var utils = require('../utils');
 
 
-describe('actions/source', function() {
+describe.only('actions/source', function() {
```
basti1302: forgotten only

bripkens: Ah, damn it. We really need a lint step for this…

PR #40, opened 2018-03-07
fix(tracing): Fix support for mysql2 versions prior to 1.5.0

pglombardo:
This fixes `Cannot read property 'prototype' of undefined` when using mysql2 packages
earlier than version 1.5.0.

### DIFF
--- a/src/tracing/instrumentation/mysql.js
+++ b/src/tracing/instrumentation/mysql.js
@@ -23,7 +23,9 @@ function instrumentMysql(mysql) {
 
 function instrumentMysql2(mysql) {
   instrumentConnection(mysql.Connection.prototype);
-  instrumentPool(Object.getPrototypeOf('mysql.Pool'));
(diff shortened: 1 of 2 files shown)


src/tracing/instrumentation/mysql.js
```diff
@@ -23,7 +23,9 @@ function instrumentMysql(mysql) {
 
 function instrumentMysql2(mysql) {
   instrumentConnection(mysql.Connection.prototype);
-  instrumentPool(mysql.Pool.prototype);
+  if ( typeof mysql.Pool !== 'undefined' && mysql.Pool ) {
```
bripkens: You can shorten this to 

```javascript
if (mysql.Pool) {
  // …
}
```

as this will cover all truthy cases. You could even shorten it to:

```javascript
mysql.Pool && instrumentPool(mysql.Pool.prototype);
```

PR #42, opened 2018-03-22
Load Tests for the Instana nodeJS sensor

orangecoding:
- Adding a server taking load from jMeter
- Instana nodeJS sensor can be enabled to track the additional load (if any)


bripkens:
Merged manually via da15c0ac2531212274586263416e0322cb9c2a04.

Thank you!

PR #52, opened 2018-04-21
Added test case proving 128bit traceId compatibility 

CodingFabian:
also already separated trace and span id generation

### DIFF
--- a/src/tracing/cls.js
+++ b/src/tracing/cls.js
@@ -43,7 +43,6 @@ exports.startSpan = function startSpan(spanName, traceId, spanId) {
   };
 
   var parentSpan = exports.ns.get(currentSpanKey);
-  var randomId = tracingUtil.generateRandomSpanId();
 
   // If specified, use params
   if (traceId && spanId) {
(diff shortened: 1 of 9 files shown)


src/tracing/cls.js
```diff
@@ -55,9 +54,9 @@ exports.startSpan = function startSpan(spanName, traceId, spanId) {
     span.p = parentSpan.s;
   // last resort, use newly generated Ids
   } else {
-    span.t = randomId;
+    span.t = tracingUtil.generateRandomTraceId();
   }
-  span.s = randomId;
+  span.s = tracingUtil.generateRandomSpanId();
```
pglombardo: This will change the behavior of newly started root spans having the same TraceID and SpanID - is that ok?

CodingFabian: yes this is ok. We could also keep this, but that would make the code path messier.
root spans are identified by missing parent.

bripkens:
LGTM except for the failing test cases :)

PR #54, opened 2018-04-25
Reduce memory footprint when using HTTP agents with `maxSockets: Infinity`.

bripkens:


CodingFabian:
thumbs up from my side. our tests already showed a significant improvement with the hacked solution. maybe to a before and after comparison run for documentation purpose?

bripkens:
Good idea, here we go (execution of `TEST=httpCallSequence ./runLoadTest.bash`):

## Before

before


## After

after

CodingFabian:
👍

PR #59, opened 2018-06-05
fix(tracing): manage tracing context correctly in ioredis instrumentation

basti1302:


run-tests-with-docker.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -e
+
+# We aren't properly shutting down these services. Therefore
+# the internal data store might get corrupted. Completely clean up
+# existing images to ensure that we will not hunt ghost bugs.
+docker-compose kill
```
bripkens: I assume the idea is to execute `run-tests-with-docker.sh` every time one would execute `npm test`? Since this will terminate and remove the containers, won't this be a lot slower?

test/tracing/ioredis/app.js
```diff
@@ -65,6 +68,51 @@ app.get('/values', function(req, res) {
...
+      request('http://127.0.0.1:' + agentPort, function(httpErr, httpRes) {
+        if (httpErr) {
+          log('HTTP call failed', httpErr);
+          return reject(httpErr);
+        }
+        return resolve(httpRes.body + ';' + redisRes);
+      });
+    });
+  }).then(function(result) {
+    res.send(result);
```
bripkens: How about we send information about the current tracing context as part of the HTTP response to the tests? Then we could verify that we have the correct span context in this location (which should be HTTP entry).

Could work similar to this:

https://github.com/instana/nodejs-sensor/blob/master/test/tracing/bluebird/app.js#L101
https://github.com/instana/nodejs-sensor/blob/master/test/tracing/bluebird/test.js#L70-L74

run-tests-with-docker.sh
basti1302: Yes, that's the idea. I'm a bit surprised about this question because this script is just literally  following the advice that was in CONTRIBUTING.md before (written by you in 2016, I think :-) ). 

Btw, on my machine this takes ~12 seconds for spinning up the containers, while the full test suite takes ~ 2 minutes.

I can also remove this from this PR if you prefer and put it back into my .git/info/exclude to keep the script locally.

bripkens: Discussed this face to face. We will keep this change for now. Should once discover any problems, one can always just execute a part of the script.

PR #69, opened 2018-07-13
feat(tracing): Record express.js path templates when tracing HTTP ent…

bripkens:
…ries

### DIFF
--- a/src/tracing/instrumentation/express.js
+++ b/src/tracing/instrumentation/express.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var shimmer = require('shimmer');
+var methods = require('methods');
 
 var requireHook = require('../../util/requireHook');
 var tracingUtil = require('../tracingUtil');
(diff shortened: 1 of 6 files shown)


src/tracing/instrumentation/express.js
```diff
@@ -95,3 +104,51 @@ function annotateHttpRootSpanWithError(err) {
...
+  // DO NOT REMOVE UNUSED PARAMETERS
+  // express.js checks parameter count to decide what kind of handler function
+  // it should invoke.
+  return function(err, req, res, next) { // eslint-disable-line no-unused-vars
+    annotateHttpRootSpanWithPathTemplate(req);
+    return fn.apply(this, arguments);
+  };
+}
+
+function annotateHttpRootSpanWithPathTemplate(req) {
```
basti1302: I think a better name would be `annotateHttpEntrySpanWithPathTemplate`, as it is not necessarily a root span, right?

bripkens: Great point. Changed!

PR #70, opened 2018-07-15
feat(tracing): Record fastify path templates when tracing HTTP entries.

bripkens:


src/util/requireHook.js
```diff
@@ -1,25 +1,87 @@
...
+  Module._load = origLoad;
+  executedHooks = {};
+  byModuleNameTransformers = {};
 };
 
-exports.teardown = function() {
-  Module._load = origLoad || Module._load;
+exports.onModuleLoad = function on(moduleName, fn) {
+  var eventListeners = byModuleNameTransformers[moduleName] = byModuleNameTransformers[moduleName] || [];
+  eventListeners.push(fn);
```
basti1302: We could get rid of the local `eventListeners` var and instead push into `byModuleNameTransformers[moduleName] ` directly, I suppose (makes the two lines a bit easier to grok also, I think).

src/util/requireHook.js
```diff
@@ -1,25 +1,87 @@
...
+  return new RegExp(arr.reduce(buildFileNamePatternReducer, ''));
+};
+
+function buildFileNamePatternReducer(agg, pathSegment, i, arr) {
+  if (agg.length > 0) {
+    agg += '\\' + path.sep;
+  }
+  agg += pathSegment;
+  if (i + 1 === arr.length) {
+    agg += '$';
```
basti1302: If we terminate the regex with `$`, should we maybe also start it with `^`? 

Also, the final `$` could simply be added directly in the `buildFileNamePattern` function instead of here in the reducer function, that is, 

`return new RegExp(arr.reduce(buildFileNamePatternReducer, '') + '$');`

This way we don't need to check `i` versus array length...

src/util/requireHook.js
bripkens: That would mean like this?

```
(byModuleNameTransformers[moduleName] = byModuleNameTransformers[moduleName] || []).push(fn);
```

I don't particularly like this style, I'll gladly change it to this if you prefer?

bripkens: clarified face to face and done.

src/util/requireHook.js
bripkens: done!

> should we maybe also start it with ^?

This is not possible, because this part is dynamic, i.e. an application could have it's code under `/opt`, `/app`, `C:\`…. The idea here is to always to a (very specific) suffix matching.

bripkens: I changed the express.js test case and example so that this is clearer.

PR #72, opened 2018-07-17
fix(tracing): fix result handling in PostgreSQL instrumentation

basti1302:


src/tracing/instrumentation/pg.js
```diff
@@ -72,7 +72,7 @@ function instrumentedQuery(ctx, originalQuery, config, values, callback) {
       span.transmit();
 
       if (originalCallback) {
-        cls.ns.bind(originalCallback).apply(this, error, res);
+        cls.ns.bind(originalCallback).apply(this, [error, res]);
```
bripkens: Can we use `arguments` here just to be safe for the future? Related: We should return whatever the original function returns (to be safe).

bripkens: Similar in line 84, i.e. the call to `originalQuery`. Which might require some restructuring, but at least it would guarantee that we do not swallow a potential fourth parameter.

basti1302: Like this? 33c179619fd2c3b4a0ae9e68cf38586a0483f883

bripkens: Jep, looks good. Though we should note that `Array.prototype.slice.call(arguments)` results in a v8 deoptimization. If possible, we should avoid doing this. If we need to copy this, then it is better to do something like this:

https://github.com/instana/weasel/blob/master/lib/hooks/eventHandlers.js#L36-L40

PR #76, opened 2018-07-24
feat(tracing): report uncaught exceptions 

basti1302:
* finish the currently active span and mark it erroneous
* report the uncaught exception as an event/issue
 
_Update: I addressed the concerns about the process still being able to accept new incoming requests, so this is ready for re-review._


CONFIGURATION.md
```diff
@@ -61,6 +61,26 @@ require('instana-nodejs-sensor')({
 });
 ```
 
+## Disabling Reporting Uncaught Exceptions
```
pglombardo: "Disable Uncaught Exception Reporting" might be better wording

pglombardo:
Slick stuff @basti1302.  And nice test coverage too.

CONFIGURATION.md
```diff
@@ -61,6 +61,26 @@ require('instana-nodejs-sensor')({
...
+The Instana Node.js sensor has the ability to report uncaught exceptions. By default, a Node.js process will be terminated by an uncaught exception (see [Node.js docs](https://nodejs.org/api/process.html#process_event_uncaughtexception)). If uncaught exception reporting is enabled, the Instana Node.js sensor will register a listener for the `uncaughtException` event and take the following actions when an uncaught exception occurs:
+
+* Report this as an incident to Instana, including the uncaught exception and its stack trace.
+* Finish the currently active span and mark it as an error (if automatic tracing is enabled).
+
+It will then rethrow the original exception to terminate the Node.js process. (Keeping a Node.js process running after an uncaught exception has occurred is strongly discouraged, as the process might be in an inconsistent state, see [Node.js docs](https://nodejs.org/api/process.html#process_warning_using_uncaughtexception_correctly).)
+
+A timeout of one second is applied, so if one or both of these actions have not completed after one second, the process will be terminated anyway.
```
bripkens: During this one second, will HTTP servers or message consumers continue to accept requests / consume messages? If so, I think we are influencing Node.js applications. Normally, message consumption would stop right away. The same is true for HTTP servers. If they don't stop consuming / accepting, then application code might run into inconsistent state and fail. This could, in the worst case, result in inconsistent / wrong data.

The only way to circumvent this that I could think of would be forcing synchronous HTTP requests to be used for data transmission to the agent. This will ensure that application code never gets any processing time in order to handle requests / messages.

basti1302: Very good point. I looked into doing synchronous requests briefly, I can look into that again more closely to see if it could be used here. I agree, if that's not possible we should definitely do this opt-in/disabled by default.

basti1302: @bripkens 

It seems the only possible option would be to use something like https://github.com/ForbesLindesay/sync-rpc, which does `child_process.spawnSync` to start a worker that does the asynchronous work for your main process. This is quite a hack and something I'd rather not do in an `uncaughtException` handler. My preferred route would be switch back to opt-in and more explicitly state the risks.

Thoughts?

basti1302: Another idea about not keeping the process alive and able to accept more requests - I think we might get a pretty good result when using core idea of sync-rpc (spawning another process) trimmed down to our actual needs. sync-rpc uses `spawnSync` to wait for the result of the asynchronous action. We actually do not care about the results of our two http requests, we can trigger them in a fire-and-forget style.

So, here's the new plan: Spawn a child process via `spawn` instead of `spawnSync`and then terminate the main process immediately. The child process can then take as much time as it wants to talk to the agent and the main process terminates synchronously in the `uncaughtException` handler.

basti1302:
I updated this once more to address @bripkens concerns:

- I added an optional dependency to [netlinkwrapper](https://github.com/JacobFischer/netlinkwrapper), which wraps a small C++ TCP library - this enables us to make synchronous HTTP requests in the `uncaughtException` handler. This handler now executes completely synchronous, blocking the event loop to avoid processing more incoming requests while the application might already be in an inconsistent state. 
- I still made the feature disabled by default again, because
    - it still is true that a any mechanism that restarts the Node.js process once it has died will kick in a bit later if this is activated
    - it also might change the behaviour - in particular, since we now rethrow the error synchronously, other `uncaughtException` handlers that have been registered after our handler are not executed.

But both of these drawbacks risks are much clearer/easier to judge as opposed to the concern of accepting more incoming requests while an uncaught exception might already have compromised the consistency of the application's internal state.

I think this as close to a reasonable tradeoff as we might get. I do not think that there is a happy one-size-fits-all solution for this feature requirement.

package.json
```diff
@@ -50,12 +50,14 @@
...
     "semver": "5.4.1",
+    "serialize-error": "^2.1.0",
     "shimmer": "1.1.0"
   },
   "optionalDependencies": {
+    "@risingstack/v8-profiler": "5.7.11",
     "event-loop-stats": "1.0.0",
     "gcstats.js": "1.0.0",
-    "@risingstack/v8-profiler": "5.7.11"
+    "netlinkwrapper": "basti1302/netlinkwrapper#v1.0.0-node-10-fix"
```
basti1302: See https://github.com/JacobFischer/netlinkwrapper/pull/4

CONFIGURATION.md
```diff
@@ -61,6 +62,30 @@ require('instana-nodejs-sensor')({
...
+});
+```
+
+The [Node.js docs](https://nodejs.org/api/process.html#process_event_uncaughtexception) advise that the "correct use of 'uncaughtException' is to perform synchronous cleanup of allocated resources (e.g. file descriptors, handles, etc) before shutting down the process." Instana adheres to this and only executes synchronous actions before rethrowing the error.
+
+The downside of this is that it might prolong the time it takes for the application process to finally terminate after the uncaught exception. Since the actions are synchronous, no other incoming requests will be accepted by your application during this time, that is, the process will be unresponsive to incoming request (HTTP, WebSockets, ...). This is an important safeguard, since the application might already be in an inconsistent state, so it would not be safe to accept/process any more requests in this process. However, this has two drawbacks:
+
+* If you have a mechanism in place that restarts the Node.js process once it has crashed, enabling uncaught exception tracking might lead to more failed requests, because the application process is kept alive for a bit longer, delaying the automatic restart.
```
bripkens: It would be great if we could give some indication for how long the process will be kept alive. At least this is some information I would want to have when making this decision.

CONFIGURATION.md
```diff
@@ -61,6 +62,30 @@ require('instana-nodejs-sensor')({
 });
 ```
 
+## Reporting Uncaught Exceptions
+The Instana Node.js sensor has the ability to report uncaught exceptions. By default, a Node.js process will be terminated by an uncaught exception (see [Node.js docs](https://nodejs.org/api/process.html#process_event_uncaughtexception)). If uncaught exception reporting is enabled, the Instana Node.js sensor will register a listener for the `uncaughtException` event and take the following actions when an uncaught exception occurs:
```
bripkens: Lovely explanation!

src/agentConnection.js
```diff
@@ -241,3 +244,98 @@ function sendData(path, data, cb, ignore404) {
...
+      port = parseInt(port, 10);
+    } catch (nonNumericPortError) {
+      logger.warn('Detected non-numeric port configuration value %s, uncaught exception will not be reported.', port);
+      return;
+    }
+  }
+
+  try {
+    netLink.connect(port, agentOpts.host);
+    netLink.blocking(false);
```
bripkens: Shouldn't this be `true`? (I didn't read the docs)

src/agentConnection.js
```diff
@@ -241,3 +244,98 @@ function sendData(path, data, cb, ignore404) {
...
+  try {
+    var payload = JSON.stringify(data);
+    var payloadLength = buffer.fromString(payload, 'utf8').length;
+  } catch (payloadSerializationError) {
+    logger.warn('Could not serialize payload, uncaught exception will not be reported.',
+      { error: payloadSerializationError });
+    return;
+  }
+
+  netLink.write(
```
bripkens: Is it possible to configure a timeout for this? If not, what are the timeouts (we might want to document them above)?

src/util/uncaughtExceptionHandler.js
```diff
@@ -0,0 +1,105 @@
...
+  var eventPayload = createEventForUncaughtException(uncaughtError);
+  agentConnection.reportUncaughtExceptionToAgentSync(eventPayload, spans);
+}
+
+
+function createEventForUncaughtException(uncaughtError) {
+  var eventText = JSON.stringify(serializeError(uncaughtError));
+  return {
+    title: 'A Node.js process terminated abnormally due to an uncaught exception.',
+    text: eventText,
```
bripkens: We can use markdown for event texts and should leverage it to make this data more readable.

src/util/uncaughtExceptionHandler.js
```diff
@@ -0,0 +1,105 @@
...
+function logAndRethrow(err) {
+  // Remove all listeners now, so the final throw err won't trigger other registered listeners a second time.
+  var registeredListeners = process.listeners(uncaughtExceptionEventName);
+  if (registeredListeners) {
+    registeredListeners.forEach(function(listener) {
+      process.removeListener(uncaughtExceptionEventName, listener);
+    });
+  }
+  // eslint-disable-next-line max-len
+  logger.error('The Instana Node.js sensor caught an otherwise uncaught exception to generate a respective Instana event for you. Instana will now rethrow the error to terminate this process, otherwise the application would be left in an inconsistent state, see https://nodejs.org/api/process.html#process_warning_using_uncaughtexception_correctly. The next line on stderr will look as if Instana crashed your application, but actually the original error came from your application code, not from Instana. Since we rethrow the original error, you should see its stacktrace below (depening on how you operate your application and how logging is configured.)');
```
bripkens: ❤️

CodingFabian: In java you can manipulate stacktraces to prevent such problems. maybe that possible in js as well?

src/agentConnection.js
basti1302: No, you don't won't the socket to be in blocking mode - this has nothing to do with the call blocking the event loop, it's a OS socket level thing (see https://www.scottklement.com/rpg/socktut/nonblocking.html for an explanation). We want to write to the socket and then end the process. If we can't write, we want to simply fail early instead of blocking indefinitely.

src/agentConnection.js
basti1302: No, that is not possible. As far as I understand it, writing to a socket that is not in blocking mode can not fail due to a timeout. The write command simply puts the data into the socket send buffer and then returns. After that, if and when the send buffer is emptied or if and when the peer receives the data is not the concern of the Node.js process anymore, that is handled by the network layer. 

The other thing for which a timeout might apply is the connect call. This just fails immediately when the connection can not be made.

src/util/uncaughtExceptionHandler.js
basti1302: @CodingFabian we never have access to the stack trace I'm talking about there. We need to terminate the Node.js process and for this, we rethrow the error. The stack trace will then include the line from which we rethrow it, there's no way around that.

src/util/uncaughtExceptionHandler.js
basti1302: Good point!

src/util/uncaughtExceptionHandler.js
CodingFabian: yes thats what I meant. In java you can rethrow the original exception, so the line that rethrows would not appear
https://stackoverflow.com/questions/42754270/re-throwing-exception-in-nodejs-and-not-losing-stack-trace
seems like its not really possible in js

CONFIGURATION.md
basti1302: A few milliseconds. Basically as long as it needs to open a new TCP connection to the agent (or fail doing so) and write the required data into the socket buffer. I'll add something there.

src/util/uncaughtExceptionHandler.js
basti1302: Formatted error in UI:

bildschirmfoto 2018-09-03 um 13 35 39

bripkens: 👍

PR #77, opened 2018-08-03
feat(tracing): Trace TraceContext headers

bripkens:


basti1302:
This has gotten stale and as far as I know there is no immediate need (read feature request from any customer) for this, closing for now.

PR #78, opened 2018-09-04
feat(tracing): instrument MSSQL

basti1302:
Adds MSSQL (Microsoft SQL Server) instrumentation (supports [mssql](https://www.npmjs.com/package/mssql), version >= 4 via [tedious driver](https://www.npmjs.com/package/tedious)).


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
+ - Add MSSQL (Microsoft SQL Server) instrumentation (supports [mssql](https://www.npmjs.com/package/mssql), version >= 4 via [tedious driver](https://www.npmjs.com/package/tedious)).
```
bripkens: Is `tedious` something that is configurable or why is this explicitly mentioned?

src/tracing/instrumentation/mssql.js
```diff
@@ -0,0 +1,227 @@
...
+
+function instrumentRequest(Request) {
+  shimmer.wrap(Request.prototype, 'query', shimMethod.bind(null, instrumentedRequestMethod));
+  shimmer.wrap(Request.prototype, 'execute', shimMethod.bind(null, instrumentedRequestMethod));
+  shimmer.wrap(Request.prototype, 'batch', shimMethod.bind(null, instrumentedRequestMethod));
+  shimmer.wrap(Request.prototype, 'bulk', shimMethod.bind(null, instrumentedBulk));
+}
+
+
+function shimMethod(instrumentedFunction, originalFunction) {
```
bripkens: Nice pattern!

src/tracing/index.js
```diff
@@ -8,6 +8,24 @@ var tracingEnabled = false;
...
+  './instrumentation/fastify',
+  './instrumentation/httpClient',
+  './instrumentation/httpServer',
+  './instrumentation/ioredis',
+  './instrumentation/kafka',
+  './instrumentation/mongodb',
+  './instrumentation/mssql',
+  './instrumentation/mysql',
+  './instrumentation/pg',
+  './instrumentation/redis',
```
bripkens: I think we forgot to extend this definition of entry / exits: https://github.com/instana/nodejs-sensor/blob/master/src/tracing/cls.js#L11

(and maybe we could change it in some way so that it is easier to remember?)

src/tracing/instrumentation/mssql.js
```diff
@@ -0,0 +1,227 @@
...
+  return function() {
+    if (isActive && cls.isTracing()) {
+      var originalArgs = new Array(arguments.length);
+      for (var i = 0; i < arguments.length; i++) {
+        originalArgs[i] = arguments[i];
+      }
+      if (originalArgs.length >= 2 && typeof originalArgs[1] === 'function') {
+        originalArgs[1] = cls.ns.bind(originalArgs[1]);
+      }
+      cls.ns.set('com.instana.mssql.stmt', originalArgs[0]);
```
bripkens: Currently every `cls.ns.set` is added to a span for cleanup purposes. Also, when I understand correctly, then there is no guarantee that a prepared statement is executed within the same async context as the one used to prepare it. So as far as I can tell, this won't work?

As an alternative, couldn't we register a property `__instanaStatement` or similar on the `PreparedStatement` so that we have access to this later on? That way we also don't need to worry about cls cleanup.

test/tracing/mssql/app.js
```diff
@@ -0,0 +1,552 @@
...
+      'CREATE PROCEDURE testProcedure' +
+      '    @username nvarchar(40)' +
+      'AS' +
+      '    SET NOCOUNT ON;' +
+      '    SELECT name, email' +
+      '    FROM UserTable' +
+      '    WHERE name = @username;'
+    );
+  })
+  .then(function() {
```
bripkens: Can we extend this app setup with the creation of the prepared statement that is then used within one of the HTTP handlers? That way we could verify the comment I made above.

CHANGELOG.md
basti1302: Yes, there is one alternative driver [msnodesqlv8](https://www.npmjs.com/package/msnodesqlv8) (which only works on Windows, though). Since we instrument the mssql package and not the driver, in theory it should also work with msnodesqlv8 but I haven't tested it at all, that's why the tedious driver is mentioned explicitly. The motivation for this PR is an actual customer requirement with the combination mssql/tedious.

src/tracing/instrumentation/mssql.js
basti1302: Hm, I think the `cls.ns.bind(originalArgs[1])` makes it work, even if the context is switched. I'll look into the `__instanaStatement` approach though, I agree that it would be nicer.

src/tracing/index.js
basti1302: Oh, bummer. I think `kafka` is also missing. For this PR I would just add mssql there.

To improve this I would 
- remove this list and the logic to set `span.k` from `cls.js` completely and make it a responsibility of the particular instrumentations to set `span.k`, probably as a non-optional argument to `startSpan`.
- Maybe add a validation check to `startSpan` which rejects/drops if `k` is not provided.
- Add expectations for `span.k` to all instrumentation tests.

But this should be a separte PR.

test/tracing/mssql/app.js
basti1302: Funny, only saw this comment now but that's exactly what I did in the last commit.

And turns you were somewhat right: cls.ns.set actually throws in that setup because there hasn't been any previous call to cls.ns.bind et al.

basti1302:
All comments are addressed IMO, waiting for Travis, then merging.

src/tracing/index.js
bripkens: image

bripkens:
👍

PR #79, opened 2018-09-05
feat(tracing): support both mongdb >= 3.0.6 and < 3.0.6

basti1302:
MongoDB completely rewrote their Node.js APM support, removing the operationIdGenerator/operationId concept in the process. This broke our MongoDB instrumentation. This change fixes it for mongodb package versions >= 3.0.6, while keeping the behaviour for older versions the same.


package.json
```diff
@@ -76,8 +76,8 @@
     "kafka-node": "^1.0.7",
     "lodash": "^4.17.4",
     "mocha": "2.3.3",
-    "mongodb": "2.2.33",
-    "mongoose": "4.7.1",
+    "mongodb": "3.1.4",
+    "mongoose": "5.2.13",
```
basti1302: Testing with latest versions on Travis but I tested it manually with the versions we used previously. I guess having automated tests with multiple package versions would be awesome, but quite a lot of effort to build.

src/tracing/instrumentation/mongodb.js
```diff
@@ -55,54 +62,74 @@ function instrument(mongodb) {
...
+    return;
+  }
+
+  var traceId;
+  var parentSpanId;
+  // event.operationId is only present prior to mongodb package version 3.0.6, see above.
+  if (event.operationId) {
+    traceId = event.operationId.traceId;
+    parentSpanId = event.operationId.parentSpanId;
+    supportsOperationIds = true;
```
basti1302: As soon as we have seen one operationId attached to an event, we know that this is a version < 3.0.6.

src/tracing/instrumentation/mongodb.js
```diff
@@ -55,54 +62,74 @@ function instrument(mongodb) {
...
+      filter: stringifyWhenNecessary(event.command.filter),
+      query: stringifyWhenNecessary(event.command.query)
+    }
   };
+
+  if (event.operationId) {
+    event.operationId.traceId = span.t;
+    event.operationId.parentSpanId = span.p;
+  }
+  requests[getUniqueRequestId(event)] = span;
```
basti1302: Before this PR, we always unnecessarily wrapped the span in another object, this wrapping has been removed.

package.json
bripkens: This is something that has been on my mind as well for quite some time. We already have instrumentations which only work in specific versions, but we aren't properly testing these. Would be great to attempt such a test harness rather sooner than later.

test/tracing/mongodb_test.js
```diff
@@ -99,7 +102,9 @@ describe('tracing/mongodb', function() {
     });
   });
 
-  it('must trace find requests with cursors', function() {
+  // correlating multiple operations (bulk writes, find with getMore) is broken in mongodb APM support since
+  // version 3.0.6, see https://groups.google.com/forum/#!topic/node-mongodb-native/uBae-HO0zw8
```
bripkens: Some time ago, I was contemplating to switch instrumentation to https://github.com/mongodb-js/mongodb-core, i.e. the underlying mongodb driver. Nothing to do right now, but an option to keep in mind. It might be more stable to instrument the methods within that driver.

basti1302: It just seems kind of a shame to resort to monkey patching when this is actually one of the rare libs that offer an official APM interface. Also, the monkey patching approach has its own set of downsides, so I'm not sold on the idea that this move would be beneficial.

PR #80, opened 2018-09-12
Fix mysql trace context

basti1302:


src/tracing/instrumentation/mysql.js
```diff
@@ -102,56 +102,65 @@ function instrumentedQuery(ctx, originalQuery, statementOrOpts, valuesOrCallback
     }
   }
```
basti1302: This looks like a large diff, but I basically only changed two things: 
- Wrap everything in `cls.ns.runAndReturn(function() {` ... `}` (to make cls add a new node to its tree, so that the current parent span data is not overwritten
- bind the original callback to cls: `originalCallback = cls.ns.bind(originalCallback);`

PR #82, opened 2018-09-20
feat(tracing): support secrets

basti1302:


src/tracing/instrumentation/httpServer.js
```diff
@@ -124,3 +129,19 @@ function getExistingTraceId(req, fallback) {
...
+function filterParams(queryString) {
+  if (!queryString ||
+      queryString === '' ||
+      !(typeof queryString === 'string')) {
+    return queryString;
+  }
+  return queryString.split('&').filter(function(param) {
+    var key = param.split('=')[0];
+    if (key) {
+      return !agentOpts.secretsMatcher(key);
```
bripkens: Maybe rename `secretsMatcher` to something that hints at the response? Wasn't clear to me when reading this whether true of false means to hide the data. So possibly something like `isAllowedKey` or something?

basti1302: You still got it backwards, which might prove your point, I guess. It is called `isSecret` in Java land. If it returns true, the key must be scrubbed. (Note the `!` in line 143).

basti1302: Renamed it to `isSecret`.

PR #88, opened 2018-10-09
Add support for non-bunyan loggers

SerayaEryn:
This adds support for non-bunyan loggers that do not have `logger.child()`, `logger.addStream()` or `logger.level()` like [winston](https://github.com/winstonjs/winston).

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -9,6 +9,9 @@ var parentLogger;
 exports.init = function(config) {
   if (config.logger) {
     parentLogger = config.logger.child({module: 'instana-nodejs-logger-parent'});
+  } else if (config.nonBunyanLogger) {
+    parentLogger = config.nonBunyanLogger;
+    return;
   } else {
     parentLogger = bunyan.createLogger({name: 'instana-nodejs-sensor'});
   }
@@ -27,6 +30,10 @@ exports.init = function(config) {
 exports.getLogger = function(moduleName) {
   if (!parentLogger) {
     exports.init({});
+  }
+
+  if (typeof parentLogger.child !== 'function') {
+    return parentLogger;
   }
 
   var logger = parentLogger.child({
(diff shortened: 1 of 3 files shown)


src/logger.js
```diff
@@ -8,7 +8,10 @@ var parentLogger;
 
 exports.init = function(config) {
   if (config.logger) {
-    parentLogger = config.logger.child({module: 'instana-nodejs-logger-parent'});
+    parentLogger = config.logger;
```
basti1302: If I understand your change correctly, your intention would be to use `config.nonBunyanLogger` to pass in a logger object. Any particular reason to change the above line, handling the Bunyan case?

basti1302:
Hi Denis,

supporting other loggers is definitely a worthwile feature, thanks for bringing this up and submitting a PR! 

I'm not 100% sure about the best configuration API to offer this. Here are some thoughts: 

1. Having a separate configuration property `config.nonBunyanLogger` for this feels a bit cumbersome. I think auto-detecting if `config.logger` is a Bunyan logger should be possible (for example, by checking for the existence of a `child` function) and might make using this a bit smoother. WDYT?
1. If its not a Bunyan logger, we should probably validate if the usual logging function (`debug`, `info`, `warn`, `error`) exist on the object received as `config.logger`.
1. Whatever we decide on, I think we should add a test for it to `logger_test.js`.
1. Finally, this new feature would need to be documented in CONFIGURATION.md.

It probably makes sense to come to a decision regarding (1.) before tackling (3.) and (4.). If you would rather not invest time in (3.) and (4.), I'm happy to take over those chores. 

I'd like to hear your thoughts on (1.), though.

src/logger.js
SerayaEryn: This was changed accidentally. i removed this change.

SerayaEryn:
Hi @basti1302,

I agree, using the `config.logger` option is better. 

I can add some test cases for the new feature and document it.

basti1302:
Cool, looking forward to it :-)

basti1302:
Looks great, thanks! 🎉 

This will go out with the next version.

PR #93, opened 2018-10-24
feat(tracing): instrument amqplib (RabbitMQ)

basti1302:


.travis.yml
```diff
@@ -23,7 +23,7 @@ before_script:
   - sudo /etc/init.d/mysql stop
   - sudo /etc/init.d/postgresql stop
   - docker-compose kill && docker-compose rm -f && docker-compose up -d
-  - sleep 180
+  - sleep 45
```
basti1302: This was changed earlier in an attempt to make the kafka tests less flaky, however, we later found the reason for the flakiness and this timeout does not help, so we can just change it back to the original value.

PR #103, opened 2018-11-14
fix(tracing): improve kafka tracing continuity

basti1302:


basti1302:
this was included in #104

PR #117, opened 2019-01-14
Update ES Database to include msearch, and mget calls as known DB type calls

DtRWoS:
Logging of query content wont work due to the unknown length and structure of the "multiple" actions


basti1302:
Hi there,

thanks for the contribution. I would like to also add tests for the newly instrumented methods, but it might take a few days until I get around to do that.

Kind regards

  Bastian

basti1302:
This has landed in 1.60.0.

PR #148, opened 2019-06-18
don't fail on errors with circular references

sklose:
we noticed some libraries (Microsoft's @azure/xxxx) throw error objects that have cyclic references and if we attach that error object to the span ...

```typescript
span.log({ event: "error", "error.object": e });
span.setTag("message", e.message);
span.setTag(Tags.ERROR, true);
```

... then it will cause the following error which results in an unhandled promise rejection.

```
TypeError: Converting circular structure to JSON
    at JSON.stringify (\u003canonymous\u003e)
    at sendData (/usr/src/node_modules/@instana/collector/src/agentConnection.js:219:22)
    at Object.sendSpans (/usr/src/node_modules/@instana/collector/src/agentConnection.js:192:3)
    at Timeout.transmitSpans (/usr/src/node_modules/@instana/core/src/tracing/spanBuffer.js:76:24)
```

### DIFF
--- a/packages/collector/src/agentConnection.js
+++ b/packages/collector/src/agentConnection.js
@@ -216,7 +216,19 @@ function sendData(path, data, cb, ignore404) {
     ignore404 = false;
   }
 
-  var payload = JSON.stringify(data);
+  const circularReplacer = () => {
+    const seen = new WeakSet();
+    return (_, value) => {
+      if (typeof value === "object" && value !== null) {
+        if (seen.has(value)) {
+          return;
+        }
+        seen.add(value);
+      }
+      return value;
+    };
+  };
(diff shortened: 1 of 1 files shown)


basti1302:
Thanks a lot. Merged via e687988022938598d7b31cff3a4d0fc57607bbc1.

By the way, your commit had a tiny, but crucial flaw. 

```
var payload = JSON.stringify(data, circularReplacer);
```

needs to be 

```
var payload = JSON.stringify(data, circularReplacer());
```

But nevermind, I've fixed that and added a test for the scenario you described. Thanks for contributing! 🎉

basti1302:
This has landed in `1.68.2`.

PR #149, opened 2019-06-19
fix(tracing): don't fail on span data with circular references

basti1302:
Succeeds #148


basti1302:
This has landed in `1.68.2`.

PR #153, opened 2019-06-24
fix(tracing): fix require cycle when client code also uses cls-bluebird

basti1302:
By requiring cls-bluebird before registering the require hook for
bluebird.

Supercedes #152


basti1302:
This has landed in 1.68.4.

PR #154, opened 2019-06-24
fix(tracing): fix instrumentation of https client for Node.js 8.9.0

basti1302:
Node had backported the refactoring to use _http_client directly in
https (instead of going through http) from Node.js 9.0.0 to
Node.js 8.9.0, only to revert that immediately in Node.js 8.9.1 a few
days later.

Well, there are folks out there using _exactly_ 8.9.0, so this fix makes
HTTPS client calls tracing possible for that Node.js version.


basti1302:
This has landed in 1.68.4.

PR #164, opened 2019-08-19
basePath is deprecated. Use prefix instead.

beep-boop-beep:
See: https://www.fastify.io/docs/latest/Server/#prefix

Fix for:
https://github.com/instana/nodejs-sensor/issues/163

### DIFF
--- a/packages/core/src/tracing/instrumentation/frameworks/fastify.js
+++ b/packages/core/src/tracing/instrumentation/frameworks/fastify.js
@@ -97,5 +97,5 @@ function annotateHttpEntrySpanWithPathTemplate(app, opts) {
     return;
   }
 
-  span.data.http.path_tpl = (app.basePath || '') + (opts.url || opts.path || '/');
+  span.data.http.path_tpl = (app.prefix || '') + (opts.url || opts.path || '/');
 }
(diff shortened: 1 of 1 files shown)


packages/core/src/tracing/instrumentation/frameworks/fastify.js
```diff
@@ -97,5 +97,5 @@ function annotateHttpEntrySpanWithPathTemplate(app, opts) {
     return;
   }
 
-  span.data.http.path_tpl = (app.basePath || '') + (opts.url || opts.path || '/');
+  span.data.http.path_tpl = (app.prefix || '') + (opts.url || opts.path || '/');
```
basti1302: We still need to support older versions of fastify, too. I guess we could check for the existence of `app.prefix` and use that if it exists, otherwise fall back to `app.basePath`?

beep-boop-beep: prefix and basePath do both exist but where changed to getters:
https://github.com/fastify/fastify/blob/7eaaabcbc376d3fdf6e14da146ffb3f5f301a7a3/fastify.js#L187

We have to check if basePath is a getter (which emits the warning). If it is, we have to use the prefix getter.

beep-boop-beep:
i have updated the PR:
https://github.com/instana/nodejs-sensor/pull/164/commits/1762816ed7ea3bfe3873adf404408241d05cfc13

basti1302:
Awesome! Merged via efb7063265757a566a3c52868c02b4998fdba50d and released as 1.71.2.

PR #233, opened 2020-05-11
added upper case level for warn and error

asafarian:
This change is needed for some users, using a special format of log levels in the upper case.
WARN and ERROR was added to be recognized.

This is a change for non-standard usage of log levels.

See Ticket:
https://support.instana.com/hc/en-us/requests/13370


asafarian:
probably not needed, closed, therefore.

PR #239, opened 2020-06-16
test(tracing): remove optional sharp dependency

basti1302:
installing it is unreliable in CI in Node.js 6


basti1302:
This is actually nonsense, installing _any_ native add-on has become unreliable on CircleCI from some reason.

PR #245, opened 2020-07-08
feat(profiling): programmatic profiling support

dmelikyan:


packages/autoprofile/compat/build-env/docker-compose.yml
```diff
@@ -9,9 +9,9 @@ services:
...
-      context: .
-      dockerfile: Dockerfile-win
-    volumes:
-      - ../..:/opt/autoprofile
+#  win-node-prebuild:
+#    build:
+#      context: .
+#      dockerfile: Dockerfile-win
+#    volumes:
+#      - ../..:/opt/autoprofile
```
basti1302, line 17: @dmelikyan Is that intentionally commented out in this PR?

dmelikyan: compiling for windows on linux it isn't working yet, more time needed to make it work, if ever.

PR #258, opened 2020-09-23
fix(metrics): move event-loop-lag to dependencies instead of optionalDependencies

### LINKED ISSUE #257: @instana/collector at ^1.106.2 depends on optional dependencies unsafely
If you install your node modules with:

```
yarn install --production --ignore-optional --frozen-lockfile
```

Then the resulting installation will not run, as `@instana/collector fails to require the `event-loop-lag` module, which is marked as optional, but in the code, it is required just as `const lag = require('event-loop-lag')(1000);`

This means our production build cannot start if we use the `@instana/collector` package.

ThisIsMissEm:
This ensures installation whilst using `--ignore-scripts --ignore-optional --production` will install correctly.

Fixes #257

### DIFF
--- a/packages/shared-metrics/package-lock.json
+++ b/packages/shared-metrics/package-lock.json
@@ -30,6 +30,29 @@
         "js-tokens": "^4.0.0"
       }
     },
+    "@instana/core": {
+      "version": "1.106.2",
+      "resolved": "https://registry.npmjs.org/@instana/core/-/core-1.106.2.tgz",
+      "integrity": "sha512-ROqPpighGgLkR/g7+/qprdDa9ahP2ZnKAi8DrvvBV34/uL0nL3c701K5ZCXKZIJW9t/Fdgx3OBj1rf+RHmrNfg==",
+      "requires": {
+        "async-hook-jl": "^1.7.6",
+        "cls-bluebird": "^2.1.0",
+        "emitter-listener": "^1.1.2",
+        "lru-cache": "^5.1.1",
+        "methods": "^1.1.2",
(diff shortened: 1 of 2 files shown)


packages/shared-metrics/package-lock.json
```diff
@@ -30,6 +30,29 @@
...
+        "shimmer": "1.2.0"
+      },
+      "dependencies": {
+        "semver": {
+          "version": "5.5.1",
+          "resolved": "https://registry.npmjs.org/semver/-/semver-5.5.1.tgz",
+          "integrity": "sha512-PqpAxfrEhlSUWge8dwIp4tZnQ25DIOthpiaHNIthsjEFQD6EvqUKUDM7L8O2rShkFccYo1VjJR0coWfNkCubRw=="
+        }
+      }
+    },
```
basti1302, line 55: Running `npm install` in `packages/shared-metrics` will add this (and all of `@instana/core`'s transitive dependencies, but due to the way this monorepo is managed with lerna, this is actually not supposed to be here, even though `@instana/core` is a dependency of `@instana/shared-metrics`. I'll take care of cleaning that up.

basti1302 (approved):
Hey Emelia,

thanks a ton for identifying this and providing the fix. Apparently I messed that up in 1211193f416422ecf0f6fd52b13a930971c4c71e when extracting the code that uses event-loop-lag from `@instana/collector` into its own sub-package (`@instana/shared-metrics`). :-( 

I need to add a bit of cleanup to the changes here and will merge it then. I'll publish a new release later today.

basti1302:
Merged via #259.

PR #259, opened 2020-09-24
fix(metrics): move event-loop-lag from optionalDeps to dependencies

basti1302:
This ensures installation whilst using
`--ignore-scripts --ignore-optional --production` will install correctly.


basti1302:
This has landed in 1.106.3.

PR #277, opened 2020-12-26
feat(tracing): update handling of W3C trace context headers

basti1302:


packages/collector/test/tracing/misc/w3c_trace_context/app.js
```diff
@@ -54,24 +61,52 @@ if (!downstreamPort) {
...
-    log(`Listening  on port: ${port}`);
+if (useHttp2) {
+  // HTTP 2
+
+  const http2 = require('http2');
+
+  const sslDir = path.join(__dirname, '..', '..', '..', 'apps', 'ssl');
+
+  const server = http2.createSecureServer({
+    key: fs.readFileSync(path.join(sslDir, 'key')),
```
willianpc, line 72: > Maybe I am being a bit paranoid, but we should keep an eye on this (or any other we may have) blocking IO piece of code.
> I think this should be fine for now. As long as we don't get complaints from customers, it should be ok.

nvm, this is test code...

willianpc (approved):
LGTM

PR #278, opened 2021-01-06
feat: do not instrument npm or yarn

basti1302:
If a process is started with `npm start` or `yarn start`, two separate
Node.js processes will be started. The first is `node /path/to/npm`
(or `node /path/to/yarn`). This process will kick off another process
with the actual application (for example, `node /usr/src/app`).

With this change, even when NODE_OPTIONS is set to
"--require /usr/src/app/node_modules/@instana/collector/src/immediate",
only the actual application will be instrumented, but not the npm/yarn
process.

### DIFF
--- a/packages/collector/dummy-app/package-lock.json
+++ b/packages/collector/dummy-app/package-lock.json
@@ -5,36 +5,36 @@
   "requires": true,
   "dependencies": {
     "@instana/autoprofile": {
-      "version": "1.100.1",
-      "resolved": "https://registry.npmjs.org/@instana/autoprofile/-/autoprofile-1.100.1.tgz",
-      "integrity": "sha512-IvCQsyUyN0nkfHbVwEYTtXw42+9DEcK72lLmHN7womrp33ehRnRevCWE/MuJxNaPgnTVbHpeDzVexGfbPw+MiQ==",
+      "version": "1.110.5",
+      "resolved": "https://registry.npmjs.org/@instana/autoprofile/-/autoprofile-1.110.5.tgz",
+      "integrity": "sha512-klGnLhHyP/gp8o8Oo/+P74g1vJHfiCXdlQOLY0LemCEFfLbHkR68DGy0Hm1ahPCpMa/37LzWSs5hE14VWf9Zzw==",
       "optional": true,
       "requires": {
+        "detect-libc": "^1.0.3",
         "nan": "^2.14.0",
-        "node-abi": "*",
+        "node-abi": "^2.19.1",
         "node-gyp": "^6.1.0"
       }
     },
     "@instana/collector": {
-      "version": "1.100.1",
-      "resolved": "https://registry.npmjs.org/@instana/collector/-/collector-1.100.1.tgz",
-      "integrity": "sha512-GhwpaiKoMq5Qb5oKfMJQBo531cMjq/U6HpSdM0i7bLtBSSx+Rk5IBj/YR+JDdXzHVMGaxsPxW2O08MUcn7Uk1A==",
+      "version": "1.110.5",
+      "resolved": "https://registry.npmjs.org/@instana/collector/-/collector-1.110.5.tgz",
+      "integrity": "sha512-jmyOUakmmbvjdxWWQsMB18xLicmQHxscw4+enkjwQU1+pHmrGkN748MuWoLOmCbqsV10mpkUe+Cmby/i0IfMlg==",
       "requires": {
-        "@instana/autoprofile": "^1.100.1",
-        "@instana/core": "^1.100.1",
-        "@instana/shared-metrics": "^1.100.1",
+        "@instana/autoprofile": "1.110.5",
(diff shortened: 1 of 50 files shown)


packages/collector/dummy-app/package-lock.json
```diff
@@ -5,36 +5,36 @@
   "requires": true,
   "dependencies": {
     "@instana/autoprofile": {
-      "version": "1.100.1",
-      "resolved": "https://registry.npmjs.org/@instana/autoprofile/-/autoprofile-1.100.1.tgz",
-      "integrity": "sha512-IvCQsyUyN0nkfHbVwEYTtXw42+9DEcK72lLmHN7womrp33ehRnRevCWE/MuJxNaPgnTVbHpeDzVexGfbPw+MiQ==",
+      "version": "1.110.5",
```
basti1302, line 8: The changes in this file are unrelated to the actual change. I just updated the Instana package to latest for the test application.

packages/collector/test/immediate/test.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
```
basti1302, line 1: This is the integration test for the change in `@instana/collector`. It starts an application under test via `npm start` and verifies that only the actual application under test is instrumented.

packages/collector/test/immediate/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "immediate",
+  "version": "1.0.0",
+  "description": "test application",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/instana/nodejs-sensor.git"
+  },
+  "main": "../apps/express",
```
basti1302, line 10: TODO: This doesn't make sense yet. We need an uninstrumented app here, and ../apps/express is instrumented manually.

packages/core/src/util/excludedFromInstrumentation.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
```
basti1302, line 1: This file contains the main logic for excluding process from auto instrumentation.

packages/collector/test/immediate/test.js
```diff
@@ -30,4 +34,59 @@ describe('collector/src/immediate', function() {
       expect(pidsThatContactedTheAgent[controls.getPid()]).to.exist;
     });
   });
+
+  describe('exclude processes from auto instrumentation', () => {
```
basti1302, line 38: This is the integration test for the change in `@instana/collector`. It starts an application under test via `npm start` and verifies that only the actual application under test is instrumented.

packages/core/src/util/excludedFromInstrumentation.js
```diff
@@ -0,0 +1,43 @@
...
+
+// If a process is started with `npm start` or `yarn start`, two separate processes will be started. The first is
+// `node /path/to/npm` (or `node /path/to/yarn`). This process will kick off another process with the actual application
+// (for example, `node /usr/src/app`). When
+// NODE_OPTIONS="--require /usr/src/app/node_modules/@instana/collector/src/immediate" is set, we would instrument both
+// processes, that is, npm/yarn as well as the actual application. Attempting to instrument the npm or yarn process has
+// no value and also creates confusing log output, so we exclude them here explicitly.
+const excludeList = [
```
willianpc, line 9: As a personal preference, I would use "or", so the array would no longer be necessary. Unless I missed something:
```javascript
const excludePattern = /^.*\/(yarn|npm)$/;
```

Also, I added `$` to restrict the pattern to the exact name `npm` and `yarn`. The way it is, something like `yarna` or `npmagic` will not pass.

Hope it makes sense!

packages/google-cloud-run/src/index.js
```diff
@@ -1,10 +1,15 @@
 'use strict';
 
+const { util: coreUtil } = require('@instana/core');
 const { environment: environmentUtil } = require('@instana/serverless');
 
-environmentUtil.validate();
+let isExcludedFromInstrumentation = coreUtil.excludedFromInstrumentation && coreUtil.excludedFromInstrumentation();
```
willianpc, line 6: Same as before, `const` should be enough :)

willianpc (approved):
LGTM.
Most comments are minimum stuff, except the regex that may need some closer look.
Nevertheless I am approving the PR.

Cheers!

packages/core/src/util/excludedFromInstrumentation.js
basti1302: Yep, that absolutely makes sense. I've changed that accordingly.

packages/google-cloud-run/src/index.js
basti1302: Absolutely. Instead of changing that one place, I added a new eslint rule for similar cases and ran eslint --fix to quickyl fix all unnecessary usages of let throughout the whole code base. There were a bunch.

PR #279, opened 2021-01-13
fix(tracing): avoid false positive warning about initializing too late

basti1302:
Users that also use the package `@contrast/agent` (for example, by starting their app via `node -r @contrast/agent ...`) would see the warning 

>  It seems you have initialized the @instana/collector package too late. Please check our documentation on that, in particular https://www.instana.com/docs/ecosystem/node-js/installation/#installing-the-nodejs-collector-package and https://www.instana.com/docs/ecosystem/node-js/installation/#common-pitfalls. Tracing might only work partially with this setup, that is, some calls will not be captured.

We can be a bit smarter about this and avoid this warning heuristically: If we see that `@contrast/agent` has been loaded, we can exclude the packages that `@contrast/agent` loads from our check.

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -23,6 +23,7 @@ exports.registerAdditionalInstrumentations = function registerAdditionalInstrume
 
 exports.preInit = function preInit() {
   var preliminaryConfig = normalizeConfig();
+  exports.util.hasThePackageBeenInitializedTooLate();
   exports.util.requireHook.init(preliminaryConfig);
   exports.tracing.preInit(preliminaryConfig);
 };
--- a/packages/core/src/util/initializedTooLateHeuristic.js
+++ b/packages/core/src/util/initializedTooLateHeuristic.js
@@ -1,7 +1,14 @@
 'use strict';
 
+let logger;
+logger = require('../logger').getLogger('util/initializedTooLateHeuristic', newLogger => {
+  logger = newLogger;
+});
+
 let firstCall = true;
-const patterns = [
+let hasBeenInitializedTooLate = false;
+
+let patterns = [
   /\/@elastic\/elasticsearch\/index.js/,
   /\/@google-cloud\/pubsub\/build\/src\/publisher\/index.js/,
   /\/@google-cloud\/pubsub\/build\/src\/subscriber.js/,
@@ -9,7 +16,6 @@ const patterns = [
   /\/@hapi\/call\/lib\//,
   /\/amqplib\/lib\//,
   /\/aws-sdk\/lib\/aws.js/,
-  /\/bluebird\/js\/release\//,
   // deliberately not including bunyan because we depend on bunyan ourselves
   /\/elasticsearch\/src\/elasticsearch.js/,
   /\/express\/index.js/,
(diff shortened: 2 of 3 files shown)


packages/core/src/index.js
```diff
@@ -23,6 +23,7 @@ exports.registerAdditionalInstrumentations = function registerAdditionalInstrume
 
 exports.preInit = function preInit() {
   var preliminaryConfig = normalizeConfig();
+  exports.util.hasThePackageBeenInitializedTooLate();
```
basti1302, line 26: By coincidence the other changes in this commit uncovered that this call was missing from `preInit`. We need to run our initialized-too-late check in `preInit` too to avoid another class of false positives.

packages/core/src/util/initializedTooLateHeuristic.js
```diff
@@ -34,22 +40,46 @@ const patterns = [
...
+        'Found @contrast/agent in the modules that have already been loaded. @instana/core will therefore exclude bluebird, request and winston from the check for modules that have been loaded before @instana/core.'
+      );
+    }
+
+    for (let i = 0; i < loadedModules.length; i++) {
       for (let j = 0; j < patterns.length; j++) {
         if (patterns[j].test(loadedModules[i])) {
           hasBeenInitializedTooLate = true;
-          break outer;
+          logger.debug(
```
basti1302, line 79: Instead of stopping the analysis when a problematic module has been found, continue with it and log every problematic module.

ThisIsMissEm:
Perhaps do it based on `process.execArgv` and then scan for "-r" 

```
const argvRequired = process.execArgv.reduce((modules, arg, index, argv) => {
  if (arg === "-r" || arg === "--require") {
    modules.push(argv[index+1])
  }
  return modules;
}, [])
```

Then that gives you a list of modules that were required by node command line arguments.

basti1302:
> Perhaps do it based on `process.execArgv` and then scan for "-r"
> 
> ```
> const argvRequired = process.execArgv.reduce((modules, arg, index, argv) => {
>   if (arg === "-r" || arg === "--require") {
>     modules.push(argv[index+1])
>   }
>   return modules;
> }, [])
> ```
> 
> Then that gives you a list of modules that were required by node command line arguments.

Hm, that's a good idea. We would probably also need to cater for the ESM case that `@contrast/agent` advertises in their README (`--loader @contrast/agent/esm.mjs`). I'm also not sure if it would catch cases where the app was started with an equivalent env var `NODE_OPTIONS=--require @contrast/agent`, but that would be easy to find out.

Of course, theoretically, users of both `@contrast/agent` and `@instana/collector` could also not use any `--require` or similar, but simply do this in their entrypoint Node.js file:

```
require('@contrast/agent');
require('@instana/collector)();

.... more code
```

Thus, ultimately, inspecting the already loaded modules sounds like it could be the most reliable approach.

WDYT?

ThisIsMissEm:
hmm, I wasn't aware of the `--loader` syntax; I'd be inclined to think anyone using `NODE_OPTIONS` is totally asking for pain though, but that's a different matter.

From `@instana/collector`'s entrypoint you can find preloaded modules with: 
```
Object.values(require.cache).filter((m) => m.loaded)
```

And if `require.main == undefined` then you know you've been required via a preload too

willianpc (approved):
LGTM

basti1302:
> hmm, I wasn't aware of the `--loader` syntax; I'd be inclined to think anyone using `NODE_OPTIONS` is totally asking for pain though, but that's a different matter.
> 
> From `@instana/collector`'s entrypoint you can find preloaded modules with:
> 
> ```
> Object.values(require.cache).filter((m) => m.loaded)
> ```
> 
> And if `require.main == undefined` then you know you've been required via a preload too

That's somewhat similar to what we do already here: https://github.com/instana/nodejs-sensor/blob/main/packages/core/src/util/initializedTooLateHeuristic.js#L46, although more elegant – I like it! I'll try to incorporate it into the PR  👍  Thanks for the suggestion.

PR #280, opened 2021-01-15
Prodreq 35124/aws sdk2.0 sqs/initial setup

willianpc:
WIP: Just to have the PR going through the pipeline

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - fix: run original function in runInAsyncContext/runPromiseInAsyncContext when tracing is not active.
+- Introduction of AWS SQS instrumentation for sending and reading messages.
 
 ## 1.113.0
 - Instrument MongoDB native driver unified topology.
--- a/packages/collector/package.json
+++ b/packages/collector/package.json
@@ -117,6 +117,10 @@
     {
       "name": "Emelia Smith",
       "email": "ThisIsMissEm@users.noreply.github.com"
+    },
+    {
+      "name": "Willian Carvalho",
+      "email": "willian.carvalho@instana.com"
     }
   ],
   "bugs": {
--- a/packages/collector/test/tracing/cloud/aws/sqs/receiveMessage.js
+++ b/packages/collector/test/tracing/cloud/aws/sqs/receiveMessage.js
@@ -1,0 +1,249 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2021
+ */
+
+'use strict';
+
+const instana = require('../../../../../')();
+const express = require('express');
+const port = process.env.APP_RECEIVER_PORT || 3216;
+const agentPort = process.env.INSTANA_AGENT_PORT || 80;
+const request = require('request-promise');
+const { sendToParent } = require('../../../../../../core/test/test_util');
+
+const app = express();
+
+const delay = require('../../../../../../core/test/test_util/delay');
+const { sqs } = require('./sqsUtil');
+const queueURL = process.env.AWS_SQS_QUEUE_URL;
+const logPrefix = `AWS SQS Message Receiver (${process.pid}):\t`;
+
+let hasStartedPolling = false;
+
+const receiveParams = {
+  AttributeNames: ['SentTimestamp'],
+  MaxNumberOfMessages: 10,
+  MessageAttributeNames: ['All'],
+  QueueUrl: queueURL,
+  VisibilityTimeout: 5,
+  // Please keep this value above 5, as tests can fail if not all messages are received
+  WaitTimeSeconds: 7
+};
+
+function log() {
+  /* eslint-disable no-console */
+  const args = Array.prototype.slice.call(arguments);
+  args[0] = `${logPrefix}${args[0]}`;
+  console.log.apply(console, args);
+}
+
+app.get('/', (_req, res) => {
+  if (hasStartedPolling) {
+    res.send('OK');
+  } else {
+    res.status(500).send('Not ready yet.');
+  }
+});
+
+function receivePromise() {
+  let span;
+  return sqs
+    .receiveMessage(receiveParams)
+    .promise()
+    .then(data => {
+      if (data && data.error) {
+        log('receive message data error', data.error);
+        return;
+      } else if (!data || !data.Messages || data.Messages.length === 0) {
+        log('no messages, doing nothing');
+        return;
+      }
+
+      span = instana.currentSpan();
+      span.disableAutoEnd();
+
+      data.Messages.forEach(message => {
+        sendToParent(message);
+      });
+
+      log('got messages:', data.Messages.map(m => m.MessageId));
+
+      const messagesForDeletion = data.Messages.map(message => {
+        return {
+          Id: message.MessageId,
+          ReceiptHandle: message.ReceiptHandle
+        };
+      });
+
+      return sqs
+        .deleteMessageBatch({
+          QueueUrl: queueURL,
+          Entries: messagesForDeletion
+        })
+        .promise()
+        .then(() => delay(200))
+        .then(() => request(`http://127.0.0.1:${agentPort}`))
+        .then(() => {
+          log('The follow up request after receiving a message has happened.');
+          span.end();
+        });
+    })
+    .catch(err => {
+      log('message receiving failed', err);
+      span && span.end(1);
+    });
+}
+
+async function receiveAsync() {
+  let span;
+  try {
+    const sqsPromise = sqs.receiveMessage(receiveParams).promise();
+    const data = await sqsPromise;
+    instana.sdk.runInAsyncContext(sqsPromise.instanaAsyncContext, async () => {
+      if (data && data.error) {
+        log('receive message data error', data.error);
+        return;
+      } else if (!data || !data.Messages || data.Messages.length === 0) {
+        log('no messages, doing nothing');
+        return;
+      }
+
+      span = instana.currentSpan();
+      span.disableAutoEnd();
+
+      data.Messages.forEach(message => {
+        sendToParent(message);
+      });
+
+      log('got messages:', data.Messages.map(m => m.MessageId));
+
+      const messagesForDeletion = data.Messages.map(message => {
+        return {
+          Id: message.MessageId,
+          ReceiptHandle: message.ReceiptHandle
+        };
+      });
+
+      await sqs
+        .deleteMessageBatch({
+          QueueUrl: queueURL,
+          Entries: messagesForDeletion
+        })
+        .promise();
+
+      await delay(200);
+      await request(`http://127.0.0.1:${agentPort}`);
+      log('The follow up request after receiving a message has happened.');
+      span.end();
+    });
+  } catch (err) {
+    span && span.end(1);
+    log('error receiving message', err);
+  }
+}
+
+function receiveCallback(cb) {
+  sqs.receiveMessage(receiveParams, (err, messagesData) => {
+    const span = instana.currentSpan();
+    span.disableAutoEnd();
+
+    if (err) {
+      log(err);
+      cb();
+      span.end(1);
+    } else if (messagesData.error) {
+      log('receive message data error', messagesData.error);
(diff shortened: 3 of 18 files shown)


packages/collector/package.json
```diff
@@ -117,6 +117,10 @@
     {
       "name": "Emelia Smith",
       "email": "ThisIsMissEm@users.noreply.github.com"
+    },
+    {
+      "name": "Willian Carvalho",
+      "email": "willian.carvalho@instana.com"
```
basti1302, line 123: 🎉 👍 👍 🎉

packages/collector/test/tracing/cloud/aws/sqs/receiveMessage.js
```diff
@@ -0,0 +1,205 @@
...
+  `Receiver server started at port ${port}`;
+});
+
+function receivePromise() {
+  let span;
+  return sqs
+    .receiveMessage(receiveParams)
+    .promise()
+    .then(data => {
+      if (!data || !data.Messages || data.Messages.length === 0) {
```
basti1302, line 57: Can `data` contain something like `data.error`? I think I saw something like that in the instrumentation. We should handle that case here. Maybe just writing the error to the console, so it doesn't go unnoticed.

packages/collector/test/tracing/cloud/aws/sqs/receiveMessage.js
```diff
@@ -0,0 +1,205 @@
...
+app.listen(port, () => {
+  `Receiver server started at port ${port}`;
+});
+
+function receivePromise() {
+  let span;
+  return sqs
+    .receiveMessage(receiveParams)
+    .promise()
+    .then(data => {
```
basti1302, line 56: This then has no catch. It probably should have one.

packages/collector/test/tracing/cloud/aws/sqs/receiveMessage.js
```diff
@@ -0,0 +1,205 @@
...
+      return sqs
+        .deleteMessage(deleteParams)
+        .promise()
+        .then(() => delay(200))
+        .then(() => request(`http://127.0.0.1:${agentPort}`))
+        .then(() => {
+          log('The follow up request after receiving a message has happened.');
+          span.end();
+        })
+        .catch(err => {
```
basti1302, line 81: I guess we could omit this catch I guess and just have the outer then have a catch. Since we return this inner promise, an error in this inner promise should bubble up to the outer promise I think.

packages/collector/test/tracing/cloud/aws/sqs/receiveMessage.js
```diff
@@ -0,0 +1,205 @@
...
+          log('The follow up request after receiving a message has failed.', err);
+          span.end(1);
+        });
+    });
+}
+
+async function receiveAsync() {
+  const sqsPromise = sqs.receiveMessage(receiveParams).promise();
+  const data = await sqsPromise;
```
basti1302, line 90: similar to the promise case, this await is not wrapped in try catch. We probably should do that. WDYT?

OTOH, an error here would be caught in pollForMessages, that might be good enough

basti1302 (changes requested):
I went through the changes for a detailed review. 

In addition to the individual comments, here are a few things that need to be done before this can be merged:
- [x] more test scenarios for the integration test:
    - [x] tracing disabled
    - [x] fix CI with queue name regex
    - [x] tracing supression
    - [x] with error in sender 
    - [x] with error in receiver
    - [ ] instrument sendMessageBatch()
- [ ] manual test with a Java SQS app and/or Go
- [ ] capture the additional attributes in the SQS span we identified
- [x] check why CI is failing (my guess is because we didn't yet provide AWS credentials)
- [ ] we should also squash all changes into a single commit IMO

packages/collector/test/tracing/cloud/aws/sqs/receiveMessage.js
willianpc: I think that a try/catch wrap is a good idea. We already have a try/catch a few lines below, which doesn't make a lot of sense to have more than one try/catch combo in the same function, IMO.

packages/collector/test/tracing/cloud/aws/sqs/sendMessage.js
```diff
@@ -0,0 +1,106 @@
...
+  } catch (err) {
+    console.log(err);
+    res.status(501).send({
+      status: 'ERROR',
+      data: String(err)
+    });
+  }
+});
+
+app.post('/send-promise2', async (req, res) => {
```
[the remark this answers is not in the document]
willianpc: In the very beginning, I wasn't sure if there could be any difference between calling `.promise().then` and doing this in two steps by defining a variable with `.promise()` then using it. 
It doesn't make any difference (as expected) and is not needed. I will remove it.

packages/collector/test/tracing/cloud/aws/sqs/sqsUtil.js
```diff
@@ -0,0 +1,14 @@
...
+ * (c) Copyright Instana Inc. and contributors 2021
+ */
+
+'use strict';
+
+const AWS = require('aws-sdk');
+
+AWS.config.update({ region: 'us-east-2' });
+
+const sqs = new AWS.SQS({ apiVersion: '2012-11-05' });
```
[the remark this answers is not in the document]
willianpc: I ran the tests without this option and it worked just fine.
This option was present in [AWS SQS official example code](https://docs.aws.amazon.com/code-samples/latest/catalog/javascript-sqs-sqs_deletequeue.js.html), so I just kept it. They don't mention in the example code if this is mandatory. So I guess we can remove it. 
Any particular reason why you are asking?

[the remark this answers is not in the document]
basti1302: Uhm, no, actually I was just curious.

I could imagine that the AWS SDK will just use a default apiVersion (maybe the most recent one) if none is provided. It is probably quite irrelevant.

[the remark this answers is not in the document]
willianpc: Ok. So I removed it and tested against the real SQS and it just works. So I will keep the constructor free of any arguments.

packages/core/src/tracing/instrumentation/cloud/aws/sqs.js
```diff
@@ -0,0 +1,384 @@
...
+   * }
+   */
+  let attributes = originalArgs[0].MessageAttributes;
+
+  if (!attributes) {
+    attributes = originalArgs[0].MessageAttributes = {};
+  }
+
+  if (cls.tracingSuppressed()) {
+    propagateSuppression(attributes);
```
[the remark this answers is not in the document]
willianpc: Got it. This was copied from GCP, where there is no span.cancel() in the suppression bit. So maybe we should take a look there as well?
https://github.com/instana/nodejs-sensor/blob/main/packages/core/src/tracing/instrumentation/cloud/gcp/pubsub.js#L85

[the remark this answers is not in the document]
basti1302: Huh. You are right. Very good catch! 

For pubsub there is already a test that makes sure we do not create a span when suppression is on. And thinking about it, I now see why it works: There is a check for the parent span a few lines later (here in your code as well as in GCP pubsub). And that check will fail when suppression is on because the entry instrumentation (for example httpServer.js) will not have created a span. 

Thinking about it, we should be more explicit about this and not rely on the behaviour of the entry instrumentation. I checked other exit instrumentation (amqp, kafkaJs, httpClient) and we are not entirely consistent but most explicitly stop the processing by doing something equivalent to `return originalSendMessage.apply(ctx, originalArgs);` when suppression is detected. 

In conclusion, we should add the line here and also for pubsub. It won't change the behaviour in practice but make the code easier to understand and more explicit.

packages/collector/test/tracing/cloud/aws/sqs/test.js
```diff
@@ -36,7 +36,7 @@ mochaSuiteFn('tracing/cloud/aws/sqs', function() {
   globalAgent.setUpCleanUpHooks();
   const agentControls = globalAgent.instance;
 
-  describe('tracing enabled, no suppression', function() {
+  describe.only('tracing enabled, no suppression', function() {
```
basti1302, line 39: I use `.only` a lot locally, but usually try to not commit it. `eslint` will fail due to it. (There is rule that forbids only so it is not accidentally committed)

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - fix: run original function in runInAsyncContext/runPromiseInAsyncContext when tracing is not active.
+- Introduction of AWS SQS instrumentation for sending and reading messages
```
basti1302, line 5: Please add full stop at the end. Yes, I am very German and nitpicky, thanks for asking :-)

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -47,6 +47,35 @@ exports.updateConfig = function updateConfig(config) {
...
+    parentSpan.n === 'sqs' &&
+    options &&
+    options.headers &&
+    options.headers['User-Agent'] &&
+    options.headers['User-Agent'].toLowerCase().indexOf('aws-sdk-nodejs') > -1 &&
+    // Same regex used by AWS SDK at /lib/services/sqs.js
+    options.headers.Host.match(/^sqs\.(?:.+?)\./)
+  ) {
+    // TODO on second though, I think we need to check the target host. We only want to suppress requests to SQS, not
+    // other requests that the AWS SDK makes.
```
basti1302, line 72: Please remove the TODO comment. It has been resolved :-)

packages/collector/test/tracing/cloud/aws/sqs/receiveMessage.js
```diff
@@ -0,0 +1,249 @@
...
+          ReceiptHandle: message.ReceiptHandle
+        };
+      });
+
+      sqs.deleteMessageBatch(
+        {
+          QueueUrl: queueURL,
+          Entries: messagesForDeletion
+        },
+        (deleteErr /* _data */) => {
```
basti1302, line 176: I think `/* data */` could be removed.

packages/collector/test/tracing/cloud/aws/sqs/test.js
```diff
@@ -0,0 +1,395 @@
...
+
+if (!supportedVersion(process.versions.node)) {
+  mochaSuiteFn = describe.skip;
+} else {
+  mochaSuiteFn = describe;
+}
+
+const retryTime = config.getTestTimeout() * 2;
+
+mochaSuiteFn.only('tracing/cloud/aws/sqs', function() {
```
basti1302, line 49: The `only` needs to be removed.

packages/core/src/tracing/instrumentation/cloud/aws/sqs.js
```diff
@@ -0,0 +1,425 @@
...
+}
+
+function shimReceiveMessage(originalReceiveMessage) {
+  return function() {
+    if (isActive) {
+      const parentSpan = cls.getCurrentSpan();
+      if (parentSpan) {
+        logger.warn(
+          // eslint-disable-next-line max-len
+          `Cannot start a AWS SQS entry span when another span is already active. Currently, the following span is active: ${JSON.stringify(
```
basti1302, line 229:
```suggestion
          `Cannot start an AWS SQS entry span when another span is already active. Currently, the following span is active: ${JSON.stringify(
```

CHANGELOG.md
basti1302, line 5:
```suggestion
- Introduction of AWS SQS instrumentation for sending and reading messages.
```

packages/collector/test/tracing/cloud/aws/sqs/test.js
basti1302, line 49:
```suggestion
mochaSuiteFn('tracing/cloud/aws/sqs', function() {
```

basti1302 (approved):
That looks good! 🎉 🎉 

Some very minor nitpicks, see inline comments.

PR #283, opened 2021-01-21
fix: improve heuristic for detecting yarn/npm

basti1302:


packages/core/src/util/excludedFromInstrumentation.js
```diff
@@ -6,11 +6,11 @@
 // NODE_OPTIONS="--require /usr/src/app/node_modules/@instana/collector/src/immediate" is set, we would instrument both
 // processes, that is, npm/yarn as well as the actual application. Attempting to instrument the npm or yarn process has
 // no value and also creates confusing log output, so we exclude them here explicitly.
-const excludePattern = /^.*\/(yarn|npm)$/i;
+const excludePattern = /^.*\/(?:npm(?:\.js)?|npm-cli(?:\.js)?|yarn(?:\.js)?|yarn\/lib\/cli(?:\.js)?)$/i;
```
willianpc, line 9: Nice. I didn't know about `?:`. Cool addition to the code.
For future reference to whomever reads this and also don't know, this is a non-capturing group, meaning that whatever is grouped is not preserved for potential further use (with `$1, $2...`).

willianpc (approved):
LGTM

PR #286, opened 2021-01-25
feat(metrics): rebuilding native add-ons via node-gyp as opt-in

basti1302:
This feature has been on by default previously and is now off by default
and needs to be enabled explicitly via
`INSTANA_REBUILD_NATIVE_ADDONS_ON_DEMAND=true`.

This also adds an option to disable copying precompiled native add-ons
(via `INSTANA_COPY_PRECOMPILED_NATIVE_ADDONS=false`).

### DIFF
--- a/packages/shared-metrics/src/util/nativeModuleRetry.js
+++ b/packages/shared-metrics/src/util/nativeModuleRetry.js
@@ -17,10 +17,16 @@ const { fork } = require('child_process');
 const detectLibc = require('detect-libc');
 
 const retryMechanisms = [];
-if (!process.env.INSTANA_DEV_DISABLE_PRECOMPILED_NATIVE_ADDONS) {
+if (
+  !process.env.INSTANA_COPY_PRECOMPILED_NATIVE_ADDONS ||
+  process.env.INSTANA_COPY_PRECOMPILED_NATIVE_ADDONS.toLowerCase() !== 'false'
+) {
   retryMechanisms.push('copy-precompiled');
 }
-if (!process.env.INSTANA_DEV_DISABLE_REBUILD_NATIVE_ADDONS) {
+if (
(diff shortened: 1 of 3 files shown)


packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -17,10 +17,16 @@ const { fork } = require('child_process');
...
-if (!process.env.INSTANA_DEV_DISABLE_PRECOMPILED_NATIVE_ADDONS) {
+if (
+  !process.env.INSTANA_COPY_PRECOMPILED_NATIVE_ADDONS ||
+  process.env.INSTANA_COPY_PRECOMPILED_NATIVE_ADDONS.toLowerCase() !== 'false'
+) {
   retryMechanisms.push('copy-precompiled');
 }
-if (!process.env.INSTANA_DEV_DISABLE_REBUILD_NATIVE_ADDONS) {
+if (
+  process.env.INSTANA_REBUILD_NATIVE_ADDONS_ON_DEMAND &&
```
willianpc, line 27: When the env var is not set (which would be all cases as this option is new), would it make sense to add a warning for customers that the rebuild is now off by default? Or it doesn't matter much for them?
I am not sure if this is a big deal for them, which is why I am asking.
Otherwise, the documentation and CHANGELOG should be enough, I believe.

willianpc (approved):
LGTM.
Please, check my question about a potential warning.

packages/shared-metrics/src/util/nativeModuleRetry.js
basti1302: That feature being on by default caused some confusion because people were unexpectedly seeing verbose output from node-gyp in their Node.js processes. The feature hasn't been available for long and I'm pretty sure no one actively relies on it. In the worst case, some gc and event loop metrics would be missing. I think printing a warning might be too much for this case.

PR #287, opened 2021-01-29
fix(sdk): run original fn in SDK wrappers when tracing is not active

basti1302:
This applies to runInAsyncContext, runPromiseInAsyncContext.


basti1302:
@willianpc I know you have the same changes in the production code in the other branch, but I wanted to create the test for this right away 🤷

willianpc (approved):
LGTM

PR #288, opened 2021-02-15
AWS S3 instrumentation

willianpc:


packages/collector/test/tracing/cloud/aws/s3/test.js
```diff
@@ -0,0 +1,200 @@
...
+    requestMethods.forEach(requestMethod => {
+      describe(`attempt to get result with: ${requestMethod}`, () => {
+        availableOperations.forEach(operation => {
+          it(`should not trace (${operation})`, async () => {
+            await appControls.sendRequest({
+              method: 'GET',
+              path: `/${operation}/${requestMethod}`
+            });
+            return retry(() => agentControls.getSpans(), retryTime)
+              .then(() => agentControls.getSpans())
```
basti1302, line 152: We need a delay here. Checking for the span immediately after sending the request does not prove we did not create any spans. The idea is:

* send the request
* wait for some time (say 500 ms) with `delay`
* then check for spans

If there are no spans after 500 ms, we can be reasonably sure that we did not create any spans.

Also, there are two agentControls.getSpans() calls here. The first one should be the delay instead.

packages/collector/test/tracing/cloud/aws/s3/test.js
```diff
@@ -0,0 +1,200 @@
...
+            await appControls.sendRequest({
+              method: 'GET',
+              path: `/${operation}/${requestMethod}`,
+              headers: {
+                'X-INSTANA-L': '0'
+              }
+            });
+
+            return retry(() => agentControls.getSpans(), retryTime)
+              .then(() => agentControls.getSpans())
```
basti1302, line 189: Same with the duplicated getSpans and the missing delay here, see above.

packages/collector/test/tracing/cloud/aws/s3/test.js
```diff
@@ -0,0 +1,200 @@
...
+
+    requestMethods.forEach(requestMethod => {
+      describe(`attempt to get result with: ${requestMethod}`, () => {
+        availableOperations.forEach(operation => {
+          it(`should not trace (${operation})`, async () => {
+            await appControls.sendRequest({
+              method: 'GET',
+              path: `/${operation}/${requestMethod}`,
+              headers: {
+                'X-INSTANA-L': '0'
```
basti1302, line 184: Minor nitpick: The `sendRequest` method actually supports an attribute `suppressTracing` which should be used instead of "manually" adding `X-INSTANA-L: 0` as a header. I have probably not been consistent with that at all, so I can see where this is coming from.

packages/core/src/tracing/instrumentation/cloud/aws/s3.js
```diff
@@ -0,0 +1,215 @@
...
+    const span = cls.startSpan('s3', EXIT);
+    span.ts = Date.now();
+    span.stack = tracingUtil.getStackTrace(instrumentedMakeRequest);
+    span.data.s3 = buildS3Data(originalArgs[0], originalArgs[1]);
+
+    const _originalCallback = originalArgs[2];
+
+    originalArgs[2] = cls.ns.bind(function(err, data) {
+      finishSpan(err, data, span);
+      setImmediate(() => finishSpan(err, data, span));
```
basti1302, line 118: Pretty sure we don't need this with setImmediate here. The synchronous finishSpan in line 117 should be enough. This was only relevant for the SQS entry.

packages/core/src/tracing/instrumentation/cloud/aws/s3.js
```diff
@@ -0,0 +1,215 @@
...
+    originalPromise = cls.ns.bind(function patchedPromise() {
+      span.data.s3 = buildS3Data(this.operation, this.params);
+
+      return new Promise((resolve, reject) => {
+        _promise
+          .apply(this, arguments)
+          .then(data => {
+            resolve(data);
+            finishSpan(null, data, span);
+            setImmediate(() => finishSpan(null, data, span));
```
basti1302, line 164: remove this line

packages/core/src/tracing/instrumentation/cloud/aws/s3.js
```diff
@@ -0,0 +1,215 @@
...
+          })
+          .catch(err => {
+            reject(err);
+            finishSpan(err, null, span);
+            setImmediate(() => finishSpan(err, null, span));
```
basti1302, line 169: remove this line

packages/core/src/tracing/instrumentation/cloud/aws/s3.js
```diff
@@ -0,0 +1,215 @@
...
+
+  span.d = Date.now() - span.ts;
+  span.transmit();
+}
+
+function addErrorToSpan(err, span) {
+  if (err) {
+    span.ec = 1;
+    if (err.code) {
+      span.data.s3.error = err.code;
```
basti1302, line 191: Do other S3 instrumentations capture the code or rather an error message?

packages/core/src/tracing/instrumentation/cloud/aws/s3.js
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  return cls.ns.runAndReturn(function() {
+    const span = cls.startSpan('s3', EXIT);
+    span.ts = Date.now();
+    span.stack = tracingUtil.getStackTrace(instrumentedRequestPromise);
+    const _promise = originalPromise;
+
+    originalPromise = cls.ns.bind(function patchedPromise() {
+      span.data.s3 = buildS3Data(this.operation, this.params);
```
basti1302, line 156: I would have expected `ctx.operation` here, not `this.operation`. Same with `this.params`. To what does `this` refer here?

packages/core/src/tracing/instrumentation/cloud/aws/s3.js
```diff
@@ -0,0 +1,215 @@
...
+
+      return new Promise((resolve, reject) => {
+        _promise
+          .apply(this, arguments)
```
basti1302, line 160: Another `this` that confuses me a bit.

basti1302 (changes requested):
This looks good! Some remarks and minor change requests inline.

packages/core/src/tracing/instrumentation/cloud/aws/s3.js
willianpc: Not sure if I got the question right. the way it is right now, which seems to be the standard for all AWS lib, err.code gives a string. eg: UnexpectedParameter. I can maybe send a mix of both `code` and `message`, or maybe even better: `err.message || err.code`. Did I get your point right? @basti1302

packages/collector/test/tracing/cloud/aws/s3/test.js
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    function verifyHttpExit(spans, parentSpan, controls, withError) {
+      return expectExactlyOneMatching(spans, [
+        span => expect(span.t).to.equal(parentSpan.t),
+        span => expect(span.p).to.equal(parentSpan.s),
+        span => expect(span.k).to.equal(constants.EXIT),
+        span => expect(span.f.e).to.equal(String(controls.getPid())),
+        span => expect(span.f.h).to.equal('agent-stub-uuid'),
+        span => expect(span.n).to.equal(withError ? 's3' : 'node.http.client')
```
basti1302, line 118: This is weird and a bit confusing. Why is this checking for an s3 span in the error test cases?

basti1302 (approved):
LGTM!

PR #291, opened 2021-02-22
fix(lambda): fix timeouts

basti1302:
Lambda functions would previously time out when
(a) the Lambda callback API was used,
(b) `context.callbackWaitsForEmptyEventLoop = false` was not set, and
(c) the Instana back end responded too slowly.

The reason was that the HTTP request to the back end was still pending
and thus the event loop was not empty. Destroying the request manually
fixes this.


willianpc (approved):
LGTM

PR #292, opened 2021-02-22
fixed false positive test

willianpc:
I changed the way that the tests should fail when testing calls with error by adding a non existent key into the S3 parameters.
Also removed `.only` that should not have been pushed upstream.


basti1302 (approved):
LGTM

PR #293, opened 2021-02-24
S3 CI error fix

willianpc:
The error was not actually in the tests, but the test Express application that was forwarding S3 errors to client calls.

### DIFF
--- a/packages/collector/test/tracing/cloud/aws/s3/app.js
+++ b/packages/collector/test/tracing/cloud/aws/s3/app.js
@@ -109,7 +109,11 @@ const S3Api = {
             span.disableAutoEnd();
 
             if (err) {
-              log(`failed on /${operation}/${method} when receiving response from AWS API`);
+              log(
+                `${
+                  withError ? 'successfully failed' : 'failed'
+                } on /${operation}/${method} when receiving response from AWS API`
+              );
               span.end(1);
               return reject(err);
             } else {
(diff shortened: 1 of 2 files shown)


packages/collector/test/tracing/cloud/aws/s3/app.js
```diff
@@ -132,6 +135,7 @@ const S3Api = {
 
           promise
             .then(data => {
+              log(` /${operation}/${method} - received data from AWS SDK`);
```
basti1302, line 138: There's an extra space at the start of string. Accidental or deliberate?

packages/collector/test/tracing/cloud/aws/s3/app.js
```diff
@@ -188,7 +195,7 @@ availableOperations.forEach(operation => {
         const data = await S3Api.runOperation(operation, method, withError);
         res.send(data);
       } catch (err) {
-        res.status(500).send({
+        res.status(200).send({
```
basti1302, line 198: I wonder why we need to mask an error like this since the test already uses `appControls.sendRequest` with `simple: withError === false`. Shouldn't that take care of the HTTP 500 on the client side?

basti1302 (approved):

packages/collector/test/tracing/cloud/aws/s3/app.js
willianpc: Accidental. thanks!

packages/collector/test/tracing/cloud/aws/s3/app.js
willianpc: As per our talk on Slack, you are right. The real problem was the lack of handling for undefined object when attempting to log `options.Bucket`.

PR #294, opened 2021-02-24
AWS DynamoDB

willianpc:


packages/collector/test/tracing/cloud/aws/dynamodb/app.js
```diff
@@ -0,0 +1,250 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2021
+ */
+
+'use strict';
+
+const instana = require('../../../../../')();
+const agentPort = process.env.INSTANA_AGENT_PORT || 80;
```
basti1302, line 9: 80 makes little sense I think? The default agent port is 42699.

packages/collector/test/tracing/cloud/aws/dynamodb/app.js
```diff
@@ -0,0 +1,250 @@
...
+    return new Promise(async (resolve, reject) => {
+      let span;
+      let promise;
+      let promiseData;
+
+      switch (method) {
+        case methods.CALLBACK:
+          dynamoDB[operation](options, (err, data) => {
+            span = instana.currentSpan();
+            span.disableAutoEnd();
```
basti1302, line 128: This should only be necessary for messaging instrumentations, and only when _receiving_ a message. I assume all of the operations we instrument here are DB exits, right?

packages/collector/test/tracing/cloud/aws/dynamodb/app.js
```diff
@@ -0,0 +1,250 @@
...
+    TableName: tableName,
+    Item: itemKey
+  },
+  updateItem: {
+    TableName: tableName,
+    Key: itemKey,
+    AttributeUpdates: {
+      author: {
+        Value: {
+          S: 'Neil Gaiman'
```
basti1302, line 87: Great literary taste! 👍

packages/collector/test/tracing/cloud/aws/dynamodb/app.js
```diff
@@ -0,0 +1,250 @@
...
+      let promiseData;
+
+      switch (method) {
+        case methods.CALLBACK:
+          dynamoDB[operation](options, (err, data) => {
+            span = instana.currentSpan();
+            span.disableAutoEnd();
+
+            if (err) {
+              span.end(1);
```
basti1302, line 131: Related to the questionable disableAutoEnd mentioned above, I think all the manual span.end calls should not be necessary at all.

packages/collector/test/tracing/cloud/aws/dynamodb/app.js
```diff
@@ -0,0 +1,250 @@
...
+            }
+          });
+          break;
+        case methods.PROMISE:
+          promise = dynamoDB[operation](options).promise();
+
+          promise
+            .then(data => {
+              span = instana.currentSpan();
+              span.disableAutoEnd();
```
basti1302, line 157: same here

packages/collector/test/tracing/cloud/aws/dynamodb/app.js
```diff
@@ -0,0 +1,250 @@
...
+              }
+              span.end(1);
+              reject(err);
+            });
+          break;
+        case methods.ASYNC:
+          try {
+            const data = await dynamoDB[operation](options).promise();
+            span = instana.currentSpan();
+            span.disableAutoEnd();
```
basti1302, line 185: same here

packages/collector/test/tracing/cloud/aws/dynamodb/test.js
```diff
@@ -0,0 +1,226 @@
...
+        span => expect(span.t).to.equal(parent.t),
+        span => expect(span.p).to.equal(parent.s),
+        span => expect(span.f.e).to.equal(String(appControls.getPid())),
+        span => expect(span.f.h).to.equal('agent-stub-uuid'),
+        span => expect(span.error).to.not.exist,
+        span => expect(span.ec).to.equal(withError ? 1 : 0),
+        span => expect(span.async).to.not.exist,
+        span => expect(span.data).to.exist,
+        span => expect(span.data.dynamodb).to.be.an('object'),
+        span => expect(span.data.dynamodb.op).to.exist
```
basti1302, line 151: Can we pass the expected op into this method to verify the actual value and not only that the annotation _exists_?

Also, are there more annotations that we could/should be capturing and that we should test for? (`table` for example, I think I have seen that below in the instrumentation code).

packages/collector/test/tracing/cloud/aws/dynamodb/util.js
```diff
@@ -0,0 +1,39 @@
...
+const interval = 1000;
+const MAX_WAIT_TIME = 10000;
+
+exports.checkTableExistence = function checkTableExistence(tableName, expectsToExist = true) {
+  let timer = 0;
+  return new Promise(resolve => {
+    const intervalId = setInterval(() => {
+      timer += interval;
+      if (timer >= MAX_WAIT_TIME) {
+        resolve(`Timeout after ${timer} ms`);
```
basti1302, line 20: Shouldn't we rather `reject` here instead of `resolve`?

packages/collector/test/tracing/cloud/aws/s3/app.js
```diff
@@ -138,11 +138,12 @@ const S3Api = {
           break;
         case methods.PROMISE:
           promise = s3[operation](options).promise();
-          span = instana.currentSpan();
-          span.disableAutoEnd();
 
           promise
             .then(data => {
+              span = instana.currentSpan();
+              span.disableAutoEnd();
```
basti1302, line 145: Seems I missed those when reviewing S3 earlier, but the same applies: disableAutoEnd and manually ending spans should not be necessary anywhere in the S3 or DynamoDB tests.

packages/core/src/tracing/instrumentation/cloud/aws/dynamodb.js
```diff
@@ -0,0 +1,179 @@
...
+
+const shimmer = require('shimmer');
+const requireHook = require('../../../../util/requireHook');
+const cls = require('../../../cls');
+const { EXIT, isExitSpan } = require('../../../constants');
+const tracingUtil = require('../../../tracingUtil');
+
+const operationsInfo = {
+  createTable: { op: 'create' },
+  // deleteTable: { op: 'delete' },
```
basti1302, line 16: What is the reason for leaving that one out?

packages/core/src/tracing/instrumentation/cloud/aws/dynamodb.js
```diff
@@ -0,0 +1,179 @@
...
+const methodList = Object.keys(operationsInfo);
+
+let isActive = false;
+
+exports.isActive = function() {
+  return isActive;
+};
+
+exports.init = function init() {
+  requireHook.onModuleLoad('aws-sdk', instrumentAWS);
```
basti1302, line 35: We are doing that in multiple places now, that is we add multiple hooks for `'aws-sdk'`. I haven't formed an opinion on that yet but I wonder if we should only have _one_ `requireHook.onModuleLoad('aws-sdk', ...);`  call in the entire code base and branch out to the DynamoDB and S3 instrumentation from there.

packages/core/src/tracing/instrumentation/cloud/aws/dynamodb.js
```diff
@@ -0,0 +1,179 @@
...
+    return originalMakeRequest.apply(this, arguments);
+  };
+}
+
+function instrumentedMakeRequest(ctx, originalMakeRequest, originalArgs) {
+  if (!isOperationShimmable(originalArgs[0])) {
+    return originalMakeRequest.apply(ctx, originalArgs);
+  }
+
+  if (cls.tracingSuppressed()) {
```
basti1302, line 70: nit: this could be merged with the `if` above. 

```
if (!isOperationShimmable(originalArgs[0]) || cls.tracingSuppressed()) {
    return originalMakeRequest.apply(ctx, originalArgs);
  }
```

But that's mostly a matter of taste I guess.

packages/core/src/tracing/instrumentation/cloud/aws/dynamodb.js
```diff
@@ -0,0 +1,179 @@
...
+    // Data attribs: op and table
+    span.data.dynamodb = buildSpanData(originalArgs[0], originalArgs[1]);
+
+    if (typeof originalArgs[2] === 'function') {
+      // callback case
+      const _originalCallback = originalArgs[2];
+
+      originalArgs[2] = cls.ns.bind(function(err, data) {
+        /**
+         * DynamoDB can send error data as a sccessful HTTP response.
```
basti1302, line 93:
```suggestion
         * DynamoDB can send error data as a successful HTTP response.
```

basti1302, line 93: Also "can send error data as a successful HTTP" makes me shudder. Hey AWS, why, oh why??

packages/core/src/tracing/instrumentation/cloud/aws/dynamodb.js
```diff
@@ -0,0 +1,179 @@
...
+            return err;
+          });
+      });
+    }
+
+    return request;
+  });
+}
+
+function finishSpan(err, _data, span) {
```
basti1302, line 145: Looks like _data is not used here?

basti1302 (changes requested):
This already looks pretty good! 👍

packages/core/src/tracing/instrumentation/cloud/aws/dynamodb.js
willianpc: I thought about the same, but decided to move on this way for now, and to discuss into detail with you after.

packages/collector/test/tracing/cloud/aws/dynamodb/util.js
willianpc: > Shouldn't we rather `reject` here instead of `resolve`?

My thinking was that even though a timeout occurred, I still would like to continue with the tests, rather than halting everything.
I can change it back if you think it makes more sense to just halt everything, assuming that the table was not created/deleted. I don't have a strong opinion here.

packages/core/src/tracing/instrumentation/cloud/aws/dynamodb.js
willianpc: > What is the reason for leaving that one out?

`deleteTable` is not instrumented in Java or Go instrumentations. There is a discussion that started on Slack about this topic. We will bring it to the Dev Tracer meeting ASAP.

packages/collector/test/tracing/cloud/aws/dynamodb/util.js
basti1302: Wouldn't there be follow up failures anyway if the table was not actually created? But in general, I do not have a strong opinion here.

packages/core/src/tracing/instrumentation/cloud/aws/dynamodb.js
basti1302: Ah, so for consistency. Gotcha, that makes sense.

packages/core/src/tracing/instrumentation/cloud/aws/dynamodb.js
basti1302: We decided to postpone this to later. The current approach is fine as it is.

packages/collector/test/tracing/cloud/aws/dynamodb/test.js
```diff
@@ -0,0 +1,237 @@
...
+        span => expect(span.t).to.equal(parent.t),
+        span => expect(span.p).to.equal(parent.s),
+        span => expect(span.f.e).to.equal(String(appControls.getPid())),
+        span => expect(span.f.h).to.equal('agent-stub-uuid'),
+        span => expect(span.error).to.not.exist,
+        span => expect(span.ec).to.equal(withError ? 1 : 0),
+        span => expect(span.async).to.not.exist,
+        span => expect(span.data).to.exist,
+        span => expect(span.data.dynamodb).to.be.an('object'),
+        span => expect(span.data.dynamodb.op).to.equal(operationsInfo[operation])
```
basti1302, line 162: I still would love to see an expectation for `span.data.dynamodb.table` here.

basti1302 (approved):
Looks good!

PR #295, opened 2021-03-04
refactor(serverless): reduce overhead for Lambda tracing

basti1302:
The execution time we add to do native tracing and in-process metric
collection for AWS Lambda functions is already low but we can optimize
this further by not waiting for a response from the back end.

Instead, we only wait until the request has been sent. This shaves
around 30% off the execution time that native tracing adds to a Lambda.

### DIFF
--- a/packages/serverless/src/backend_connector.js
+++ b/packages/serverless/src/backend_connector.js
@@ -30,7 +30,7 @@ let requestHasFailed = false;
 let warningsHaveBeenLogged = false;
 
 const needsEcdhCurveFix = semver.gte(process.version, '8.6.0') && semver.lt(process.version, '10.0.0');
-const legacyTimeoutHandling = semver.lt(process.version, '10.0.0');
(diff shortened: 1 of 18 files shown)


packages/serverless/src/backend_connector.js
```diff
@@ -155,7 +155,18 @@ function send(resourcePath, payload, destroySocketAfterwards, callback) {
     },
     rejectUnauthorized: !disableCaCheck
   };
+
   if (!legacyTimeoutHandling) {
+    // The timeout specified here in the request options will kick in if *connecting* to the socket takes to long.
```
willianpc, line 160: small typo: `too long`

willianpc (approved):
LGTM.
Thanks for adding comments explaining the changes. It helped me a lot to understand what is going on.

packages/serverless/src/backend_connector.js
basti1302: fixed

PR #298, opened 2021-03-09
Lambda Extension Support 

hmadison:
There is still a bit of clean-up and polish needed for this to be merged, but I would like an extra set of eyes on my approach given my lack of familiarity with the structure of this code-base.


packages/aws-lambda/.gitignore
```diff
@@ -3,3 +3,4 @@ instana-aws-lambda.tgz
 instana-aws-lambda-auto-wrap.tgz
 instana-serverless.tgz
 npm-debug*
+include/
```
basti1302, line 6: I think we should actually commit the extension binary to the repo instead of ignoring it, see below.

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -58,19 +60,23 @@ cat <<EOF >> package.json
...
 npm install $PACKAGE_NAMES
 VERSION=$(jq -r .version node_modules/@instana/aws-lambda/package.json)
 echo "downloaded version $VERSION"
 rm -f package.json package-lock.json
 cd ..
 
-echo "step 3/5: creating local zip file with layer contents"
+echo "step 4/7: Add extension to layer"
+mkdir -p extensions
+cp ../include/instana-lambda-extension ./extensions
```
basti1302, line 72: I think we should rather copy the extension binary into this repository and commit it instead of using something that is in .gitignore. Updating the extension would then also be a commit that is visible in the history. When including this from a file that only exists locally we do not get reproducible builds. Also, the Lambda layer is built on the CI server so you would need to get the extension binary there somehow anyway.

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -100,7 +106,6 @@ while read -r region; do
 
 done <<< "$REGIONS"
 
-echo "step 5/5: cleaning up"
+echo "step 7/7: cleaning up"
 rm -rf $TMP_ZIP_DIR
-rm -rf $ZIP_NAME
-
```
basti1302, line 106: Just as a reminder for ourselves, this needs to be reverted at some point, shortly before merging this.

I guess it was for troubleshooting/inspecting the resulting zip.

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -27,24 +27,26 @@ EOF
 }
 
 PACKAGE_NAMES="@instana/aws-lambda instana-aws-lambda-auto-wrap"
-LAYER_NAME=instana-nodejs
+LAYER_NAME=hmadison-instana-nodejs
```
basti1302, line 30: Just as a reminder for ourselves, this needs to be reverted at some point, shortly before merging this.

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -27,24 +27,26 @@ EOF
...
 LICENSE=MIT
-ZIP_PREFIX=instana-nodejs-layer
+ZIP_PREFIX=hmadison-instana-nodejs-layer
```
basti1302, line 32: Just as a reminder for ourselves, this needs to be reverted at some point, shortly before merging this.

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -27,24 +27,26 @@ EOF
...
-echo "step 1/5: fetching AWS regions (skipping, using fixed list of regions for now)"
+echo "step 1/7: fetching AWS regions (skipping, using fixed list of regions for now)"
 
 # Actually, this should give us the regions where the Lambda service is provided:
 # REGIONS=$(aws ssm get-parameters-by-path --path /aws/service/global-infrastructure/services/lambda/regions --output text --query "Parameters[].Value" | tr '\t' '\n')
 # But for some reason, publishing to all of these regions does not work. In particular, the
 # following regions either require special authorization/subscription status or don't support Lambdas: ap-east-1 me-south-1 ap-northeast-3
 
-REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
+REGIONS=$'us-east-2'
```
basti1302, line 43: Just as a reminder for ourselves, this needs to be reverted at some point, shortly before merging this.

packages/serverless/src/backend_connector.js
```diff
@@ -156,9 +167,9 @@ function send(resourcePath, payload, destroySocketAfterwards, callback) {
     rejectUnauthorized: !disableCaCheck
   };
   if (!legacyTimeoutHandling) {
-    options.timeout = backendTimeout;
+    options.timeout = layerExtensionHasFailed ? backendTimeout : layerExtensionTimeout;
```
basti1302, line 170: nit: This occurs multiple times, maybe we can extract it into a function?

packages/serverless/src/backend_connector.js
```diff
@@ -156,9 +167,9 @@ function send(resourcePath, payload, destroySocketAfterwards, callback) {
...
   }
-  if (proxyAgent) {
+  if (proxyAgent && layerExtensionHasFailed) {
```
basti1302, line 172: Does the extension already respect the proxy settings?

basti1302 (commented):
That already looks pretty good! 🎉 

Some minor comments below.

packages/serverless/src/backend_connector.js
hmadison: Yes!

packages/aws-lambda/layer/bin/publish-layer.sh
hmadison: I updated the script to change the name/region list based off of the `BUILD_NUMBER` environment variable Jenkins sets.

basti1302 (approved):

PR #300, opened 2021-03-12
fix: update to shimmer@1.2.1 to be able to patch async functions

basti1302:


willianpc (approved):
I missed that the version was updated on NPMjs.
I only checked the tags on Github, where version 1.2.1 is not present.
Thanks for doing this!

basti1302:
> I missed that the version was updated on NPMjs.
> I only checked the tags on Github, where version 1.2.1 is not present.
> Thanks for doing this!

Yep, weird, this confused me at well, otherwise I would have updated this dep ages ago.

PR #301, opened 2021-03-17
Instrumentation of Bull Framework

willianpc:


packages/collector/src/tracing/instrumentation/process/childProcess.js
```diff
@@ -99,15 +99,17 @@ function shimFork(original) {
           message.job &&
           typeof message.job === 'object' &&
           message.job.opts &&
-          typeof message.job.opts === 'object' &&
           entrySpan &&
           entrySpan.k === 1 &&
           processIdentityProvider
         ) {
+          if (typeof message.job.opts !== 'object') {
```
basti1302, line 106: @willianpc `typeof null` is `object`. So if for some reason we get `message.job.opts = null` (unlikely, I know) this will result in an NPE a few lines later. I'd rather write this as 
```
if (message.job.opts == null) {
 message.job.opts = {};
}
```

willianpc: good catch. will do

packages/collector/src/tracing/instrumentation/process/childProcess.js
```diff
@@ -57,6 +60,76 @@ function shimSpawn(original) {
...
+  return function() {
+    // args: modulePath, args, options
+    const _args = Array.prototype.slice.call(arguments);
+    const modulePath = _args[0];
+    const args = _args[1];
+
+    if (typeof modulePath === 'string' && bullMasterProcessMatch.test(modulePath) && args && args.execArgv) {
+      if (!selfPath.immediate) {
+        logger.warn(
+          "Detected a child_process.spawn of 'Bull processor', but the path to @instana/collector is not available, " +
```
basti1302, line 75: nitpick:

```suggestion
          "Detected a child_process.fork of 'Bull processor', but the path to @instana/collector is not available, " +
```

packages/collector/src/tracing/instrumentation/process/childProcess.js
```diff
@@ -57,6 +60,76 @@ function shimSpawn(original) {
...
+            'so this Bull processor instance will not be instrumented.'
+        );
+      } else {
+        logger.debug(
+          `Detected a child_process.spawn of Bull, instrumenting it by adding --require ${selfPath.immediate}.`
```
basti1302, line 80: nitpick:

```suggestion
          `Detected a child_process.fork of Bull, instrumenting it by adding --require ${selfPath.immediate}.`
```

packages/collector/src/tracing/instrumentation/process/childProcess.js
```diff
@@ -57,6 +60,76 @@ function shimSpawn(original) {
...
+        if (
+          //
+          message && //
+          message.cmd === 'start' &&
+          message.job &&
+          typeof message.job === 'object' &&
+          message.job.opts &&
+          entrySpan &&
+          entrySpan.k === 1 &&
+          processIdentityProvider
```
basti1302, line 104: This check makes no sense, because we access `processIdentityProvider` already way before we get here (line 83). I probably added it and we refactored this part later.

packages/collector/test/tracing/messaging/bull/sender.js
```diff
@@ -0,0 +1,94 @@
...
+'use strict';
+
+require('../../../../')();
+const logPrefix = `Bull (${process.pid}):\t`;
+const Queue = require('bull');
+const redisServer = process.env.REDIS_SERVER || 'redis://127.0.0.1:6379';
+const queueName = process.env.BULL_QUEUE_NAME || 'nodejs-team';
+const express = require('express');
+const port = process.env.APP_SENDER_PORT || 3215;
+const bullJobName = process.env.BULL_JOB_NAME || 'steve';
```
basti1302, line 15: I always meant to ask: Who is Steve? :-)

packages/collector/test/tracing/messaging/bull/test.js
```diff
@@ -0,0 +1,425 @@
...
+              method: 'POST',
+              path: urlWithParams
+            });
+
+            return retry(() => verifyResponseAndJobProcessing({ response, testId, isRepeatable, isBulk }), retryTime)
+              .then(() => delay(config.getTestTimeout() / 4))
+              .then(() => agentControls.getSpans())
+              .then(spans => {
+                if (spans.length > 0) {
+                  fail(`Unexpected spans (AWS SQS suppressed: ${stringifyItems(spans)}`);
```
basti1302, line 303: left over copy/paste error (AWS SQS)

packages/core/src/tracing/instrumentation/process/process.js
```diff
@@ -0,0 +1,104 @@
...
+    const traceContext = ipcMessage.job.opts.instanaTracingContext;
+    delete ipcMessage.job.opts.instanaTracingContext;
+
+    if (
+      !traceContext ||
+      !traceContext.X_INSTANA_T ||
+      !traceContext.X_INSTANA_S ||
+      ipcMessage.job.opts.X_INSTANA_L === '0'
+    ) {
+      return originalProcessEmit.apply(this, arguments);
```
basti1302, line 56: Wouldn't we need to also do `delete ipcMessage.job.opts.X_INSTANA_L` here before calling `originalProcessEmit.apply`? OTOH, there are explicit tests for that I believe, so maybe this is handled somewhere else?

packages/core/src/util/normalizeConfig.js
```diff
@@ -67,6 +68,7 @@ const defaults = {
   tracing: {
     enabled: true,
     automaticTracingEnabled: true,
+    activateImmediately: false,
```
basti1302, line 71: We probably should consider adding this config option (and the corresponding env var) to our docs.

basti1302 (approved):
Excellent work! All looks good to me, except for some rather trivial nitpicks.

packages/collector/test/tracing/messaging/bull/sender.js
willianpc: well, it's a Job called Steve... got it?

packages/core/src/tracing/instrumentation/process/process.js
willianpc: That's a good catch. At the end, this check was not necessary, because the suppression is always handled in `bull.js#instrumentedProcessJob`.
In that function, `X_INSTANA_L` is already removed, and `X_INSTANA_S` and `X_INSTANA_T` are not even added to the `opts` object, which make the check of their existence enough already.

packages/collector/test/tracing/messaging/bull/sender.js
basti1302: Ouch! 😂   Good one.

PR #302, opened 2021-03-22
fix(opentracing): default to type entry when no parent is referenced

basti1302:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+- Update to shimmer@1.2.1.
```
basti1302, line 4: This is unrelated to this PR but I forgot to add it in a previous PR :-/

packages/core/src/tracing/opentracing/Span.js
```diff
@@ -53,7 +53,7 @@ function Span(tracer, name, fields) {
     data: {
       service: serviceName,
       sdk: {
-        type: 'local',
+        type: 'entry',
```
willianpc, line 56: Would it work if we did `type: parentContext ? 'local' : 'entry'`?
It's sort of personal preference, but on the other hand, seems easier to read, as one doesn't need to read the code further to realize that the `type` may change based on whether `parentContext` exists.
But again, this is mostly personal preference.

willianpc (approved):
LGTM.
I added a small comment with a personal preference, so feel free to do it or not.
Cheers!

packages/core/src/tracing/opentracing/Span.js
basti1302: Absolutely, good call.

PR #303, opened 2021-03-23
fix(metrics): register gc stats loader listener immediately

basti1302:
This fixes an issue where GC metrics would not be available when the
dependency gcstats.js was already present in node_modules and the
activation of the gc metrics module would happen too late.

### DIFF
--- a/packages/collector/test/metrics/app/test.js
+++ b/packages/collector/test/metrics/app/test.js
@@ -65,6 +65,10 @@ describe('snapshot data and metrics', function() {
         expect(findMetric(allMetrics, ['gc', 'minorGcs'])).to.exist;
         expect(findMetric(allMetrics, ['gc', 'majorGcs'])).to.exist;
 
+        const gc = aggregated.gc;
+        expect(gc).to.be.an('object');
+        expect(gc.statsSupported).to.be.true;
+
         expect(findMetric(allMetrics, ['healthchecks'])).to.exist;
         expect(findMetric(allMetrics, ['heapSpaces'])).to.exist;
         expect(findMetric(allMetrics, ['http'])).to.exist;
--- a/packages/shared-metrics/src/gc.js
+++ b/packages/shared-metrics/src/gc.js
@@ -16,6 +16,10 @@ const incrementalMarkingsWindow = slidingWindow.create(windowOpts);
 const processWeakCallbacksWindow = slidingWindow.create(windowOpts);
 const gcPauseWindow = slidingWindow.create(windowOpts);
 
+let gcStats;
+let activateHasBeenCalled = false;
+let hasBeenActivated = false;
+
 exports.payloadPrefix = 'gc';
 exports.currentPayload = {
   minorGcs: 0,
@@ -23,6 +27,13 @@ exports.currentPayload = {
   incrementalMarkings: 0,
   weakCallbackProcessing: 0,
   gcPause: 0
+};
+
+exports.activate = function activate() {
+  activateHasBeenCalled = true;
+  if (gcStats) {
+    actuallyActivate();
+  }
 };
 
 const nativeModuleLoader = require('./util/nativeModuleRetry')({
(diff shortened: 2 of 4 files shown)


packages/collector/test/metrics/app/test.js
```diff
@@ -65,6 +65,10 @@ describe('snapshot data and metrics', function() {
         expect(findMetric(allMetrics, ['gc', 'minorGcs'])).to.exist;
         expect(findMetric(allMetrics, ['gc', 'majorGcs'])).to.exist;
 
+        const gc = aggregated.gc;
+        expect(gc).to.be.an('object');
+        expect(gc.statsSupported).to.be.true;
```
basti1302, line 70: This test tests the case where the native add-on (gcstats.js in this case) is available immediately. Unfortunately, this test never checked for the supported flag, so the tests missed the issue that this PR fixes.

packages/shared-metrics/src/gc.js
```diff
@@ -37,37 +47,42 @@ const nativeModuleLoader = require('./util/nativeModuleRetry')({
 
 let senseIntervalHandle;
 
-exports.activate = function activate() {
-  nativeModuleLoader.on('loaded', gcStats => {
```
basti1302, line 41: This was the main issue: We previously only added the listeners to the nativeModuleLoader once `activate` was called (which happens when a connection to the back end is established). This is of course far too late, especially if gcstats.js is already available in node_modules. 

Thus, we only registered the listener long after the `loaded` event would have been emitted, causing us to miss that event entirely and never executing the handler for `loaded`.

packages/shared-metrics/src/gc.js
```diff
@@ -37,37 +47,42 @@ const nativeModuleLoader = require('./util/nativeModuleRetry')({
...
-        minorGcWindow.addPoint(1);
-        majorGcWindow.addPoint(1);
-        incrementalMarkingsWindow.addPoint(1);
-        processWeakCallbacksWindow.addPoint(1);
-      }
-      exports.currentPayload.usedHeapSizeAfterGc = stats.after.usedHeapSize;
-    });
-    startSensing();
-  });
+nativeModuleLoader.on('loaded', gcStats_ => {
```
basti1302, line 50: Now we register the listener immediately.

packages/shared-metrics/src/gc.js
```diff
@@ -37,37 +48,46 @@ const nativeModuleLoader = require('./util/nativeModuleRetry')({
...
+      incrementalMarkingsWindow.addPoint(1);
+    } else if (type === 8) {
+      processWeakCallbacksWindow.addPoint(1);
+    } else if (type === 15) {
+      minorGcWindow.addPoint(1);
+      majorGcWindow.addPoint(1);
+      incrementalMarkingsWindow.addPoint(1);
+      processWeakCallbacksWindow.addPoint(1);
+    }
+    exports.currentPayload.usedHeapSizeAfterGc = stats.after.usedHeapSize;
```
willianpc, line 87: ~~Is `stats.after.usedHeapSize` a function or an attribute that will be read only once?~~
~~As this is inside of the `on` event, if this is an attribute, I am honestly not entirely sure if Node will update an exported value or what could be the implications with that. Could this be an issue?~~

I saw that something similar is done already at line 94 to 97, so I think this just works as intended.

willianpc (approved):
LGTM.
I added some general questions not necessarily related to the code, but we can discuss on Slack another time :)

packages/shared-metrics/src/gc.js
basti1302: We read the exports.currentPayload once per second here: https://github.com/instana/nodejs-sensor/blob/7abc5a5fd3c74bd3a9592decf3efa700f3d3a292/packages/core/src/metrics/index.js#L58 before processing it further.

packages/shared-metrics/src/gc.js
```diff
@@ -37,37 +48,46 @@ const nativeModuleLoader = require('./util/nativeModuleRetry')({
...
-  });
+nativeModuleLoader.on('loaded', gcStats_ => {
+  gcStats = gcStats_;
+  exports.currentPayload.statsSupported = true;
+  if (activateHasBeenCalled) {
+    actuallyActivate();
+  }
+});
+
+nativeModuleLoader.on('failed', () => {
```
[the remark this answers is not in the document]
basti1302: You mean remove the handlers for those events? We do not remove them. We could probably use [`once`](https://nodejs.org/api/events.html#events_emitter_once_eventname_listener) instead of `on` for that. Though I don't think it is an issue.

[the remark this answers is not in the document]
basti1302: Switched to `once` now because it is more appropriate for this use case.

PR #311, opened 2021-03-29
chore: run "npm run update-deps"

basti1302:

PR #313, opened 2021-03-30
Update root dependencies

basti1302:
Follow up to #311


rtrys:
@basti1302 [node-gyp@8.0.0](https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md#v800-2021-04-03) is released and here is replacing deprecated [request](https://github.com/request/request), would you mind update to [node-gyp@8.0.0](https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md#v800-2021-04-03) in this PR?

basti1302:
> @basti1302 node-gyp@8.0.0 is released and here is replacing deprecated request, would you mind update to node-gyp@8.0.0 in this PR?

This PR is about the updating the packages in the root package.json and package-lock.json. Those are only used for development purposes (integration tests etc.). None of those dependencies end up in the packages we release to the npm registry.

About node-gyp@8, I didn't look into it but since the release notes say they dropped support for Python 2 I am pretty sure we cannot move to that version soon. We need to support a lot of customers with a wide range of setups and I'm pretty sure there are still quite a few scenarios where `@instana/collector` is installed when Python 2 is the default Python.

rtrys:
thanks for explain me, I'm going to create an issue to request the update and no forget

basti1302:
This has become stale and will not be merged. Most parts of this are now in #335 and [this branch](https://github.com/instana/nodejs-sensor/tree/update-root-dependencies-2).

PR #319, opened 2021-04-07
Calling getCls() from collector module instead of relative path to cls

willianpc:
The cls module is not found, as reported here: https://github.com/instana/nodejs-sensor/issues/318

### DIFF
--- a/packages/collector/src/tracing/instrumentation/process/childProcess.js
+++ b/packages/collector/src/tracing/instrumentation/process/childProcess.js
@@ -7,14 +7,7 @@
 
 const processIdentityProvider = require('../../../pidStore');
 
-/**
- * We cannot call cls directly from the core by using relative path with require('../../../../../core/src/tracing/cls')
- * otherwise customers will get a MODULE NOT FOUND error.
- * Instead, we have to rely on the @instana/core module and use the tracing.getCls function, which must be invoked
- * in the `exports.init()` function;
- */
 const getCls = require('@instana/core').tracing.getCls;
-let cls;
 const coreChildProcess = require('child_process');
 const shimmer = require('shimmer');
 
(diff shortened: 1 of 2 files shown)


packages/collector/src/tracing/instrumentation/process/childProcess.js
```diff
@@ -6,7 +6,15 @@
 'use strict';
 
 const processIdentityProvider = require('../../../pidStore');
-const cls = require('../../../../../core/src/tracing/cls');
+
+/**
+ * We cannot call cls directly from the core by using relative path with require('../../../../../core/src/tracing/cls')
+ * otherwise customers will get a MODULE NOT FOUND error.
+ * Instead, we have to rely on the @instana/core module and use the tracing.getCls function, which must be invoked
+ * in the `exports.init()` function;
```
basti1302, line 14: I don't think we need this explanation here.

packages/collector/src/tracing/instrumentation/process/childProcess.js
```diff
@@ -24,6 +32,7 @@ logger = require('../../../logger').getLogger('tracing/child_process', newLogger
 // There is also ./edgemicro.js, which is responsible for instrumenting the code that is used to spawn the individual
 // edgemicro workers.
 exports.init = function() {
+  cls = getCls();
```
basti1302, line 35: It  might be better to call `getCls` where `cls`  is actually used (around line 103), instead of calling it in init.

We also nee to add a null check there, because `getCls` might not return a value in some cases.

See `packages/collector/src/uncaught/index.js` for an example.

basti1302 (changes requested):

basti1302:
Please release a fix for this as soon as possible :-)

willianpc:
Force merging as Bastian is on Holiday, and because his requested changes were already applied.

PR #320, opened 2021-04-08
Added Module._load patch to collector

willianpc:
[skip ci]

### DIFF
--- a/packages/collector/src/index.js
+++ b/packages/collector/src/index.js
@@ -5,6 +5,19 @@
 
 'use strict';
 
+const Module = require('module');
+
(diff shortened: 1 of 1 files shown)


willianpc:
@basti1302 what do you hink?
This is a small PoC / proposal to prevent sibling packages in the monorepo to require each other.
Of course, in this case the protection is specifically from the collector towards the core, but we can either replicate the change to other packages or think of something else to reuse the code.

basti1302:
IMO this can be solved by static analysis and should not require a check at runtime. I've put up #321 as an alternative, checking this with an eslint plugin.

PR #321, opened 2021-04-12
chore: add lint rule to disallow relative imports across package boundary

basti1302:


basti1302:
@willianpc I think we should rather do something like this instead of #320. WDYT?

willianpc:
> @willianpc I think we should rather do something like this instead of #320. WDYT?

Looks great. And it doesn't require any code in the API.

willianpc (approved):
LGTM

PR #322, opened 2021-04-13
Typechecking core/tracing folder except instrumentation

willianpc:
The type checking is limited to the `core/tracing` folder only, without recursion. However, all related sub folders that are required by modules located in the root had to be automatically typed as well (this is how TS works, even though if we limit the type checking to a specific folder or module).


packages/core/src/tracing/cls.js
```diff
@@ -155,238 +169,290 @@ class InstanaPseudoSpan extends InstanaSpan {
...
+function call(fn) {
+  fn();
+}
 
-  // If the client code has specified a trace ID/parent ID, use the provided IDs.
-  if (traceId) {
-    span.t = traceId;
-    if (parentSpanId) {
-      span.p = parentSpanId;
+module.exports = {
```
willianpc, line 179: @basti1302 
This is a "big" change for the cls module, so I am tagging you earlier, in case you disagree/have questions/suggestions.
Basically, this has no effect to whomever require the cls module.
The difference is that TS recognizes automagically all the type annotations of each exported method already, so I don't have to do the job twice.
Why twice you may ask? Because we are exporting method by method, but from the importers perspective, this is one module (cls) with lots of methods inside. In this case, TS does not understand (at least in the way we are doing this) that all methods belong to the whole cls module. This makes me:
 1. Annotate types for each individual method
 2. Annotate a "cls typedef" where I need to repeat the signature of each method and tell TS: hey this is the type that this module represents. 
 3. Then, because of item 2, for each module that requires cls, I need to say: hey use this typedef in order to know what is inside of cls

We can also, as always jump into a call to discuss

willianpc: one note: right now, I am running CI tests to get problems, and fixing them

packages/core/src/tracing/cls.js
```diff
@@ -10,39 +10,72 @@ const tracingUtil = require('./tracingUtil');
...
- * Usage:
- *   cls.ns.get(key);
- *   cls.ns.set(key);
- *   cls.ns.run(function() {});
+/**
+ * This type is to be used across the code base, as we don't have a public explicit type for a span.
+ * Also, it is often that we simply create a literal object and gradually throw few properies on it and
+ * handle them as spans. This type has all span properties, but they are all optional, so we can safely
+ * type these literal objects.
+ * TODO: move InstanaSpan and InstanaPseudoSpan to their own file and make them publicly accessible?
```
willianpc, line 54: @basti1302 I would like to discuss this with you. Not sure if there is a reason for the InstanaSpan and InstanaPseudoSpan to be private within the module

packages/core/src/tracing/index.js
```diff
@@ -203,13 +239,20 @@ exports.enableSpanBatching = function enableSpanBatching() {
 exports._getAndResetTracingMetrics = function _getAndResetTracingMetrics() {
   return {
     pid:
+      // TODO: This is not instrumented yet -> /nodejs-sensor/packages/collector/src/pidStore/index.js
```
basti1302, line 242: Not instrumented yet? Or did you mean not _typed_ yet?

packages/core/src/tracing/opentracing/Span.js
```diff
@@ -9,11 +9,66 @@ const opentracing = require('opentracing');
...
+ */
+
+/**
+ * @typedef {Object} InnerSpanData
+ * @property {string} service
+ * @property {InnerSpanDataSdk} sdk
+ */
+
+/**
+ * @typedef {Object} InnerSpan
```
basti1302, line 32: *question:* Why is it **Inner**Span. What does the "Inner" refer to?

packages/core/src/tracing/sdk/sdk.js
```diff
@@ -151,14 +180,20 @@ module.exports = exports = function(isCallbackApi) {
     completeSpan(error, span, tags);
   }
 
+  /**
+   * @param {string} name
+   * @param {string} tags
+   * @param {Function | string} callback
+   * @returns
```
basti1302, line 187: *question:* What does an empty `@returns` annotation do? There are a couple of them in this PR.

packages/core/src/tracing/spanHandle.js
```diff
@@ -165,22 +182,35 @@ SpanHandle.prototype.end = function end(errorCount) {
 };
 
 /**
+ * TODO: make it as a class
```
basti1302, line 185: Yes, we should. Is this in the scope for this PR or something for later?

packages/core/src/tracing/spanHandle.js
```diff
@@ -9,6 +9,7 @@ const constants = require('./constants');
 
 /**
  * Provides very limited access from client code to the current active span.
+ * @param {import('./cls').InstanaBaseSpan} _span
  */
 function SpanHandle(_span) {
```
basti1302, line 14: SpanHandle should probably also become a class, same as NoopSpanHandle.

packages/core/src/tracing/spanHandle.js
```diff
@@ -150,6 +166,7 @@ SpanHandle.prototype.disableAutoEnd = function disableAutoEnd() {
 
 /**
  * Finishes as span that has been switched to manual-end-mode before.
+ * @param {boolean | number} errorCount
```
basti1302, line 169: *issue:* I think this should always be a number. Do we pass in a boolean somewhere? That might be wrong.

packages/core/src/tracing/tracingHeaders.js
```diff
@@ -313,6 +348,30 @@ function readW3cTraceContext(headers) {
...
 }
 
+/**
+ * @typedef {Object} InstanaAncestor
+ * @property {string} t
+ * @property {string} p
+ */
+
+/**
+ * @typedef {Object} TraceResult
```
basti1302, line 358: *suggestion (non-blocking):*

This object represents the incoming headers (Instana headers and W3C trace context in a format that has already been processed and normalized and provides individual instrumentations (like httpServer or http2Server)  with the means to decide how to continue the trace and correlate it with a parent span.

I think we can find a better name than `TraceResult` -- that sounds like the result of tracing something, which is not accurate. We did not yet trace anything at this point. 

Better names might be one of:
- `TracingHeaders`,
- `TraceHeaders`,
- `TraceContext`.



WDYT?

basti1302 (changes requested):
Excellent work! Thanks a lot for doing this 🙇 🙇 

I left a couple of comments for discussion and also requested some minor changes.

packages/core/src/tracing/cls.js
willianpc: Ok. I won't touch this right now, for the sake of not making this PR even fatter. But will do in the next iteration.

packages/core/src/tracing/cls.js
willianpc: From my experience, it *eventually* happens that a code review becomes more complex to be done depending on the changes, which seems to be totally acceptable.
I would probably raise a concern too, if it was a recurring situation, instead of bringing that up in the very first time that such thing happens, but that's how I think, of course.
We can discuss in person which level of "pickiness" we want for our CR sessions so to prevent undesired code changes that one does not like to look at.

About the question, I think that it could be done in the way that you mentioned. I honestly cannot remember why I chose to do it in this particular way, since the PR is sitting there for quite some time.
My suspicion is that the work to do such thing should not be too big, so I will give it a try, and if I stumble into any unforeseen issue, I let you know.

packages/core/src/tracing/constants.js
```diff
@@ -42,23 +42,25 @@ exports.SDK = {
   INTERMEDIATE: 'intermediate'
 };
 
-/*
+/**
  * Determine if <span> is an entry span (server span).
+ * @param {import('./cls').InstanaBaseSpan} span
```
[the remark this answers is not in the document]
willianpc: Oops, good catch again. Thanks!

packages/core/src/tracing/spanHandle.js
willianpc: I would leave it to another PR, so we keep the context of typing. Even though there were some code changes, but still they were related to the typing.

packages/core/src/tracing/spanHandle.js
willianpc: Line 175 to 177:
```javascript
  if (errorCount === true) {
    errorCount = 1;
  }
```

packages/core/src/tracing/tracingHeaders.js
willianpc: changed it to `TracingHeaders`

packages/core/src/tracing/cls.js
```diff
@@ -10,39 +10,72 @@ const tracingUtil = require('./tracingUtil');
...
+/**
+ * @typedef {Function} GetFromFunction
+ * @returns {GetFromProperties}
+ */
+
+/**
+ * @description This object is actually here: /nodejs-sensor/packages/collector/src/pidStore/index.js
+ * @typedef {Object} PIDStore
+ * @property {GetFromFunction} getFrom
+ */
```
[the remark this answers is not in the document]
willianpc: I explicitly added a comment saying that PIDStore belongs to the collector, which I am not instrumenting here.
Touching something else means that Typescript will look for all dependencies, even though we don't clearly say that we don't want to instrument a folder. This may or may not happen in this case, but from my experience so far, it becomes a rabbit hole very easily.

If we add this interface in the /core area: This is not actually Typescript, so creating a file for an interface means that this file will only exist because of the interface. In a "normal" TS project, such file would be deleted during the build.
Having that said, where would you recommend that the interface is created?

packages/core/src/tracing/constants.js
```diff
@@ -42,23 +42,25 @@ exports.SDK = {
...
+ * @returns {boolean}
  */
 exports.isExitSpan = function isExitSpan(span) {
   return span && span.k === exports.EXIT;
 };
 
-/*
+/**
  * Determine if <span> is an intermediate span (local span).
+ * @param {import('./cls').InstanaBaseSpan} span
```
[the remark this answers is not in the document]
willianpc: Fixed.

packages/core/src/tracing/index.js
willianpc: Oops, I was probably thinking about something else.

packages/core/src/tracing/opentracing/Span.js
willianpc: At the time I was confused, as we have more than one definition of a span in different places.
I don't remember exactly why I picked this name. Probably because it is referenced inside another span or some other object.
Basically we have Span class, InstanaSpan, InstanaPseudoSpan, and in many cases we have arbitrary objects with pieces of a span.
Would you suggest a different name? `OpenTracingSpan` maybe?

basti1302 (approved):

packages/core/src/tracing/cls.js
```diff
@@ -246,147 +316,192 @@ exports.putPseudoSpan = function putPseudoSpan(spanName, kind, traceId, spanId)
...
+  isTracing,
+  setTracingLevel,
+  tracingLevel,
+  tracingSuppressed,
+  getAsyncContext,
+  enterAsyncContext,
+  leaveAsyncContext,
+  runInAsyncContext,
+  runPromiseInAsyncContext
+};
```
basti1302, line 507: Thanks for iterating on that. ❤️

packages/core/src/tracing/opentracing/Span.js
basti1302: Only saw that now: OpenTracingSpan sounds good but we can also rename later.

PR #324, opened 2021-04-16
fix(fargate, google-cloud-run): filter environment variables secrets

basti1302:


willianpc (approved):
LGTM.
I could not spot any problem with the code.
The change looks straight forward, after I took the time to understand what exactly was going on.

PR #328, opened 2021-04-30
Instrumenting sqs-consumer

willianpc:
We are adding extra SQS instrumentation by instrumenting https://github.com/BBC/sqs-consumer.
Our SQS instrumentation requires some manual work when it comes to receive messages.
SQS-Consumer lib abstracts the polling/reading/deleting messages which is exactly where this manual code should be added, and by instrumenting the library we allow customers to safely use sqs-consumer and still get a proper trace and correlation between spans.

### DIFF
--- a/packages/collector/test/tracing/cloud/aws/sqs/test.js
+++ b/packages/collector/test/tracing/cloud/aws/sqs/test.js
@@ -46,13 +46,13 @@ if (!supportedVersion(process.versions.node)) {
 
 const retryTime = config.getTestTimeout() * 2;
 
-mochaSuiteFn('tracing/cloud/aws/sqs', function() {
+mochaSuiteFn('tracing/cloud/aws/sqs', function () {
   this.timeout(config.getTestTimeout() * 3);
 
   globalAgent.setUpCleanUpHooks();
   const agentControls = globalAgent.instance;
 
-  describe('tracing enabled, no suppression', function() {
+  describe('tracing enabled, no suppression', function () {
     const senderControls = new ProcessControls({
       appPath: path.join(__dirname, 'sendMessage'),
       port: 3215,
@@ -110,6 +110,30 @@ mochaSuiteFn('tracing/cloud/aws/sqs', function() {
             ]);
           }, retryTime);
         }
+      });
+    });
+
+    describe('sqs-consumer API', () => {
+      const sqsConsumerControls = new ProcessControls({
+        appPath: path.join(__dirname, 'sqs-consumer'),
+        port: 3216,
+        useGlobalAgent: true,
+        env: {
+          AWS_SQS_QUEUE_URL: queueURL
+        }
+      });
+
+      ProcessControls.setUpHooksWithRetryTime(retryTime, sqsConsumerControls);
+
+      const apiPath = '/send-callback';
+
+      it('receives message', async () => {
+        const response = await senderControls.sendRequest({
+          method: 'POST',
+          path: apiPath
+        });
+
+        return verify(sqsConsumerControls, response, apiPath, false);
       });
     });
 
--- a/packages/core/src/tracing/instrumentation/cloud/aws/sqs.js
+++ b/packages/core/src/tracing/instrumentation/cloud/aws/sqs.js
@@ -43,12 +43,13 @@ let isActive = false;
 
 exports.init = function init() {
   requireHook.onModuleLoad('aws-sdk', instrumentSQS);
+  requireHook.onModuleLoad('sqs-consumer', instrumentSQSConsumer);
 };
 
 function instrumentSQS(AWS) {
   // /aws-sdk/lib/service.js#defineMethods
-  shimmer.wrap(AWS.Service, 'defineMethods', function(original) {
-    return function(svc) {
+  shimmer.wrap(AWS.Service, 'defineMethods', function (original) {
+    return function (svc) {
       const patchedMethod = original.apply(this, arguments);
 
       if (
@@ -68,7 +69,7 @@ function instrumentSQS(AWS) {
 }
 
 function shimSendMessage(originalSendMessage) {
-  return function() {
+  return function () {
     if (isActive) {
       const originalArgs = new Array(arguments.length);
       for (let i = 0; i < originalArgs.length; i++) {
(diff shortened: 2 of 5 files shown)


packages/core/src/tracing/instrumentation/cloud/aws/sqs.js
```diff
@@ -252,10 +255,23 @@ function instrumentedReceiveMessage(ctx, originalReceiveMessage, originalArgs) {
...
+     * As we use message attributes to store Instana headers, if the customer does not set this attribute to All,
+     * we cannot see the Instana headers, so we need to explicitly add them.
+     */
+    const receveingParams = originalArgs[0];
+
+    if (!receveingParams.MessageAttributeNames) {
+      receveingParams.MessageAttributeNames = [];
+    }
+
+    receveingParams.MessageAttributeNames.push('X_INSTANA_L', 'X_INSTANA_S', 'X_INSTANA_T');
```
basti1302, line 269: Just checked the docs on that:
> When using ReceiveMessage, you can send a list of attribute names to receive, or you can return all of the attributes by specifying All or .* in your request. You can also use all message attributes starting with a prefix, for example bar.*.

Three questions: 
1) What is the default behaviour if `MessageAttributeNames` is not set at all? Does it default to `All` (receive all attributes by default) or `[]` (receive no attributes by default). If it is the former, we should not add our `X-INSTANA-...` attributes because that would drastically change the behaviour then: Instead of receiving all attributes, only our trace correlation headers would be received. If the default behaviour is the latter (do not receive any attributes), then setting that parameter and adding our attribute names is mandatory and correct, of course.

2) Would we want to omit adding the `X_INSTANA_...` values if we see that `All` or `.*` is already set? (Not sure if it makes a difference to add our headers explicitly, probably not?)

3) If I read the docs correctly, when we add our attribute names, we could also just add one header name prefix, `X_INSTANA.*` instead of explicitly adding all three individually. WDYT?

packages/core/src/tracing/instrumentation/cloud/aws/sqs.js
```diff
@@ -321,6 +337,7 @@ function instrumentedReceiveMessage(ctx, originalReceiveMessage, originalArgs) {
         );
 
         promise.instanaAsyncContext = cls.getAsyncContext();
+        sqsConsumerAsyncContext = promise.instanaAsyncContext;
```
basti1302, line 340: Since this variable exists in the scope of our instrumentation module: What happens if two receive-messages calls happen at the same time? Or rather, if a second `receiveMessage` call is started while an earlier `receiveMessage` call is still in progress? Would we run the risk of overwriting the `sqsConsumerAsyncContext` of an ongoing call with a different async context?

In case this is actually a problem: Do we have access to the `promise` object in `instrumentedSQSConsumerExecuteHandler`? Or can we attach it to some object that we have access to in `instrumentedSQSConsumerExecuteHandler`?

basti1302 (approved):
LGTM in general. 👍

packages/collector/test/tracing/cloud/aws/sqs/test.js
```diff
@@ -144,13 +168,13 @@ mochaSuiteFn('tracing/cloud/aws/sqs', function() {
...
       if (withError === 'sender') {
-        expect(response.data).to.equal("MissingRequiredParameter: Missing required key 'MessageBody' in params");
+        expect(response.data).to.match(/MissingRequiredParameter: Missing required key 'MessageBody' in params/);
       } else {
         return retry(() => {
           if (isBatch) {
             verifyResponseAndBatchMessage(response, receiverControls);
           } else {
-            verifyResponseAndMessage(response, receiverControls);
+            return verifyResponseAndMessage(response, receiverControls);
```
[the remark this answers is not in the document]
willianpc: you are right damn. my intention was to add `await`, not return... and even with `await` I am not entirely sure if it will solve the issue.
I will review it, thanks!

packages/core/src/tracing/instrumentation/cloud/aws/sqs.js
willianpc: 1. As far as I can see, it's empty array, at least is the default in sqs-consumer. So I understood this addition as an improvement of the instrumentation. When instrumenting SQS, I added `['All']` in the test app, which  may be the case we missed this.

2. I will add this logic. I did not want to write much, which is why I didn't add this check, but it makes sense.

3. I did not know that, and it makes total sense to me, thanks

packages/core/src/tracing/instrumentation/cloud/aws/sqs.js
willianpc: I thought about that too, but I assumed that this should not be a problem because:

1. Either this will run in a single thread or in a child process, which would then be a copy of the instrumentation (or am I wrong and they are shared anyway?)
2. The way SQS works is by long polling, which makes all messages to arrive together anyway, in the same context. There is no concurrent message processing, at least afaik.

But I will double check if this is a problem (it could be as I will have to fix tests).
I need to double check if we have access to the promise from executeHandler. From what I remember, we don't as sqs-consumer almost always return `await promise`... I need to go back to the code, as it may be that a lot of extra work needs to be done

packages/collector/test/tracing/cloud/aws/sqs/test.js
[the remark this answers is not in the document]
basti1302: `verifyResponseAndMessage` is synchronous. No await needed. (We are already inside the `retry` at this point.)

packages/core/src/tracing/instrumentation/cloud/aws/sqs.js
basti1302: > Either this will run in a single thread or in a child process

I don't follow. AFAICT there is nothing in sqs-consumer that would guarantee (or even make it likely) that child processes or threads are used.  

> The way SQS works is by long polling

What if an app receives from multiple queues? Wouldn't that be multiple long polling requests that happen in parallel?

willianpc: I managed to implement a local reference to the async context in a new version.
Could you please take a look? Essentially, I patched `sqsConsumer.sqs.receiveMessage.promise` from the `receiveMessage` function from SQS-Consumer, and added promise.instanaAsyncContext to ctx (sqs this reference).
Not the global instana async context is gone

packages/core/src/tracing/instrumentation/cloud/aws/sqs.js
willianpc: Just for history purposes:
 * aws-sdk will not display any message attributes if not explicitly defined in the MessageAttributeNames array.

PR #329, opened 2021-05-04
fix(tracing): fix TypeError in @elastic/elasticsearch instrumentation

basti1302:
This fixes `TypeError: Cannot read property 'indexOf' of undefined\n at processParams (/usr/src/app/node_modules/@instana/core/src/tracing/instrumentation/database/elasticsearchModern.js:222:16`.


willianpc (approved):
LGTM

PR #330, opened 2021-05-05
fix(events): handle nulls properly when reporting exceptions/rejections

basti1302:


packages/collector/src/uncaught/index.js
```diff
@@ -183,11 +186,18 @@ function onUnhandledRejection(reason) {
 }
 
 function createEventForUnhandledRejection(reason) {
-  return createEvent(reason, 'An unhandled promise rejection occured in a Node.js process.', 5);
+  return createEvent(reason, 'An unhandled promise rejection occured in a Node.js process.', 5, true);
```
willianpc, line 189: Reminder for myself: create enum for severity and remove magic numbers

willianpc (approved):
lgtm.
We are using magic numbers for severity, and also we have a string as errorType in `throwUncaughtError(errorType)`, but both things were already there and are not part of the PR, so I will handle them when I start adding types in these places.

packages/collector/src/uncaught/index.js
basti1302: Just for context: These numbers are dictated by the back end, those are not severity levels we use in `@instana/collector`. Of course we can add some kind of enum for them but the severity value we send to the agent with the event JSON payload needs to be an actual number with the values we use here.

PR #331, opened 2021-05-06
Removed duplicated instrumentation of receiveMessage

willianpc:
The way it was implemented, we were running sqs.receiveMessage twice, causing slowness in the message polling, and also losing some messages that were ending up in the dead letter queue.
Some extra discussion that is relevant: https://github.com/instana/nodejs-sensor/pull/328#discussion_r623862433
Basically, the concern is that a shared object is used in two different queues, so we avoid that by creating an object inside the sqs reference based on  the queue name.


basti1302 (approved):
LGTM! 🎉

PR #332, opened 2021-05-06
fix(tracing): let instrumented grpc ServiceClient inherit from Client

basti1302:
This fixes an issue with [`mali`](https://www.npmjs.com/package/mali) up to version `0.20.0` which relies on `ServiceClient.super_` to refer to `Client`.


willianpc (approved):
LGTM. No extra comments.

PR #334, opened 2021-05-25
AWS Refactoring + AWS Lambda invoke function instrumentation

willianpc:
This PR includes the instrumentation of invoke and invokeAsync Lambda functions, but also has the AWS instrumentations refactoring squeezed in.

1. SQS remains untouched due to its particularly more complex instrumentation
2. A new single point of instrumentation was added
3. A class InstanaAWSProduct was created, which should be extended by specific AWS Products. eg: S3
4. The following AWS products suffered refactoring to be part of the new format:
    * S3
    * Kinesis
    * Lambda (Brand new instrumentation)
    * DynamoDB
1. A new test app was created in order to test multiple AWS products being used in the same place
2. A new integration test was created for the multi AWS product app
    * The test contains one operation from each of the mentioned products at item 4
    * The test does a round robin between the methods Callback, Async and Promise, in oder to reduce the number of tes cycles (something we may reuse in the individual AWS tests and other cloud dependent tests as well)

### DIFF
--- a/packages/collector/test/tracing/cloud/aws/lambda/app.js
+++ b/packages/collector/test/tracing/cloud/aws/lambda/app.js
@@ -1,0 +1,124 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2021
+ */
+
+'use strict';
+
+require('../../../../../src')();
+
+const AWS = require('aws-sdk');
+const express = require('express');
+const logPrefix = `AWS Lambda (${process.pid}):\t`;
+AWS.config.update({ region: 'us-east-2' });
+const lambda = new AWS.Lambda();
+const functionName = process.env.AWS_LAMBDA_FUNCTION_NAME || 'team-nodejs-invoke-function';
+const log = require('@instana/core/test/test_util/log').getLogger(logPrefix);
+
+const operations = {
+  invoke: {
+    FunctionName: functionName,
+    // "Event"|"RequestResponse"|"DryRun"
+    InvocationType: 'RequestResponse',
+    Payload: '{"ok": true}'
+  },
+  invokeAsync: {
+    FunctionName: functionName,
+    InvokeArgs: '{"ok": true}'
+  }
+};
+
+/**
+ * @param {'invoke' | 'invokeAsync'} op
+ * @param {(err: *, data: *) => void} cb
+ * @param {boolean} withError
+ * @param {boolean} ctx
+ */
+function execOperation(op, cb, withError = false, ctx = false) {
+  /** @type {import('aws-sdk').Lambda.Types.InvocationRequest | import('aws-sdk').Lambda.Types.InvokeAsyncRequest} */
+
+  const options = { ...operations[op] };
+  if (withError) {
+    options.InvalidParameter = 1;
+  }
+
+  if (ctx && op === 'invoke') {
+    const base64Value = Buffer.from('{"Custom": {"awesome_company": "Instana"}}').toString('base64');
+    options.ClientContext = base64Value;
+  }
+
+  if (typeof cb === 'function') {
+    lambda[op](options, (err, data) => {
+      cb(err, {
+        data,
+        clientContext: options.ClientContext
+      });
+    });
+  } else {
+    const p = lambda[op](options).promise();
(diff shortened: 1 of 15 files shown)


packages/collector/test/tracing/cloud/aws/lambda/app.js
```diff
@@ -0,0 +1,113 @@
...
+function execOperation(op, cb, withError = false, ctx = false) {
+  /** @type {import('aws-sdk').Lambda.Types.InvocationRequest | import('aws-sdk').Lambda.Types.InvokeAsyncRequest} */
+
+  const options = { ...operations[op] };
+  if (withError) {
+    options.InvalidParameter = 1;
+  }
+
+  if (ctx && op === 'invoke') {
+    options.ClientContext = 'eyJzb21lX2NvbnRlbnQiOiAxfQ==';
```
basti1302, line 46: Just for my understanding: What is encoded in this client context in the test application? Ah, nevermind, it is just a test value "{"some_content": 1}". Maybe you could spell that out in clear text and then encode it to base64 on the fly, so it becomes obvious? (This relates to another comment around client context further down.)

packages/core/src/tracing/instrumentation/cloud/aws/lambda.js
```diff
@@ -0,0 +1,133 @@
...
+      };
+
+      // invokeAsync doesn't have the option ClientContext
+      if (originalArgs[0] === 'invoke') {
+        if (params.ClientContext != null) {
+          clientContextContentBase64 = Buffer.from(params.ClientContext, 'base64').toString();
+
+          try {
+            clientContextContentJSON = JSON.parse(clientContextContentBase64);
+            clientContextContentJSON.Custom = instanaCustomHeaders.Custom;
```
basti1302, line 55: What if the existing client context already has the attribute `Custom`? Apparently this is some kind of "standard" attribute that others could be using as well (https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html). If I read this correctly, we would  overwrite the existing `Custom` value unconditionally?

I think we need to check if `clientContextContentJSON.Custom` already exists and add our attributes to it if it does.

packages/collector/test/tracing/cloud/aws/lambda/test.js
```diff
@@ -0,0 +1,169 @@
...
+              });
+              return verify(appControls, response, apiPath, operation, withError);
+            });
+          });
+        });
+      });
+    });
+    function verify(controls, response, apiPath, operation, withError) {
+      return retry(() => {
+        return agentControls.getSpans().then(spans => verifySpans(controls, spans, apiPath, operation, withError));
```
basti1302, line 69: Suggestion: Since we manipulate the client context in our instrumentation, it would be great if we somehow could verify in the test that that content the test application put in there is still in there after our instrumentation manipulated it. I'm not sure how to achieve that - maybe if we add the client context to the http response after executing the lambda invocation (see above).

packages/collector/test/tracing/cloud/aws/lambda/app.js
```diff
@@ -0,0 +1,113 @@
...
+  res.status(500).send({
+    status: 'failed',
+    error: err
+  });
+}
+
+function httpSuccess(res, data) {
+  res.send({
+    status: 'ok',
+    data: data
```
basti1302, line 71: If we add `options.ClientContext` to the HTTP response, we should be able to formulate expectations around the the modified client context in the test. Maybe worth a try.

packages/core/test/test_util/log.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+'use strict';
+
+/**
+ * Returns a log function
+ * @param {string} logPrefix
+ * @returns {(...args: *) => void}
+ */
+exports.getLogger = function (logPrefix) {
```
basti1302, line 13: 👍 👍 👍

basti1302 (changes requested):
This looks great! I have just one blocking comment about the modifications of the client context value (and some related questions/suggestions around testing that), everything else looks good. Great work refactoring the AWS SDK instrumentations to a unified approach 👍 👍

basti1302 (approved):

PR #337, opened 2021-05-27
chore: update @types/node to latest 14.x

basti1302:
Apparently the aliases `node:events`, `node:http` etc. have been removed
with
https://github.com/DefinitelyTyped/DefinitelyTyped/commit/ec6522b3d8f1b15251aaed2f3208f7e697a7ee96#diff-50ed3ba474b514ee7443ee72b00abc59b143f94f70bb8a0a83c3293ec21f4540
so the core modules need to be referred to without the prefix now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6298,9 +6298,9 @@
       }
     },
     "@types/node": {
-      "version": "14.14.43",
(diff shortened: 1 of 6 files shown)


willianpc (approved):
LGTM.
Thanks for this update

PR #338, opened 2021-05-28
feat(lambda): detect Lambda cold starts & timeouts

basti1302:


packages/aws-lambda/src/wrapper.js
```diff
@@ -93,8 +95,13 @@ function shimmedHandler(originalHandler, originalThis, originalArgs, _config) {
         alias,
         runtime: 'nodejs',
         functionName: context.functionName,
-        functionVersion: context.functionVersion
+        functionVersion: context.functionVersion,
+        reqId: context.awsRequestId
```
basti1302, line 99: This is unrelated to this PR but I figured that having the request ID in the span would be actually quite helpful for correlating a trace in Instana with cloudwatch logs, so I added it.

packages/aws-lambda/test/integration_test/test.js
```diff
@@ -1010,74 +1024,84 @@ function registerTests(handlerDefinitionPath) {
...
-        expect(span.ec).to.equal(1);
-      } else if (error === false) {
-        expect(span.data.lambda.error).to.not.exist;
-        expect(span.ec).to.equal(0);
-      } else {
-        throw new Error(`Unknown error expectation type, don't know how to verify: ${error}`);
-      }
-      verifyHeaders(span);
-    });
+    const checks = [
```
basti1302, line 1027: Most of this diff comes from migrating from the `expectExactlyOneMatching(spans, callback)` syntax to `expectExactlyOneMatching(spans, [expectations])` syntax.

packages/aws-lambda/test/integration_test/test.js
```diff
@@ -1010,74 +1024,84 @@ function registerTests(handlerDefinitionPath) {
...
+    } else if (error === false) {
+      checks.push(span => expect(span.data.lambda.error).to.not.exist);
+      checks.push(span => expect(span.ec).to.equal(0));
+    } else {
+      throw new Error(`Unknown error expectation type, don't know how to verify: ${error}`);
+    }
+    if (expectations.expectColdStart === true) {
+      checks.push(span => expect(span.data.lambda.coldStart).to.be.true);
+    } else if (expectations.expectColdStart === false) {
+      checks.push(span => expect(span.data.lambda.coldStart).to.not.exist);
```
basti1302, line 1077: I believe this is actually the only new check that has been added here.

packages/aws-lambda/test/integration_test/test.js
```diff
@@ -1010,74 +1024,84 @@ function registerTests(handlerDefinitionPath) {
...
+      span => expect(span.async).to.not.exist,
+      span => expect(span.data.http).to.be.an('object'),
+      span => expect(span.data.http.method).to.equal('GET'),
+      span => expect(span.data.http.url).to.equal(downstreamDummyUrl),
+      span =>
+        expect(span.data.http.header).to.deep.equal({
+          'x-downstream-header': 'yes'
+        }),
+      verifyHeaders
+    ]);
```
basti1302, line 1104: Again, no actual change, just converting this to the preferred syntax.

willianpc (approved):
LGTM. I added some minor considerations you may want to check, but no show stopper.
Cheers!

PR #339, opened 2021-05-31
AWS Lambda Correlation with Lambda invoke function

willianpc:


packages/aws-lambda/src/triggers.js
```diff
@@ -44,6 +44,9 @@ exports.enrichSpanWithTriggerData = function enrichSpanWithTriggerData(event, sp
     span.data.lambda.trigger = 'aws:sqs';
     extractEventFromSQS(event, span);
     return;
+  } else if (isInvokeFunction(context)) {
+    span.data.lambda.trigger = 'aws:sdk.invoke';
```
basti1302, line 48: I just double checked and Andrii added the trigger as `aws:lambda.invoke`. I wasn't aware of this earlier, but we should follow Go's example and use the same string. (That string is also a recognized trigger in our back end processing pipeline now.)

https://github.com/instana/go-sensor/blob/f06ddbbdb8958b0ed33a0c743e1cdbb97fe20b00/instrumentation/instalambda/handler.go#L162

packages/aws-lambda/test/integration_test/test.js
```diff
@@ -813,6 +825,91 @@ function registerTests(handlerDefinitionPath) {
...
+        .then(spans =>
+          expectExactlyOneMatching(spans, [
+            span => expect(span.n).to.equal('aws.lambda.entry'),
+            span => expect(span.k).to.equal(constants.ENTRY),
+            span => expect(span.data.http).to.not.exist
+          ])
+        ));
+  });
+
+  describe('triggered by AWS SDK Lambda invoke function with parent span from an empty context', function () {
```
basti1302, line 898: The title looks copied/pasted - maybe it should mention x-instana-l or trace suppression.

basti1302 (approved):
LGTM! We need to change the trigger name (I just learned this now), everything else looks splendid. 👍 👍 👍

PR #340, opened 2021-06-02
WIP: Testing CI for node 15

willianpc:


basti1302 (approved):
LGTM! 👍

packages/collector/test/tracing/protocols/http/client/test.js
```diff
@@ -155,7 +155,15 @@ function registerTests(useHttps) {
               span => expect(span.ec).to.equal(1),
               span =>
                 expect(span.data.http.url).to.match(/ha-te-te-peh:\/\/999\.0\.0\.1(?:\/)?:not-a-port\/malformed-url/),
-              span => expect(span.data.http.error).to.match(/Protocol .* not supported./),
+              span => {
+                if (semver.gte(process.version, '16.2.0')) {
+                  expect(span.data.http.error).to.equal('Invalid URL');
+                } else if (semver.gte(process.version, '16.0.0')) {
+                  expect(span.data.http.error).to.match(/Invalid URL: /);
```
willianpc, line 162: Can't we just match `Invalid URL` for Node 16.* and we remove one if statement? Not a big deal though.
Maybe even ignore case, just in... case? (awful pun)

willianpc (commented):
LGTM.
I have one question and a small suggestion, but nothing that would prevent a merge, IMO.

packages/autoprofile/precompile/build-single-addon.js
```diff
@@ -29,8 +30,17 @@ module.exports = exports = function buildSingleAddOn(abi, version) {
     mkdirpSync(addonDir);
   }
 
-  execSync(`node node_modules/node-gyp/bin/node-gyp.js rebuild --target=${version} --arch=x64`, {
-    cwd: path.join(__dirname, '..'),
+  let darwinNodeJs10Fix = '';
+  if (platform === 'darwin' && semver.gte(version, '10.0.0') && semver.lt(version, '12.0.0')) {
+    darwinNodeJs10Fix = '--build_v8_with_gn=false';
```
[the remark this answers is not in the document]
basti1302: It failed with an error without this extra option. Something like this:

```
**gyp: name 'build_v8_with_gn' is not defined while evaluating condition 'build_v8_with_gn == "true"' in binding.gyp while trying to load binding.gyp**
gyp ERR! configure error 
```

I took the workaround from here: https://github.com/nodejs/node-gyp/issues/1457#issuecomment-395490546

Since we no longer publish prebuild binaries for Darwin (this was also changed in this PR) it probably does not matter much. The Darwin binaries are only used for local testing now.

packages/collector/test/tracing/protocols/http/client/test.js
basti1302: Right, good call.

basti1302 (approved):

PR #341, opened 2021-06-08
test(ci): run test suite for Node.js 8 and Node.js 10 in alternating fashion

basti1302:
The mechanism I introduced in #340/343f69a to run the full test suite for Node.js 8 and 10 in an alternating fashion
on every other build was nonsense. This PR fixes that.


willianpc (approved):
LGTM

PR #342, opened 2021-06-08
Applying reusable verification functions and improving execution time

willianpc:
Essentially, this change does the following:

 * Refactor AWS tests
 * Makes `verify*` functions reusable, where most common cases are covered
 * Reduces the number of creation/deletion of AWS entities (streams, tables, buckets) to one time
 * Runs AWS tests in parallel with Promise.all, when these tests don't depend on a sequence of operations to run

### DIFF
--- a/packages/collector/test/tracing/cloud/aws/lambda/test.js
+++ b/packages/collector/test/tracing/cloud/aws/lambda/test.js
@@ -5,16 +5,15 @@
 
 'use strict';
 
-// const semver = require('semver');
 const path = require('path');
 const { expect } = require('chai');
 const { fail } = expect;
-const constants = require('@instana/core').tracing.constants;
 const supportedVersion = require('@instana/core').tracing.supportedVersion;
 const config = require('../../../../../../core/test/config');
-const { expectExactlyOneMatching, retry, stringifyItems, delay } = require('../../../../../../core/test/test_util');
+const { retry, stringifyItems, delay } = require('../../../../../../core/test/test_util');
 const ProcessControls = require('../../../../test_util/ProcessControls');
 const globalAgent = require('../../../../globalAgent');
+const { verifyHttpRootEntry, verifyExitSpan } = require('@instana/core/test/test_util/common_verifications');
 
 const SPAN_NAME = 'aws.lambda.invoke';
 const functionName = 'wrapped-async';
@@ -23,6 +22,8 @@ let mochaSuiteFn;
 const withErrorOptions = [false, true];
 const requestMethods = ['Callback', 'Promise'];
 const availableOperations = ['invoke', 'invokeAsync'];
+
+const getNextCallMethod = require('@instana/core/test/test_util/circular_list').getCircularList(requestMethods);
 
 if (!supportedVersion(process.versions.node)) {
   mochaSuiteFn = describe.skip;
@@ -47,19 +48,18 @@ mochaSuiteFn('tracing/cloud/aws/lambda', function () {
     });
     ProcessControls.setUpHooksWithRetryTime(retryTime, appControls);
     withErrorOptions.forEach(withError => {
-      requestMethods.forEach(requestMethod => {
-        describe(`getting result with: ${requestMethod}; with error: ${withError ? 'yes' : 'no'}`, () => {
-          availableOperations.forEach(operation => {
-            it(`operation: ${operation}`, async () => {
-              const withErrorOption = withError ? '&withError=1' : '';
-              const apiPath = `/${operation}/${requestMethod}`;
-              const response = await appControls.sendRequest({
-                method: 'GET',
-                path: `${apiPath}?ctx=1${withErrorOption}`,
-                simple: withError === false
-              });
-              return verify(appControls, response, apiPath, operation, withError);
+      describe(`getting result with error: ${withError ? 'yes' : 'no'}`, () => {
+        availableOperations.forEach(operation => {
+          const requestMethod = getNextCallMethod();
+          it(`operation: ${operation}/${requestMethod}`, async () => {
+            const withErrorOption = withError ? '&withError=1' : '';
+            const apiPath = `/${operation}/${requestMethod}`;
+            const response = await appControls.sendRequest({
+              method: 'GET',
+              path: `${apiPath}?ctx=1${withErrorOption}`,
+              simple: withError === false
             });
+            return verify(appControls, response, apiPath, operation, withError);
           });
         });
       });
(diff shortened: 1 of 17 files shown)


packages/core/test/test_util/common_tests.js
```diff
@@ -0,0 +1,131 @@
...
+ * @param {string} apiPath
+ * @param {string} pid
+ * @param {Array.<(span: InstanaBaseSpan) => void>} [extraTests]
+ * @returns
+ */
+exports.verifyHttpEntry = function verifyHttpEntry(expect, spans, apiPath, pid, extraTests) {
+  /** @type {Array.<(span: InstanaBaseSpan) => void>} */
+  const tests = [
+    (/** @type {InstanaBaseSpan} */ span) => {
+      expect(span.p).to.not.exist;
```
basti1302, line 27: This is only true for root entry spans. If this method is meant to check for HTTP root entry spans, it should be named accordingly (verifyHttpRootEntry).

Alternatively: We could add an optional parentSpan to the arguments. If it is present, we can test

```
expect(span.t).to.equal(parentSpan.t);
expect(span.p).to.equal(parentSpan.s);
```

If it is not present, we test

```
expect(span.p).to.not.exist;
```

The optional parentSpan argument would be a bit ugly with the current positional arguments, so this might only be a good idea if we switch to named arguments (see above).

Then again, maybe it is better to add explicit methods for both cases, like verifyHttpRootEntry and verifyHttpNonRootEntry (which might both delegate to the same method which then takes an optional parent span).

packages/collector/test/tracing/cloud/aws/lambda/test.js
```diff
@@ -72,32 +72,8 @@ mochaSuiteFn('tracing/cloud/aws/lambda', function () {
       }, retryTime);
     }
     function verifySpans(controls, response, spans, apiPath, operation, withError) {
-      const httpEntry = verifyHttpEntry(spans, apiPath, controls);
-      verifyLambdaExit(spans, response, httpEntry, operation, withError);
-    }
-    function verifyHttpEntry(spans, apiPath, controls) {
```
basti1302, line 78: Getting rid of all those repetitive implementations is 👍 👍 👍 👍

packages/core/test/test_util/common_tests.js
```diff
@@ -0,0 +1,131 @@
...
+      expect(span.p).to.equal(parent.s);
+    },
+    (/** @type {InstanaBaseSpan} */ span) => {
+      expect(span.f.e).to.equal(pid);
+    },
+    (/** @type {InstanaBaseSpan} */ span) => {
+      expect(span.f.h).to.equal('agent-stub-uuid');
+    },
+    (/** @type {InstanaBaseSpan} */ span) => {
+      expect(span.error).to.not.exist;
```
basti1302, line 81: This check should be removed completely. I know it was present in all the original verifyXxx methods but it serves no purpose (see my comment in cls.js).

packages/core/test/test_util/common_tests.js
```diff
@@ -0,0 +1,131 @@
...
+    },
+    (/** @type {InstanaBaseSpan} */ span) => {
+      expect(span.ec).to.equal(withError ? 1 : 0);
+    },
+    (/** @type {InstanaBaseSpan} */ span) => {
+      expect(span.async).to.not.exist;
```
basti1302, line 87: This check should be removed completely. I know it was present in all the original verifyXxx methods but it serves no purpose (see my comment in cls.js).

packages/core/src/tracing/cls.js
```diff
@@ -69,6 +69,8 @@ function init(config, _processIdentityProvider) {
  * @property {{s?: number, d?: number}} [b]
  * @property {*} [f]
  * @property {*} [gqd]
+ * @property {*} [error]
+ * @property {*} [async]
```
basti1302, line 73: I can see where this is coming from (I think) but we definitely should not add these two attributes to this type! 

They have been deprecated and removed from the production code a long time ago and they should never be present on any spans we produce, so we should not allow for them in this type.

When removing them from the code base I also added expect(span.error).to.not.exist and expect(span.async).to.not.exist to a lot of tests just to make sure I hadn't forgotten to remove them everywehere. Then, when copying those verification methods around, that pattern spread to a lot of test cases. My bad, I screwed that up, I should have removed those checks everywhere like two years ago. :-( 

You probably added those two here to get the new verifyXxx methods to typecheck. Instead of adding these two attributes to the type, we should instead remove the checks for the non-existence, at least in the new type checked verifyXxx methods. We know that we do not create these attributes anywhere, keeping these two checks around has no value whatsoever.

packages/core/test/test_util/common_tests.js
```diff
@@ -0,0 +1,131 @@
+/*
```
basti1302, line 1: I would prefer a file name without `_tests`. This sounds like it is a set of tests instead of a set of utility methods. How about something along the lines of `common_verifications`, `common_checks` or maybe `span_verifications` (so far we only have verifyXxxSpan methods here).

basti1302 (commented):
I know this is WIP but I was curious and wanted to check it out, so I left some early review comments. :-) 

This is shaping up nicely! 👍

packages/core/test/test_util/circular_list.js
```diff
@@ -0,0 +1,14 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2021
+ */
+
+'use strict';
+
+exports.getCircularList = function getCircularList(arr) {
```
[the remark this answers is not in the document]
willianpc: I read somewhere that, for readability purposes, using the actual object/function as part of a module is better.
It was actually associated to ES6 modules, but the same applies here.
The basic idea is that whoever imports that function doesn't have to rename it to something else, even though in most cases I have seen, people tend to use the same name anyways.
However, by exporting the named object as part of the module and not the module itself, makes it 100% guaranteed that the object will be imported with its own name at all times.

For better illustration in both cases (ES6 modules and CommonJS):

```javascript
// inside of my-module, the function is called somethingElse()
import * as myFunction from './my-module'
```
against:

```javascript
// https://i.kym-cdn.com/entries/icons/facebook/000/020/147/drake.jpg
import somethingElse from './my-module'
```

And in our case, with CommonJS

```javascript
// inside of my-module, the function is called somethingElse()
const myFunction = require('./my-module')
```
against:

```javascript
// https://i.kym-cdn.com/entries/icons/facebook/000/020/147/drake.jpg
const { somethingElse } = require('./my-module')
```

I have seen in our code base that the "export as module" is largely used. I won't make a fuss about this, so if you want to keep the current style, I will gladly change it.
Just wanted to point out why I did it the way I did.

packages/core/test/test_util/common_tests.js
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * @param {Chai.ExpectStatic} expect
+ * @param {Array.<InstanaBaseSpan>} spans
+ * @param {string} apiPath
+ * @param {string} pid
+ * @param {Array.<(span: InstanaBaseSpan) => void>} [extraTests]
+ * @returns
+ */
+exports.verifyHttpEntry = function verifyHttpEntry(expect, spans, apiPath, pid, extraTests) {
```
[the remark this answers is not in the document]
willianpc: IMO, the object as only parameter is good in pretty much any aspect (especially when it comes to the destructuring feature). But I would like to have a proper typing in place, so it doesn't become a "throw whatever you feel like in here". Right now, we don't have an explicit and clear plan to add types to the tests, so I would like to postpone this change. But we should do it! (both the object as param and typing)

packages/core/test/test_util/common_tests.js
willianpc: My thinking was towards parent HTTP entry, so I think it makes sense to rename it, as you said.
This is the most common scenario, as we discussed in the Power Point presentation I presented to you: HTTP Server ENTRY -> INSTRUMENTED_MODULE EXIT -> INSTRUMENTED_MODULE ENTRY -> HTTP Client EXIT.

I suggest to keep it as is for now (no optional argument, renaming the function properly), and we discuss again when the next wave of refactoring comes up.

packages/core/test/test_util/circular_list.js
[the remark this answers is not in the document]
basti1302: Hm, good point. Never thought about it that way. I just tend to favor the other style I mentioned because it looks kinda neat for single purpose modules. But I don't have any strong opinion about it whatsoever. Both styles are present in the code base so by all means, there is no need to change this module here.

packages/core/test/test_util/common_tests.js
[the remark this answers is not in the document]
basti1302: > But I would like to have a proper typing in place, so it doesn't become a "throw whatever you feel like in here". 

I fully agree.

packages/core/test/test_util/common_verifications.js
```diff
@@ -0,0 +1,143 @@
...
+ * @property {string} pid
+ * @property {Array.<(span: InstanaBaseSpan) => void>} [extraTests]
+ */
+
+/**
+ * @param {HttpRootEntryOptions} options
+ * @returns {InstanaBaseSpan}
+ */
+// eslint-disable-next-line
+exports.verifyHttpRootEntry = function verifyHttpRootEntry({ expect, spans, apiPath, pid, extraTests }) {
```
basti1302, line 50: Meant to ask that in the last review, but I forgot: Why do these functions receive `expect` as a parameter. It should be the same expect function all the time. Why do we not just import `expect` here in this file?

basti1302 (approved):
This looks great, just one small question left. 👍

packages/core/test/test_util/common_verifications.js
willianpc: That completely escaped from my mind. I will do this change :D
Thanks!

PR #343, opened 2021-06-08
feat(tracing): redact embedded credentials from URLs

### LINKED ISSUE #327: Automatically mask basic-auth url credentials
In the context of a node runtime, some application dependencies that are using a legacy basic-auth in the URL may get exposed in (plain-text) within an account. 

For example, a call such as: 
`https://username:password@some.domain.com/api/health `

May expose the `username:password`.  We'd like a regex to detect this well known pattern and mask the sensitive information in the agent. This is already happening within other runtimes (java/go) 

Instana pointed to this area for a fix: 
https://github.com/instana/nodejs-sensor/blob/main/packages/core/src/tracing/instrumentation/protocols/httpClient.js#L137

basti1302:
That is, remove the user:password part from URLs like http://user:password@example.org.

Closes #327.


packages/core/src/tracing/instrumentation/protocols/http2Client.js
```diff
@@ -12,10 +12,7 @@ const constants = require('../../constants');
 const httpCommon = require('./_http');
 const readSymbolProperty = require('../../../util/readSymbolProperty');
 const tracingUtil = require('../../tracingUtil');
-const urlUtil = require('../../../util/url');
-
-const discardUrlParameters = urlUtil.discardUrlParameters;
-const filterParams = urlUtil.filterParams;
+const { filterParams, sanitizeUrl } = require('../../../util/url');
```
basti1302, line 15: remark: `discardUrlParameters` has been renamed to `sanitizeUrl` because it now discards query parameters and also redacts credentials embedded into the URL.

basti1302:
This is currently work in progress, unit tests for `sanitizeUrl` still need to be added.

packages/core/src/util/url.js
```diff
@@ -7,11 +7,22 @@
...
+
 /**
  * @param {string} url
  * @returns {string}
  */
-exports.discardUrlParameters = function discardUrlParameters(url) {
+exports.sanitizeUrl = function sanitizeUrl(url) {
+  // Remove embedded credentials like the user:password part in https://user:password@example.org.
+  // Do RegExp#test before doing RegExp#exec because a non-match is the hot path.
+  if (embeddedCredentialsRegex.test(url)) {
```
willianpc, line 19: Personally, in general I would go for the standard lib URL module when it comes to parse URL.
But I don't spot anything wrong with the RegExp either, so the solution looks great. It takes me a moment to understand the regexp, but that's all :)

willianpc (approved):
LGTM

packages/core/src/util/url.js
basti1302: Oh, yes, that's absolutely right. Don't know why I didn't think of that before. I'll rebuild this with `url`, this will be much safer/less error-prone 👍

PR #347, opened 2021-06-23
Memcached connection property

willianpc:
One small leftover caught while discussions something else with Cedric.


basti1302 (approved):
LGTM

PR #349, opened 2021-06-29
Type checking for the metrics module and its dependencies

willianpc:
This PR adds types to the shared metrics module.

A couple of notes:

 * Typescript will automatically require typing of dependencies of the files we want to add types, which requires typing of several unplanned files to be typed. For this reason:
     *  The use of `*` (any) was necessary to prevent an even longer propagation of undesirable files to be typed
     * The same was also true for a few additions of `@ts-ignore`
     * It is already expected that many types we are adding will have to be revisited anyway, once we have 90% or more type checking coverage in our code, so this is not a problem
     * The initial goal of this PR was to add types to the public API, but due to the unexpected type requirement propagation, the PR was limited to shared metrics
 * As this PR was becoming too big, the work will be split into type checking modules. Hopefully we can type check the public API in the next PR, but it may be necessary to split it by other dependencies again


packages/shared-metrics/src/libuv.js
```diff
@@ -9,9 +9,13 @@ const path = require('path');
 
 const lag = require('event-loop-lag')(1000);
 
+/** @type {*} */
 let eventLoopStats;
 
 exports.payloadPrefix = 'libuv';
+
+// This line can probly be removed. On the other hand, this is a 6 year old code, so we should review it anyway.
```
basti1302, line 17: @willianpc I don't really understand what this comment is about.

basti1302 (approved):
LGTM.

packages/shared-metrics/src/libuv.js
willianpc: Sorry, that was not clear at all, and I may either remove the comment or make it as a TODO and add more info.

So, TS complains that this object is read only (check line 40), thus, it cannot be assigned a value.
What I had in mind was that maybe we could review the code together at some point and see if the `defineProperty` is still needed, and if so, if we can remove the "forbidden" assignment of `{}` to the object.
For the moment, I am just ignoring the warning and trying to focus on the typing only. Even though I changed some code, I try my best to avoid it at this stage.

Is it more clear?

basti1302: Absolutely 👍   Thanks for the explanation.

PR #350, opened 2021-06-30
chore(ci): pull the mssql server image from mcr.microsoft.com

basti1302:
Apparently Microsoft has removed its images from Dockerhub and moved to
their own registry. While we are at it, also update to the latest
available mssql image.


willianpc (approved):
LGTM

PR #354, opened 2021-07-05
chore(metrics): add pre-compiled binaries for darwin to source controls

basti1302:
The pre-compiled binaries are only removed from .gitignore and added to source control. They will still not be included in the package released to the npm registry(compare packages/shared-metrics/package.json#files).

Having them in .gitignore made the test packages/collector/test/nativeModuleRetry/test.js fail unless you had built the precompiled binaries locally first.


package.json
```diff
@@ -26,7 +26,8 @@
     "update-deps": "lerna exec '../../node_modules/.bin/ncu --upgrade' && lerna exec 'npm i --package-lock-only' && npm run reinstall-deps",
     "test": "lerna run test --stream",
     "test:debug": "lerna run test:debug --stream",
-    "lint": "node misc/is_eol || (tsc && lerna run lint --stream && eslint misc && eslint dockerfile-examples)",
+    "lint": "node misc/is_eol || (npm run typecheck && lerna run lint --stream && eslint misc && eslint dockerfile-examples)",
+    "typecheck": "tsc",
```
basti1302, line 30: This is an unrelated change I had lying around locally. The only thing it changes is that type checking via tsc is now available as a separate npm script.

willianpc (approved):
LGTM

PR #355, opened 2021-07-05
update glob-parent and other dependencies

basti1302:
deps: run "npm run update-deps" to update dependencies

In particular, this updates glob-parent (dependency of eslinst) to 5.1.2
in all packages to address a bogus dependabot alert.


willianpc (approved):
LGTM

PR #357, opened 2021-07-12
small w3c fixes

basti1302:


willianpc (approved):
Types!!! Thanks for that!
and, LGTM

PR #358, opened 2021-07-13
AWS SQS test improvements

willianpc:
### Motivation
Our tests randomly fail many times around SQS tests. So we wanted to at least reduce the number of fails

### Potential Cause

SQS consumes messages in its own way, meaning that we may request a number of messages to be retrieved, but a smaller amount is retrieved instead. 

On top of that, we test messages sent in a batch, causing an eventual "leak" of messages from a test to the next, compromising the expected result.

A similar problem happens when we test SQS Consumer, which almost always leaves some messages unread or messages are not deleted as expected. These messages are consumed in the next CI build, which may also compromise the expected results.

### Solution
Each form of message has its own queue now:
 * `team_nodejs`: single messages sent by our sendMessage test app and received by our receiveMessage app
 * `team_nodejs-batch`: same as above, but messages are sent in batches
 * `team_nodejs-consumer`: messages are sent by our sendMessage app and received by SQS Consumer.

Additionally:
 * Queues are always deleted at the end of the test
 * In case of an interrupted test (preventing queue deletion) each queue is purged before the start of the test

### Additional Info

 * Both queue deletion and queue purge take 60 seconds to be completed, even though the promise returns in a few seconds after triggering one of these operations. The AWS backend will throw an error if a new purge or a new queue creation is called within these 60 seconds
 * The usage of uuid to generate queue names was discarded at this point, as that would potentially cause unnecessary queues to exist in case of a failure while deleting them at the end of tests. eg: a new push upstream was done, so the current test is aborted, leaving existing queues without deletion.

### DIFF
--- a/packages/collector/test/tracing/cloud/aws-sdk/v2/s3/app.js
+++ b/packages/collector/test/tracing/cloud/aws-sdk/v2/s3/app.js
@@ -19,13 +19,7 @@ const s3 = new AWS.S3();
 const app = express();
 const port = process.env.APP_PORT || 3215;
 
-function log() {
-  /* eslint-disable no-console */
-  const args = Array.prototype.slice.call(arguments);
-  args[0] = `${logPrefix}${args[0]}`;
-  console.log.apply(console, args);
-  /* eslint-enable no-console */
-}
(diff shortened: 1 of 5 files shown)


basti1302 (approved):
LGTM

PR #360, opened 2021-07-14
 feat(tracing): trace correlation for SNS-to-SQS

basti1302:


packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/sendNonInstrumented.js
```diff
@@ -16,15 +16,15 @@ exports.sendMessageWithLegacyHeaders = function sendMessageWithLegacyHeaders(que
       MessageBody: 'message sent via callback function',
       QueueUrl: queueURL,
       MessageAttributes: {
-        X_INSTANA_ST: {
+        X_InSTaNa_St: {
```
basti1302, line 19: ...just to make sure we read the incoming headers case insensitive

packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/test.js
```diff
@@ -138,20 +138,29 @@ mochaSuiteFn('tracing/cloud/aws-sdk/v2/sqs', function () {
...
-          await sendMessageWithLegacyHeaders(queueURL, '1234', '5678');
-          return verifyLegacy();
+          const traceId = '1234';
+          const spanId = '5678';
+          await sendMessageWithLegacyHeaders(queueURL, traceId, spanId);
+          return verifySingelSqsEntrySpanWithParent(traceId, spanId);
         });
 
-        function verifyLegacy() {
-          return retry(async () => {
```
basti1302, line 146: renamed to verifySingelSqsEntrySpanWithParent

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v2/sqs.js
```diff
@@ -355,53 +356,103 @@ function instrumentedReceiveMessage(ctx, originalReceiveMessage, originalArgs) {
...
-}
-
 /**
- * Flattens the AWS SQS MessageAttribute format into a basic object structure.
+ * Reads all trace context relevant message attributes from an incoming message and provides them in a normalized format
+ * for later processing.
  */
-function convertAttributesFromSQS(sqsAttributes) {
-  const attributes = {};
+function readTracingAttributes(sqsAttributes) {
```
basti1302, line 363: This works a little different now than before. Instead of copying/flattening all incoming attributes and using that flattened copy, we now only read the three attributes we care about and put them into a normalized structure of the shape: `{traceId, parentId, level}`.  

All code that is executed after that then only needs to work with that uniform structure and does not need to care about legacy header names etc.

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v2/sqs.js
```diff
@@ -355,53 +356,103 @@ function instrumentedReceiveMessage(ctx, originalReceiveMessage, originalArgs) {
...
-  const spanT = readAttribCaseInsensitive(attributes, sqsAttributeNames.TRACE_ID, sqsAttributeNames.LEGACY_TRACE_ID);
-  const spanP = readAttribCaseInsensitive(attributes, sqsAttributeNames.SPAN_ID, sqsAttributeNames.LEGACY_SPAN_ID);
-
-  if (spanT) {
-    span.t = spanT;
-  }
-
-  if (spanP) {
-    span.p = spanP;
+  if (tracingAttributes.traceId && tracingAttributes.parentId) {
```
basti1302, line 453: This actually changes the behavior slightly. Before, if only one of traceId or parentId was present, we would use that single attribute. Actually, a trace ID without a parent ID or vice versa is invalid and should be ignored IMO, so now this check only succeeds if both traceID and parentID are there. In practice, this should not make any difference, because any well behaved upstream tracer will either send both IDs or no ID at all.

willianpc (approved):
LGTM

packages/aws-lambda/src/triggers.js
```diff
@@ -290,7 +290,14 @@ function readTraceCorrelationFromSqs(event) {
...
 
     const sqsMessageBody = event.Records[0].body;
-    if (typeof sqsMessageBody === 'string' && sqsMessageBody.startsWith('{')) {
+    // Parsing the message body introduces a tiny overhead which we want to avoid unless we are sure that the incoming
+    // message actually has tracing attributes. Thus some preliminary, cheaper checks are executed first.
+    if (
+      typeof sqsMessageBody === 'string' &&
+      sqsMessageBody.startsWith('{') &&
+      sqsMessageBody.includes('"Type":"Notification"') &&
+      /"X_INSTANA_/i.test(sqsMessageBody)
```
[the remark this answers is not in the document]
basti1302: Good point, I'll move it to a constant in module scope (for both places where this is defined). 👍

PR #363, opened 2021-07-16
feat(lambda): W3C trace context support for Lambda tracing

basti1302:


packages/aws-lambda/test/spec_compliance/lambda.js
```diff
@@ -0,0 +1,33 @@
...
+// assumed root dir of the Lambda.
+require('../../src/metrics/rootDir').root = require('path').resolve(__dirname, '..', '..');
+
+const fetch = require('node-fetch');
+
+const downstreamDummyUrl = process.env.DOWNSTREAM_DUMMY_URL;
+
+exports.handler = instana.wrap(async event => {
+  const { headers, httpMethod, path } = event;
+  console.log(`-> ${httpMethod} ${path} ${JSON.stringify(headers)}`);
```
willianpc, line 24: are these `console.log`s intended to stay here?

packages/aws-lambda/test/spec_compliance/lambda.js
```diff
@@ -0,0 +1,33 @@
...
+  const downstreamResponse = await fetch(downstreamDummyUrl);
+  const downstreamResponseBody = await downstreamResponse.json();
```
willianpc, line 26: Is there any chance that one of these `await` statements could be rejected?

willianpc (approved):
LGTM.
Added two questions out of curiosity. Unless the await statement can be rejected, which in this case you may consider adding a try/catch around them :)

packages/aws-lambda/test/spec_compliance/lambda.js
basti1302: They should not fail but if they do, packages/aws-lambda/test/runtime_mock#runHandler will catch that error and the test will fail with a nice, clean error message.

packages/aws-lambda/test/spec_compliance/lambda.js
basti1302: Removed them.

PR #367, opened 2021-08-06
fix(deps): upgrade tar

gebhardtr:
Per CVEs


basti1302:
Hi Richard,

which CVE are you referring to? Do you have a link?

Note that the dependencies declared in the package.json file in the root directory are only used for development purposes, they are not part of the packages published to the npm registry. The dependencies of the released packages are determined by the package.json files in the subdirectory of packages dir.

That said, there is a dependency to node-tar in [packages/shared-metrics/package.json](https://github.com/instana/nodejs/blob/59ce34dd0a944f4ed3f59f3e19b03f42912839d8/packages/shared-metrics/package.json#L62). We could bump this to ^5.0.7 (which is mentioned in this CVE: https://nvd.nist.gov/vuln/detail/CVE-2021-32803, not sure if that is the one you are referring to?).

But to be honest, it hardly matters. The dependency is already declared as a version range using a caret, so installing 5.0.7 is already possible. Also, that CVE does not apply at all to our usage of tar.

gebhardtr:
Thanks @basti1302!

Yes,

https://nvd.nist.gov/vuln/detail/CVE-2021-32804
https://nvd.nist.gov/vuln/detail/CVE-2021-32803

I'll update all references, thanks again.

basti1302 (approved):
LGTM

basti1302:
I've moved the commits over to https://github.com/instana/nodejs/pull/368. Our CI is currently configured to not run builds from forks and I would like one green CI run before merging.

Thanks for this!

basti1302:
This has landed in 1.130.1.

PR #368, opened 2021-08-09
fix(deps): upgrade tar

basti1302:
Motivated by
* https://nvd.nist.gov/vuln/detail/CVE-2021-32804
* https://nvd.nist.gov/vuln/detail/CVE-2021-32803

Note that those CVEs are actually not applicable to the usage of
node-tar in any Instana package, versions are solely bumped to avoid
false positives in automatic dependency checks.


basti1302:
Originally provided via https://github.com/instana/nodejs/pull/367.

basti1302:
This has landed in 1.130.1.

PR #376, opened 2021-08-25
chore: add issue template

basti1302:


basti1302:
* Compare https://github.com/instana/go-sensor/issues/new/choose to get a good idea on the UX (I mostly copied their configs)
* A rough preview of the bug template: https://github.com/instana/nodejs/blob/issue-template/.github/ISSUE_TEMPLATE/bug.yaml (rendering is not 100% accurate though)

PR #379, opened 2021-08-30
fix(tracing): fix vendored version of emitter-listener

basti1302:
Use different property names as the original to avoid conflicts even
when the unfixed version of emitter-listener is still installed by other
packages (like cls-hooked for example) and wraps an emitter after we
have wrapped it.

### DIFF
--- a/packages/core/src/tracing/clsHooked/emitter-listener.js
+++ b/packages/core/src/tracing/clsHooked/emitter-listener.js
@@ -4,6 +4,20 @@
  *
  * Original license: BSD 2 Clause
  * (see https://github.com/othiym23/emitter-listener/blob/7586fba839cf87774d5df224ce479c3b7e2b9964/package.json#L26)
+ *
+ * A note on the usage of the properties __wrappedInstana, __unwrapInstana, __wrapped  and __unwrap: The original
+ * package (https://github.com/othiym23/emitter-listener/blob/af156cef522f1daa11523b2d94b6f7fd1e859ca5/listener.js)
+ * attaches properties named __wrapped and __unwrap to the event emitter object. Even with this fixed and vendored
+ * version of emitter-listener, the issue fixed in https://github.com/othiym23/emitter-listener/pull/7 can still
+ * persist if other packages install the unfixed emitter-listener package as a dependency, and if the unfixed version
+ * of wrapEmitter gets called _after_ this version.
+ *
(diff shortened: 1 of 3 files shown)


packages/core/src/tracing/clsHooked/emitter-listener.js
```diff
@@ -4,6 +4,17 @@
  *
  * Original license: BSD 2 Clause
  * (see https://github.com/othiym23/emitter-listener/blob/7586fba839cf87774d5df224ce479c3b7e2b9964/package.json#L26)
+ *
+ * A note on the usage of the properties __wrappedIn, __unwrapIn, __wrapped  and __unwrap: The original package
+ * (https://github.com/othiym23/emitter-listener/blob/af156cef522f1daa11523b2d94b6f7fd1e859ca5/listener.js) attaches
+ * properties named __wrapped and __unwrap to the event emitter object. Even with this this fixed and vendored
```
willianpc, line 10: `this` is written twice in a row. Not sure if it was intentional as the sentence would more or less still make sense with or without them.

packages/core/test/tracing/clsHooked/emitter-listener_test.js
```diff
@@ -129,7 +129,7 @@ describe('emitter-listener', () => {
     expect(ee.removeListener.__wrapped, 'removeListener is not wrapped').to.not.exist;
 
     expect(function () {
-      ee.__unwrap();
+      ee.__unwrapIn();
```
willianpc, line 132: Not sure if `In` is supposed to mean "in something" or an abbreviation of "Instana".
If that is the case, I would suggest that we actually write "Instana" to avoid interpretation of "in something".
If it really means "in", what exactly do you mean by that? In what?

willianpc (approved):
LGTM.
I have two considerations, both related to interpretation in a comment and about method naming.
As none of them is a deal breaker, I am approving the PR.

packages/core/test/tracing/clsHooked/emitter-listener_test.js
basti1302: Good point, it was actually meant as an abbreviation of Instana, but I see how that is confusing. I'll change it accordingly.

packages/core/src/tracing/clsHooked/emitter-listener.js
basti1302: That was a typo, thanks. Fixed now.

PR #380, opened 2021-09-01
chore(profiling): add missing pre-compiled darwin binary

basti1302:
In commit 66ed11a7ff95a861c1e63958b476770f4c2bff52 the pre-compiled
binaries for darwin in packages/shared-metrics were added to source
control and the corresponding packages/shared-metrics/addons/.gitignore
file was removed. However, I failed to also remove
packages/autoprofile/addons/.gitignore, so when we rebuilt binaries for
Node.js 16/ABI version 93 we missed the fact that the latest Darwin
build for autoprofile was not under source control.

Note that the pre-compiled darwin binaries folder is only removed from
.gitignore and the missing file for ABI version 93 is added to source
control. Pre-compiled binaries for Darwin will not be included in the
package released to the npm registry
(see packages/autoprofile/package.json#files). To keep packages small,
only pre-compiled binaries for Linux (glibc and muslc) are included in
released packages.

Note: The missing precompiled binary made the test
packages/collector/test/profiling/test.js and tests in
packages/autoprofile fail on Darwin under Node.js 16, unless you had
build the native add-on locally first.


basti1302:
Gnaaa. Build is failing with 

```
@instana/core: > npm audit --production
@instana/autoprofile: # npm audit report
@instana/autoprofile: tar  <=4.4.17 || 5.0.0 - 5.0.9 || 6.0.0 - 6.1.8
@instana/autoprofile: Severity: high
@instana/autoprofile: Arbitrary File Creation/Overwrite via insufficient symlink protection due to directory cache poisoning using symbolic links - https://npmjs.com/advisories/1779
@instana/autoprofile: Arbitrary File Creation/Overwrite via insufficient symlink protection due to directory cache poisoning using symbolic links - https://npmjs.com/advisories/1780
```

This is a new vulnerability only reported yesterday. We need to update the tar dependency but I would do that in a separate PR.

kirrg001 (approved):
Tests are green now :) LGTM

PR #381, opened 2021-09-01
fix(tests): added missing pg host to run-tests.sh script

kirrg001:
no issue

- POSTGRES_HOST is "localhost" by default and won't match the expections e.g. [here](https://github.com/instana/nodejs/blob/v1.132.2/packages/collector/test/tracing/database/pg_native/test.js#L289)
- If you have not exported this env variable in your shell manually, you will run into pg assertion problems

Works in Circle CI because it is defined in the config file.


basti1302 (approved):
LGTM

PR #382, opened 2021-09-02
fix(security): updated tar dependency

kirrg001:
no issue

- npm audit failed for all node versions
- 2 packages effected

> @instana/autoprofile: # npm audit report
> @instana/autoprofile: tar  <=4.4.17 || 5.0.0 - 5.0.9 || 6.0.0 - 6.1.8
> @instana/autoprofile: Severity: high
> @instana/autoprofile: Arbitrary File Creation/Overwrite via insufficient symlink protection due to directory cache poisoning using symbolic links - https://n> pmjs.com/advisories/1779
> @instana/autoprofile: Arbitrary File Creation/Overwrite via insufficient symlink protection due to directory cache poisoning using symbolic links - https://n> pmjs.com/advisories/1780
> @instana/autoprofile: fix available via `npm audit fix`
> @instana/autoprofile: node_modules/tar

**FYI Cannot update node-gyp to 8.x because of python2 drop. See conversation**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -238,6 +238,7 @@
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/@aws-crypto/ie11-detection/-/ie11-detection-1.0.0.tgz",
       "integrity": "sha512-kCKVhCF1oDxFYgQrxXmIrS5oaWulkvRcPz+QBDMsUr2crbF4VGgGT6+uQhSwJFdUAQ2A//Vq+uT83eJrkzFgXA==",
+      "dev": true,
       "requires": {
         "tslib": "^1.11.1"
       },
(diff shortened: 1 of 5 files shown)


packages/autoprofile/package.json
```diff
@@ -33,7 +33,7 @@
     "detect-libc": "^1.0.3",
     "nan": "^2.14.2",
     "node-abi": "^2.30.0",
-    "node-gyp": "^7.1.2"
+    "node-gyp": "^8.2.0"
```
basti1302, line 36: I'm not 100% sure if we want to do this because node-gyp 8 removes support for Python 2 and only supports Python 3 going forward. Of course, Python 2 is EOL since over 18 months or so but I could imagine that some systems out there are still on Python 2.

Is the move to node-gyp 8 strictly necessary to get rid of the audit warning for node-tar? If so, we can consider it. Since `@instana/autoprofile` is an optional dependency of `@instana/collector`, it should not break anything anyway.

kirrg001: Wasn't aware of python 2. That's defo too breaky. Something for a new major.
I'll push a different fix in some mins

kirrg001:
npm audit passed for all versions as far as I can see. Hope it passes the tests too. We'll see 👯

kirrg001:
@basti1302 Tests passed. 14 & 16 failed firstly, just re-run them and now they pass. Probably just a random failure.

basti1302 (approved):
LGTM, thanks! 🙇

PR #383, opened 2021-09-02
fix(docs): added missing jq dependency to contributor requirements

kirrg001:
no issue

- `lerna run audit --stream` failed because of a missing global dependency

@basti1302 @willianpc Open for discussion where to put the requirement 👍

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,8 @@
 # Contributing
 
-## Local Set Up
+## Requirements
+
+`brew install jq` for OSX (for other systems please look up [here](https://stedolan.github.io/jq/)) is required to run `npm run audit` or `lerna audit run`.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,5 +1,9 @@
 # Contributing
 
+## Requirements
+
+- `brew install jq` for OSX (for other systems please look up [here](https://stedolan.github.io/jq/)) is required to run `lerna audit run`
```
basti1302, line 5:
```suggestion
- `brew install jq` for OSX (for other systems please look up [here](https://stedolan.github.io/jq/)) is required to run `npm run audit` or `lerna audit run`.
```

Looks good to me. Do we want to pull up the section about the postgres packages (first paragraph in "Local Set Up") up here, feels like both should be listed under "Requirements" if we have this section. WDYT?

basti1302 (approved):
LGTM

CONTRIBUTING.md
kirrg001: Yeah was wondering about this as well. Let's combine the sections and simply use "Requirements".

PR #385, opened 2021-09-03
chore(ci): removed circle ci build slot reduction

kirrg001:
refs https://github.com/instana/nodejs/commit/343f69a765160306f1db2deaa9926b5cc9b82420
refs https://github.com/instana/nodejs/commit/52981fa5eb84b4ab4f586e3ef57c3f3f85229276

We jumped on a bigger CircleCI plan. The restrictions of max five concurrent jobs is no longer in place, so actually this run either Node 8 or Node 10 shenanigans are no longer needed — we are no longer blocking other teams with our test suite (which we did, before changing the plan).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,30 +144,6 @@ shared: &shared
     # error.
     - run: npm run lint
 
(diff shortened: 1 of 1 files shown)


basti1302 (approved):
LGTM

PR #386, opened 2021-09-03
test(refactor): moved env variables to collector package

kirrg001:
no issue

Reasons:
- env variables had to be declared in 2 places
- `cd packages/collector && npm run test` fails if you havent set the env variables manually in shell
- simply moved them to an init file for the collector test env for now (could use dotenv later if there are more use cases)


packages/collector/test/initEnv.js
```diff
@@ -0,0 +1,35 @@
+/*
```
willianpc, line 1: +1. for DRYing stuff <3

.circleci/config.yml
```diff
@@ -21,6 +21,10 @@ shared: &shared
     - POSTGRES_PASSWORD=nodepw
     - POSTGRES_DB=nodedb
     - MEMCACHED=localhost:11211
+    - MSSQL_HOST=localhost
+    - MSSQL_PORT=1433
+    - MSSQL_USER=sa
+    - MSSQL_PW=stanCanHazMsSQL1
```
kirrg001, line 27: These were missing at all in the circle ci config :)

.circleci/config.yml
```diff
@@ -86,7 +90,7 @@ shared: &shared
     # For more recent Node.js versions (>= 10) we use the latest 1.x as stated in package.json. We can't use grcp@1.10.1
     # in Node.js 10 since it fails to build under Node.js >= 10. We still want to explicitly test the combination of
     # Node.js 8 with grpc@1.10.1 for reasons.
-    - run: '([[ $(node -v) =~ ^v8.*$ ]] && npm install grpc@1.10.1) || true'
+    - run: "([[ $(node -v) =~ ^v8.*$ ]] && npm install grpc@1.10.1) || true"
```
kirrg001, line 93: Consistency styling update

basti1302 (approved):
LGTM

.circleci/config.yml
basti1302: Probably not required because the default values work for CircleCI, but +1 to add them here as well for consistency. 👍

basti1302 (approved):

.circleci/config.yml
kirrg001: FYI Tests failed in CircleCi because of the missing envs :)

basti1302: Ah, right! Now that we set them in the code if they are not present, they are absolutely required here. Didn't think that through properly 👍

PR #387, opened 2021-09-07
Opentelemetry exporter

willianpc:
Creating the PR in case someone wants to check the current status / progression


packages/collector/src/agentConnection.js
```diff
@@ -377,6 +377,11 @@ function sendData(path, data, cb, ignore404) {
           statusCodeError.statusCode = res.statusCode;
           cb(statusCodeError);
           return;
+        } else {
+          const statusCodeError = new Error(res.statusMessage);
```
basti1302, line 381: This `else` branch looks like it breaks the `ignore404` mechanism. Is this still needed now that the OTel exporter uses `backend_conector` from `@instana/serverless`? What's the reason behind this change?

packages/opentelemetry-exporter/src/index.ts
```diff
@@ -0,0 +1,211 @@
...
+      f: {
+        e: process.pid
+      },
+      data: {}
+    };
+
+    const traceId = this._buildTraceOrSpanId(span.spanContext().traceId, true);
+
+    if (traceId.length === 32) {
+      result.t = traceId.substr(16);
```
basti1302, line 126: We need to set the additional attribute `span.lt` when we limit the trace ID from 32 characters to 16 characters; we need to preserve the original long trace ID in `span.lt` (*l*ong *t*race ID), see here https://github.com/instana/agent/blob/master/agent-optional-parent/agent-open-telemetry/src/main/java/com/instana/agent/main/impl/otel/OpenTelemetryTraceEndpoint.java#L137-L138 and https://github.com/instana/nodejs/blob/main/packages/core/src/tracing/tracingHeaders.js#L364-L370.

packages/opentelemetry-exporter/src/index.ts
```diff
@@ -0,0 +1,211 @@
...
+      // always uses the IDs from the traceparent header, thus we mark the span with span.tp accordingly.
+      result.tp = true;
+    }
+
+    result.data.service = span.resource && span.resource.attributes && span.resource.attributes['service.name'];
+    result.data.operation = span.name;
+
+    if (span.spanContext().traceState) {
+      // TODO: should we stringify and parse this guy?
+      result.data.trace_state = span.spanContext().traceState;
```
basti1302, line 176: I see that this is in the agent code, too, but I wonder why we even capture the tracestate header in full? We should discuss that with Miel and the others I think. 🤔

packages/opentelemetry-exporter/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "keywords": [
+    "opentelemetry",
+    "exporter",
+    "nodejs",
+    "tracing"
+  ],
+  "author": "Willian Carvalho <willian.carvalho@instana.com>",
+  "homepage": "https://github.com/instana/nodejs/blob/main/packages/opentelemetry-exporter/README.md",
+  "license": "MIT",
+  "main": "lib/opentelemetry-exporter.js",
```
basti1302, line 14: Looks like the output folder is configured to be `dist` so I think this needs to be `lib/opentelemetry-exporter.js`.

packages/opentelemetry-exporter/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "directories": {
+    "lib": "src",
+    "test": "test"
+  },
+  "files": [
+    "src"
```
basti1302, line 20: How do we make sure the `dist` folder is included in the released package? Should we add the `dist` folder to the `files` array?

Hint/Note: We could and should test the packaging aspects by running `npm pack` locally and inspecting the resulting tar.gz file plus running `npm install` with that tar.gz file in a temporary directory before actually publishing to the npm registry.

packages/opentelemetry-exporter/src/index.ts
```diff
@@ -0,0 +1,211 @@
...
+    if (endpointUrl) {
+      process.env[instanaEndpointUrlEnvVar] = endpointUrl;
+    }
+
+    if (agentKey) {
+      process.env[instanaAgentKeyEnvVar] = agentKey;
+    }
+
+    instanaEnvironment.validate();
+    instanaBackendConnector.init();
```
basti1302, line 90: I think we should provide an `identityProvider` here to make sure `span.f` and the x-instana-host header are set correctly in backend_connector.

packages/serverless/src/backend_connector.js
```diff
@@ -105,7 +110,11 @@ exports.sendMetrics = function sendMetrics(metrics, callback) {
 };
 
 exports.sendSpans = function sendSpans(spans, callback) {
-  send('/traces', spans, false, callback);
+  try {
+    send('/traces', spans, false, callback);
```
basti1302, line 114: Just for my curiosity: why would `send` throw a synchronous exception? And if it potentially does, we need to handle that in all `sendXxx` methods or put the try-catch directly into `send`. But I actually think `send` should not throw an exception in the first place.

basti1302 (commented):
👋  left some early comments.

packages/collector/src/agentConnection.js
willianpc: I will remove the PR as this is WIP and many things are not yet concluded

packages/opentelemetry-exporter/package.json
willianpc: > How do we make sure the dist folder is included in the released package

`"prepublishOnly": "npm run build"`

> Hint/Note: We could and should test the packaging aspects by running npm pack locally and inspecting the resulting tar.gz file plus running npm install with that tar.gz file in a temporary directory before actually publishing to the npm registry.

Ok, that's a very good point. Thanks for the hint!

packages/opentelemetry-exporter/src/index.ts
willianpc: True. I missed that info when I started "converting" the Java code into JS.
Thanks!

packages/serverless/src/backend_connector.js
willianpc: Otel fails silently with probably some empty `catch`. I added try/catch in our code temporarily to be able to see these silent errors.

packages/opentelemetry-exporter/package.json
willianpc: Good catch, thanks.
I still did not decide if I will keep TS in there. 
It doesn't make a lot of difference, and it is just "out of the standard" of our code.
So the `main` property will change anyways :)

packages/collector/src/agentConnection.js
willianpc: But it can be removed. I thought this was something that was missing in the code anyways, because `sendData` gives a callback with no errors when there was actually an error.
I will just rollback the changes.

packages/opentelemetry-exporter/package.json
basti1302: My point was not about the prepublish script but about which files will be included in the package and I think dist is missing in the `files` array in packge.json. But I get it now, this is all more preliminary.

PR #388, opened 2021-09-08
Opentelemetry exporter

willianpc:
## OpenTelemetry Exporter

A few remarks:

1. The current implementation of OpenTelemetry seems not to be able to properly instrument `node-fetch`, so I changed my test app to use `request-promise` instead
2. During some debug session, I found out that OpenTelemetry calls some Google API internally. Probably to collect usage data. Just something that I thought worth sharing
3. We agreed on copying `uninstrumentedHttp.js` from the `core` package to the `serverless` package in order to remove its dependency from the core package

### DIFF
--- a/packages/opentelemetry-exporter/.ncurc.json
+++ b/packages/opentelemetry-exporter/.ncurc.json
@@ -1,0 +1,3 @@
+{
+  "reject": ["mocha", "eslint-plugin-mocha"]
+}
--- a/packages/opentelemetry-exporter/README.md
+++ b/packages/opentelemetry-exporter/README.md
@@ -1,0 +1,76 @@
+# @instana/opentelemetry-exporter
+
+An [OpenTelemetry exporter](https://opentelemetry.io/docs/js/exporters/) to Instana specific span format.
+
+## Installation
+
+    $ npm i --save @instana/opentelemetry-exporter
+
+## Instana and OpenTelemetry Versions
+
+Even though the [Instana Node.js SDK supports several versions of Node.js](https://www.instana.com/docs/ecosystem/node-js/#supported-nodejs-versions),
+users of the Instana Exporter must take into consideration the
+[versions supported by the OpenTelemetry instrumentator](https://github.com/open-telemetry/opentelemetry-js#supported-runtimes).
+
+Making sure that the Node.js version used in your application fulfills both Instana Exporter and OpenTelemetry SDK versions
+is particularly important for Instana customers who wish to migrate from the Instana Collector to OpenTelemetry.
+
+## Usage
+
+The Instana Node.js OpenTelemetry exporter for serverless works just like any OpenTelemetry exporter.
+Once you have your application properly set to be monitored by the OpenTelemetry SDK, the injected tracing module
+expects an exporter. In our case, the Instana exporter must be imported and instantiated to be used by the
+OpenTelemetry tracing.
+
+The code sample below demonstrates how the tracing module could look like with the Instana exporter:
+
+```javascript
+'use strict';
+
+const process = require('process');
+require('@opentelemetry/api');
+const opentelemetry = require('@opentelemetry/sdk-node');
+const { getNodeAutoInstrumentations } = require('@opentelemetry/auto-instrumentations-node');
+const { Resource } = require('@opentelemetry/resources');
+const { SemanticResourceAttributes } = require('@opentelemetry/semantic-conventions');
+
+// Import the Instana OpenTelemetry Exporter
+const { InstanaExporter } = require('@instana/opentelemetry-exporter');
+
+// Instantiate the Instana Exporter.
+// Make sure to provide the proper agent key and backend endpoint URL.
+// You can provide the agent key and backend endpoint URL via the following environment variables:
+// * INSTANA_AGENT_KEY
+// * INSTANA_ENDPOINT_URL
+//
+// Alternatively, you can pass these values as an argument to the constructor (see bellow), although it is strongly
+// recommended that such sensitive data is not hard coded in the code base.
+// Eg: const instanaTraceExporter = new InstanaExporter({ agentKey: 'agent_key', endpointUrl: 'endpoint_url' });
+const instanaTraceExporter = new InstanaExporter();
+
+const nodeAutoInstrumentations = getNodeAutoInstrumentations();
+
+const sdk = new opentelemetry.NodeSDK({
+  resource: new Resource({
+    [SemanticResourceAttributes.SERVICE_NAME]: 'my-service'
+  }),
+  // Configure OpenTelemetry to use the Instana Exporter
+  traceExporter: instanaTraceExporter,
+  instrumentations: [nodeAutoInstrumentations]
+});
+
+sdk
+  .start()
+  .then(() => console.log('Tracing initialized'))
+  .catch(error => console.log('Error initializing tracing', error));
+
+process.on('SIGTERM', () => {
+  sdk
+    .shutdown()
+    .then(() => console.log('Tracing terminated'))
+    .catch(error => console.log('Error terminating tracing', error))
+    .finally(() => process.exit(0));
+});
+```
+Assuming that your main application is already importing the tracing module, this is all you have to do.
+Your spans will be properly exported from your OpenTelemetry tracer to the Instana backend.
--- a/packages/opentelemetry-exporter/package.json
+++ b/packages/opentelemetry-exporter/package.json
@@ -1,0 +1,64 @@
+{
+  "name": "@instana/opentelemetry-exporter",
+  "version": "1.133.0",
+  "description": "OpenTelemetry Instana Exporter",
+  "keywords": [
+    "opentelemetry",
+    "exporter",
+    "nodejs",
+    "tracing"
+  ],
+  "author": "Willian Carvalho <willian.carvalho@instana.com>",
+  "homepage": "https://github.com/instana/nodejs/blob/main/packages/opentelemetry-exporter/README.md",
+  "license": "MIT",
+  "main": "src/index.js",
+  "directories": {
+    "lib": "src",
+    "test": "test"
+  },
+  "files": [
+    "src"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/instana/nodejs.git"
+  },
+  "scripts": {
+    "start": "node test/app.js",
+    "debug": "node --inspect-brk test/app.js",
+    "test": "echo \"$(node -v)\" | grep v8 > /dev/null || npm run test:mocha",
+    "test:mocha": "mocha --sort --reporter mocha-multi --reporter-options spec=-,xunit=../../test-results/opentelemetry-exporter/results.xml $(find test -iname '*test.js' -not -path '*node_modules*')",
+    "test:debug": "WITH_STDOUT=true npm run test:mocha",
+    "audit": "bin/prepare-audit.sh && npm audit --production; AUDIT_RESULT=$?; git checkout package-lock.json; exit $AUDIT_RESULT",
+    "lint": "eslint src test",
+    "verify": "npm run lint && npm test",
+    "prettier": "prettier --write 'src/**/*.js' 'test/**/*.js'"
+  },
+  "bugs": {
+    "url": "https://github.com/instana/nodejs/issues"
+  },
+  "dependencies": {
+    "@instana/serverless": "1.133.0",
+    "@opentelemetry/core": "0.25.0"
+  },
+  "devDependencies": {
+    "@instana/core": "1.133.0",
+    "eslint-plugin-mocha": "^9.0.0",
+    "mocha": "^7.2.0",
+    "chai-spies": "1.0.0",
+    "@opentelemetry/resources": "0.25.0",
+    "@opentelemetry/sdk-trace-base": "0.25.0",
+    "@opentelemetry/semantic-conventions": "0.25.0",
+    "@opentelemetry/node": "0.24.0",
+    "@opentelemetry/auto-instrumentations-node": "0.24.0",
+    "@opentelemetry/sdk-node": "0.24.0",
+    "@opentelemetry/instrumentation-http": "0.24.0",
+    "@opentelemetry/instrumentation": "0.24.0",
+    "@opentelemetry/sdk-trace-node": "0.24.1-alpha.4",
+    "@opentelemetry/api": "1.0.3",
+    "eslint": "^7.30.0",
+    "eslint-config-airbnb-base": "^14.2.1",
+    "eslint-plugin-import": "^2.23.4",
+    "prettier": "^2.3.2"
+  }
+}
--- a/packages/opentelemetry-exporter/src/index.js
+++ b/packages/opentelemetry-exporter/src/index.js
@@ -1,0 +1,296 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2021
+ */
+
+'use strict';
+
+const { backendConnector: instanaBackendConnector, environment: instanaEnvironment } = require('@instana/serverless');
+const { ExportResultCode } = require('@opentelemetry/core');
+const { diag } = require('@opentelemetry/api');
+const instanaEndpointUrlEnvVar = 'INSTANA_ENDPOINT_URL';
+const instanaAgentKeyEnvVar = 'INSTANA_AGENT_KEY';
+
+/** @typedef {import('@opentelemetry/sdk-trace-base').ReadableSpan} ReadableSpan */
+/** @typedef {import('@instana/core/src/tracing/cls').InstanaBaseSpan} InstanaBaseSpan */
+
+/** Span Kind
+  Default value. Indicates that the span is used internally.
+  INTERNAL = 0,
+
+  Indicates that the span covers server-side handling of an RPC or other
+  remote request.
+  SERVER = 1,
+
+  Indicates that the span covers the client-side wrapper around an RPC or
+  other remote request.
+  CLIENT = 2,
+
+  Indicates that the span describes producer sending a message to a
+  broker. Unlike client and server, there is no direct critical path latency
+  relationship between producer and consumer spans.
+  PRODUCER = 3,
+
+  Indicates that the span describes consumer receiving a message from a
+  broker. Unlike client and server, there is no direct critical path latency
+  relationship between producer and consumer spans.
+  CONSUMER = 4
+*/
+const SPAN_KIND_INTERNAL = 0;
+const SPAN_KIND_SERVER = 1;
+const SPAN_KIND_CLIENT = 2;
+const SPAN_KIND_PRODUCER = 3;
+const SPAN_KIND_CONSUMER = 4;
+const SERVER_KIND = 'server';
+const CLIENT_KIND = 'client';
+const PRODUCER_KIND = 'producer';
+const CONSUMER_KIND = 'consumer';
+const INTERNAL_KIND = 'internal';
+/**
+ * https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L314
+ * Status Code
+ *
+ * Default
+ * UNSET = 0,
+ * The operation has been validated by an Application developer or
+ * Operator to have completed successfully.
+ * OK = 1,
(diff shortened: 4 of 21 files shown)


packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2021
+ */
+
+'use strict';
+
+const { backendConnector: instanaBackendConnector, environment: instanaEnvironment } = require('@instana/serverless');
+const { ExportResultCode } = require('@opentelemetry/core');
```
willianpc, line 9: The dependency from the `opentelemetry/core` is very small, as I am only using the enum of result codes.
I wonder if we could hard code these values to remove this dependency, having the risk that the default enum values can change.
On the other hand, a customer would probably be adding `opentelemetry/core` anyways. 
Having that said, we could also maybe use peerDependencies.

Any thoughts?

packages/opentelemetry-exporter/.ncurc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "reject": ["mocha", "eslint-plugin-mocha"]
+}
```
kirrg001, line 3: What is ncurc.json? :)

basti1302: https://github.com/raineorshine/npm-check-updates#configuration-files

See also https://github.com/instana/nodejs/blob/main/package.json#L31

packages/opentelemetry-exporter/README.md
```diff
@@ -0,0 +1,67 @@
+# @instana/opentelemetry-exporter
```
kirrg001, line 1: https://github.com/open-telemetry/opentelemetry-js#supported-runtimes

Maybe something to mention in our docs too? We support 6 & 8, Otel doesn't officially.

packages/opentelemetry-exporter/src/index.js
basti1302: Oof, good question. If we specify this as a dependency in package.json, we also prescribe a version number. Currently we pin it to 0.25.0. That is, if a customer uses 0.24.0 or 0.26.0, it would be installed two times anyway. Even if we go with a range, that still might happen.

Thus, the risk that the enum values that we use and the ones that the customer's app is using diverge exists no matter what we do, I don't see a way around that. But it should hopefully be a rather unlikely risk. 

I really don't have a strong opinion, but I would tend to leave the dependency in and not use a peer dependency.

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+const instanaEndpointUrlEnvVar = 'INSTANA_ENDPOINT_URL';
+const instanaAgentKeyEnvVar = 'INSTANA_AGENT_KEY';
+
+/** @typedef {import('@opentelemetry/sdk-trace-base').ReadableSpan} ReadableSpan */
+/** @typedef {import('@instana/core/src/tracing/cls').InstanaBaseSpan} InstanaBaseSpan */
+
+// The exporter is based on the parser from the Agent:
+// eslint-disable-next-line max-len
+// https://github.com/instana/agent/blob/51da097e99e9d05ac5c0b5b7dd9275a12c20685d/agent-optional-parent/agent-open-telemetry/src/main/java/com/instana/agent/main/impl/otel/OpenTelemetryTraceEndpoint.java#L69
```
basti1302, line 18: _thought:_ This is a link to a private repo from our public code base. There is nothing wrong with that in principle, but it feels a bit odd to me, just saying.

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+    // If endpoint URL and agent key are not provided, we don't set them in an attempt to use env vars, if they were set
+    if (endpointUrl) {
+      process.env[instanaEndpointUrlEnvVar] = endpointUrl;
+    }
+    if (agentKey) {
+      process.env[instanaAgentKeyEnvVar] = agentKey;
+    }
+
+    if (!process.env[instanaEndpointUrlEnvVar] || !process.env[instanaAgentKeyEnvVar]) {
+      throw new Error('The Instana endpoint URL and/or the agent key were not provided');
```
basti1302, line 85: I wonder if we should just put the exporter into a non-active mode instead of crashing the process. There are valid arguments for both behaviors, so it is a bit a matter of taste. The other serverless collectors (for Lambda, Fargate etc.) simply do nothing if the required configuration is not in place.

People often integrate observability packages (like `@instana/collector` or `@instana/opentelemetry-exporter` but only use them in certain environments (like production) but not when testing locally, for example. Crashing the process when the env vars are not present is quite inconvenient in that scenario.

`instanaEnvironment.validate();` (called below) already logs warnings if either of the two or both are missing.

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    instanaEnvironment.validate();
+    instanaBackendConnector.init();
```
basti1302, line 89: _question:_ Do we want to pass in an `identityProvider` here to set a different `Host` header than `nodejs-serverless`?

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+    this._isShutdown = true;
+    return Promise.resolve();
+  }
+
+  /**
+   * @param {Array.<InstanaBaseSpan>} spans
+   * @param {(result: import('@opentelemetry/core').ExportResult) => void} resultCallback
+   */
+  _sendSpans(spans, resultCallback) {
+    instanaBackendConnector.sendSpans(spans, (err /* , _responseBody */) => {
```
basti1302, line 122: The backend connector's init method would need to be called with `propagateErrorsUpstream = true` otherwise the `sendSpans` method will not pass errors to the given callback. (default for propagateErrorsUpstream is false)

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+      n: 'otel',
+      f: {
+        e: process.pid
+      },
+      data: {}
+    };
+
+    const traceId = this._buildTraceOrSpanId(span.spanContext().traceId, true);
+
+    if (traceId.length === 32) {
```
basti1302, line 148: In reality, there should only be lengths of 16 or 32, but we still could check `(traceId.length >= 16)` instead of `(traceId.length === 32)` to be extra sure.

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    const hasParent = typeof span.parentSpanId === 'string' && span.parentSpanId.length > 0;
+
+    if (hasParent) {
+      result.p = this._buildTraceOrSpanId(span.parentSpanId);
+    }
+
+    result.s = this._buildTraceOrSpanId(span.spanContext().spanId);
+    result.ts = this._getTimeInMicroseconds(span.startTime);
```
basti1302, line 162: We use milliseconds. The method name suggests that _getTimeInMicroseconds returns microseconds. Do we need to convert? Or maybe the method name is misleading and it actually returns milliseconds?

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+    result.d = this._getTimeInMicroseconds(span.duration);
```
basti1302, line 163: same here, we might need conversion between µs and ms.

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+      // If an OTel entry span continues an ongoing trace (which is equivalent to the original span having a parent), it
+      // always uses the IDs from the traceparent header, thus we mark the span with span.tp accordingly.
+      result.tp = true;
+    }
+
+    result.data.service = span.resource && span.resource.attributes && span.resource.attributes['service.name'];
+    result.data.operation = span.name;
+
+    if (span.spanContext().traceState) {
+      // TODO: should we stringify and parse this guy?
```
basti1302, line 198: I don't think we should.

packages/shared-metrics/package-lock.json
```diff
@@ -1,3213 +1,3275 @@
...
-			"integrity": "sha512-W9tKgmSn0DpSatfri0nx52Joq5hVXgeLiqR/5G0sZNDoLZFOr/xjBUDcShCOGNsBnEMNo1KAMBkTej1Hm62HTw==",
-			"dev": true,
-			"requires": {
-				"flat": "^4.1.0",
-				"lodash": "^4.17.15",
-				"yargs": "^13.3.0"
-			}
-		}
-	}
+  "name": "@instana/shared-metrics",
```
basti1302, line 2: This diff comes from a mistake in an earlier PR. Please rebase this PR on main and discard all changes to this file in this PR.

test-suite-ports.md
```diff
@@ -16,4 +16,4 @@ For this reason, we maintain an overview of the ports used by the integration te
 | metrics-util         | N.A. (no integration tests, only unit tests) |
 | serverless           | N.A. (no integration tests, only unit tests) |
 | shared-metrics       | app under test: 7215, mock agent: 7211 |
-
+| opentelemetry-exporter | app under test: 3215, mock back end: 10443 |
```
basti1302, line 19: Port 3215 is already in use by `collector`, can we use a different port?

packages/opentelemetry-exporter/test/test.js
```diff
@@ -0,0 +1,115 @@
...
+
+  describe('Communication to the backend', () => {
+    const backendPort = 10433;
+    const appControls = new Control({
+      backendPort,
+      startBackend: true,
+      otelAppPath: './test/app',
+      env: {
+        INSTANA_DISABLE_CA_CHECK: 'true',
+        PORT: 3215,
```
basti1302, line 79: Please use a different port, 3215 is already in use by the `collector` package.

basti1302 (changes requested):
Great work! I added a couple of comments and request for changes inline.

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+    if (hasParent) {
+      result.p = this._buildTraceOrSpanId(span.parentSpanId);
+    }
+
+    result.s = this._buildTraceOrSpanId(span.spanContext().spanId);
+    result.ts = this._getTimeInMicroseconds(span.startTime);
+    result.d = this._getTimeInMicroseconds(span.duration);
+    let isEntrySpan = false;
+
+    switch (span.kind) {
```
kirrg001, line 166: I was testing an example Otel Instrumention ([pg](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node)).

I received 4 spans (3x express + 1x pg) in the backend.
The span kind for pg query is "ENTRY". I thought this has to be EXIT. Maybe I am just confused :)

packages/opentelemetry-exporter/README.md
willianpc: I am not entirely sure to be honest.
Essentially, this package is a span exporter that actually depends on the OTel SDK, so being compatible to their supported runtimes is something that their users inevitably have to respect anyway.
The only dependency from our stack is actually the communication to the backend, through the `@instana/serverless` package.
I see no harm at mentioning the version compatibility, but as far as I can see, it doesn't mean much.

kirrg001: My thinking was that our customers start to use Otel and might use an older version (as a hint for our customers).
But maybe that is not a use case for our customers. Also not sure :)

packages/opentelemetry-exporter/src/index.js
willianpc: Good catch. I will pay more attention before adding links to the code.
We should not have private repo links in the a public repo.

packages/opentelemetry-exporter/src/index.js
willianpc: Makes total sense. I guess I added the error in the very beginning of the coding and planned to changed later, buuuut, changes never came

packages/opentelemetry-exporter/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "url": "https://github.com/instana/nodejs/issues"
+  },
+  "dependencies": {
+    "@instana/serverless": "1.133.0",
+    "@opentelemetry/core": "0.25.0"
+  },
+  "devDependencies": {
+    "@instana/core": "1.133.0",
+    "eslint-plugin-mocha": "^9.0.0",
+    "mocha": "^7.2.0",
```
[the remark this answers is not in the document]
basti1302: There was a reason for having mocha and eslint in each package but I am not sure what the reason was :-) We can try to clean this up, but I'd it is out of scope for this PR.

packages/opentelemetry-exporter/src/index.js
basti1302: But on the other hand: Do we even want to report an error to OTel's resultCallback? What would be the consequence.

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+
+  /**
+   * @param {Array.<ReadableSpan>} spans
+   * @param {(result: import('@opentelemetry/core').ExportResult) => void} resultCallback
+   */
+  export(spans, resultCallback) {
+    const validSpans = spans.map(this._transform.bind(this));
+
+    if (!this._isShutdown) {
+      this._sendSpans(validSpans, resultCallback);
```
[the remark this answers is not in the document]
basti1302: If I understand this correctly, we would behave more or less like the exporter you linked to because we forward any error into the resultCallback.

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
...
+   * @returns {InstanaBaseSpan}
+   */
+  _transform(span) {
+    /** @type {import('@instana/core/src/tracing/cls').InstanaBaseSpan} */
+    const result = {
+      n: 'otel',
+      f: {
+        e: process.pid
+      },
+      data: {}
```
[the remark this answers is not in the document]
basti1302: Discussed offline: A part of the conversion from OTel format to Instana happens in the back end, so naturally the JSON we send will look a bit different from traditional Instana spans.

packages/opentelemetry-exporter/src/index.js
kirrg001: My bad. I was sending wrong data to the backend while I was playing with the code locally. Sorry!

packages/opentelemetry-exporter/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+const process = require('process');
+require('@opentelemetry/api');
+const opentelemetry = require('@opentelemetry/sdk-node');
+const { getNodeAutoInstrumentations } = require('@opentelemetry/auto-instrumentations-node');
+const { Resource } = require('@opentelemetry/resources');
+const { SemanticResourceAttributes } = require('@opentelemetry/semantic-conventions');
+
+// Import the Instana OpenTelemetry Exporter
+const { InstanaExporter } = require('@instana/opentelemetry-exporter');
```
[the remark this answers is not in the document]
basti1302: Discussed offline: We could discuss a retry behavior for backendConnector at some point, but this is certainly out of scope for this PR.

packages/opentelemetry-exporter/package.json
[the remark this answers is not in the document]
willianpc: Good question.

My understanding of monorepo projects is that even though lerna manages dependencies in the root project, the sub project should still have all dependencies in its own package.json (the story is slightly different for package-lock.json.
This is because later on, when the package is published , lerna will no longer be present, so the package "is on its own".


But when it comes to dev dependencies, well, this may or may not be true. Most people will just use packages and I don't think many people would install dev dependencies or even run tests themselves, so having dev dependencies in the package.json may not make sense.

Does this make sense? Or am I missing something?
- Cheers

packages/opentelemetry-exporter/src/index.js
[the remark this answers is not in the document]
willianpc: Yes it does. And it is not always very nice from an exporter implementer point of view, because the exporter silently fail into an empty `catch` or something.
I had to add temporary extra `try/catch` statements into our code to debug things sometimes.

packages/opentelemetry-exporter/README.md
willianpc: I am adding a small section with links to both our docs and OTel docs where supported versions are described.
I also add a note concerning Instana customers who wish to switch to Otel.

packages/opentelemetry-exporter/README.md
```diff
@@ -0,0 +1,67 @@
...
+process.on('SIGTERM', () => {
+  sdk
+    .shutdown()
+    .then(() => log('Tracing terminated'))
+    .catch(error => log('Error terminating tracing', error))
+    .finally(() => process.exit(0));
+});
+```
+Assuming that your main application is already importing the tracing module, this is all you have to do.
+Your spans should be properly exported from your OpenTelemetry tracer to the Instana backend.
```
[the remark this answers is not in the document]
willianpc: Thank you! This is more important to me than you can imagine

packages/opentelemetry-exporter/src/index.js
willianpc: As per our talk on Slack, this is not a deal breaker, so we can discuss this further on.
Will keep as is for now :)

packages/opentelemetry-exporter/src/index.js
willianpc: I will add this test. But from my experience so far, the exporter always fails silently to prevent the original code to crash. This is why I had to add temporary try/catch statements when I started the creation of the exporter.
I don't quite understand what the error reported via `resultCallback` does, but definitely does not crash the application.

packages/opentelemetry-exporter/src/index.js
willianpc: So, this bit of code was pretty much copied from the Java code, and actually the traceId is always 32 hex strings long (see `_buildTraceOrSpanId`).
This verification could actually be removed, since it is guaranteed that the traceId will be 32 hex strings no matter what.

Does this make sense?
^ @basti1302

packages/opentelemetry-exporter/src/index.js
willianpc: I am using the right function now, and with the right name.
Thanks for point it out!

test-suite-ports.md
willianpc: Good catch. 
I also added the collector to the list with the port 3215.

packages/opentelemetry-exporter/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+sdk
+  .start()
+  .then(() => log('Tracing initialized'))
+  .catch(error => log('Error initializing tracing', error));
+
+process.on('SIGTERM', () => {
+  sdk
+    .shutdown()
+    .then(() => log('Tracing terminated'))
```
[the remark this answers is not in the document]
willianpc: Nice catch! I replaced it by `console.log`
Thanks!

packages/opentelemetry-exporter/src/index.js
```diff
@@ -0,0 +1,288 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2021
+ */
+
+'use strict';
+
+const { backendConnector: instanaBackendConnector, environment: instanaEnvironment } = require('@instana/serverless');
```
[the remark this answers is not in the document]
willianpc: @kirrg001 The backend connector accepts any logger. I am passing `diag` from Otel to it.
Thanks!

packages/opentelemetry-exporter/src/index.js
basti1302: Ah, I didn't see that. Makes sense, yes. 👍

basti1302 (approved):
This already looked great before the additional changes, looks even better now. 

This is ready to be merged IMO! 🎉 🎉 👍 👍

kirrg001 (approved):

PR #389, opened 2021-09-09
fix(test): make version check more lenient

basti1302:
Previously, this test would break when one of the dependencies would
release a new version.


kirrg001 (approved):
LGTM 👯

willianpc (approved):
Thanks for the quick fix!

PR #390, opened 2021-09-13
test(feat): added ability to upload local core modification to lambda

kirrg001:
no issue

- [x] tested on lambda & release rocks

@basti1302 As you said on Friday removing "INSTANA_DISABLE_LAMBDA_EXTENSION" will still send data to the target backend is absolutely true. I am not sure why I had trouble in the first try, maybe because I was just confused? If you want to I can add this env to the deploy script. Let me know :)

Please merge both commits and do not squash :)

### DIFF
--- a/packages/aws-lambda/.gitignore
+++ b/packages/aws-lambda/.gitignore
@@ -2,4 +2,5 @@ node_modules/
 instana-aws-lambda.tgz
 instana-aws-lambda-auto-wrap.tgz
 instana-serverless.tgz
+instana-core.tgz
 npm-debug*
--- a/packages/aws-lambda/lambdas/bin/create-zip-util
+++ b/packages/aws-lambda/lambdas/bin/create-zip-util
@@ -25,6 +25,7 @@ function createZip {
       #   "instana-aws-lambda-auto-wrap": "file:../../instana-aws-lambda-auto-wrap.tgz"
       # }
       # ...then those need to be removed:
+      npm uninstall --production -S @instana/core
       npm uninstall --production -S @instana/aws-lambda
       npm uninstall --production -S @instana/serverless
       npm uninstall --production -S instana-aws-lambda-auto-wrap
@@ -37,6 +38,13 @@ function createZip {
       # }
       # but only doing an npm install will use a stale copy of the tgz files from npm's cache, so we force npm to use a
       # fresh copy by installing the tgz files again explicitly.
+      #
+      # Furthermore we have to ensure that instana packages, which uses other instana packages, also will use the local version.
+      # We can achieve that by installing our local core version first and ensuring npm will use the root installation for sub packages
+      # We have to remove the lock file otherwise the integrity will fail.
+      rm package-lock.json
+
+      npm --loglevel=warn --production install -S "../../instana-core.tgz"
       npm --loglevel=warn --production install -S "../../instana-aws-lambda-auto-wrap.tgz"
       npm --loglevel=warn --production install -S "../../instana-aws-lambda.tgz"
       npm --loglevel=warn --production install -S "../../instana-serverless.tgz"
@@ -48,8 +56,10 @@ function createZip {
       #   "@instana/aws-lambda": "file:../../instana-aws-lambda.tgz"
       # }
       # this needs to be removed and then we install the latest published package from npm
+      npm uninstall --production -S @instana/core
       npm uninstall --production -S @instana/aws-lambda
       npm uninstall --production -S @instana/serverless
+      npm install --production -S @instana/core
       npm install --production -S @instana/aws-lambda
       npm install --production -S @instana/serverless
     else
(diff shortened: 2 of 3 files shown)


packages/aws-lambda/lambdas/wrapped-async/package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "@instana/aws-lambda": "../../instana-aws-lambda.tgz",
     "@instana/serverless": "../../instana-serverless.tgz",
+    "@instana/core": "../../instana-core.tgz",
```
kirrg001, line 18: There are 2 more lambda folders which make use of the generated tgz files. Should I add the core package there too? @basti1302

packages/serverless/package-lock.json
```diff
@@ -72,6 +72,21 @@
...
+				"async-hook-jl": "^1.7.6",
+				"cls-bluebird": "^2.1.0",
+				"lru-cache": "^6.0.0",
+				"methods": "^1.1.2",
+				"opentracing": "^0.14.5",
+				"redis-commands": "^1.7.0",
+				"semver": "7.3.5",
+				"shimmer": "1.2.1"
+			}
+		},
```
basti1302, line 89: *issue:* This declaration is wrong here. package-lock.json files in the individual packages must not declare dependencies to other `@instana` packages as those are managed by lerna. I would suggest to revert this file.

basti1302 (changes requested):
Thanks for fixing this. Mostly LGTM but there is one (probably accidental) change in the package-lock.json file of the serverless package that needs to be fixed before merging.

.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 npm-debug*
 packages/collector/test/bin/edgemicroResources/cli/edgemicro
 nodejs-license-report.csv
+.DS_Store
```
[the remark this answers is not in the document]
kirrg001: Interesting. Wasn't aware of this technique :) Thanks. Will change

packages/aws-lambda/.gitignore
```diff
@@ -2,4 +2,6 @@ node_modules/
 instana-aws-lambda.tgz
 instana-aws-lambda-auto-wrap.tgz
 instana-serverless.tgz
-npm-debug*
\ No newline at end of file
+instana-core.tgz
+npm-debug*
+lambdas/zip/*
```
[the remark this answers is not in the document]
kirrg001: 👍  I'll test that again :)

packages/aws-lambda/lambdas/bin/create-zip-util
```diff
@@ -37,6 +38,13 @@ function createZip {
       # }
       # but only doing an npm install will use a stale copy of the tgz files from npm's cache, so we force npm to use a
       # fresh copy by installing the tgz files again explicitly.
+      #
+      # Furthermore we have to ensure that instana packages, which uses other instana packages, also will use the local version.
+      # We can achieve that by installing our local core version first and ensuring npm will use the root installation for sub pacakges
```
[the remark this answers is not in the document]
kirrg001: Whooops

packages/serverless/package-lock.json
kirrg001: Good catch! Was probably touched when I was adding symbolic links in the beginning of my test.
🙈

packages/aws-lambda/.gitignore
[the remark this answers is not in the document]
kirrg001: I'll removed it. Works without. I don't remember why I had trouble with this file. 
Thanks!

packages/aws-lambda/lambdas/wrapped-async/package.json
kirrg001: @basti1302 Any opinion on that? :)

Packages are: just-creates-spans and timeout-detection

They don't seem to me like they would need to use a local core modification, but I am not sure

kirrg001:
Resolved comments 👍

packages/aws-lambda/lambdas/wrapped-async/package.json
basti1302: That's a good question and I am not 100% sure, but I think the changes in `create-zips.sh` should be good enough. If we do this right, it should not matter which `@instana` dependencies are declared in the package.json files in the individual lambdas (or how they are declared), because that is completely managed by the script when deploying the Lambdas. WDYT?

kirrg001: Yeah I think we don't need the core dependency here as e.g. aws-lambda uses it as sub dependency, which then should use the local core installation because of the `create-zips.sh` modification.
I'll re-test and remove that line here :)

kirrg001: Works!Removed it

basti1302 (approved):

PR #396, opened 2021-09-29
Development dependencies / EOL alert

willianpc:


packages/shared-metrics/src/dependencies.js
```diff
@@ -269,3 +310,83 @@ function sortByDistance(distances, dependency1, dependency2) {
...
+    });
+  }
+}
+
+/**
+ * @param {string} mainPackageJsonPath
+ * @param {Object.<string, *>} packages
+ */
+function detectInstalledDevDependencies(mainPackageJsonPath, packages) {
+  if (downstreamConnection) {
```
kirrg001, line 340: Could you please add a note here that serverless is using shared metrics package but downstreamConnection will be null and that's why it will be skipped. Thanks :)

packages/shared-metrics/src/dependencies.js
```diff
@@ -269,3 +310,83 @@ function sortByDistance(distances, dependency1, dependency2) {
...
+            for (let i = 0; i < devDepKeys.length; i++) {
+              if (devDepKeys[i] in packages) {
+                installedDevDeps.push(devDepKeys[i]);
+              }
+            }
+
+            if (installedDevDeps.length > 0) {
+              downstreamConnection.sendEvent(
+                {
+                  title: 'Devopment dependencies installed',
```
kirrg001, line 361: Devopment -> Development

packages/collector/src/announceCycle/agentready.js
```diff
@@ -142,3 +146,39 @@ function fireMonitoringEvent() {
...
+      if (err) {
+        logger.debug('Node.js version EOL', err);
+      }
+    }
+  );
+}
+
+function detectEOLNodeVersion() {
+  // This verificarion must be updated from time to time
+  // Check https://endoflife.software/programming-languages/server-side-scripting/nodejs for EOL versions
```
basti1302, line 175: The page referred to in the comment looks outdated (it still lists Node.js 10 as a current release). We should link to https://nodejs.org/en/about/releases/ instead, just as in the issue description. We could also throw in another link to https://github.com/nodejs/Release#end-of-life-releases.

An additional comment to also update the EOL check in packages/collector/test/agentCommunication_test.js when updating this one might be helpful for our future selves.

packages/shared-metrics/src/dependencies.js
```diff
@@ -269,3 +310,83 @@ function sortByDistance(distances, dependency1, dependency2) {
...
+      if (readFileErr) {
+        logger.debug(`Error reading ${mainPackageJsonPath}`, readFileErr);
+        cb(readFileErr, null);
+      } else {
+        mainPackageJsonContents = contents;
+        cb(null, mainPackageJsonContents);
+      }
+    });
+  }
+}
```
basti1302, line 333: _suggestion:_ We already have facilities to read and JSON-parse the main package.json files in applicationUnderMonitoring.js: https://github.com/instana/nodejs/blob/d8c48eaa9aa78e9fde2d680f841627ec5a583fac/packages/core/src/util/applicationUnderMonitoring.js#L69-L80

It is currently a part of the `getMainPackageJsonStartingAtDirectory` that can not be used independently. Could you extract lines 69-80 of applicationUnderMonitoring into its own exported method and use it here?

packages/shared-metrics/src/dependencies.js
```diff
@@ -269,3 +310,83 @@ function sortByDistance(distances, dependency1, dependency2) {
...
+        try {
+          const { devDependencies } = JSON.parse(data.toString());
+
+          if (devDependencies) {
+            const installedDevDeps = [];
+            const devDepKeys = Object.keys(devDependencies);
+
+            for (let i = 0; i < devDepKeys.length; i++) {
+              if (devDepKeys[i] in packages) {
+                installedDevDeps.push(devDepKeys[i]);
```
basti1302, line 354: _remark:_ Theoretically, this could lead to false positives. Assuming the application uses package-aaa as a devDependency and package-bbb as a production dependency, but package-bbb also depends transitively on package-aaa -- then this check would detect a devDependency although only production dependencies have been installed. My gut feeling is that this might be a purely academic concern, since applications typically have devDependencies with a larger scope (eslint, mocha, babel, ...) which usually should not be transitive production dependecies.

basti1302 (approved):
Great stuff! There are two minor issues that need to be addressed (marked as _issues_) IMO, otherwise looks splendid.

packages/collector/src/announceCycle/agentready.js
```diff
@@ -142,3 +146,39 @@ function fireMonitoringEvent() {
...
+      text:
+        'This version no longer receives updates or security fixes and might contain unfixed vulnerabilities.\n\n' +
+        'Please consider upgrading Node.js to an active version.\n\n' +
+        'For a list of active versions visit ' +
+        '[https://nodejs.org/en/about/releases/](https://nodejs.org/en/about/releases/)',
+      plugin: 'com.instana.forge.infrastructure.runtime.nodejs.NodeJsRuntimePlatform',
+      id: pidStore && typeof pidStore.getEntityId === 'function' ? pidStore.getEntityId() : undefined,
+      timestamp: Date.now(),
+      duration: ELEVEN_MINUTES,
+      severity: 5
```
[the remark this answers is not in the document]
willianpc: Exactly, I opted by using the type only. If you provide a value not covered by `ProblemSeverity`, `tsc` will complain.
BUT, to make it even more clear, you are right. We can have some constants defined somewhere.
I am not sure about the enum, to be fair, because I don't think we use such thing very often, or at all, maybe?
The magic number issue is technically gone, IF you use some fancy IDE or build the code (TS screaming at you). But by just batting an eye to it, it really looks ugly.
Let me review this and see if an enum fits well. Otherwise, I will just create constants somewhere (not sure where yet, as the collector package has no constants.js)

packages/collector/src/announceCycle/agentready.js
```diff
@@ -142,3 +146,39 @@ function fireMonitoringEvent() {
...
+      }
+    }
+  );
+}
+
+function detectEOLNodeVersion() {
+  // This verificarion must be updated from time to time
+  // Check https://endoflife.software/programming-languages/server-side-scripting/nodejs for EOL versions
+  const isEOL =
+    semver.lt(process.versions.node, '12.0.0') || semver.satisfies(process.versions.node, '>=13.0.0 <14.0.0');
```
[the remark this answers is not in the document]
willianpc: Thanks for catching this. I may have missed the info about v15 when I read the EOL table.
I think I will also move this test to an exported function somewhere, so when we need to update the EOL list, we don't have to do it in multiple places

packages/collector/src/announceCycle/agentready.js
willianpc: Right, indeed this page is not reliable, and nothing is supposed to beat the official release page anyway.
I will apply the changes. Thanks!

packages/shared-metrics/src/dependencies.js
willianpc: good idea. leave it to me!

packages/collector/src/announceCycle/agentready.js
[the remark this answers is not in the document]
basti1302: I really have no strong opinion here. As I said, from my perspective it is okay as it is. Constants that are centrally defined could also be very much good enough if you want to make it a bit more self-explaining.

packages/shared-metrics/src/dependencies.js
```diff
@@ -269,3 +310,83 @@ function sortByDistance(distances, dependency1, dependency2) {
...
+              }
+            }
+
+            if (installedDevDeps.length > 0) {
+              downstreamConnection.sendEvent(
+                {
+                  title: 'Devopment dependencies installed',
+                  text: `The following **devDependencies** are installed:\n\n * ${installedDevDeps.join('\n * ')}`,
```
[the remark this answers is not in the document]
willianpc: Makes total sense. Does 10 sound a reasonable number?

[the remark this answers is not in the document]
basti1302: Yes, 10 sounds good.

PR #397, opened 2021-09-29
test(frameworks): added bookshelf, sequilize and typeorm tests for parameterized queries

kirrg001:
**Background:**
Sequilize does not support parameterized queries yet. 
We have added 3 new popular frameworks to our test suite to proof how they behave with queries and parameters.
Furthermore if one of these frameworks report a support request again, we can easily check what's wrong.

- [x] finish code changes & improve the code
- [x] code quality check
- [x] CI green? (waiting...)


packages/shared-metrics/test/dependencies/test.js
```diff
@@ -126,16 +126,6 @@ describe('dependencies', function () {
           expectVersion(deps.fastify, '^3.20.2');
           expectVersion(deps.express, '^4.17.1');
           expectVersion(deps.koa, '^2.13.1');
-
```
kirrg001, line 129: Would like to discuss this change.

This test fails when you add/remove dev or prod dependencies to collector package.

I currently cannot see an advantage to check whether specific dependencies of the collector package exists or not, because it fails every time you change them, because we link the collector to the current local folder.

I think as long as we check whether @instana is inside and the deps of the customer, we are good here?

Any other opinions?

basti1302: The purpose of this check is to verify the mechanism to limit the collected dependencies to a certain number. Since the dependencies are sorted in a specific way before the limit is applied, checking that dependencies up to a specific position in that sorted list are actually included and others are not included is one way of checking whether the sorting/limiting works as intended. This comes at the cost of having to update this test when dependencies in collector change.

I currently don't have a good idea how to make it less fragile and still verifying that sorting/limiting works as intended.

In the end, removing those checks and only checking if we collected 300 dependencies is probably also good enough, though.

kirrg001: > In the end, removing those checks and only checking if we collected 300 dependencies is probably also good enough, though.

Agree 👍

basti1302 (approved):
LGTM

PR #398, opened 2021-10-04
tests(ci-cache): Fixed circle ci caching keys

kirrg001:
no issue

- we set CACHE_VERSION to **4** in circle ci env variables

- **problems**:
  - with the previous technique it was possible that pull requests update a package's cache build
    and because of the fallback cache key the main branch used this build, which can cause
    tests to fail because we have and will have tests, which check the existence of dependencies
    for e.g. customers project metrics
  - we were not able to easily purge the project cache without adding a commit + changing v3 to v4 for the circle ci config file

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,46 +33,46 @@ shared: &shared
 
     - restore_cache:
         keys:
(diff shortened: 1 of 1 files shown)


basti1302 (approved):
LGTM

PR #407, opened 2021-10-22
chore: add script to break down test suite runtimes on CI

basti1302:
refs 37526


kirrg001:
That is really cool <3

PR #411, opened 2021-10-29
fix(tracing): reject trace ID/parent span ID with all bytes zero

basti1302:
This is to be considered invalid as per
* https://www.w3.org/TR/2020/REC-trace-context-1-20200206/#trace-id and
* https://www.w3.org/TR/2020/REC-trace-context-1-20200206/#parent-id

This also adds four new test cases to the spec compliance test suite:
* malformed traceparent header
* all zeroes trace ID/parent ID
* sampled flag=0
* more than 32 tracestate list-members


willianpc (approved):
LGTM 👍🏽 
Maybe it was already covered before, but do we also need to add tests for sampled flag?

basti1302:
> LGTM 👍🏽 Maybe it was already covered before, but do we also need to add tests for sampled flag?

Oh, good catch. I've added those. I have also put this back to WIP for now since there will be another change around limiting the tracestate length, which can go into this PR.

PR #412, opened 2021-11-03
Fixed TypeError: Cannot read property 'opts' of null

willianpc:
### For context

Bull swallows errors and emits them with `on.('error')` which is why we didn't have any errors on our side. A test was added to catch these errors.

After the code was fixed, some issues showed up, probably because they were "hidden" behind the silently handled error thing. These errors were all around concurrent jobs to be processed.
They affected bulked jobs and repeatable jobs that were being processed in a short period of time (eg: less than 150 ms).

### DIFF
--- a/packages/core/src/tracing/instrumentation/messaging/bull.js
+++ b/packages/core/src/tracing/instrumentation/messaging/bull.js
@@ -10,6 +10,7 @@ const cls = require('../../cls');
 const { ENTRY, EXIT, isExitSpan } = require('../../constants');
 const requireHook = require('../../../util/requireHook');
 const tracingUtil = require('../../tracingUtil');
+const { getFunctionArguments } = require('../../../util/function_arguments');
 
 let logger = require('../../../logger').getLogger('tracing/bull', newLogger => {
   logger = newLogger;
@@ -32,10 +33,7 @@ function instrumentBull(Bull) {
 function shimJobCreate(originalJobCreate) {
   return function () {
     if (isActive) {
-      const originalArgs = new Array(arguments.length);
-      for (let i = 0; i < originalArgs.length; i++) {
-        originalArgs[i] = arguments[i];
-      }
+      const originalArgs = getFunctionArguments(arguments);
 
       return instrumentedJobCreate(this, originalJobCreate, originalArgs);
     }
@@ -80,7 +78,7 @@ function instrumentedJobCreate(ctx, originalJobCreate, originalArgs) {
     // or is root if no parent present
     const span = cls.startSpan(exports.spanName, EXIT);
     span.ts = Date.now();
-    span.stack = tracingUtil.getStackTrace(instrumentedJobCreate);
+    span.stack = tracingUtil.getStackTrace(instrumentedJobCreate, 1);
     span.data.bull = {
       sort: 'exit',
       queue: queueName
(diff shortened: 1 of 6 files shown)


packages/core/src/tracing/instrumentation/messaging/bull.js
```diff
@@ -227,9 +212,21 @@ function instrumentedProcessJob(ctx, originalProcessJob, originalArgs) {
     const spanT = attributes.X_INSTANA_T;
     const spanP = attributes.X_INSTANA_S;
 
+    const parentSpan = cls.getCurrentSpan();
+
+    if (parentSpan && parentSpan.p === spanP && parentSpan.t === spanT) {
```
basti1302, line 217: Should we also maybe check parentSpan.n here? If I read this right, we want to allow starting a new span if the parent span is also a bull span that represents a concurrent Bull job invocation. 

If that interpretation is correct, would it might make sense to add `parentSpan.n !== 'bull'` to the list of checks?

basti1302 (approved):
LGTM! 👍

packages/core/src/tracing/instrumentation/messaging/bull.js
willianpc: Good point. I think it makes sense.
I will add this verification!
Thanks

packages/core/src/tracing/instrumentation/messaging/bull.js
```diff
@@ -227,9 +212,21 @@ function instrumentedProcessJob(ctx, originalProcessJob, originalArgs) {
     const spanT = attributes.X_INSTANA_T;
     const spanP = attributes.X_INSTANA_S;
 
+    const parentSpan = cls.getCurrentSpan();
+
+    if ((parentSpan && parentSpan.p === spanP && parentSpan.t === spanT) || (parentSpan && parentSpan.n !== 'bull')) {
+      // We allow a new entry span even if there is already an active entry span, because repeatable and bulked jobs can
+      // run in parallel if concurrency when enabled.
+      // But here we check if the job parent span data is the same as the existent span. In this case, we don't
```
kirrg001, line 220: > run in parallel if concurrency when enabled.

Should it be "is enabled"?

> But here we check if the job parent span data is the same as the existent span

I wasn't able to understand the use case here. 

I had a look at bull's [codebase](https://github.com/OptimalBits/bull/blob/019d6125bea8a195d7690d52707b4e7089a92f16/lib/queue.js#L874) and it looks to me they execute `processJobs` n times (number of concurrencies) and if a job repeatable, it (including the concurrency config) is executed again at a time t.

Does that mean if concurrency is enabled, we don't want to create multiple entry spans for a specific job?

Thanks :)

willianpc: Ah, thanks. It was indeed `is`, not `when`.
This condition was transitioned from `blocking instrumentation if ANY type of span already exists` to `allow instrumenting Bull spans that are not exactly the same as the existing one, if any`.

The reason for the change is: when concurrency is enabled, repeatable and bulked jobs can be sent in parallel, depending on the number of child process are being used. When this happens, they can be received at the same time, which was causing some spans to be lost (as we were blocking ANY kind of existing span).

The condition here is more to have _some_ level of filtering in regarding of spans to be instrumented.

I added this explanation in the PR description. Maybe I wasn't clear enough.

kirrg001: I missed the removal of https://github.com/instana/nodejs/pull/412/files#diff-9e34bfdddf61bfb8623b8f23fec0279bee4aae47c0d2d2845493ba61182bcff4L166

I just noticed!

Okay now it makes sense. And thanks for the details <3

PR #414, opened 2021-11-04
(chore) Running CircleCI builds with parallelism

willianpc:
This was a tricky one, because the CircleCI documentation is fairly vague.
Finding information online is also not so easy.
Enabling parallelism is not complicated, but if you want to take the best out of it, you need to enable test splitting by time, ~and that's when things get "fun"~ which we gave up doing as the benefit wasn't as good as expected + the effort to make it work was not worthy (adding junit is needed, which would not have been a trivial change, especially because of our support to Node < 10).

Please, let me know if you have questions, and help me improve the scripts wherever you find relevant :)


package.json
```diff
@@ -118,7 +118,9 @@
     "memcached": "2.2.2",
     "mkdirp": "^1.0.4",
     "mocha": "^8.4.0",
+    "mocha-junit-reporter": "^2.0.2",
```
willianpc, line 121: We need junit because it logs file names. Junit is also recommended by CircleCI for Node.js (and probably general JS) projects.

bin/ci_glob_setup.sh
```diff
@@ -0,0 +1,27 @@
...
+export CI_CORE_TEST_FILES=$(circleci tests glob "packages/core/test/**/*test.js" | circleci tests split --split-by=timings | sed 's/packages\/core\///g') \
+export CI_GOOGLE_CLOUD_RUN_TEST_FILES=$(circleci tests glob "packages/google-cloud-run/test/**/*test.js" | circleci tests split --split-by=timings | sed 's/packages\/google-cloud-run\///g') \
+export CI_LEGACY_SENSOR_TEST_FILES=$(circleci tests glob "packages/legacy-sensor/test/**/*test.js" | circleci tests split --split-by=timings | sed 's/packages\/legacy-sensor\///g') \
+export CI_METRICS_UTIL_TEST_FILES=$(circleci tests glob "packages/metrics-util/test/**/*test.js" | circleci tests split --split-by=timings | sed 's/packages\/metrics-util\///g') \
+export CI_SERVERLESS_TEST_FILES=$(circleci tests glob "packages/serverless/test/**/*test.js" | circleci tests split --split-by=timings | sed 's/packages\/serverless\///g') \
+export CI_SHARED_METRICS_TEST_FILES=$(circleci tests glob "packages/shared-metrics/test/**/*test.js" | circleci tests split --split-by=timings | sed 's/packages\/shared-metrics\///g')
```
basti1302, line 27: Would it be worth creating these env vars in a dynamic fashion? It would work automatically when we add more packages later.

Something like this could work (not tested completely, since the circleci command is not available locally):


```
set -xeo pipefail

cd `dirname $BASH_SOURCE`/..

for package in packages/*/ ; do
  echo $package
  # remove leading "packages/"
  package_name=${package#packages/}
  # remove trailing "/"
  package_name=${package_name%/}
  # replace all "-" by "_"
  package_name=${package_name//-/_}
  # convert lower case to upper case
  package_name="${package_name^^}"

  declare CI_${package_name}_TEST_FILES=$(circleci tests glob "${package}test/**/*test.js" | circleci tests split --split-by=timings | sed "s|${package}||g" ) 
  export CI_${package_name}_TEST_FILES
done
```

basti1302 (approved):
Awesome! 👊

package.json
basti1302: Or phrased the other way around: If we want to use mocha-junit-reporter over the xunit reporter, then we should remove the xunit reporter. Having both mocha-junit-reporter and xunit seems redundant.

You mention file names as the reason for this new reporter. I agree that it would be convenient to have them, but for what exactly are they required? My guess is they are required for the split-by=timings feature, how exactly does that work. I don't see the `results-junit.xml` files being fed back into the `circleci tests split --split-by=timings` command, does that command use those files by default?

basti1302: This discussion is now obsolete since we agreed on not going for the split-by=timings variant. Too little value for too much hassle (plus additional maintenance overhead down the road)

PR #415, opened 2021-11-05
fix(dependency): updated tar to 6.x in shared-metrics

kirrg001:
refs 75234

Tar 6 has already dropped Node support < 10
See https://github.com/npm/node-tar/blob/065e3850dfcfa439fd7d4bce0407ba616f85f576/package.json#L48

We currently receive security updates in tar 5.x, but that can stop quickly, because Node 8 support has stopped in 2020 already.

We should just update tar to 6.x in shared-metrics because:

Node 8  (and lower) is officially not supported for two more packages:
- semver 7.x
- lru-cache@6.x

(and there are probably tons of more, which just don't make use of the engines keyword)

If you set `npm config set engine-strict true`, you will see that you cannot install collector
with Node 8 (without the tar update). That menas, it is already a problem.

`tar` is used in nativeModuleRetry#copyPrecompiled (https://github.com/instana/nodejs/blob/dc2160bddab5a25d0b6bab25c5d10bee8b5f1bd5/packages/shared-metrics/src/util/nativeModuleRetry.js#L75), but this piece of code is not covered in our tests.

Where is nativeModuleRetry script used? Only in shared-metrics.
- in gc.js
- in lubuv.js

Options:
1) ~Add tests for copyPrecompiled to ensure that Node 8 still works~ There is a test!
2) Disallow copyPrecompiled for Node < 10
3) Do nothing, because the execution of `tar.x` is catched in the codebase and the user won't experience anything bad. But (!) if anything goes wrong with requiring tar in Node 8 or before `tar.x` is executed and it's not catched, we can bring down the customers application. We'd need to try/catch the require statement + tar.x

`tar` is not a binary installation, that's why I'd not add it as optional dependency, because it is not possible to fail except the customer uses engine-strict.

----

Other usages of `tar`:
- in tests
- in dummy apps
- in docker examples
- in subdependencies of dev dependencies
  - these should be updated in a separate and general version update
  
## TODO
- [x] ci passes 
- ~[ ] consider adding a test!~ there is a test!
- [x] Option 2

### DIFF
--- a/packages/shared-metrics/src/libuv.js
+++ b/packages/shared-metrics/src/libuv.js
@@ -47,5 +47,7 @@ function sense() {
     stats.statsSupported = true;
     return stats;
   }
-  return {};
+  return {
+    statsSupported: false
+  };
 }
--- a/packages/shared-metrics/src/util/nativeModuleRetry.js
+++ b/packages/shared-metrics/src/util/nativeModuleRetry.js
@@ -11,8 +11,8 @@ const EventEmitter = require('events');
 const copy = require('recursive-copy');
 const fs = require('fs');
 const os = require('os');
+const semver = require('semver');
 const path = require('path');
-const tar = require('tar');
 const detectLibc = require('detect-libc');
 
 /**
@@ -171,6 +171,20 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
 
     logger.info(`Found a precompiled version for ${opts.nativeModuleName} ${label}, unpacking.`);
 
+    /**
+     * tar@6 has dropped support for Node < 10
+     * It might work to require tar@6 or to execute commands with tar@6 and Node < 10,
+     * but we don't want to risk that a customers application fails - especially if tar@6 adds
+     * breaking changes. We decided to disallow this feature.
+     */
+    if (semver.lt(process.version, '10.0.0')) {
+      logger.info(`Skipped copying precompiled version for ${opts.nativeModuleName} ${label} with Node < 10.`);
+      callback(false);
+      return;
+    }
+
+    const tar = require('tar');
+
     tar
       .x({
         cwd: os.tmpdir(),
@@ -203,10 +217,10 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
             // is, node_modules/${opts.nativeModuleName}). Node.js' module loading infrastructure
             // (lib/internal/modules/cjs/loader.js and lib/internal/modules/package_json_reader.js) have built-in
             // caching on multiple levels (for example, package.json locations and package.json contents). If Node.js
-            // has tried unsuccessfully to load a module or read a package.json from a particular path, it will remember
-            // and not try to load anything from that path again (a `false` will be put into the cache for that cache
-            // key). Instead, we force a new path, by adding precompiled to the module path and use the absolute path to
-            // the module to load it.
+            // has tried unsuccessfully to load a module or read a package.json from a particular path,
+            // it will remember and not try to load anything from that path again (a `false` will be
+            // put into the cache for that cache key). Instead, we force a new path, by adding precompiled
+            // to the module path and use the absolute path to the module to load it.
             opts.loadFrom = targetDir;
             callback(true);
           }
--- a/packages/collector/test/nativeModuleRetry/test.js
+++ b/packages/collector/test/nativeModuleRetry/test.js
@@ -7,6 +7,7 @@
 
 const async = require('async');
 const { expect } = require('chai');
+const semver = require('semver');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
@@ -38,7 +39,7 @@ describe('retry loading native addons', function () {
         let foundAtLeastOneUnsupported;
         for (let i = 0; i < allMetrics.length; i++) {
           if (allMetrics[i].data.libuv) {
-            expect(allMetrics[i].data.libuv.statsSupported).to.not.exist;
+            expect(allMetrics[i].data.libuv.statsSupported).to.be.false;
             foundAtLeastOneUnsupported = true;
             break;
           }
@@ -50,11 +51,16 @@ describe('retry loading native addons', function () {
         const libuv = aggregated.libuv;
         expect(libuv).to.exist;
         expect(libuv).to.be.an('object');
-        expect(libuv.statsSupported).to.be.true;
-        expect(libuv.min).to.be.a('number');
-        expect(libuv.max).to.be.a('number');
-        expect(libuv.sum).to.be.a('number');
-        expect(libuv.lag).to.be.a('number');
+
+        if (semver.lt(process.version, '10.0.0')) {
+          expect(libuv.statsSupported).to.be.false;
+        } else {
+          expect(libuv.statsSupported).to.be.true;
+          expect(libuv.min).to.be.a('number');
+          expect(libuv.max).to.be.a('number');
+          expect(libuv.sum).to.be.a('number');
+          expect(libuv.lag).to.be.a('number');
+        }
       }
     },
     {
@@ -67,7 +73,7 @@ describe('retry loading native addons', function () {
         let foundAtLeastOneUnsupported;
         for (let i = 0; i < allMetrics.length; i++) {
           if (allMetrics[i].data.libuv) {
-            expect(allMetrics[i].data.gc.statsSupported).to.not.exist;
+            expect(allMetrics[i].data.gc.statsSupported).to.be.false;
             foundAtLeastOneUnsupported = true;
             break;
           }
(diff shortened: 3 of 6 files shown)


kirrg001:
@basti1302 Could you please give this a pre review? Let me know if you agree or disagree :) Thanks!

packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -171,51 +170,68 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
...
-        // See below for the reason why we append 'precompiled' to the path.
-        const targetDir = path.join(opts.nativeModulePath, 'precompiled');
+    let tar;
+
+    // tar 6.x supports Node >= 10
+    // Ensure we don't bring down customers application
+    try {
+      tar = require('tar');
+    } catch (err) {
+      logger.warn('Requiring tar@6.x has failed.', err);
```
basti1302, line 180: *suggestion:* We might want to mention what the user-observable consequence of the failure to require tar@6 is.

Maybe something like 

```
logger.warn(`Using the precompiled build for ${opts.nativeModuleName} ${label} is not supported on Node.js ${process.version}.`);
```

packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -171,51 +170,68 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
...
-      .then(() => {
-        // See below for the reason why we append 'precompiled' to the path.
-        const targetDir = path.join(opts.nativeModulePath, 'precompiled');
+    let tar;
+
+    // tar 6.x supports Node >= 10
+    // Ensure we don't bring down customers application
+    try {
+      tar = require('tar');
+    } catch (err) {
```
basti1302, line 179: *question:* How about trying to require and then catching the error we explicitly check for the Node.js version and only attempt to require tar if we are on a supported Node.js version? The try-catch way probably works as well, I am not sure which one is actually better. WDYT?

packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -171,51 +170,68 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
...
+            }
+          );
+        })
+        .catch(tarErr => {
+          logger.warn(`Unpacking the precompiled build for ${opts.nativeModuleName} ${label} failed.`, tarErr);
+          callback(false);
+        });
+    } catch (err) {
+      logger.warn(`Tar extraction for precompilation has failed for ${opts.nativeModuleName}.`, err);
+      callback(false);
```
basti1302, line 233: *question:* What makes this new try-catch block around the whole extraction necessary? Is this for the case when requiring node-tar@6 on an older Node.js version actually works but calling functions from it does not? In that case, it might be an argument to simply not requiring in Node.js < 10 in the first place.

basti1302 (commented):
The general direction looks good in my opinion. I would maybe lean towards disabling the copy-precompiled feature outright in Node.js < 10, but that's totally up for discussion.

I'll will follow up with a more detailed answer to the questions and aspects you brought up [here](https://github.com/instana/nodejs/pull/415#issue-1045588841).

basti1302:
> We should just update tar to 6.x in shared-metrics because:
> 
> Node 8 (and lower) is officially not supported for two more packages:
> 
> * semver 7.x
> * [lru-cache@6.x](mailto:lru-cache@6.x)

I would leave these other two packages out of the picture for this PR. We use semver much more widely, so I think we cannot update this to a newer version right now.

By the way, when we update node-tar, we might want to remove it from https://github.com/instana/nodejs/blob/main/packages/shared-metrics/.ncurc.json as well.

> If you set `npm config set engine-strict true`, you will see that you cannot install collector with Node 8 (without the tar update). That menas, it is already a problem.

Interesting, I didn't know that. But you say it already fails to install even before updating to tar@6? That's good, so that update does not create a new problem then. Besides, people running on EOL Node.js versions can then just not make use of the engine-strict option, that seems fair enough.

> `tar` is used in nativeModuleRetry#copyPrecompiled (
> 
> https://github.com/instana/nodejs/blob/dc2160bddab5a25d0b6bab25c5d10bee8b5f1bd5/packages/shared-metrics/src/util/nativeModuleRetry.js#L75
> 
> ), but this piece of code is not covered in our tests.

I beg to differ :-) 

Unfortunately, the tests covering this are not straight forward to find. They are in the collector package, because they require the whole integration test shebang with the agent stub etc. When I wrote those tests, that test infrastructure was only available in the collector package :-/.  Actually, I _think_, we should be able to move those tests to shared-metrics now, because by now there is a precedent for having an integration test with the agent stub in the shared-metrics package -> https://github.com/instana/nodejs/blob/main/packages/shared-metrics/test/dependencies/test.js (this did not exist when the native retry feature and tests were added).

> Where is nativeModuleRetry script used? Only in shared-metrics.
> 
> * in gc.js
> * in lubuv.js

That's correct, these are the only two usages.

> 1. Add tests for copyPrecompiled to ensure that Node 8 still works

Those exist. 👍 

> 2. Disallow copyPrecompiled for Node < 10

That's what I had in mind originally, I think it is a bit more straightforward then (3.).

> 3. Do nothing, because the execution of `tar.x` is catched in the codebase and the user won't experience anything bad. But (!) if anything goes wrong with requiring tar in Node 8 or before `tar.x` is executed and it's not catched, we can bring down the customers application. We'd need to try/catch the require statement + tar.x
> I'd go with 3) and just add more catches

Also possible, assuming the catch will also catch outright "syntax errors" (because tar@6 uses syntax constructs that 
 are  not available in Node.js 8/10) when requiring it. 
 
> `tar` is not a binary installation, that's why I'd not add it as optional dependency, because it is not possible to fail except the customer uses engine-strict.

I agree with that, tar should not be an optional dependency. 👍 
 
> Other usages of `tar`:
> 
> * in tests
> * in dummy apps
> * in docker examples
> * in subdependencies of dev dependencies
>   * these should be updated in a separate and general version update

Do we explicitly require the tar package in any of our tests? If so, we might need to provide a different version of node-tar in our dev-dependencies, maybe in the root package, in case we want to keep running the tests that depend on it in older Node.js versions. To be honest, I am not sure about those other usages right now.

basti1302:
> * [x]  ci passes

Hu, that's interesting. I would have expected https://github.com/instana/nodejs/blob/main/packages/collector/test/nativeModuleRetry/test.js to fail for Node.js 8 and 10? 🤔 🤔

kirrg001:
> > * [x]   ci passes
> 
> Hu, that's interesting. I would have expected https://github.com/instana/nodejs/blob/main/packages/collector/test/nativeModuleRetry/test.js to fail for Node.js 8 then, but it didn't? 🤔 🤔

I ran the tests and the fn was not called in any of the tests. Or maybe I have missed something 🤔

kirrg001:
> That's what I had in mind originally, I think it is a bit more straightforward then (3.).

I have no strong opinion if 2 or 3. We can go with 3 :) Its defo an easier check than the try catch statements. My thinking was to try our best to still offer the feature to as many people we can.

kirrg001:
> Node 8 (and lower) is officially not supported for two more packages:
> semver 7.x
> lru-cache@6.x
> I would leave these other two packages out of the picture for this PR. We use semver much more widely, so I think we cannot update this to a newer version right now.

Definitely :) I was summing up that these two packages will make the installation of the collector fail if engine-strict is enabled.

>  I would have expected https://github.com/instana/nodejs/blob/main/packages/collector/test/nativeModuleRetry/test.js to fail for Node.js 8 then, but it didn't

Oh. I haven't seen this test. So yeah there is a test and copyPrecompiled is covered. That is good! Thanks for the hint!
I have tested Node 8 with tar 6 in a script in the beginning and it was working fine. That's why I thought option 2) is the best approach.

> Do we explicitly require the tar package in any of our tests?

No. I'd update our dev dependencies in a separate PR as part of the stability roadmap. Ok for you?

kirrg001:
Discussed: We go with option 2!

basti1302:
> > Do we explicitly require the tar package in any of our tests?
> 
> No. I'd update our dev dependencies in a separate PR as part of the stability roadmap. Ok for you?

Sure, absolutely 👍   I was just curious.

packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -11,8 +11,8 @@ const EventEmitter = require('events');
 const copy = require('recursive-copy');
 const fs = require('fs');
 const os = require('os');
+const semver = require('semver');
```
kirrg001, line 14: FYI

semver@7 does not support Node < 10 anymore.
But I ignored it for this PR because we already require semver@7 in different places (production code).
I'll check if we should protect the code against semver@7 and  lru-cache@6.x in a different PR.

packages/shared-metrics/src/libuv.js
```diff
@@ -47,5 +47,7 @@ function sense() {
     stats.statsSupported = true;
     return stats;
   }
-  return {};
+  return {
+    statsSupported: false
```
kirrg001, line 51: FYI I have added consistency changes for libuv and gc to have the `statsSupported` key always available in the metrics result. Before these optimisations the key did not exist for some cases.

Especially in this line, I tried to read the actual `exports.currentPayload.statsSupported` value, which is set on failed event, but it is impossible to access it without refactoring the code.

basti1302: The decision to not transmit statsSupported = false might have been deliberate back in the days (to save a few bytes on the wire). But I agree, it is probably not worth it and having the value always present as a boolean makes sense.

basti1302 (approved):
LGTM

PR #416, opened 2021-11-15
WIP: Poc async local storage

willianpc:
This branch has the latest changes from main from December 1st. at 10:40 so there are a huge amount of changes in it.
The only relevant change is the new file `/packages/core/src/tracing/clsHooked/async_local_storage_context.js`.
Everything else are silly changes, like improving `@ts-nocheck` comments and the activation of the new async context in `packages/core/src/tracing/clsHooked/index.js`


packages/serverless/package-lock.json
```diff
@@ -1,4191 +1,8 @@
 {
 	"name": "@instana/serverless",
 	"version": "1.137.2",
-	"lockfileVersion": 2,
+	"lockfileVersion": 1,
```
kirrg001, line 4: Have you switched to Node 16 locally? (NPM 8)
This looks wrong to me :)

packages/core/src/tracing/clsHooked/index.js
```diff
@@ -36,9 +36,14 @@ const semver = require('semver');
...
  * Node < v8 - uses AsyncWrap and async-hooks-jl
+ * Node 12.17 - 16.6 - uses native AsyncLocalStorage. See below:
+ * There is a bug introduced in Node 16.7 which breaks Async LS: https://github.com/nodejs/node/issues/40693
  * Node >= v8 - uses native async-hooks
  */
-if (process && semver.gte(process.versions.node, '8.0.0')) {
+if (process && semver.satisfies(process.versions.node, '12.17 - 16.6')) {
+  // module.exports = require('./context');
+  module.exports = require('./async_local_storage_context');
+} else if (process && semver.gte(process.versions.node, '8.0.0')) {
```
kirrg001, line 46: FYI I have not removed this code yet. If this branch get's "v2" branch is a base, I'll take care after merge to remove this check + the old context file + usage. If that is okay with you

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,365 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2021
+ */
+
+// This evolved from a copy of
+// Jeff-Lewis, fix(destroy): destroy based on asyncId only (3cf7539  on Jul 25, 2017)
+// https://github.com/Jeff-Lewis/cls-hooked/blob/066c6c4027a7924b06997cc6b175b1841342abdc/context.js
```
kirrg001, line 8: I guess we can remove this commit now?

packages/serverless/package-lock.json
willianpc: Yeah, I got lost in the version switch craziness while testing the storage in different versions.
Will fix it. Thanks!

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,361 @@
...
+    assert.ok(context, 'context must be provided for entering');
+    storage.enterWith(context);
+    this.sharedContext = context;
+  }
+
+  /**
+   * @param {InstanaCLSContext} context
+   */
+  exit(context) {
+    assert.ok(context, 'context must be provided for exiting');
```
basti1302, line 308: Should we set this.sharedContext = null here?

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,361 @@
...
+   *
+   * This function assumes that the returned promise (if any) is CLS-friendly or wrapped already.
+   * @param {Function} fn
+   * @param {InstanaCLSContext} ctx
+   */
+  runPromiseOrRunAndReturn(fn, ctx) {
+    let isPromise = false;
+    let valueOrPromise;
+    const context = ctx || this.createContext();
+    storage.enterWith(context);
```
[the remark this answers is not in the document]
willianpc: I need more investigation to prove why this is needed, as I really may have forgotten or didn't even really know, as a few changes involved trial and error, and this could be one of them.

All I can say for now is that if I remove this call to `enterWith`, there are several errors when I run the CI build. I can point that some of them happen in aws-lambda tests, but also in different places, like here: https://app.circleci.com/pipelines/github/instana/nodejs/2564/workflows/62a5d950-0bcf-49ee-841b-703d652f0f3a/jobs/11061/tests#failed-test-1

I will go back to this piece of the code and see if I can replace the `enterWith` call with a more satisfactory implementation.

Thanks

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,361 @@
...
+   * This function assumes that the returned promise is CLS-friendly or wrapped already.
+   * @param {Function} fn
+   * @param {InstanaCLSContext} ctx
+   */
+  runPromise(fn, ctx) {
+    const context = ctx || this.createContext();
+
+    // This updates the context in the promise, otherwise we return the context parent
+    // TODO: revisit this: do I need this enterWith?
+    storage.enterWith(context);
```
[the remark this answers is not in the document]
willianpc: Theoretically, there is no need to call storage.exit, as this is done automatically by the async LS.
From my experience so far, this call is really not needed, as all tests just pass without it.

About replacing individual calls to `storage.enterWith` by `this.enter`, this sounds about right. I will apply this change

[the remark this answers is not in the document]
basti1302: > Theoretically, there is no need to call storage.exit, as this is done automatically by the async LS.

You are probably right, I'm still thinking too much in terms of cls-hooked where this always needed to be symmetrical. Just ignore the comments asking for matching exit calls, they probably make no sense. And you are absolutely right, if those would be needed we would see that in the tests.

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    // const self = this;
+
+    return function clsBind() {
+      storage.enterWith(context);
+      try {
+        return fn.apply(this, arguments);
+      } finally {
+        // self.exit(context);
```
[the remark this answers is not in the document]
basti1302: See above, my bad. I probably just have not yet internalized how AsyncLocalStorage handles things enough.

packages/core/src/tracing/clsHooked/async_local_storage_context.js
basti1302: Nevermind, that would probably defeat the purpose of having it in the first place.

willianpc:
I have created a new PR, cleaner and forked from the `v2` branch.
I have added all requested changes and answered questions from you in there.
Please take a look at it when you have a chance: https://github.com/instana/nodejs/pull/430

PR #417, opened 2021-11-15
Pinned lru-cache and semver versions

kirrg001:
refs 75234

- lru cache has not released a new version after 6.0.0 yet
- lru cache 6.x has dropped support for Node < 10 ([see](https://github.com/isaacs/node-lru-cache/blob/master/package.json#L31))
- with this small improvement we just ensure that if lru-cache releases a breaking change for 6 & 8, our customers won't suffer from it
- semver has a sub dependency to lru-cache
  - see https://github.com/npm/node-semver/blob/v7.3.3/package.json#L39
  - vs https://github.com/npm/node-semver/blob/v7.3.5/package.json#L39

### DIFF
--- a/packages/autoprofile/package-lock.json
+++ b/packages/autoprofile/package-lock.json
@@ -1136,12 +1136,13 @@
 			}
 		},
 		"lru-cache": {
-			"version": "6.0.0",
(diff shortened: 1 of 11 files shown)


basti1302 (approved):
LGTM!

PR #418, opened 2021-11-16
[WIP] Clean up dev dependencies + bump versions

kirrg001:
refs 75234
refs 80052

See advantages:
https://github.com/lerna/lerna#common-devdependencies

- [ ] common dependencies
- [ ] version updates
- [ ] removal of unused dependencies 
- [ ] https://github.com/lerna/lerna/issues/3029
- [ ] https://github.com/lerna/lerna/issues/1079 (not sure about mocha being in root because it's CLI dependency?)
- [ ] --legacy-peer-deps required, because of NPM bug (usually only happens when you move deps to root)
---

before:
real	6m47.953s
user	5m49.277s
sys	3m34.443s

after:


kirrg001:
Also closing this PR after thinking about it another time.

https://github.com/lerna/lerna/issues/3029 is in our way. It is impossible to move common dependencies to the root if we always have to take care that lerna does not mess up our lock file.

On top of that we are running into https://github.com/npm/cli/issues/4084. It forces us to use legacy peer deps option, which we don't want, because this will install a wrong dependency tree. This might lead to wrong assumptions in our test env.

I think we should only do a few smaller clean up steps and then leave it as it is:

- ensure major versions (such as mocha, semver etc) of dev dependencies are equal
- general update of dev dependencies
- add some lerna dependency docs (how to use it - some workflows should be documented IMO)
-  removal of unused dependencies
- research if there is more

kirrg001:
Conclusion:

- we have the option add rules how to use lerna with common dev dependencies on root
- we look at it with a pair of new eyes (in a new PR)

PR #419, opened 2021-11-17
prepare automated release process via lerna/conventional-commits

basti1302:


CONTRIBUTING.md
```diff
@@ -110,20 +110,22 @@ Later on, the original package-lock.json is restored when `git checkout package-
...
 - Acquire an OTP token for 2fa.
 - Ensure you are logged in with npm in your shell (`npm whoami`)
-- Run either
-    - `NPM_CONFIG_OTP={your token} lerna publish --force-publish="*" patch && lerna bootstrap`, or
-    - `NPM_CONFIG_OTP={your token} lerna publish --force-publish="*" minor && lerna bootstrap`.
+- Run `NPM_CONFIG_OTP={your token} lerna publish && lerna bootstrap`.
+
+Lerna will determine if this is going to be a major, minor or patch version from the commit comments of all commits since the last release. It will also automatically update the CHANGELOG.md files in the root of the repository and also in all individual packages, based on those commits.
+
+If you want to review the changes that lerna is about to apply (for example the CHANGELOG files), you can execute `lerna version --no-git-tag-version --no-push` first.
```
basti1302, line 119: I would like to do this in two steps: Let's publish the next 2 or 3 releases still manually and if we encounter no issues there, we can switch to having it done automatically via the Github action (see below). Thus, for now, only the instructions on how to do a release are updated.

bin/needs-release.js
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env node
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2021
+ */
+
+'use strict';
```
basti1302, line 7: This file is used in the Github action to determine if a new release should be published. Otherwise we might publish releases that actually contain no real changes at all (only build changes etc.)

package.json
```diff
@@ -33,9 +33,10 @@
     "audit": "lerna run audit --stream && npm audit --production",
     "bootstrap": "lerna bootstrap",
     "bump": "lerna publish",
+    "commit": "cz",
     "doctoc": "doctoc README.md packages/collector/README.md",
-    "lint": "node misc/is_eol || (npm run typecheck && lerna run lint --stream && eslint misc && eslint dockerfile-examples)",
-    "postinstall": "npm run bootstrap",
+    "lint": "node misc/is_eol || (eslint .*.js *.js && npm run typecheck && lerna run lint --stream && eslint misc && eslint dockerfile-examples)",
```
basti1302, line 38: This change (linting additional files and directories) is unrelated to the main intent of this PR.

package.json
```diff
@@ -103,7 +107,7 @@
     "graphql": "^14.7.0",
     "graphql-subscriptions": "^1.2.1",
     "graphql-tools": "^7.0.5",
-    "grpc": "^1.24.2",
```
basti1302, line 106: I have moved grpc to optional dependencies simply because it takes ages to install (lot's of C++ code that needs to be compiled), simply to speed up the `npm install` step in the Github action (via `--no-optional`). It is quite a hack. Let me know if y'all think this is a bad idea. :-)

package.json
```diff
@@ -33,9 +33,10 @@
     "audit": "lerna run audit --stream && npm audit --production",
     "bootstrap": "lerna bootstrap",
     "bump": "lerna publish",
+    "commit": "cz",
```
basti1302, line 36: this is for `npm run commit`, which brings up an interactive CLI wizard to compose a commit message

.github/workflows/release.yaml
```diff
@@ -0,0 +1,33 @@
+# This github action automatically publishes a release when we merge to the main branch.
+# THIS IS CURRENTLY A POC. It does not yet actually publish a release, but merely performs a dry run.
+name: auto-release
+on:
+  push:
+    branches:
+      - main
```
kirrg001, line 7: **For tomorrow's meeting:**
We have the option to either let the Github action determine if a new release should happen (dangerous?) or we keep the decision within the team by merging develop branch and into main which will trigger a full release automatically.

basti1302: Yep, that's one of the items on my list 👍 :-)

basti1302: Outcome: For now we want to keep a manual trigger for actually publishing a release. We still need to decide if we want to realize that via two branches (development vs. main) or by manually triggering a job to publish a release. IMO we can leave that undecided for now, as this PR (and the GH action) will not actually release anything.

basti1302:
This is ready for review now I think. I'll fix the commits on main (or our commitlint config) today to make the commit messages and our commitlint rules agree with each other.

package.json
kirrg001: I think it's smart, but IMO I'd not use this, because it's really hard to understand why we are doing this + it's a misusage.

Two other options/ideas:
1. Try https://github.com/actions/cache if the installation takes too long
2. Switch to https://www.npmjs.com/package/@grpc/grpc-js

package.json
```diff
@@ -141,7 +145,7 @@
     "request-promise": "^4.2.2",
     "request-promise-native": "^1.0.5",
     "rimraf": "^3.0.2",
-    "sequelize": "^6.6.2",
```
kirrg001, line 144: _Suggestion_: Version updates not part of this PR (or as separate commit, which can be merged as a single change)

package.json
kirrg001: Do you want to move this out of the PR?

kirrg001 (commented):
The PR still has a couple of todo's and I think there are some small changes needed before we can merge it e.g. comment out "- run: bin/needs-release.js" for now? 

I'll wait till the PR is in a final shape and review it again then :)

package.json
basti1302: Option (2.) does not work for us -- we instrument the "old" grpc package, thus we need that package in our dev dependencies (for the integration test). Instrumenting the "new" package `@grpc/grpc-js` is on the list I think, but not done yet, plus it would not help to get rid of the old dependency.

I'll look into (1.) 👍

package.json
basti1302: Ah, good catch, I wanted to drop that commit again, but forgot. I only did this because the build failed at the `npm audit` stage because of the the ansi-regex vulnerability and didn't even get to the commitlint step on CircleCI. I'll definitely remove that again.

package.json
basti1302: I would just keep it in this one as it is a really small-ish change unless you feel strongly about having it in a separate PR. The resulting commit history after merging this PR will be the same. (It is a separate commit already.)

package.json
basti1302: For now I have removed that commit, so grpc is back in the normal devDependencies.

package.json
basti1302: I have dropped that commit.

basti1302:
> The PR still has a couple of todo's and I think there are some small changes needed before we can merge it e.g. comment out "- run: bin/needs-release.js" for now?
> 
> I'll wait till the PR is in a final shape and review it again then :)

I have removed bin/needs-release.js again, good call. The remaining TODOs in the Github action are actually meant to stay there for now. But we can also throw out the whole Github action. It was just meant to give us a convenient preview on how `lerna publish` / `lerna version` will behave once we move this to CI.

package.json
kirrg001: 👍 Works for me

commitlint.config.js
```diff
@@ -0,0 +1,18 @@
...
+
+// Helpful links:
+// - Conventional Commits: https://www.conventionalcommits.org/en/v1.0.0/
+// - commitlint: https://github.com/conventional-changelog/commitlint
+// - Configuration: https://github.com/conventional-changelog/commitlint/blob/master/docs/reference-rules.md
+
+module.exports = {
+  extends: ['@commitlint/config-conventional'],
+  rules: {
+    'body-max-line-length': [0, 'always', 100] // disables the body-max-line-length rule
```
kirrg001, line 16: <3

kirrg001 (approved):
LGTM!

PR #420, opened 2021-11-19
chore: switched to Node@16 (NPM 8)

kirrg001:
refs 80768

- nvm install 16.13.0
- nvm alias default 16.13.0
- nvm use 16.13.0

Or simply: `nvm use`

Reasons:
1. We wanted to switch to Node16 anyway as we would like to stay on a newer local node version.
2. We would like to tryout NPM workspaces. Lerna is not maintained anymore.

https://github.com/lerna/lerna/issues/2975
https://github.com/lerna/lerna/issues/2703


basti1302 (approved):
LGTM

kirrg001:
A full re-installation of dependencies is **37 seconds faster with Node 16.**
Not a lot, but with #418 we will roughly reduce it by another minute.

PR #421, opened 2021-11-19
[DRAFT] Experimental NPM workspaces (move away from lerna in v2)

kirrg001:
NPM workspaces are so simple :heart:

These are my first findings (they might be not 100% correct, but will be corrected if needed):

- feels like NPM is very smart under the hood and just recognizes that you are in a workspace context.
- single lock file in the root folder - so cool!
- default npm commands! (such as `npm i` -> will just install all dependencies in root + workspaces)
- out of the box functionality such as `npm uninstall mocha --workspaces` (this has not worked with lerna, it was a hard)
- the lerna bug I was running into is no longer a problem because the whole lock structure is different! Lerna copies the package.json, removes local package references and generates the lock file. With NPM we only have one lock file. No one is manipulating it. Still NPM is doing symlinks to local packages. I am not 100% sure (with our use cases) that we might run into problems. 
- no more need to run hacky package lock scripts for audit
- high hopes that all the lerna frustration goes away

**Downside: Does not work well with lerna, because it's just too different.** But you can still use lerna commands such as lerna version or lerna publish as far as I can see. Which is the most important fact.

This needs more brain and testing, but the first impression is really good!

- [x] NPM workspaces works with Node >= 10
- [ ] fix tests
- [x] create v2 branch and change PR base
- [x] move all dev dependencies to root
- [ ] add docs for us!
- [x] cleanup lock cache ci
- [ ] enable skipped tests
- [ ] push knex downgrade to v2 branch directly
- [x] `lerna version --no-git-tag-version --no-push` test it
- [ ] `lerna publish` test it
- [ ] update contributiing docs!
- [ ] add command to switch from v1 to v2 (in v2)
- [ ] add command to switch from v2 to v1 (in v1)
- [x] ci cache with collector-v2!

## NPM workspaces limitations

- https://github.com/npm/cli/issues/4056 
- https://github.com/npm/cli/issues/4084 (--legacy--peer-deps is a workaround, but it means dependencies with peer dependencies won't use the right version)


package.json
```diff
@@ -30,21 +33,19 @@
     }
   ],
   "scripts": {
-    "audit": "lerna run audit --stream && npm audit --production",
-    "bootstrap": "lerna bootstrap",
-    "bump": "lerna publish",
+    "audit": "npm audit --production",
```
kirrg001, line 36: This is so cool and simple!
It automatically audits every package in our workspace!

kirrg001:
I'd right now not continue with playing with this. I'd suggest to touch this area again when NPM workspaces has less problems (see PR description).

On top of the two "blockers" some tests where failing in shared-metrics package, because the way NPM workspaces installs dependencies doesn't make the package happy. We rely a lot on the fact that dependencies are physically available in `shared-metrics/node_modules`, but NPM installs everything in the root (as best as it can). But for this package we'd want to "nohoist" dependencies ([see](https://github.com/npm/rfcs/issues/287)).

I am still convinced that the switch will be great. Happy to switch over as soon as we can. But for know I'd stick with lerna.

PR #423, opened 2021-11-25
fix(tracing): require @elastic/elasticsearch/api in a safe way

basti1302:
The instrumentation for @elastic/elasticsearch needs to load an additional module from the
instrumented package (the api module). We previously assumed that we can safely make this require
call because we only execute it when the application under monitoring executes
require('@elastic/elasticsearch'). Thus, we know that @elastic/elasticsearch is in the node_module
tree. However, when our package is added via --require, the module load path list is different and
does not include the node_modules directory from the application under test. In this scenario the
aforementioned require('@elastic/elasticsearch/api') call would fail in our instrumentation. To
solve that, we now let the requirehook pass the module filename back to the caller, so we can
construct a valid path for require from that in the instrumentation.


packages/core/src/tracing/instrumentation/database/elasticsearchModern.js
```diff
@@ -30,8 +31,9 @@ exports.init = function init() {
   requireHook.onModuleLoad('@elastic/elasticsearch', instrument);
 };
 
-function instrument(es) {
-  const ESAPI = require('@elastic/elasticsearch/api');
+function instrument(es, esModuleFilename) {
+  const pathToEsApi = path.join(esModuleFilename, '..', 'api');
+  const ESAPI = require(pathToEsApi);
```
basti1302, line 36: This change solves the problem at hand. However, we could run into similar issues in the future, it is a non-obvious trap to fall into. I wonder how we can steer our future selves in the right direction to not simply `require` additional modules from the instrumented packages, but instead construct the path based on the filename the requireHook returns.

Thoughts?

packages/core/src/tracing/instrumentation/database/elasticsearchModern.js
```diff
@@ -436,13 +438,13 @@ function forEachApiAction(ESAPI, fn) {
   forEachKeyRecursive(esApiFromBuildApi, [], fn);
 }
 
-function forEachKeyRecursive(obj, path, fn) {
+function forEachKeyRecursive(obj, apiPath, fn) {
```
basti1302, line 441: This needed to be renamed because we now have the identifier `path` in the outer scope already (because we require the `path` module`).

packages/core/src/tracing/instrumentation/database/elasticsearchModern.js
willianpc: I would think that requiring a package into the customer's code by itself sounds like a yellow flag already, but I wouldn't be able to predict the impacts of doing it.

When it comes to potential preventions, I don't have any concrete idea at this point, but a combination of a custom lint rule + unit test + having a special function in our API to require modules inside instrumentations seems like an interesting path.

For instance, let's say we have a function `sensistiveRequire` or whatever name fits, we could at least have a try/catch wrap around it, which would scream at us when potential errors happen, so we could more easily debug and fix customer problems.

The next step would be to create a custom lint rule (which I honestly know little about it, and I am not even sure if this is actually possible) that would shout at us if we use `require` directly into instrumentations. Again, I am not sure if this is possible, as we DO require normally anyway, so Idk how we could differentiate one from the other.

Last but not least, try to write some test that evaluates the require somehow. Either by providing a different path for node_modules or something else.

Sorry for not providing a more concrete idea, but maybe this is some useful brainstorming starting point...

basti1302: The special function for such a require and the custom lint rule sound like good ideas. 👍

basti1302: I've extracted the require to a function in tracingUtil now and also added an eslint rule in a separate commit.

packages/opentelemetry-exporter/package.json
```diff
@@ -53,11 +53,11 @@
   ],
   "dependencies": {
     "@instana/serverless": "1.137.1",
+    "@opentelemetry/api": "1.0.3",
```
basti1302, line 56: @willianpc After implementing the new eslint rule, it complained about this line: https://github.com/instana/nodejs/blob/1d3b29806f1e423f12a4b5cb7340ef584f2991f5/packages/opentelemetry-exporter/src/index.js#L10

There we import `@opentelemetry/api` in production code, but so far we only had it in our `devDependencies`. This actually works anyway, because we depend on `@opentelemetry/core` already and that one depends on  `@opentelemetry/api` in turn. So we always have the `api` package available implicitly, even when `@instana/opentelemetry-exporter` is installed as a dependency or `npm install --production` is used etc.

I know we discussed the OTel dependencies to some extent here already: https://github.com/instana/nodejs/pull/388/files#r712224844 - but not specifically the `api` module (or I cannot remember).

Thoughts? Should we rather leave `@opentelemetry/api` in the devDependencies and just disable the new rule for that line in index.js? That would also work for me. Or do we want to move it from devDependencies to dependencies?

packages/core/test/.eslintrc.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
   },
 
   rules: {
+    'instana/no-unsafe-require': 'off',
```
basti1302, line 24: The rule is enabled globally and disabled for all `test` folders.

packages/opentelemetry-exporter/package.json
willianpc: This is wrong then. Could you move `@opentelemetry/api` to dependencies?
Essentially, this is the otel exporter own logger, so it makes sense to be in production, as our `instanaBackendConnector` will forward any message to this one.
I don't know how this went unnoticed. 
Thanks for the finding!

basti1302: > Could you move @opentelemetry/api to dependencies?

I actually already have (see the diff), so I guess this change is okay then.

willianpc (approved):
Looks pretty cool!

PR #424, opened 2021-11-27
fix(collector): prevent initializing @instana/collector multiple times

basti1302:
Previously, it was possible to instrument the same Node.js process multiple
times. This could happen in various ways:

* Putting `require('@instana/collector')();` into the application multiple
  times.
* Starting the app with `--require /path/to/@instana/collector/src/immediate.js`
  while also having `require('@instana/collector')();` in the application code.
* Having `require('@instana/collector')();` in the application code while being
  in a K8s cluster with the autotrace webhook.

This would make the application starting the announce loop twice. Among other
unexpected behaviors, it would also make the process send twice as many metrics
JSON blobs (two per second instead of one). This in turn would cause the agent
logging "Entity queue is full. Dropping data." frequently, because the agent's
`sense()` loop will only poll once per second from the Node.js sensor's internal
`entityData` queue, while `@instana/collector` sends twice per second.

The module @instana/collector might be installed in multiple locations and one
instance might be loaded via `--require`, so we cannot rely on a lock mechanism
in the scope of the module, but need it to be in the global scope.

refs 81681

### DIFF
--- a/packages/collector/src/index.js
+++ b/packages/collector/src/index.js
@@ -24,6 +24,13 @@ let config;
  * @param {import('./util/normalizeConfig').CollectorConfig} [_config]
  */
 function init(_config) {
+  // @ts-ignore: Property '__INSTANA_INITIALIZED' does not exist on type global
+  if (global.__INSTANA_INITIALIZED) {
+    // Prevent initializing @instana/collector multiple times for the same process.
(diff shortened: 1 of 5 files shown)


kirrg001 (approved):
LGTM 👍

PR #425, opened 2021-11-27
test(collector): update lock files of apps under test with npm@8

basti1302:
When switching to Node.js 16 and npm@8 for local development, we did not update
the two lockfiles. This only becomes apparent when running those tests locally,
because running those tests now creates a diff because those two lock files are
updated to version 2 of the lock file format.


basti1302:
CI currently fails, will take a look next week.

basti1302:
This is now included in https://github.com/instana/nodejs/pull/427

PR #427, opened 2021-11-30
chore(dependencies): move dev dependencies to root

basti1302:
This is a follow up to https://github.com/instana/nodejs/pull/418. More details about the changes and the reasoning behind them can be found in the individual commit messages.


kirrg001 (approved):
Lets do it!

PR #429, opened 2021-12-02
chore: removed uncaught exception config

kirrg001:
refs 80206

https://www.instana.com/docs/ecosystem/node-js/configuration#reporting-uncaught-exceptions

We have to update our docs when merging v2.


packages/collector/src/uncaught/index.js
```diff
@@ -61,123 +56,16 @@ exports.init = function (_config, _downstreamConnection, _processIdentityProvide
...
-          'https://github.com/nodejs/node/issues/28550 for details.'
-      );
-    }
-  } else {
-    logger.info('Reporting uncaught exceptions is disabled.');
-  }
-}
-
 exports.deactivate = function () {
-  process.removeListener(uncaughtExceptionEventName, onUncaughtException);
```
kirrg001, line 103: I was wondering why the feature got deprecated in 1.112. https://github.com/instana/nodejs/commit/5fda92a48032f06951578c3ac874d69cc089651f does not really explain it :)

I understand that with https://www.instana.com/docs/ecosystem/os-process/#abnormal-termination we can tell if an app has crashed. But I wonder how the customer will figure out the reason behind the crash? (if he doesn't use rollbar or similar)

In general we do report errors using `span.ec` and `span.error` when finishing spans and problems were detected.

basti1302: True, for the reason of the crash one would need to check the log files of the application or similar means. The two features are not fully equivalent, but the abnormal termination feature is much more general (does not work only for Node.js).

The reason it got deprecated is in part because of the following crux: To report any data from a process that is about to terminate you need **synchronous** HTTP communication, something that is simply not available in Node.js. The current work around (starting a child process and waiting for it to terminate) is not great.

In particular, it delays the process termination, which will also delay starting a new process when some form of orchestration (K8s) or process monitoring ([forever](https://www.npmjs.com/package/forever), [nodemon](https://nodemon.io/) etc.) is used. Thus, this feature prolongs the downtime of the service under monitoring.

basti1302:
I think we should also delete `packages/collector/test/uncaught/assumptions_test.js` and `packages/collector/test/uncaught/disabled_test.js`, as well as all apps (plus the `controls.js`file) inside `packages/collector/test/uncaught/apps/`, except for `server.js` (which is still used).

All of those tests were related to the uncaught exception feature, they no longer serve any purpose now. 

WDYT?

basti1302 (approved):
The current changes all look good to me 👍  I think we should delete a few more related files in this PR, see above.

PR #430, opened 2021-12-02
V2 async localstorage

willianpc:
This new PR has all changes requested by the team, as well as it attempts to answer some questions in the [previous PR](https://github.com/instana/nodejs/pull/416).
It is also cleaner and forked from `v2` branch instead of `main`.

### DIFF
--- a/packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/receiveMessage.js
+++ b/packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/receiveMessage.js
@@ -99,49 +99,54 @@ async function receiveAsync() {
   try {
     const sqsPromise = sqs.receiveMessage(receiveParams).promise();
     const data = await sqsPromise;
-    instana.sdk.runInAsyncContext(sqsPromise.instanaAsyncContext, async () => {
-      if (data && data.error) {
-        log('receive message data error', data.error);
-        return;
-      } else if (!data || !data.Messages || data.Messages.length === 0) {
-        log('no messages, doing nothing');
-        return;
-      }
 
-      data.Messages.forEach(message => {
-        sendToParent(message);
+    return new Promise(resolve => {
+      instana.sdk.runInAsyncContext(sqsPromise.instanaAsyncContext, async () => {
+        if (data && data.error) {
+          log('receive message data error', data.error);
+          return resolve();
+        } else if (!data || !data.Messages || data.Messages.length === 0) {
+          log('no messages, doing nothing');
+          return resolve();
+        }
+
+        data.Messages.forEach(message => {
+          sendToParent(message);
+        });
+
+        log(
+          'got messages:',
+          data.Messages.map(m => m.MessageId)
+        );
+
+        span = instana.currentSpan();
+        span.disableAutoEnd();
+
+        const messagesForDeletion = data.Messages.map(message => {
+          return {
+            Id: message.MessageId,
+            ReceiptHandle: message.ReceiptHandle
+          };
+        });
+
+        await sqs
+          .deleteMessageBatch({
+            QueueUrl: queueURL,
+            Entries: messagesForDeletion
+          })
+          .promise();
+
+        await delay(1000);
+        await request(`http://127.0.0.1:${agentPort}`);
+        log('The follow up request after receiving a message has happened.');
+        span.end();
+        return resolve();
       });
-
-      log(
-        'got messages:',
-        data.Messages.map(m => m.MessageId)
-      );
-
-      span = instana.currentSpan();
-      span.disableAutoEnd();
-
-      const messagesForDeletion = data.Messages.map(message => {
-        return {
-          Id: message.MessageId,
-          ReceiptHandle: message.ReceiptHandle
-        };
-      });
-
-      await sqs
-        .deleteMessageBatch({
-          QueueUrl: queueURL,
-          Entries: messagesForDeletion
-        })
-        .promise();
-
-      await delay(1000);
-      await request(`http://127.0.0.1:${agentPort}`);
-      log('The follow up request after receiving a message has happened.');
-      span.end();
     });
   } catch (err) {
     span && span.end(1);
     log('ERROR receiving/deleting message', err);
+    return Promise.reject(err);
   }
 }
 
--- a/packages/collector/test/tracing/control_flow/pubsub_async_iterator/app.js
+++ b/packages/collector/test/tracing/control_flow/pubsub_async_iterator/app.js
@@ -36,13 +36,13 @@ asyncIterator.pullValue().then(event1 => {
   // Chronologically, everything inside the then-handler this happens after cls.ns.set (see below). Due to custom
   // queueing in pubsub_async_iterator, the cls context would get lost though (unless we fix it).
   log(event1);
-  valuesReadFromCls.push(cls.ns.get('key'));
+  valuesReadFromCls.push(cls.ns.get('key', true));
   asyncIterator.pullValue().then(event2 => {
     log(event2);
-    valuesReadFromCls.push(cls.ns.get('key'));
+    valuesReadFromCls.push(cls.ns.get('key', true));
     asyncIterator.pullValue().then(event3 => {
       log(event3);
-      valuesReadFromCls.push(cls.ns.get('key'));
+      valuesReadFromCls.push(cls.ns.get('key', true));
     });
   });
 });
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -349,17 +349,19 @@ function setCurrentSpan(span) {
 
 /**
  * Get the currently active span.
+ * @param {boolean} [fallbackToSharedContext=false]
  * @returns {InstanaSpan}
  */
-function getCurrentSpan() {
-  return ns.get(currentSpanKey);
+function getCurrentSpan(fallbackToSharedContext = false) {
+  return ns.get(currentSpanKey, fallbackToSharedContext);
 }
 
-/*
+/** *
  * Get the reduced backup of the last active span in this cls context.
+ * @param {boolean} [fallbackToSharedContext=false]
  */
-function getReducedSpan() {
-  return ns.get(reducedSpanKey);
+function getReducedSpan(fallbackToSharedContext = false) {
+  return ns.get(reducedSpanKey, fallbackToSharedContext);
 }
 
 /**
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -356,7 +356,7 @@ function getCurrentSpan(fallbackToSharedContext = false) {
   return ns.get(currentSpanKey, fallbackToSharedContext);
 }
 
-/** *
+/**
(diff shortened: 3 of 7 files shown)


packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/receiveMessage.js
```diff
@@ -99,49 +99,54 @@ async function receiveAsync() {
...
-        return;
-      } else if (!data || !data.Messages || data.Messages.length === 0) {
-        log('no messages, doing nothing');
-        return;
-      }
 
-      data.Messages.forEach(message => {
-        sendToParent(message);
-      });
+    return new Promise(resolve => {
```
willianpc, line 103: This change makes sure that a promise is returned and "awaited" before the loop triggers another queue polling.

packages/collector/test/tracing/control_flow/pubsub_async_iterator/app.js
```diff
@@ -36,13 +36,13 @@ asyncIterator.pullValue().then(event1 => {
   // Chronologically, everything inside the then-handler this happens after cls.ns.set (see below). Due to custom
   // queueing in pubsub_async_iterator, the cls context would get lost though (unless we fix it).
   log(event1);
-  valuesReadFromCls.push(cls.ns.get('key'));
+  valuesReadFromCls.push(cls.ns.get('key', true));
```
willianpc, line 39: @basti1302 , this change is related to your suggestion in the old PR: https://github.com/instana/nodejs/pull/416#discussion_r760925276
There were other changes, so I will tag you in all of them.

packages/core/src/tracing/cls.js
```diff
@@ -349,17 +349,19 @@ function setCurrentSpan(span) {
 
 /**
  * Get the currently active span.
+ * @param {boolean} [fallbackToSharedContext=false]
  * @returns {InstanaSpan}
  */
-function getCurrentSpan() {
-  return ns.get(currentSpanKey);
+function getCurrentSpan(fallbackToSharedContext = false) {
```
willianpc, line 355: @basti1302 , again, about the new optional param from https://github.com/instana/nodejs/pull/416#discussion_r760925276

willianpc: There is one more in the same place, doing the same thing, so I won't tag you there again, as they are pretty much the same thing, in the same file

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,324 @@
...
+    return unset.bind(null, context, key, value);
+  }
+
+  /**
+   * Retrieves a value by key from the current CLS context (or a parent context), assuming the key/value pair has
+   * been set earlier in this context.
+   * @param {string} key
+   * @param {boolean} [fallbackToSharedContext=false]
+   */
+  get(key, fallbackToSharedContext = false) {
```
willianpc, line 88: @basti1302 again... https://github.com/instana/nodejs/pull/416#discussion_r760925276

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,324 @@
...
+    const context = ctx || this.createContext();
+
+    /**
+     * We need the new context to be set as the active context before the promise is called.
+     * Otherwise, the current active context will cause the new span transmission to fail.
+     * Unlike our old code, AsyncLocalStorage.run does not update the active context, so we have to do it manually with
+     * storage.enterWith. This is done by `this.enter`, which calls `enterWith`.
+     * The same happens in the old code, and for the same reasons.
+     */
+    this.enter(context);
```
willianpc, line 160: Please, @basti1302 let me know if this change and explanation in the comments satisfies your questioning here: https://github.com/instana/nodejs/pull/416#discussion_r760927718

The call to `enterWith` stays, but the code was cleaned a lot and I no longer update the context twice.

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,324 @@
...
+   * is thrown if the given function does not return a promise.
+   *
+   * This function assumes that the returned promise (if any) is CLS-friendly or wrapped already.
+   * @param {Function} fn
+   * @param {InstanaCLSContext} ctx
+   */
+  runPromiseOrRunAndReturn(fn, ctx) {
+    let isPromise = false;
+    const context = ctx || this.createContext();
+    this.enter(context);
```
willianpc, line 182: @basti1302 same thing here, about https://github.com/instana/nodejs/pull/416#discussion_r760927718

packages/core/src/tracing/instrumentation/protocols/graphql.js
```diff
@@ -98,7 +98,7 @@ function traceQueryOrMutation(
   operationDefinition,
   operationName
 ) {
-  const activeEntrySpan = cls.getCurrentSpan();
+  const activeEntrySpan = cls.getCurrentSpan(true);
```
willianpc, line 101: @basti1302 also related to the graphql change from your comment https://github.com/instana/nodejs/pull/416#discussion_r760925276.
There is also the same change at line 168.

packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/receiveMessage.js
willianpc: This problem is also present in the current CLS. But as we manually control the active context in the `exit()` method, the [this._set.pop()](https://github.com/instana/nodejs/blob/main/packages/core/src/tracing/clsHooked/context.js#L294) call removes the active context (which contains the SQS exit span, if I remember correctly now).
This `.pop()` sets the active context to null (in the particular case of this test, there is nothing else in the `this._set` array), so the SQS instrumentation doesn't complain about an existent `currentSpan()` when it attempts to instrument the entry span.

packages/core/src/tracing/cls.js
```diff
@@ -349,17 +349,19 @@ function setCurrentSpan(span) {
...
+  return ns.get(currentSpanKey, fallbackToSharedContext);
 }
 
-/*
+/** *
```
basti1302, line 359: _nitpick_: this looks odd, probably a typo
```suggestion
/**
```

packages/core/src/tracing/clsHooked/async_local_storage_context.js
basti1302: Yep, that makes sense and looks good 👍

basti1302 (approved):
LGTM

basti1302:
Awesome work! 🎉 🎉

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,324 @@
...
+     * we wish to use it as a parent context for the next instrumentation.
+     * For instance, we need a shared context in the GraphQL instrumentation when HTTP is used as a client entry point.
+     * In this case, the HTTP server context is properly managed by the Async Localstorage and the context is exited by
+     * it. But we need this context as the parent of the GraphQL instrumentation.
+     */
+    this.sharedContext = null;
+    this.name = name;
+    this.active = null;
+    this.id = -1;
+    this._contexts = new Map();
```
kirrg001, line 55: I guess we can remove `this._contexts` as it was replaced by async local storage keeping track of the context.

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,324 @@
...
+    wrapEmitter(emitter, attach, bind);
+  }
+
+  /**
+   * @param {InstanaCLSContext} context
+   */
+  enter(context) {
+    assert.ok(context, 'context must be provided for entering');
+    storage.enterWith(context);
+    this.sharedContext = context;
```
kirrg001, line 264: Wondering:

This class is a singleton. I do not fully understand how we can rely on `sharedContext` being set to the correct context.

1. GraphQL request mutation comes in. There are x subscribers
2. GraphQL pub/sub process is activated. We have to call `enter` in the GraphQL Subscription Instrumentation to somehow set the remembered/previous context.
3. We hope that no one else calls `enter` in between, otherwise we'd loose our shared context. As far as I can see there are other use cases where `enter` can be called in between (from other instrumentations), because GraphQL subscription instrumentation calling `enter` and the actual subscription tracing call in the GraphQL instrumentation is async!
4. GraphQL subscription is executed to inform the clients.
5. We hope `sharedContext` is still the correct one.

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -0,0 +1,324 @@
...
+   * as soon as the the promise resolves/is rejected. Asynchronous work started in this promise will happen in that new
+   * context.
+   *
+   * If the given function does not create a then-able, an error will be thrown.
+   *
+   * This function assumes that the returned promise is CLS-friendly or wrapped already.
+   * @param {Function} fn
+   * @param {InstanaCLSContext} ctx
+   */
+  runPromise(fn, ctx) {
```
kirrg001, line 150: As far as I can see `runPromise` and `runPromiseOrRunAndReturn` **with args** are only called via our SDK from customers? I might be wrong. Wondering if we manually should test https://www.instana.com/docs/ecosystem/node-js/api/#restoring-the-async-context-manually these scenarios. Maybe I have not found where these use cases are covered.

basti1302: - The amqp instrumentation uses `runPromise`: https://github.com/instana/nodejs/blob/edb5d71a51a4c6ec3342e04f2cd30314772084d1/packages/core/src/tracing/instrumentation/messaging/amqp.js#L234
- The bull instrumentation uses `runPromise`:  https://github.com/instana/nodejs/blob/main/packages/core/src/tracing/instrumentation/messaging/bull.js#L182
- Our Lambda wrapper uses `runPromiseOrRunAndReturn`: https://github.com/instana/nodejs/blob/edb5d71a51a4c6ec3342e04f2cd30314772084d1/packages/aws-lambda/src/wrapper.js#L82

packages/core/src/tracing/clsHooked/async_local_storage_context.js
basti1302: I agree, that seems to be unused.

packages/core/src/tracing/clsHooked/async_local_storage_context.js
kirrg001: > with args

:)

kirrg001: Discussed: covered in tests

packages/core/src/tracing/clsHooked/async_local_storage_context.js
kirrg001: Discussed: valid concern, but "this approach" is also used in the older context implementations.
We would like to change this, but need time & brain for that. I'll make that clear as part of a comment.

kirrg001 (approved):

PR #431, opened 2021-12-09
chore: removed legacy context

kirrg001:
refs 80206

Was not able to find any other usages regarding legacy ctx.


basti1302 (approved):
LGTM. Let's get rid of this 🎉

PR #432, opened 2021-12-09
chore: removed disableAutomaticTracing legacy config

kirrg001:
refs 80206

I found this config. It looks legacy to me. Need to update the docs for all config changes before releasing v2

https://www.instana.com/docs/ecosystem/node-js/configuration/

Alternativ: Deprecated the config first, but somehow this config already looks deprecated. Maybe the docs were not updated to be clear?


basti1302 (approved):
LGTM 👍

PR #434, opened 2021-12-09
chore: updated dev deps

kirrg001:
refs 75234

- no major updates
- trying to update as much as possible
- apollo/* can't be updated
  - running into a lot of peer dependency errors
  - mostly because they want graphql 15
- **aws-sdk fails when updating from 3.2x to 3.4x**
  - need to investigate in a different PR
- mocha/multi can't be updated because of a peer dependency problem with mocha
  - can be updated in v2
 - mali 0.2x needs @grpc/proto-loader 0.3 | 0.4 | 0.5
   - it was a coincidence that it worked, but not sure why. looks like a subdependency installed a non 0.6 version
- @google-cloud/pubsub > 2.12 does not work with Node 8
  - ignore and do not investigate -> update in v2
  
**Please look at all three commits.**


basti1302 (approved):
LGTM! Thanks! 🙇

PR #435, opened 2021-12-13
chore: updated otel dependencies

kirrg001:
refs 75234

OTel has released 1.x versions. See https://github.com/open-telemetry/opentelemetry-js/blob/main/CHANGELOG.md#101--experimental-0270

IMO I would not update to it now, but maybe add a ticket to ensure our otel exporter still works with 1.x?

### DIFF
--- a/packages/opentelemetry-exporter/package-lock.json
+++ b/packages/opentelemetry-exporter/package-lock.json
@@ -10,18 +10,18 @@
 			"license": "MIT",
 			"dependencies": {
 				"@opentelemetry/api": "1.0.3",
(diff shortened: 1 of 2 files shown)


basti1302 (approved):
LGTM. I would also vote to wait for @willianpc with the 1.x updates. 👍

kirrg001:
Good idea!

willianpc:
For history purposes, I created this branch https://github.com/instana/nodejs/pull/450/files where I confirm that OTel is working with version 1.x.x
I will take it from there :)
Cheers

PR #436, opened 2021-12-13
chore: added experimental dependabot config

kirrg001:
refs 75234

- update dependencies more often, but ensure it's not noisy or annoying

"Once a week it will look for patch+minor dependency updates (dev+prod), but it won't create more than 2 pull requests"

Need to see it in action.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: /
+    target-branch: main
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 2
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: /
+    target-branch: main
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 2
+    ignore:
+      - dependency-name: "*"
```
basti1302, line 10: Since this is in an `ignore` section, wouldn't that ignore _all_ depenencies and thereby effectively disable the dependabot PRs? But maybe I'm just reading this wrong.

kirrg001: I got this from an issue
See https://github.com/dependabot/dependabot-core/issues/2219

As far as I understood:
- you need to specify a dependency-name e.g. `lodash` and then tell dependabot which semver type to ignore
- but we want all dependencies

kirrg001: Wondering if we also want to ignore patch updates.

basti1302 (approved):
LGTM

PR #437, opened 2021-12-13
Updated @aws-sdk & added support for 3.4x

kirrg001:
no issue

- fails when updating from 3.2 to 3.4

@aws-sdk/smithly-client >= 3.36.0 changed how the dist structure gets delivered, which breaks our instrumentation for all three packages

https://github.com/aws/aws-sdk-js-v3/blob/main/packages/smithy-client/CHANGELOG.md#3360-2021-10-08

- [x] update versions to next minor
- [x] fixed bug & added support for both versions
- [x] test multiple versions to ensure quality

**Please review commit per commit.**


basti1302 (approved):
LGTM! Thanks for finding and fixing this 👍 🎉

PR #440, opened 2021-12-15
chore: fixed tests for @elasic/elasticsearch 7.16

kirrg001:
This no longer works because of https://github.com/elastic/elasticsearch-js/compare/v7.15.0...v7.16.0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L11

They only allow to require js files

---

PR contains cleanup commits and commits for better readability.

### DIFF
--- a/packages/collector/test/tracing/database/elasticsearch_modern/test.js
+++ b/packages/collector/test/tracing/database/elasticsearch_modern/test.js
@@ -48,34 +48,13 @@ mochaSuiteFn('tracing/elasticsearch (modern client)', function () {
     // TODO: Add newer versions, currently we only test against 7.9
     {
       versionRange: '^7.9.1',
-      apiVersion: '7.9',
       instrumentationFlavor: 'transport'
     },
     {
       versionRange: '7.9.0',
-      apiVersion: '7.9',
       instrumentationFlavor: 'api'
     }
-
-    // To test older client versions, you also need to set up a matching version of the Elasticsearch Docker container.
-    // The legacy client (package `elasticsearch`) supported using different server/API versions with the same client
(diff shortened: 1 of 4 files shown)


packages/collector/test/tracing/database/elasticsearch_modern/test.js
```diff
@@ -31,58 +32,54 @@ mochaSuiteFn('tracing/elasticsearch (modern client)', function () {
   this.timeout(Math.max(config.getTestTimeout() * 4, 30000));
 
   before(() => {
-    originalEsVersion = require('@elastic/elasticsearch/package.json').version;
+    // NOTE: We cannot use `require(@elastic/elasticsearch/package.json), because of
+    //       https://github.com/elastic/elasticsearch-js/blob/v7.16.0/package.json#L11
```
basti1302, line 36: _nitpick_: IMO we do not strictly need an explanation here.

packages/collector/test/tracing/database/elasticsearch_modern/test.js
```diff
@@ -31,58 +32,54 @@ mochaSuiteFn('tracing/elasticsearch (modern client)', function () {
...
+        });
 
         globalAgent.setUpCleanUpHooks();
         const agentControls = globalAgent.instance;
 
         const controls = new ProcessControls({
           dirname: __dirname,
           useGlobalAgent: true,
-          env: { ES_API_VERSION: apiVersion }
+          env: {}
```
basti1302, line 82: _nitpick_: If we do not pass any environment variables, we could also remove this line completely, I think (instead of passing an empty object).

basti1302 (approved):
LGTM

PR #443, opened 2021-12-20
fix(aws-lambda): reduced lambda timeout error count when using extension 

kirrg001:
refs 82459

- this problem already started with **layer version 73 (1.124)**

Caused by https://github.com/instana/nodejs/commit/40b555d559d2cd913f8b857b888eb472df06c1d8

71 -> no errors 1.121
72 -> no errors 1.121
73 -> 1.124 -> errors
79 -> errors
80 -> errors
88 -> 1.133.0 -> errors
87 -> 1.132.0 -> errors
94 -> errors

We can see in the reproduction scenario that the timeout gets triggered with negative remaining millis.

> Heuristical timeout detection was triggered with -6328 milliseconds left.

And we run into this error:

> 2021-12-20T14:02:10.847Z [ERROR] instana-lambda-extension.transport: Error in transport.: err="Post "https://serverless-magenta-saas.instana.rocks/bundle": POST https://serverless-magenta-saas.instana.rocks/bundle giving up after 1 attempt(s): context deadline exceeded" | 2021-12-20T14:02:10.847Z [ERROR] instana-lambda-extension.transport: Error in transport.: err="Post "https://serverless-magenta-saas.instana.rocks/bundle": POST https://serverless-magenta-saas.instana.rocks/bundle giving up after 1 attempt(s): context deadline exceeded"
-- | --

**giving up after 1 attempt** (we have configured 3 allowed attempts)

**context deadline exceeded**

"Context Deadline Exceeded is an error occurring in Go when a context of an HTTP request has a deadline or a timeout set, i.e., the time after which the request should abort. This error is returned if the time of a server response is greater than the set timeout"

**instana-lambda-extension.transport: Error in transport**

See the client package - it comes from [here](https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L695)
Which is used [here](https://github.com/instana/lambda-extension/blob/main/transport/transport.go#L182)

The fixes in node.js reduce the problem from 92-94% error count to 99,95%.
We still need to apply a Go fix I guess.

We have seen "Timed out waiting for entry span." too in the logs.

## TODO:

- [x] Improve GO code

### DIFF
--- a/packages/aws-lambda-auto-wrap/src/index.js
+++ b/packages/aws-lambda-auto-wrap/src/index.js
@@ -25,6 +25,7 @@ if (!wrappedHandler) {
 }
 
 exports.handler = function instanaAutowrapHandler(event, context, callback) {
+  // NOTE: wondered why this is not protected with a try/catch, because we do not use shimmer
   return wrappedHandler(event, context, callback);
 };
 
--- a/packages/aws-lambda-auto-wrap/src/index.js
+++ b/packages/aws-lambda-auto-wrap/src/index.js
@@ -25,7 +25,6 @@ if (!wrappedHandler) {
 }
 
 exports.handler = function instanaAutowrapHandler(event, context, callback) {
-  // NOTE: wondered why this is not protected with a try/catch, because we do not use shimmer
   return wrappedHandler(event, context, callback);
 };
 
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -72,8 +72,7 @@ TMP_ZIP_DIR=tmp
 
 echo Will build Lambda layer with name \"$LAYER_NAME\".
 
-#REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
-REGIONS=$'us-east-2'
+REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
 
 if [[ -z $SKIP_DOCKER_IMAGE ]]; then
   echo Will build Docker image with name \"$DOCKER_IMAGE_NAME\".
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -72,7 +72,8 @@ TMP_ZIP_DIR=tmp
 
 echo Will build Lambda layer with name \"$LAYER_NAME\".
 
-REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
+#REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
+REGIONS=$'us-east-2'
 
 if [[ -z $SKIP_DOCKER_IMAGE ]]; then
   echo Will build Docker image with name \"$DOCKER_IMAGE_NAME\".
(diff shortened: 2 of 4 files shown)


packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -72,7 +72,8 @@ TMP_ZIP_DIR=tmp
 
 echo Will build Lambda layer with name \"$LAYER_NAME\".
 
-REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
+#REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
+REGIONS=$'us-east-2'
```
kirrg001, line 76: TODO: revert

packages/aws-lambda-auto-wrap/src/index.js
```diff
@@ -25,6 +25,7 @@ if (!wrappedHandler) {
 }
 
 exports.handler = function instanaAutowrapHandler(event, context, callback) {
+  // NOTE: wondered why this is not protected with a try/catch, because we do not use shimmer
```
kirrg001, line 28: TODO: Bastian will take a look

packages/aws-lambda/src/wrapper.js
```diff
@@ -401,16 +401,32 @@ function postHandler(entrySpan, error, result, callback) {
...
+   * `setTimeout` is not 100% reliable
+   */
+  if (remainingMillis < 200) {
+    logger.debug('Skipping heuristical timeout detection because lambda timeout exceeded already.');
+    return;
+  }
 
   if (entrySpan) {
+    // CASE: Timeout not needed, we already send the data to the backend successfully
+    if (entrySpan.transmitted) {
```
kirrg001, line 416: Alternative: kill timeout!

packages/aws-lambda/src/wrapper.js
```diff
@@ -401,16 +401,32 @@ function postHandler(entrySpan, error, result, callback) {
  * back end.
  */
 function postHandlerForTimeout(entrySpan, remainingMillis) {
```
kirrg001, line 403: Even if we would make this optional for the customer the error count can occur still if the extension does not receive a request from node. That's why we need the go fix too.

It's still an option to make this optional.

kirrg001:
Next steps:

- [ ] add opt-out for timeout feature
- [ ] tidy up PR

We will first release some node improvements regarding the lambda timeout.
But these won't fully fix the problem. In parallel we will continue with the go improvements.

packages/aws-lambda-auto-wrap/src/index.js
kirrg001: @basti1302 Did you have a chance to take a look?

basti1302: The wrapped handler ultimately calls the actual Lambda handler, that is, the customer's handler function. If we would wrap this in a try-catch, we would potentially alter the behavior: If the customer's handler function throws an error, it needs to bubble up through here to the AWS Lambda runtime. We must not catch an exception here.

Alternatively, we can theoretically discuss wrapping `loadTargetHandlerFunction` in a try-catch, because that is our own logic. But how would we handle an error? An error in there would mean we were not able to find the actual Lambda handler, which is basically an unrecoverable failure, which is why we throw the respective errors from that function.

In summary: I think the current behavior, that is, not handling any errors here and instead letting errors from `wrappedHandler` bubble up the call stack is the correct behavior.

packages/aws-lambda/src/wrapper.js
basti1302: Yes, let's add an opt-out for the heuristical timeout detection 👍

kirrg001:
FYI reverted the lambda extension binary for now
We will update the bin as soon as we have merged the go improvements

kirrg001:
I'll squash the commits after review using the following commit msg

> fix(aws-lambda): reduced lambda timeout error count when using extension

packages/aws-lambda/src/wrapper.js
```diff
@@ -174,7 +174,13 @@ function shimmedHandler(originalHandler, originalThis, originalArgs, _config) {
       });
     };
 
-    registerTimeoutDetection(context, entrySpan);
+    /**
+     * We offer the customer to disable the timeout detection
+     * See https://github.com/instana/nodejs/pull/443
+     */
+    if (!process.env.DISABLE_LAMBDA_TIMEOUT_DETECTION || process.env.DISABLE_LAMBDA_TIMEOUT_DETECTION === 'false') {
```
basti1302, line 181: All other env vars are prefixed with `INSTANA_`. Could you prefix this one, too?

kirrg001: 👍

kirrg001: Pushed

basti1302 (approved):
LGTM

PR #446, opened 2021-12-21
fix(tracing): fix vendoring of emitter-listener for legacy cls context

basti1302:


kirrg001:
Is there a ticket reference?

basti1302:
> Is there a ticket reference?

No, there is no ticket for this.

kirrg001 (approved):

kirrg001:
Good catch 🎣

PR #449, opened 2021-12-28
fix(async ls): trying to create a polyfill

willianpc:


kirrg001:
I'd suggest to close the PR, but keep the branch for a later lookup in case we need it :)

willianpc:
This PoC has been aborted.

The team doesn't see great value at maintaining the polyfill.
We will keep the strategy of falling back to the old async storage strategy when the Node.js version is smaller than 12 and higher than 16.7 up to whatever version the issue is fixed (which should be later this month of January 2022)

PR #451, opened 2021-12-30
test(lambda): add testbed based on the AWS Lambda runtime emulator

basti1302:
This adds an initial version of a testbed that can be used to test the
Lambda instrumentation (@instana/aws-lambda, instana-aws-lambda-auto-wrap
and the Lambda extension) locally.


kirrg001:
I like the idea of adding the lambda runtime docker option. Great find 👍

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -62,7 +62,15 @@ if [[ -z $CONTAINER_REGISTRY ]]; then
   CONTAINER_REGISTRY=icr.io
 fi
 if [[ -z $DOCKER_IMAGE_NAME ]]; then
```
kirrg001, line 64: Do we want to add some general docs or examples of how to use `DOCKER_IMAGE_NAME`? :) Maybe you have some local examples we can put into our docs.

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -188,59 +200,69 @@ echo "building layer with package version $VERSION"
...
-      --layer-name $LAYER_NAME \
-      --version-number $lambda_layer_version \
-      --statement-id public-permission-all-accounts \
-      --principal \* \
-      --action lambda:GetLayerVersion \
-      --output text
-  else
-    echo "   + WARNING: Lambda layer version $lambda_layer_version does not seem to be numeric, will not set permissions in region $region"
-  fi
+if [[ -z $SKIP_AWS_PUBLISH_LAYER ]]; then
```
kirrg001, line 212: The file has become big with a lot of conditions and flags. 

I am wondering if we want to add some prompts (not part of this PR)

e.g.
Are these your inputs? local build bla blub
Do you really want to publish "x y z"?
These are the regions to publish to .... ? Confirm

The prompts can be skipped for e.g. Concourse

packages/aws-lambda/test/runtime_emulator/Dockerfile
```diff
@@ -0,0 +1,18 @@
+ARG LAMBDA_BASE_IMAGE=public.ecr.aws/lambda/nodejs:14
+ARG INSTANA_LAYER=icr.io/instana/aws-lambda-nodejs:latest
```
kirrg001, line 2: Wondering why we use the remote image here

kirrg001: Oh I see it get's overriden in the build script. NVM

kirrg001 (approved):
LGTM :)

packages/aws-lambda/layer/bin/publish-layer.sh
basti1302: Good point, I added a comment for both DOCKER_IMAGE_NAME and LAYER_NAME.

packages/aws-lambda/layer/bin/publish-layer.sh
basti1302: That would be a nice improvement but I agree, it should be out of scope for this PR. There are already some checks in place that should prevent accidentally overwriting our standard layer and image with local changes, but some additional safe guards might not harm.

kirrg001: Sweet <3

PR #454, opened 2022-01-04
feat: self-disable if detected Node.js runtime version is too old

basti1302:
This change prevents @instana packages from loading in (and thereby crashing)
Node.js processes that run on an unsupported, outdated Node.js version.

refs 80206

### DIFF
--- a/packages/aws-fargate/src/index.js
+++ b/packages/aws-fargate/src/index.js
@@ -5,6 +5,17 @@
 
 'use strict';
 
+const { isNodeJsTooOld, minimumNodeJsVersion } = require('@instana/core/src/util/nodeJsVersionCheck');
+
+if (isNodeJsTooOld()) {
+  // eslint-disable-next-line no-console
+  console.error(
+    `The package @instana/aws-fargate requires at least Node.js ${minimumNodeJsVersion} but this process is ` +
+      `running on Node.js ${process.version}. This Fargate container will not be monitored by Instana.`
+  );
+  return;
+}
+
 const { util: coreUtil } = require('@instana/core');
 const { environment: environmentUtil } = require('@instana/serverless');
 
--- a/packages/google-cloud-run/src/index.js
+++ b/packages/google-cloud-run/src/index.js
@@ -5,6 +5,17 @@
 
 'use strict';
 
+const { isNodeJsTooOld, minimumNodeJsVersion } = require('@instana/core/src/util/nodeJsVersionCheck');
+
+if (isNodeJsTooOld()) {
+  // eslint-disable-next-line no-console
+  console.error(
+    `The package @instana/google-cloud-run requires at least Node.js ${minimumNodeJsVersion} but this process is ` +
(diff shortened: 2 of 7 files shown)


.eslintrc.js
```diff
@@ -35,7 +35,7 @@ module.exports = {
       [
         { pattern: '' },
         { pattern: ' \\(c\\) Copyright IBM Corp. \\d{4}' },
-        { pattern: ' \\(c\\) Copyright Instana Inc. and contributors \\d{4}' },
+        { pattern: '(?: \\(c\\) Copyright Instana Inc. and contributors \\d{4})?' },
```
basti1302, line 38: This is actually unrelated to the PR, but this is simply the first PR where this becomes relevant.

With transfer of business from the Instana Inc entity to IBM, the copyright for new files going forward should no longer require this line, so I made it optional.

packages/google-cloud-run/src/index.js
```diff
@@ -5,6 +5,16 @@
 
 'use strict';
 
+const minimumNodeJsVersion = require('@instana/core/src/util/minimumNodeJsVersion');
+if (require('@instana/core/src/util/nodeJsTooOld')(minimumNodeJsVersion)) {
```
kirrg001, line 9: Suggestion: You could require the minimum node js version inside the `nodeJsTooOld` script (both live in core)
This will make the code a little bit more readable. And you can still offer to pass in a different node version

```
if (require('@instana/core/src/util/nodeJsTooOld')()) {
   ...
}
```

kirrg001: Suggestion: You could even do

```
require('@instana/core/src/util/nodeJsTooOld')()
```

And put the console.error into the utility

kirrg001:
Super sweet!

willianpc (approved):
LGTM.
I have one small consideration about reducing import usage and DRYing the functionality, but this is no reason not to approve the PR.

packages/aws-fargate/src/index.js
```diff
@@ -5,6 +5,16 @@
 
 'use strict';
 
+const minimumNodeJsVersion = require('@instana/core/src/util/minimumNodeJsVersion');
+if (require('@instana/core/src/util/nodeJsTooOld')(minimumNodeJsVersion)) {
```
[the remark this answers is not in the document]
basti1302: I have bundled both the minimum version and the check into one file, thereby we only have one import now. I left the console.error in each entry file because the text is a bit different for each package.

packages/google-cloud-run/src/index.js
basti1302: I have bundled both the minimum version and the check into one file, thereby we only have one new module and one line with imports now. I left the console.error in each entry file because the text is a bit different for each package.

basti1302: > require('@instana/core/src/util/nodeJsTooOld')()

That would actually be too minimal IMO, the important piece is to return early and stop evaluating the rest of the module.

kirrg001 (approved):

PR #458, opened 2022-01-12
WIP: TS types for TS projects

willianpc:
This essentially enables intellisense of our API (written in Js with JSDoc type annotations) in projects written with Typescript.

This is in addition to the already in place intellisense for projects written in VanillaJS.

### What's new

A bunch of d.ts files is what is new.

Also, a few files were slightly modified to allow a proper generation of the declaration files, but nothing that should impact our tracer.

Once this is merged into master, we must keep in mind that any change in the JSDoc annotation types will trigger changes in the d.ts files as well.

### Should I care about the d.ts file content?
No.


willianpc:
This task has been aborted because:

1. The majority of the team does not like Typescript and may have a bad time at maintaining it in the future.
2. Apparently not too many customers use Typescript, which kind of makes this task not so relevant.
3. It's a bit of a hassle to have to generate the d.ts files before the commit/push and eventually having them popping up into PRs. This can cause confusion and unnecessary noise in the PRs.

willianpc:
> @willianpc FWIW, here is a customer that does use Typescript and would like to see this implemented ;) Honestly, it's rare these days to come across a package without typings

Thanks for your input, @chromey !

@basti1302 @kirrg001 FYI, as aborting this work was a democratic decision made by the team.

basti1302:
Hi @chromey please file this as a feature request at https://support.instana.com/hc/en-us/requests/new. Thank you.

chromey:
Hi @basti1302, I created https://support.instana.com/hc/en-us/requests/27044

Thank you for reconsidering this!

PR #459, opened 2022-01-13
Removed .npmignore

kirrg001:
See https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files

Either whitelist or blacklist - we should not do both. Discussed in Slack

- removed .npmignore
- cleanup files directive
- added changelog.md to tgz

If you run `npm pack` in a package, you can see that everything is still correctly packed as before


basti1302 (approved):
LGTM, thanks for the cleanup.

PR #464, opened 2022-01-25
feat(aws-lambda): added support for SSM parameter store

kirrg001:
refs 6483

Added support for AWS SSM.

Instead of defining the plain `INSTANA_AGENT_KEY` as env variable, the customer can use the parameter store.
They can set `INSTANA_SSM_PARAM_NAME=/my/ssm/path`. We try to fetch the value from the AWS API. The challenge here is that the API call is async.

**Attention**: If you set both env variables, `INSTANA_AGENT_KEY` is always stronger. Can be discussed

### DIFF
--- a/packages/aws-lambda/src/ssm.js
+++ b/packages/aws-lambda/src/ssm.js
@@ -1,14 +1,13 @@
 /*
  * (c) Copyright IBM Corp. 2022
- * (c) Copyright Instana Inc. and contributors 2022
  */
 
 'use strict';
 
 const ENV_NAME = 'INSTANA_SSM_PARAM_NAME';
-let fetchedValue;
-let envValue;
-let errorFromAWS;
+let fetchedValue = null;
+let envValue = null;
+let errorFromAWS = null;
 let initTimeoutInMs = 0;
 
 module.exports.reset = () => {
@@ -19,11 +18,9 @@ module.exports.reset = () => {
 };
 
 module.exports.isUsed = () => !!envValue;
+module.exports.getValue = () => envValue;
 
 module.exports.validate = () => {
-  // NOTE: Always reset the env value, we do not wanne cache, because it might change
-  envValue = null;
-
   const _envValue = process.env[ENV_NAME];
 
   if (!_envValue || !_envValue.length) {
@@ -31,8 +28,7 @@ module.exports.validate = () => {
   }
 
   envValue = _envValue;
-
-  return envValue;
+  return true;
 };
 
 module.exports.init = ({ logger }) => {
@@ -57,13 +53,11 @@ module.exports.init = ({ logger }) => {
      * Your code runs in an environment that includes the AWS SDK for JavaScript,
      * with credentials from an AWS Identity and Access Management (IAM) role that you manage.
      */
-    // eslint-disable-next-line
+    // eslint-disable-next-line instana/no-unsafe-require, import/no-extraneous-dependencies
     AWS = require('aws-sdk');
 
     // https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html
-    AWS.config.update({ region: process.env.AWS_REGION });
-
-    const ssm = new AWS.SSM();
+    const ssm = new AWS.SSM({ region: process.env.AWS_REGION });
     const params = {
       Name: envValue,
       // See https://docs.aws.amazon.com/cli/latest/reference/ssm/get-parameter.html#options
@@ -95,7 +89,7 @@ module.exports.init = ({ logger }) => {
   }
 };
 
-exports.waitAndGetInstanaKey = callback => {
+module.exports.waitAndGetInstanaKey = callback => {
   // CASE: We already fetched the SSM value, skip & save time
   if (fetchedValue) {
     return callback(null, fetchedValue);
@@ -106,10 +100,12 @@ exports.waitAndGetInstanaKey = callback => {
   }
 
   const endInMs = Date.now();
+  const awsTimeoutInMs = 1000;
+
   // CASE: the time between ssm lib initialisation and waitAndGetInstanaKey call
   //       (which is the end of the customers lambda handler) is already too big to wait for the AWS response
-  if (endInMs - initTimeoutInMs > (process.env.INSTANA_LAMBDA_SSM_AWS_TIMEOUT_IN_MS || 1000)) {
-    return callback('Stopped waiting for AWS SSM response.');
+  if (endInMs - initTimeoutInMs > awsTimeoutInMs) {
+    return callback(`Stopped waiting for AWS SSM response after ${awsTimeoutInMs}ms.`);
   }
 
   /**
@@ -119,7 +115,18 @@ exports.waitAndGetInstanaKey = callback => {
    *
    * In our tests it takes usually ~>150ms (remote call)
    */
-  const stopIntervalAfterMs = process.env.INSTANA_LAMBDA_SSM_TIMEOUT_IN_MS || 150;
+  let stopIntervalAfterMs = 250;
+  let ssmTimeOutEnv = process.env.INSTANA_LAMBDA_SSM_TIMEOUT_IN_MS;
+
+  if (ssmTimeOutEnv && ssmTimeOutEnv.length > 0) {
+    ssmTimeOutEnv = Number(ssmTimeOutEnv);
+
+    // NOTE: Customer could set the timeout higher than the lambda timeout, but that is up to him
+    if (!isNaN(ssmTimeOutEnv)) {
+      stopIntervalAfterMs = ssmTimeOutEnv;
+    }
+  }
+
   const start = Date.now();
 
   const waiting = setInterval(() => {
--- a/packages/aws-lambda/src/ssm.js
+++ b/packages/aws-lambda/src/ssm.js
@@ -5,6 +5,7 @@
 'use strict';
 
 const ENV_NAME = 'INSTANA_SSM_PARAM_NAME';
+const ENV_DECRYPTION = 'INSTANA_SSM_DECRYPTION';
 let fetchedValue = null;
 let envValue = null;
 let errorFromAWS = null;
@@ -60,9 +61,12 @@ module.exports.init = ({ logger }) => {
     const ssm = new AWS.SSM({ region: process.env.AWS_REGION });
     const params = {
       Name: envValue,
-      // See https://docs.aws.amazon.com/cli/latest/reference/ssm/get-parameter.html#options
-      // We always expect a string here. Never a decrypt
-      WithDecryption: false
+      /**
+       * See https://docs.aws.amazon.com/cli/latest/reference/ssm/get-parameter.html#options
+       * The value in the parameter store was either created with type "string" or "safestring"
+       * A "safestring" uses a KMS key to encrypt/decrypt the value in the store.
+       */
+      WithDecryption: process.env[ENV_DECRYPTION] === 'true'
     };
 
     logger.debug(`INSTANA_SSM_PARAM_NAME is ${envValue}.`);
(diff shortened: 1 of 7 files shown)


packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,107 @@
...
+  if (!_envValue || !_envValue.length) {
+    return false;
+  }
+
+  envValue = _envValue;
+
+  return envValue;
+};
+
+module.exports.init = ({ logger }) => {
```
kirrg001, line 25: Maybe consider adding a global timeout vsriablr

init vs waitAndGet -> timeout between these cllas

packages/aws-lambda/test/ssm_test.js
```diff
@@ -0,0 +1,156 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2019
```
kirrg001, line 3: Fix dates tomorrow

willianpc: You can remove the Instana Copyright, I guess.
It is being removed from the lint rule.
Also, as this is a new file, you should probably change 2021 to 2022 :)

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
...
+  const start = Date.now();
+
+  const waiting = setInterval(() => {
+    const end = Date.now();
+
+    if (fetchedValue) {
+      clearInterval(waiting);
+
+      callback(null, fetchedValue);
+    } else if (end - start > stopIntervalAfterMs) {
```
willianpc, line 132: Maybe `stopIntervalAfterMs` should have a max value, as it can be set by customers through `process.env.INSTANA_LAMBDA_SSM_TIMEOUT_IN_MS`.
If the value is too high, it could cause the setInterval ticks to run for a long period of time.
I don't have enough experience with AWS Lambda, but couldn't this hold the process for too long or cause some kind of unwanted behavior if the interval ticks continue for too long? I believe that "too long" here means 1 or 2 seconds, or something like this, no?

willianpc: Or maybe we can have a third condition that guarantees that the setInterval tick loop ends at some point.

willianpc (approved):
Overall, looks good to me.
I added some comments that I consider important to review, in regards of environment variables and potential setInterval long looping ticks.
Some other comments are minor suggestions that don't impact the implementation, so I leave it to you to decide if it should be changed or not.

Cheers!

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
...
+  } catch (err) {
+    logger.warn('AWS SDK not available.');
+    errorFromAWS =
+      'Could not fetch instana key from SSM parameter store using ' +
+      `"${process.env.INSTANA_SSM_PARAM_NAME}", because the AWS SDK is not available. ` +
+      `Reason: ${err.message}`;
+  }
+};
+
+exports.waitAndGetInstanaKey = callback => {
```
kirrg001, line 98: Feedback: consistent usage of `module.exports`

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
...
+  fetchedValue = null;
+  envValue = null;
+  errorFromAWS = null;
+  initTimeoutInMs = 0;
+};
+
+module.exports.isUsed = () => !!envValue;
+
+module.exports.validate = () => {
+  // NOTE: Always reset the env value, we do not wanne cache, because it might change
```
basti1302, line 24: How would the env var change during the lifetime of a Lambda handler instance? I think it can only change when a new Lambda handler with a different environment configuration has been deployed, which will start a new Lambda handler instance, and thereby also restart the underlying container/process. IMO it would be valid to cache the result of this check globally.

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
...
+
+  const _envValue = process.env[ENV_NAME];
+
+  if (!_envValue || !_envValue.length) {
+    return false;
+  }
+
+  envValue = _envValue;
+
+  return envValue;
```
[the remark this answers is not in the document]
kirrg001: Yeah consistent type returns is better.
Will change :)

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
...
+  initTimeoutInMs = Date.now();
+
+  try {
+    /**
+     * From https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html:
+     *
+     * Your code runs in an environment that includes the AWS SDK for JavaScript,
+     * with credentials from an AWS Identity and Access Management (IAM) role that you manage.
+     */
+    // eslint-disable-next-line
```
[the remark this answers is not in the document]
basti1302: You can also combine both, like `eslint-disable-next-line $ruleId`.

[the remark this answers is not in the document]
kirrg001: Sure will do 👍

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
...
+    AWS = require('aws-sdk');
+
+    // https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html
+    AWS.config.update({ region: process.env.AWS_REGION });
```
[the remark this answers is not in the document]
kirrg001: This var is a default var in AWS lambda. Customer does not need to do anything :) 
See https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html

> Lambda runtimes set several environment variables during initialization.

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+  // CASE: Customer has set INSTANA_SSM_PARAM_NAME, but we were not able to fetch the value from AWS
+  if (errorFromAWS) {
+    return callback(errorFromAWS);
+  }
+
+  const endInMs = Date.now();
+  // CASE: the time between ssm lib initialisation and waitAndGetInstanaKey call
+  //       (which is the end of the customers lambda handler) is already too big to wait for the AWS response
+  if (endInMs - initTimeoutInMs > (process.env.INSTANA_LAMBDA_SSM_AWS_TIMEOUT_IN_MS || 1000)) {
```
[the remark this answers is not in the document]
kirrg001: I only added `INSTANA_LAMBDA_SSM_AWS_TIMEOUT_IN_MS` for the test env to reduce the execution time.
We won't document the env variable. 

I think I'll remove this variable and switch to function options :)

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
...
+    return callback('Stopped waiting for AWS SSM response.');
```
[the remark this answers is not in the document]
kirrg001: Yeah good idea. Will add the value to the error message :)

packages/aws-lambda/src/ssm.js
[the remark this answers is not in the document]
basti1302: I think we should rather pass the region into the AWS.SSM() constructor instead of making this change globally. 

I think new `AWS.SSM({region: process.env.AWS_REGION});` should work as well and it might be less invasive.

See the section "Constructor Details" on https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SSM.html

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /**
+   * Inside AWS the call to `getParameter` mostly takes 30-50ms
+   * Because we initialise the fetch already before the customer's handler is called,
+   * the chance is very high that the interval is not even used.
+   *
+   * In our tests it takes usually ~>150ms (remote call)
+   */
+  const stopIntervalAfterMs = process.env.INSTANA_LAMBDA_SSM_TIMEOUT_IN_MS || 150;
```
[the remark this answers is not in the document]
kirrg001: Yeah same as above :) Will switch to function options
Thanks for feedback

packages/aws-lambda/src/ssm.js
kirrg001: > as it can be set by customers through process.env.INSTANA_LAMBDA_SSM_TIMEOUT_IN_MS.

Same as above :) Will switch to function options. Customer should not set these values

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
+/*
+ * (c) Copyright IBM Corp. 2022
+ * (c) Copyright Instana Inc. and contributors 2022
+ */
+
+'use strict';
+
+const ENV_NAME = 'INSTANA_SSM_PARAM_NAME';
```
basti1302, line 8: Is there a specific need to make the parameter name configurable? Or could we also just always fetch the same parameter name and tell users which parameter they need to set?

packages/aws-lambda/src/ssm.js
```diff
@@ -0,0 +1,143 @@
...
+    AWS = require('aws-sdk');
+
+    // https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html
+    AWS.config.update({ region: process.env.AWS_REGION });
+
+    const ssm = new AWS.SSM();
+    const params = {
+      Name: envValue,
+      // See https://docs.aws.amazon.com/cli/latest/reference/ssm/get-parameter.html#options
+      // We always expect a string here. Never a decrypt
```
basti1302, line 70: I am not 100% sure about that. Since the use case is storing the Instana agent key securely, we might also expect users to store them as a secure key?

packages/aws-lambda/src/ssm.js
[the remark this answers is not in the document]
kirrg001: Yeah great idea 👍

packages/aws-lambda/src/ssm.js
kirrg001: Good question :) 

Two reasons why I think we want to leave it as it is

1. Naming patterns. If you add a new value to the parameter store you need to choose a name.
I can imagine that companies use naming patterns such as `/company/lambda/x` 

2. Visibility. Setting INSTANA_SSM_PARAM_NAME in their env shows them that they explicitly use SSM over the default way. 

I am happy to discuss that topic on a call if you think we want to define a name for the customers :)

basti1302: OTOH, maybe letting users configure the key might be more user friendly. TBH, I am not familiar enough with SSM to know how users would expect this to work (configurable or fixed key). Feel free to ignore this comment.

packages/aws-lambda/src/ssm.js
kirrg001: Yeah you are right. Will add support for decrypt too :) 

From https://docs.aws.amazon.com/kms/latest/developerguide/services-parameter-store.html

> To decrypt the parameter value before returning it, set the WithDecryption parameter of GetParameter to true. When you use WithDecryption, Parameter Store calls the AWS KMS Decrypt operation on your behalf to decrypt the parameter value. As a result, the GetParameter request returns the parameter with a plaintext parameter value, as shown in the following example.

packages/aws-lambda/src/ssm.js
basti1302: > I am happy to discuss that topic on a call if you think we want to define a name for the customers :)

No, I think that's fine. We would still want an env var anyway to enable SSM usage (something like `INSTANA_USE_SSM=true`) and if keys can have a structure as you pointed out, letting users configure it absolutely makes sense. 👍

packages/aws-lambda/src/ssm.js
basti1302: Do we need both? Or only encrypted? If we offer both, we probably need a way for users to tell us which one it is, right?

packages/aws-lambda/src/ssm.js
kirrg001: > I think it can only change when a new Lambda handler with a different environment configuration has been deployed

Yeah I was't sure about that. Usually it works like that - a restart of a the node process is required
My thinking was that it won't hurt to re-read the value (performance wise) if we are not 100% sure

But let me just remove the two lines and cache the value :)

packages/aws-lambda/src/ssm.js
kirrg001: Added discussion to the ticket. Waiting for reply

kirrg001: Discussed: we want to support both

We might want to introduce an extra env variable e.g. `INSTANA_SSM_DECRYPTION=Boolean`

Any other suggestions?

basti1302 (approved):

PR #465, opened 2022-01-25
fix(deps): bumped node-fetch to ^2.6.7 (security patch)

kirrg001:
no issue

See https://github.com/advisories/GHSA-r683-j2x4-v87g


willianpc (approved):
LGTM.
Thanks for the fix <3

PR #467, opened 2022-01-28
chore: covered graphql v15 & fixed tests

kirrg001:
refs 85796

- we have not tested against graphql v15
- the packages we were using in our tests were no longer compatible and are no longer maintained
- there is a new package called graphql-ws

## TODO
- [ ] cleanup
- [ ] add graphql v16 installation
- [ ] `const folder = process.env.GRAPHQL_REQUIRE || 'graphql';` different solution?


packages/collector/test/tracing/protocols/graphql/test.js
```diff
@@ -24,135 +24,155 @@ const globalAgent = require('../../../globalAgent');
 
 const agentControls = globalAgent.instance;
 
-const mochaSuiteFn =
-  !supportedVersion(process.versions.node) || semver.lt(process.versions.node, '8.5.0') ? describe.skip : describe;
-
-mochaSuiteFn('tracing/graphql', function () {
+const skipNodeVersion = {
```
kirrg001, line 27: The diff in this file looks big, but not a lot has changed

1. wrap the whole execution to run v14,v15,v16
2. forward GRAPHQL_VERSION to the sub processes

packages/collector/test/tracing/protocols/graphql/test.js
```diff
@@ -24,135 +24,155 @@ const globalAgent = require('../../../globalAgent');
...
+  14: semver.lt(process.versions.node, '10.0.0'),
+  15: semver.lt(process.versions.node, '10.0.0'),
+  16: semver.lt(process.versions.node, '12.0.0')
+};
+
+describe('tracing/graphql', function () {
   this.timeout(config.getTestTimeout() * 2);
 
-  globalAgent.setUpCleanUpHooks();
+  [14, 15, 16].forEach(graphqlVersion => {
```
basti1302, line 37: I really like this way of testing multiple versions of the package. ❤️

packages/collector/test/tracing/protocols/graphql/mockVersion.js
```diff
@@ -0,0 +1,45 @@
...
+
+const originalOnFileLoad = requireHook.onFileLoad;
+requireHook.onFileLoad = function onFileLoad() {
+  if (arguments[0].toString() !== '/\\/graphql\\/execution\\/execute.js/') {
+    return originalOnFileLoad.apply(this, arguments);
+  }
+  const str = `/${GRAPHQL_REQUIRE}/execution/execute.js`;
+  const reg = new RegExp(str, '');
+  arguments[0] = reg;
+  return originalOnFileLoad.apply(this, arguments);
```
basti1302, line 44: Interesting solution. :) Maybe a comment would help, I needed a bit to wrap my head around this. Just to see if I am getting this correctly: We are effectively monkey-patching our own require-hook utility here to make it work despite the fact that we delegate all graphql-related require calls to different file system locations, right?

packages/collector/test/tracing/protocols/graphql/mockVersion.js
```diff
@@ -0,0 +1,45 @@
...
+
+  mock('graphql/language', `${GRAPHQL_REQUIRE}/language`);
+  mock('graphql/execution', `${GRAPHQL_REQUIRE}/execution`);
+  mock('graphql/utilities', `${GRAPHQL_REQUIRE}/utilities`);
+  mock('graphql/jsutils', `${GRAPHQL_REQUIRE}/jsutils`);
+  mock('graphql/validation', `${GRAPHQL_REQUIRE}/validation`);
+  mock('graphql/type', `${GRAPHQL_REQUIRE}/type`);
+  mock('graphql/polyfills', `${GRAPHQL_REQUIRE}/polyfills`);
+  mock('graphql/subscription', `${GRAPHQL_REQUIRE}/subscription`);
+  mock('graphql/error', `${GRAPHQL_REQUIRE}/error`);
```
basti1302, line 29: This might break in interesting, non-obvious ways at some point when a graphql release comes with a different set of sub packages. But I guess that is an acceptable risk.

basti1302 (approved):
Looks awesome! 👍

packages/collector/test/tracing/protocols/graphql/mockVersion.js
kirrg001: Right:) You are right, hard to understand without comments. Will add some more comments

packages/collector/test/tracing/protocols/graphql/mockVersion.js
kirrg001: Agree. Lets see if this will create trouble in the future :)

PR #470, opened 2022-02-01
fix(tracing): fix version constraint for http2 instrumentation

basti1302:
Until Node.js v8.7.0, the http2 module was behind a flag (--expose-http2).
Unconditionally requiring it on Node.js 8.4.0 - 8.7.0 will crash the Node.js
process when that flag has not been provided. Starting with Node.js v8.8.0,
the http2 module became available without that flag.

See also:
* https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#8.4.0
* https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#8.8.0

### DIFF
--- a/packages/core/src/tracing/instrumentation/protocols/http2Client.js
+++ b/packages/core/src/tracing/instrumentation/protocols/http2Client.js
@@ -24,7 +24,7 @@ let HTTP2_HEADER_PATH;
 let HTTP2_HEADER_STATUS;
 
 exports.init = function init(config) {
-  if (semver.gte(process.versions.node, '8.4.0')) {
+  if (semver.gte(process.versions.node, '8.8.0')) {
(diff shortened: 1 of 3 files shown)


packages/core/src/tracing/instrumentation/protocols/http2Client.js
```diff
@@ -24,7 +24,7 @@ let HTTP2_HEADER_PATH;
 let HTTP2_HEADER_STATUS;
 
 exports.init = function init(config) {
-  if (semver.gte(process.versions.node, '8.4.0')) {
```
kirrg001, line 27: Wondering about the context of this change. Why has it worked before? 🤔

kirrg001 (approved):
LGTM

PR #478, opened 2022-02-07
chore: upgraded mocha to v8

kirrg001:
refs 80206

- mocha@9 dropped Node 10


packages/core/test/test_util/retry.js
```diff
@@ -29,5 +29,11 @@ module.exports = function retry(fn, time, until) {
 
   return delay(time / 20)
     .then(fn)
-    .catch(() => retry(fn, time, until));
+    .catch(err => {
+      if (err.name === 'AssertionError') {
+        throw err;
+      }
```
basti1302, line 35: @kirrg001 I think this might break a lot of tests. We actually retry assertions that might only become true after a while (that is, we wait until the assertion becomes true). And we do that in quite a few tests I think. Directly failing on AssertionError without retrying it will break those tests, I think.

kirrg001: 🤔 My expectation/understanding from the retry function is to retry when the data is not available

e.g. agentControls.getSpans() throws error -> retry

But an AssertionError should not be retried 🤔 Maybe there are currently edge cases, where we misuse retry?

kirrg001: The problem with this function is that it hides the underlying problem when a test is red, which is uncomfortable

kirrg001: Ok the underlying problem is not the retry function, it's the test using a different timeout than the retry function
We might want to wrap mocha describe in the future to ensure the timeouts are set by default

basti1302: Right, the `retry` function rethrows the last error it encounters when it finally gives up, so that function on its own should not hide anything. 

But setting the Mocha timeout and the timeout for the  `retry` function to reasonable values is definitely a bit of a mess (that is, fiddly/fragile/error prone) in the code base at the moment. You can provide a specific timeout when calling `retry` (second parameter), and you can set the Mocha timeout per test/suite, as we often do, like [here](https://github.com/instana/nodejs/blob/main/packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test.js#L60). 

 Any suggestions how to improve this? I'm all ears. How would you wrap the describe function and what would the wrapper do? 
 
 Although that is probably a discussion for a different PR anyway.

basti1302 (approved):
LGTM

PR #482, opened 2022-02-09
build: do not tag 1.x images with latest

basti1302:
**DO NOT MERGE YET**

After releasing 2.0.0, only 2.x images will be tagged as latest. This change prevents 1.x images as
being tagged as latest.

- [x] change origin of pull request to v1 if v2 got merged into main 
- [x] https://github.com/instana/nodejs/pull/482#issuecomment-1085894782

### DIFF
--- a/packages/serverless/ci/pipeline.yaml
+++ b/packages/serverless/ci/pipeline.yaml
@@ -315,6 +315,9 @@ jobs:
             - bash
             - -ceux
             - |
+                # Passing in PACKAGE_VERSION=1 will use the latest 1.x package and also avoid tagging the resulting
+                # container image as latest.
+                PACKAGE_VERSION=1 \
                 BUILD_LAYER_WITH=npm \
                 CONTAINER_REGISTRY_USER=iamapikey \
                 CONTAINER_REGISTRY_PASSWORD=((concourse-icr-containers-public.password)) \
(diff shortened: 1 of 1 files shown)


kirrg001 (approved):
Great!

kirrg001:
I think we also need to remove the latest tagging in the publish-layer script when we merge v2 into main.
Maybe we forgot that because in the beginning we planned two release different layer names for different major versions.

IMO main should not tag the docker images with latest.
See https://github.com/instana/nodejs/blob/v1.140.0/packages/aws-lambda/layer/bin/publish-layer.sh#L265

Our docs say

> FROM icr.io/instana/aws-lambda-nodejs:latest as instana-layer

basti1302:
> I think we also need to remove the latest tagging in the publish-layer script when we merge v2 into main. Maybe we forgot that because in the beginning we planned two release different layer names for different major versions.
> 
> IMO main should not tag the docker images with latest. See https://github.com/instana/nodejs/blob/v1.140.0/packages/aws-lambda/layer/bin/publish-layer.sh#L265
> 
> Our docs say
> 
> > FROM icr.io/instana/aws-lambda-nodejs:latest as instana-layer

Hm, I think this change: https://github.com/instana/nodejs/pull/482/files#diff-57474cc8dddbba7d64d7180d178eab239cd0448c8c2879abc2193a7e960e2720R318-R320

together with the `if [[ $PACKAGE_VERSION == latest ]]; then` in the script you linked to should be enough to not tag v1 versions as latest once we have merged this.

PR #483, opened 2022-02-09
build(lambda): add support for publishing a layer for a specific version

basti1302:
This will serve as a preparation for adding another commit to https://github.com/instana/nodejs/pull/482, so that an 1.x Lambda layer image will not be tagged as `latest` after 2.0.0 has been published.

In contrast to the changes in #482, these changes here can be merged into main immediately and also into v2.

When that is done I will rebase #482 on this and only add a minor change to pipeline.yaml to pass PACKAGE_VERSION=1 to from pipeline.yaml to `publish-layer.sh`.

### DIFF
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -54,7 +54,11 @@ EOF
 
 cd `dirname $BASH_SOURCE`/..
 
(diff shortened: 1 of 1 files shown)


kirrg001 (approved):
LGTM

Do you create another PR to set PACKAGE_VERSION for 1.x after 2.0 got merged into main?

PR #486, opened 2022-02-18
Instrumentation of node-rdfafka/kafka-avro

willianpc:
### Node-RdKafka Limitations

These limitations impact how the tracer performs and must be explicit to customers in the documentation:

 * The **Producer as a stream** can only have span correlation if the Writable option **objectMode is set to true**, otherwise, there is no way to append the Instana headers to it.
 * The Producer, as stream or standard API **cannot** propagate span correlation when headerFormat is set to `binary`. More info here: https://github.com/Blizzard/node-rdkafka/pull/935.
 * If the option `dr_cb` is not set to true, we assume that the message was sent, but without any guarantee.

### Relevant Notes

* Kafka-Avro currently works up until Node.js v14.17 due to a dependency to an old version of Node-RdKafka.
* Because of the above, we install kafka-avro as an optional dependency. However, even in Node.js v16 and above, npm will attempt to install it, but the installation won't fail if an error occurs. This means two things:
   * The CI time will increase also for Node v16 > as there will be actually an installation attempt.
   * There may be some garbage in node_modules after the installation attempt. Some packages installed during the installation that fails seem to not being cleaned up.
 * In order to test kafka-avro, we had to add Schema Registry to docker-compose.yml and config.yml, also increasing the time of the CI build.
 * Kafka configuration was changed in order to support and communicate properly with the Schema Registry server
 
 Tasks

 - [x] increase code coverage for error cases @kirrg001 
 - [x] optimise readability in instrumention @kirrg001 
 - [x] add avro installation to the test for v14 @kirrg001 
 - [x] fix level !== 1 @kirrg001 
 - [x] objectMode @kirrg001 
 - [x] fix https://github.com/instana/nodejs/pull/486#discussion_r818908700 @kirrg001 
 - [x] fix https://github.com/instana/nodejs/pull/486#discussion_r818827112

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -129,9 +129,6 @@
         "uuid": "^8.3.2",
         "winston": "^3.3.3",
         "ws": "^7.5.6"
-      },
-      "optionalDependencies": {
-        "kafka-avro": "^3.1.1"
       }
     },
     "misc/eslint-plugin": {
@@ -14910,18 +14907,6 @@
       "integrity": "sha512-dOy+3AuW3a2wNbZHIuMZpTcgjGuLU/uBL/ubcZF9OXbDo8ff4O8yVp5Bf0efS8uEoYo5q4Fx7dY9OgQGXgAsQA==",
       "dev": true
     },
-    "node_modules/ansicolors": {
-      "version": "0.2.1",
-      "resolved": "https://registry.npmjs.org/ansicolors/-/ansicolors-0.2.1.tgz",
-      "integrity": "sha1-vgiVmQl7dKXJxKhKDNvNtivYeu8=",
-      "optional": true
-    },
-    "node_modules/ansistyles": {
-      "version": "0.1.3",
-      "resolved": "https://registry.npmjs.org/ansistyles/-/ansistyles-0.1.3.tgz",
-      "integrity": "sha1-XeYEFb2gcbs3EnhUyGT0GyMlRTk=",
-      "optional": true
-    },
     "node_modules/any-promise": {
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/any-promise/-/any-promise-1.3.0.tgz",
@@ -16014,15 +15999,6 @@
         "url": "https://github.com/sponsors/ljharb"
       }
     },
-    "node_modules/avsc": {
-      "version": "5.7.3",
-      "resolved": "https://registry.npmjs.org/avsc/-/avsc-5.7.3.tgz",
-      "integrity": "sha512-uUbetCWczQHbsKyX1C99XpQHBM8SWfovvaZhPIj23/1uV7SQf0WeRZbiLpw0JZm+LHTChfNgrLfDJOVoU2kU+A==",
-      "optional": true,
-      "engines": {
-        "node": ">=0.11"
-      }
-    },
     "node_modules/avvio": {
       "version": "5.9.0",
       "resolved": "https://registry.npmjs.org/avvio/-/avvio-5.9.0.tgz",
@@ -16133,7 +16109,7 @@
       "version": "0.21.4",
       "resolved": "https://registry.npmjs.org/axios/-/axios-0.21.4.tgz",
       "integrity": "sha512-ut5vewkiu8jjGBdqpM44XxjuCjq9LAKeHVmoVfHVzy8eHgxxq8SbAVQNovDA8mVi05kP0Ea/n/UzcSHcTJQfNg==",
-      "devOptional": true,
+      "dev": true,
       "dependencies": {
         "follow-redirects": "^1.14.0"
       }
@@ -16211,7 +16187,7 @@
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
       "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
-      "devOptional": true
+      "dev": true
     },
     "node_modules/base": {
       "version": "0.11.2",
@@ -16379,7 +16355,7 @@
       "version": "1.5.0",
       "resolved": "https://registry.npmjs.org/bindings/-/bindings-1.5.0.tgz",
       "integrity": "sha512-p2q/t/mhvuOj/UeLlV6566GD/guowlr0hHxClI0W9m7MWYkL1F0hLo+0Aexs9HSPCtR1SXQ0TD3MMKrXZajbiQ==",
-      "devOptional": true,
+      "dev": true,
       "dependencies": {
         "file-uri-to-path": "1.0.0"
       }
@@ -16442,7 +16418,7 @@
       "version": "3.7.2",
       "resolved": "https://registry.npmjs.org/bluebird/-/bluebird-3.7.2.tgz",
       "integrity": "sha512-XpNj6GDQzdfW+r2Wnn7xiSAd7TM3jzkxGXBGTtWKuSXv1xUV+azxAm8jdWZN06QTQk+2N2XB9jRDkvbmQmcRtg==",
-      "devOptional": true
+      "dev": true
     },
     "node_modules/body-parser": {
       "version": "1.19.0",
@@ -16590,7 +16566,7 @@
       "version": "1.1.11",
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
       "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
-      "devOptional": true,
+      "dev": true,
       "dependencies": {
         "balanced-match": "^1.0.0",
         "concat-map": "0.0.1"
@@ -16823,53 +16799,6 @@
       },
       "engines": {
         "node": ">=10"
-      }
-    },
-    "node_modules/bunyan": {
-      "version": "1.8.15",
-      "resolved": "https://registry.npmjs.org/bunyan/-/bunyan-1.8.15.tgz",
-      "integrity": "sha512-0tECWShh6wUysgucJcBAoYegf3JJoZWibxdqhTm7OHPeT42qdjkZ29QCMcKwbgU1kiH+auSIasNRXMLWXafXig==",
-      "engines": [
-        "node >=0.10.0"
-      ],
-      "optional": true,
-      "bin": {
-        "bunyan": "bin/bunyan"
-      },
-      "optionalDependencies": {
-        "dtrace-provider": "~0.8",
-        "moment": "^2.19.3",
-        "mv": "~2",
-        "safe-json-stringify": "~1"
-      }
-    },
-    "node_modules/bunyan-format": {
-      "version": "0.2.1",
-      "resolved": "https://registry.npmjs.org/bunyan-format/-/bunyan-format-0.2.1.tgz",
-      "integrity": "sha1-pLOw2ABwqGUnlBcmnj8A/wL7y0c=",
-      "optional": true,
-      "dependencies": {
-        "ansicolors": "~0.2.1",
-        "ansistyles": "~0.1.1",
-        "xtend": "~2.1.1"
-      }
-    },
-    "node_modules/bunyan-format/node_modules/object-keys": {
-      "version": "0.4.0",
-      "resolved": "https://registry.npmjs.org/object-keys/-/object-keys-0.4.0.tgz",
-      "integrity": "sha1-KKaq50KN0sOpLz2V8hM13SBOAzY=",
-      "optional": true
-    },
-    "node_modules/bunyan-format/node_modules/xtend": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/xtend/-/xtend-2.1.2.tgz",
-      "integrity": "sha1-bv7MKk2tjmlixJAbM3znuoe10os=",
-      "optional": true,
-      "dependencies": {
-        "object-keys": "~0.4.0"
-      },
-      "engines": {
-        "node": ">=0.4"
       }
     },
     "node_modules/busboy": {
@@ -17336,12 +17265,6 @@
       "resolved": "https://registry.npmjs.org/cint/-/cint-8.2.1.tgz",
       "integrity": "sha1-cDhrG0jidz0NYxZqVa/5TvRFahI=",
       "dev": true
-    },
-    "node_modules/cip": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/cip/-/cip-1.0.1.tgz",
-      "integrity": "sha1-FzAzSiIvhWXgl/YMM5BxiUsPva8=",
-      "optional": true
     },
     "node_modules/class-utils": {
       "version": "0.3.6",
@@ -18103,7 +18026,7 @@
       "version": "0.0.1",
       "resolved": "https://registry.npmjs.org/concat-map/-/concat-map-0.0.1.tgz",
       "integrity": "sha1-2Klr13/Wjfd5OnMDajug1UBdR3s=",
-      "devOptional": true
+      "dev": true
     },
     "node_modules/concat-stream": {
       "version": "1.6.2",
@@ -19639,19 +19562,6 @@
       "resolved": "https://registry.npmjs.org/dottie/-/dottie-2.0.2.tgz",
       "integrity": "sha512-fmrwR04lsniq/uSr8yikThDTrM7epXHBAAjH9TbeH3rEA8tdCO7mRzB9hdmdGyJCxF8KERo9CITcm3kGuoyMhg==",
       "dev": true
-    },
-    "node_modules/dtrace-provider": {
-      "version": "0.8.8",
-      "resolved": "https://registry.npmjs.org/dtrace-provider/-/dtrace-provider-0.8.8.tgz",
-      "integrity": "sha512-b7Z7cNtHPhH9EJhNNbbeqTcXB8LGFFZhq1PGgEvpeHlzd36bhbdTWoE/Ba/YguqpBSlAPKnARWhVlhunCMwfxg==",
-      "hasInstallScript": true,
-      "optional": true,
-      "dependencies": {
-        "nan": "^2.14.0"
-      },
-      "engines": {
-        "node": ">=0.10"
-      }
     },
     "node_modules/duplexer": {
       "version": "0.1.2",
@@ -21890,7 +21800,7 @@
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/file-uri-to-path/-/file-uri-to-path-1.0.0.tgz",
       "integrity": "sha512-0Zt+s3L7Vf1biwWZ29aARiVYLx7iMGnEUl9x33fbB/j3jR81u/O2LbqK+Bm1CDSNDKVtJ/YjwY7TUd5SkeLQLw==",
-      "devOptional": true
+      "dev": true
     },
     "node_modules/fill-keys": {
       "version": "1.0.2",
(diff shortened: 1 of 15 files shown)


packages/collector/test/tracing/messaging/kafka-avro/producer.js
```diff
@@ -0,0 +1,80 @@
+/*
+ * (c) Copyright IBM Corp. 2022
+ */
+
+'use strict';
+
+require('../../../../')({
+  tracing: {
+    kafka: {
+      headerFormat: 'string'
```
kirrg001, line 10: Notes from call:

Because of the [binary bug in rdkafka](https://github.com/Blizzard/node-rdkafka/pull/935) the span correlation wouldn't work at all in binary mode (which is still the default mode in kafka instrumention)

Node.js tracer has already added support for kafka string headers although the kafka header migration has not even started yet (or is not finished yet).
Setting the `headerFormat` to 'string' here means, rdkafka/avro instrumention works if the consumer tracer and the producer tracer support and use string format.

Therefor the customer needs to set `headerFormat: string`.

@willianpc as discussed: `headerFormat` is not documented yet in our public docs. We might want to wait for @basti1302 opinion here

package-lock.json
```diff
@@ -104,6 +104,7 @@
         "nats": "^1.4.12",
         "node-fetch": "^2.6.1",
         "node-nats-streaming": "^0.3.2",
+        "node-rdkafka": "^2.12.0",
```
kirrg001, line 107: I am getting a diff for package-lock.json file on Node v16.

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -0,0 +1,412 @@
...
+
+function instrumentProducer(ProducerClass) {
+  shimmer.wrap(ProducerClass.prototype, 'produce', shimProduce);
+}
+
+function instrumentConsumer(module) {
+  const className = 'KafkaConsumer';
+  const originalKafkaConsumer = module[className];
+
+  module[className] = function () {
```
kirrg001, line 42: _Suggestion_: Use named functions

Original: [Function: KafkaConsumer]
Mocked: [Function (anonymous)]

packages/collector/test/tracing/messaging/node-rdkafka/test.js
```diff
@@ -0,0 +1,290 @@
...
+  globalAgent.setUpCleanUpHooks();
+  const agentControls = globalAgent.instance;
+
+  describe('tracing enabled, no suppression', function () {
+    const producerControls = new ProcessControls({
+      appPath: path.join(__dirname, 'producer'),
+      port: 3216,
+      useGlobalAgent: true,
+      env: {
+        RDKAFKA_PRODUCER_DELIVERY_CB: getNextDeliveryCb()
```
kirrg001, line 68: We are currently testing for _"tracing enabled, no suppression"_ only with `RDKAFKA_PRODUCER_DELIVERY_CB: true`. I was wondering if we also want to test with `false`. My initial expectation here was that we have to iterate over the `producerEnableDeliveryCbOptions`.

packages/collector/test/tracing/messaging/node-rdkafka/test.js
```diff
@@ -0,0 +1,290 @@
...
+          useGlobalAgent: true,
+          env: {
+            RDKAFKA_CONSUMER_AS_STREAM: consumerMethod === 'stream' ? 'true' : 'false'
+          }
+        });
+
+        ProcessControls.setUpHooksWithRetryTime(retryTime, consumerControls);
+
+        producerApiMethods.forEach(producerMethod => {
+          [false, 'sender'].forEach(withError => {
```
kirrg001, line 88: Is it possible that something is missing in the tests regarding error cases?

See line https://github.com/instana/nodejs/pull/486/files#diff-1b77510f78eb55910153dd198bee01c7d0e63aa32d30da88adae14562db0d5f8R143

`withError` is never `receiver` as far as I can see.

Furthermore, there are three cases where we set `span.ec` in the instrumentation, but it looks like only [one case](https://github.com/instana/nodejs/pull/486/files#diff-24bbb2057190b9840c5c35b03ff3849adf0501722464d3ba476378bf0dfd8fd4R162) is covered.

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -0,0 +1,412 @@
...
+   */
+  if (!Array.isArray(eventData)) {
+    eventData = [eventData];
+  }
+
+  if (isActive) {
+    eventData.forEach(messageData => {
+      const parentSpan = cls.getCurrentSpan();
+
+      if (!parentSpan) {
```
kirrg001, line 215: I was wondering why we do not call the original emit function for both cases:

- isActive is false (https://github.com/instana/nodejs/pull/486/files#diff-24bbb2057190b9840c5c35b03ff3849adf0501722464d3ba476378bf0dfd8fd4R211)
- parentSpan is not null (https://github.com/instana/nodejs/pull/486/files#diff-24bbb2057190b9840c5c35b03ff3849adf0501722464d3ba476378bf0dfd8fd4R215)

kirrg001:
Thanks for this great work 💯 🥳

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -0,0 +1,412 @@
...
+
+function instrumentedProduce(ctx, originalProduce, originalArgs) {
+  if (cls.tracingSuppressed()) {
+    const headers = addTraceLevelSuppression(originalArgs[6]);
+    originalArgs[6] = headers;
+    return originalProduce.apply(ctx, originalArgs);
+  }
+
+  const parentSpan = cls.getCurrentSpan();
+  if (!parentSpan || constants.isExitSpan(parentSpan)) {
```
kirrg001, line 111: _General question_: What is the use case for parentSpan being already an exit span?
So there is an incoming message, which should be produced to a topic, but the parent span is already an exit span.
When does that happen?

Thanks!

packages/collector/test/tracing/messaging/node-rdkafka/test.js
willianpc: Good catch. Will do it.

packages/collector/test/tracing/messaging/node-rdkafka/test.js
willianpc: > withError is never receiver as far as I can see

The reason why `receiver` is never set is because I could not find a way to force an error when receiving a message. Yet I kept the option in the test in case we find ways to reproduce errors in the future.

>  there are three cases where we set span.ec in the instrumentation

The uncovered cases include either an event emit from deliveryCb or a received message. We can discuss together, but as far as I could investigate, I could not properly force/capture errors in these cases

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
willianpc: Nice catch! I don't know why this did not cause errors before, as it should.
But they were missing for sure.
Thanks!

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
willianpc: Essentially, the idea is to check if there isn't already a Kafka exit span being processed, or any kind of exit span for that matter.
As a rule of thumb, there should never be an ENTRY -> ENTRY span correlation, or EXIT -> EXIT. The expectation is to always have the alternated ENTRY -> EXIT -> ENTRY and so on.

If this is not clear enough, please let me know :)

package.json
```diff
@@ -170,6 +171,9 @@
     "winston": "^3.3.3",
     "ws": "^7.5.6"
   },
+  "optionalDependencies": {
+    "kafka-avro": "^3.1.1"
```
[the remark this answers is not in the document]
willianpc: That's a valid point. The slowness is significant.
Apart from that, as I mentioned in the PR description, since the installation of kafka-avro is optional, even where the installation is not possible, npm goes forward with the installation as far as it works, which causes an unnecessary amount of time spent for nothing.

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
kirrg001: Thanks!

package-lock.json
kirrg001: Looks great now :) Thanks

packages/collector/test/tracing/messaging/node-rdkafka/producer.js
```diff
@@ -0,0 +1,200 @@
...
+      producerOptions,
+      {},
+      {
+        topic: topic,
+        // This option is needed in order to use headers as Producer stream.
+        // This information needs to be very clear in the documentation.
+        // In the instrumentation we can check Buffer.isBuffer, like it's done here:
+        // https://github.com/Blizzard/node-rdkafka/blob/master/lib/producer-stream.js#L232
+        // Without this option, the stream expects only the message as a Buffer.
+        objectMode: true
```
basti1302, line 49: Should we also have a minimal test case where objectMode is false?

Of course, there would be no trace correlation but we can at least verify that we do not break anything when the option is set to false. The expectations for that test case would be that and exit and an entry are produced, but they would not be part of the same trace (different trace IDs).

package.json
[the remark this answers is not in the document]
basti1302: TBH, I'm 50/50 on including the kafka-avro tests. We (well, Willian) have invested time to build it. kafka-avro also was the request that lead to the instrumentation of node-rdkafka. OTOH, a "significant slowdown" (as Willian put) it plus the unfixable audit issue are clear downsides. 

Either way, it is good to have tested kavka-avro with this a couple of times on CI, even if we do not merge this particular integration test into main in the end.

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
basti1302: nitpick/opinion/matter of taste: I would marginally favor an early return for the cases that Kate mentioned. 

That is, something like

```
if (!isActive) {
   return originalEmit.apply(ctx, originalArgs);
}
if (parentSpan) {
   return originalEmit.apply(ctx, originalArgs);
}
if (cls.tracingSuppressed()) {
       const headers = addTraceLevelSuppression(messageData.headers);
        messageData.headers = headers;
        return originalEmit.apply(ctx, originalArgs);
}

... rest of the instrumentation logic
```

However, really just my opinion/preference, not a blocker or anything.

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -0,0 +1,417 @@
...
+  if (!Array.isArray(eventData)) {
+    eventData = [eventData];
+  }
+
+  if (isActive) {
+    eventData.forEach(messageData => {
+      const parentSpan = cls.getCurrentSpan();
+
+      if (!parentSpan) {
+        if (cls.tracingSuppressed()) {
```
basti1302, line 216: Hm, actually this case does not really make sense here. This is the consumer/entry span instrumentation right? How would there be a suppression flag in the cls context if we have not evaluated the headers yet? 

Normally, for entry instrumentations, the order is as follows: Check the incoming headers for trace level suppression/X_INSTANA_L=0. If that is present, set the suppression flag in CLS. Then, _after that_, directly before creating the span, check if suppression is active and create a span or not depending on the outcome.

Or am I misreading this?

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -0,0 +1,417 @@
...
+                }
+              }
+              level = readTraceLevelBinary(instanaHeadersAsObject);
+            }
+            removeInstanaHeadersFromMessage(messageData);
+          }
+
+          if (event === 'data' || event === 'error') {
+            cls.ns.runAndReturn(function () {
+              if (level !== '1') {
```
basti1302, line 268: We should also allow for the level header not to be present at all and assume level 1 as default. This check should probably more be like `level === 0` instead of `level !== 1`.

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -0,0 +1,417 @@
...
+  if (!traceCorrelationEnabled) {
+    return headers;
+  }
+  switch (headerFormat) {
+    case 'binary':
+      return addTraceContextHeaderBinary(headers, span);
+    case 'string':
+      return addTraceContextHeaderString(headers, span);
+    case 'both':
+    // fall through (both is the default)
```
basti1302, line 325: Question: Do we actually want to use `both` as default even though we know the binary headers will not be handled correctly? WDYT about defaulting to `string` directly as long as the issue in node-rdkafka is not fixed? 

OTOH, if the node-rdkafka issue gets fixed, we get working trace correlation automatically, even if the consumer on the other end only understands the binary Instana headers. Maybe `both` is actually the correct choice. 🤔

basti1302 (approved):
Outstanding work! I added a few minor comments and questions, but this looks great.

packages/collector/test/tracing/messaging/kafka-avro/producer.js
kirrg001: The point was not to remove something in the PR :)

How can the customer use kafka-avro when there is no general docs about headerFormat?
Refs https://github.ibm.com/instana/docs/pull/949/files

Let's discuss offline :)

package.json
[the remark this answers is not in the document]
kirrg001: How about we install kafka-avro manually in our test only for v14?
Similar to what we do in our [ES tests](https://github.com/instana/nodejs/blob/main/packages/collector/test/tracing/database/elasticsearch_modern/test.js#L606)

This would I guess resolve both problems.

packages/collector/test/tracing/messaging/node-rdkafka/test.js
kirrg001: Yeah I totally understand 

I might try to increase code coverage before we merge the PR if that is okay for you :)

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
kirrg001: I will take this over if it's okay for you @willianpc

package.json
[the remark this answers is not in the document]
kirrg001: I can take over this too :)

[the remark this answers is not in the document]
basti1302: > How about we install kafka-avro manually in our test only for v14?

I think this is a great idea! 👍

packages/collector/test/test_util/ProcessControls.js
```diff
@@ -286,6 +286,11 @@ class ProcessControls {
     }
     return new Promise(resolve => {
       this.process.once('exit', resolve);
+      this.process.once('exit', () => {
+        this.process.pid = null;
+        resolve();
+      });
+
```
basti1302, line 293: @kirrg001 did you mean to delete line 288? Right now we would call `resolve` twice.

kirrg001: Whoops 😅

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
kirrg001: I'll fix that cc @willianpc

packages/collector/test/tracing/messaging/node-rdkafka/producer.js
kirrg001: Will take over too

packages/collector/test/tracing/messaging/kafka-avro/producer.js
willianpc: I will push the change from the Go side as well. The card is sitting there for a while.

packages/collector/test/tracing/messaging/kafka-avro/consumer.js
```diff
@@ -0,0 +1,82 @@
...
+});
+
+let isReady = false;
+
+const topicName = 'kafka-avro-topic';
+
+(async () => {
+  await kafkaAvro.init();
+
+  isReady = true;
```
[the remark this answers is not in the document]
willianpc: It makes sense. Even though I faced no issues locally or in the pipelines, but clearly this change makes sense

packages/collector/test/tracing/messaging/kafka-avro/test.js
```diff
@@ -0,0 +1,244 @@
...
+const { verifyHttpRootEntry, verifyHttpExit } = require('@instana/core/test/test_util/common_verifications');
+
+let mochaSuiteFn;
+
+/**
+ * Although kafka-avro is compatible with Node v8.x.x we are limiting tests to v10 as minimum as there is an unknown
+ * issue with Circle CI where kafka-avro seems to be installed properly in version 8, but the module is not found in
+ * node_modules.
+ */
+const kafkaAvroAllowedVersions = satisfies(process.versions.node, '10 - 14.17');
```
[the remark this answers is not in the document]
willianpc: Makes sense. the correct semver test then needs to be 10 - 14. According to this cheat sheet: https://devhints.io/semver
I also tested it in https://semver.npmjs.com/ so it does the work.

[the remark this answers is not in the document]
willianpc: @kirrg001 could you please do the change to `10 - 14`?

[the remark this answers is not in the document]
kirrg001: Yes 👍

packages/collector/test/tracing/messaging/node-rdkafka/producer.js
willianpc: good catch. I missed that one

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
willianpc: Oh no, you are right. I must have copied/pasted this from somewhere else and missed it.
Thanks for pointing it out 🥇

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
willianpc: Good point!

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
kirrg001: I'll take it over 👍

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
willianpc: Trying to think from a customer's perspective, I think keeping "both" sounds correct.
As you said, if node-rdkafka is fixed, the default option of "both" becomes true again.
And we have the node-rdkafka current limitation already documented in the docs.

My 2 cents: keep it as is. 
If you folks decide to change it, we just need some sort of reminder to change it back when node-rdkafka is updated with the fix. And even then, there may be the need of a version validation to switch the default behavior.

packages/collector/test/tracing/messaging/kafka-avro/consumer.js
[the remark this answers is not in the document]
kirrg001: Will change

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
basti1302: Yes, that makes sense. Thanks for pointing this out.

kirrg001: Summary: 

- currently is "binary"
- we want to switch to "both" for rdkafka because customer does not need to set "headerFormat" explicitly
- "headerFormat" is not even documented yet
- using "both" does not hurt, because we just send two versions of the header format

kirrg001 (approved):
Done 🦾

PR #490, opened 2022-02-24
chore(async ls): updated the versions that contain the Async LS fix

willianpc:
* Fix introduced in v17.2: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V17.md#commits-5
 * Fix introduced in v16.14: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V16.md#commits
 * Semantic calculator to support in the version query: https://semver.npmjs.com/
 
 Build passed with `cimg/node:16.14.0`.
 CircleCI Node 16 is still on 16.13

### DIFF
--- a/packages/core/src/tracing/clsHooked/index.js
+++ b/packages/core/src/tracing/clsHooked/index.js
@@ -37,9 +37,11 @@ const semver = require('semver');
  * that is appropriate for the version of on nodejs that is running.
  * Node 12.17 - 16.6 - uses native AsyncLocalStorage. See below:
  * There is a bug introduced in Node 16.7 which breaks Async LS: https://github.com/nodejs/node/issues/40693
+ * Async LS fix introduced in v17.2: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V17.md#commits-5
+ * Async LS fix introduced in v16.14: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V16.md#commits
  * Node >= v8 - uses native async-hooks
  */
-if (process && semver.satisfies(process.versions.node, '12.17 - 16.6')) {
+if (process && semver.satisfies(process.versions.node, '12.17 - 16.6 || ^16.14 || >=17.2')) {
   module.exports = require('./async_local_storage_context');
 } else {
   module.exports = require('./context');


packages/core/src/tracing/clsHooked/index.js
```diff
@@ -37,9 +37,11 @@ const semver = require('semver');
  * that is appropriate for the version of on nodejs that is running.
  * Node 12.17 - 16.6 - uses native AsyncLocalStorage. See below:
  * There is a bug introduced in Node 16.7 which breaks Async LS: https://github.com/nodejs/node/issues/40693
+ * Async LS fix introduced in v17.2: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V17.md#commits-5
+ * Async LS fix introduced in v16.14: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V16.md#commits
  * Node >= v8 - uses native async-hooks
  */
-if (process && semver.satisfies(process.versions.node, '12.17 - 16.6')) {
+if (process && semver.satisfies(process.versions.node, '12.17 - 16.6 || ^16.14 || >=17.2')) {
```
kirrg001, line 44: FYI the Circle CI Node version is still v16.13.1. That means we are not testing against the shipped async fix yet.

Wondering if we want to use this in our circleci yaml

> image: circleci/node:16:14

willianpc: Good point. I don't have a strong opinion here, but it seems reasonable to update the image anyway.
In regard to testing async ls, it doesn't really matter much, as it is tested at least by Node 14, so we are sure that the implementation works.
If the concern is to be sure that Async LS works in the fixed version, then this is yet another reason to do it so.

kirrg001: > If the concern is to be sure that Async LS works in the fixed version

Yeah I would like to see that it's working :)

.circleci/config.yml
```diff
@@ -264,7 +264,7 @@ memcached: &memcached
 jobs:
   "node-16":
     docker:
-      - image: circleci/node:16
+      - image: cimg/node:16.14.0
```
willianpc, line 267: There you go, @kirrg001

kirrg001 (approved):

PR #492, opened 2022-02-28
test: added tsoa framework to integration tests

kirrg001:
no issue

https://github.com/lukeautry/tsoa

The good thing: the normal use cases using the tsoa framework work as expected.

Background: Customer's urls (such as /api/bla/blubb and /api/du/da) were grouped together under /api/ in the ui, which is unexpected.

- only a few routes are affected
- these were grouped together because path_tpl was missing
- only found one underyling reason:
  - path_tpl is missing when requests are exited too early
 
**The python sensor is always attaching the path_tpl too**, see https://github.com/instana/python-sensor/search?p=1&q=path_tpl
  
Please ignore the `build` folder. This makes it easier to review

**TBD:**
Decide if and how we want to fix the use case. 
If we keep the fix I have applied then we will send original urls and not the original path!
I think we should discuss this in a call first.


kirrg001:
I'd like to merge this into main now. I actually wanted to wait till the customer comes back with infos.
But for now we can just use the benefit of having the framework covered in our test suite.

basti1302 (approved):
LGTM

PR #496, opened 2022-03-08
fix(collector): fix export returned from init

basti1302:
The fix to prevent initializing @instana/collector multiple times
(commit [b3261b7a653b406cbe2eeaaf9050134bbeedfac9](https://github.com/instana/nodejs/commit/b3261b7a653b406cbe2eeaaf9050134bbeedfac9)) broke the export returned
from the init function. Instead of returning a reference to the init function,
to which all exports are attached, the function returned undefined for the case
when @instana/collector had already been initialized previously. This breaks the
contract of the init function and in particular it breaks users that want to use
the export to access the API of @instana/collector.


kirrg001 (approved):
Good catch! Great work!

PR #502, opened 2022-03-11
Fargate/Google Cloud Run script & docs update

basti1302:


packages/aws-fargate/images/README.md
```diff
@@ -0,0 +1,91 @@
+Node.js Fargate Container Images
```
basti1302, line 1: I moved the README one level up from packages/aws-fargate/images/instana-aws-fargate/README.md (and gave it a bit of an overhaul)

packages/aws-fargate/images/instana-aws-fargate/build-and-push.sh
```diff
@@ -39,24 +55,17 @@ npm_tag=$2
 
 ./build.sh $build_mode $npm_tag
 
-setImageTag $image_tag_prefix $build_mode
+setImageTag $image_tag_prefix $build_mode $npm_tag
```
basti1302, line 58: I refactored our most recent modifications a bit so that $image_tag now also contains the Docker tag (`:latest` or `:next`). This made it possible to remove most of the if-else snippets we added earlier. I think it is a bit more maintainable/readable now. (Same in build.sh)

packages/aws-fargate/images/test-images/Dockerfile-alpine
```diff
@@ -1,8 +1,7 @@
-ARG DOCKER_TAG=latest
-ARG INSTANA_LAYER=icr.io/instana/aws-fargate-nodejs
+ARG INSTANA_LAYER=icr.io/instana/aws-fargate-nodejs:latest
```
basti1302, line 1: The INSTANA_LAYER env var now always contains a tag. Thus, the extra env var DOCKER_TAG is no longer needed.

packages/aws-fargate/images/test-images/build-and-push-all.sh
```diff
@@ -1,39 +0,0 @@
-#!/usr/bin/env bash
```
basti1302, line 1: I rarely used this anymore, I think it is better to delete it than to keep maintaining it (or letting it become unmaintained).

packages/aws-fargate/images/test-images/restart.sh
```diff
@@ -1,36 +0,0 @@
-set -eo pipefail
```
basti1302, line 1: also unused and unmaintained

PR #506, opened 2022-03-18
chore: added missing packages for initializedTooLateHeuristic and optimised tests

kirrg001:


basti1302:
> the regex never matched

There is an integration test at https://github.com/instana/nodejs/tree/main/packages/collector/test/tracing/misc/too_late that has been green so far. What exactly do you mean with "the regex never matched"?

kirrg001:
nvm the function is working fine :D Added rdkafka and a unit test. I'd suggest to keep a single case in the integration test (it runs 26s) and rely on the unit test for the rest.

basti1302:
> nvm the function is working fine :D Added rdkafka and a unit test. I'd suggest to keep a single case in the integration test (it runs 26s) and rely on the unit test for the rest.
> 

Yes, great idea!

kirrg001:
> Yes, great idea!

Yeah 👍

packages/core/test/util/initializedTooLateHeuristic_test.js
```diff
@@ -0,0 +1,78 @@
...
+    });
+
+    require.cache['/@contrast/agent/'] = {};
+
+    const tracing = require('../../src/');
+    tracing.init();
+  });
+
+  after(() => {
+    delete require.cache['/@contrast/agent/'];
```
basti1302, line 31: _question:_ Do we also need to undo the `sinon.stub` calls on the `requireHook` module in `after` here`? I think we should do that.

packages/core/test/util/initializedTooLateHeuristic_test.js
```diff
@@ -0,0 +1,78 @@
...
+
+  it('hasBeenInitializedTooLate is true', () => {
+    const originalRequireCache = require.cache;
+
+    instrumentedModules.forEach(moduleName => {
+      // eslint-disable-next-line no-console
+      console.log(`for module ${moduleName}`);
+
+      // RESET CACHE & module
+      require.cache = originalRequireCache;
```
basti1302, line 57: _question:_ Does that actually do anything? Aren't `require.cache` and `originalRequireCache` two references to the very same object? AFAIK `require.resolve` does not create a new object for `require.cache`, it just updates the existing data structure right? (Also, our own modifications will not replace require.cache with a new object, as we only set entries and delete them.)

basti1302 (approved):
LGTM!

packages/core/test/util/initializedTooLateHeuristic_test.js
kirrg001: TODO: check if `sinon.reset` is needed

packages/core/test/util/initializedTooLateHeuristic_test.js
kirrg001: Check if we need to use clone

PR #510, opened 2022-03-22
build(deps): update transitive minimist dependency to 1.2.6 

basti1302:
Addresses https://github.com/advisories/GHSA-xvch-5gv4-984h.


kirrg001 (commented):
Suggestion: Do not use `fix`, because this is IMO not a user facing change.
The customer needs to update their sub-dependencies (or better said their lock file) manually

kirrg001 (approved):
Commented, otherwise 👍

basti1302:
> Suggestion: Do not use `fix`, because this is IMO not a user facing change. The customer needs to update their sub-dependencies (or better said their lock file) manually

Right, good call. I've rewritten the commit message to `build(deps): update transitive minimist dependency to 1.2.6`, in line with what Dependabot uses. 👍

PR #511, opened 2022-03-22
fix(collector): work around Bazel's node-patches module

basti1302:
This fixes the annouce procedure for Node.js apps that have been built
with Bazel and include its node-patches module. Specifically, the
patches to `fs.readlinkSync` break the announcement of
@instana/collector to the Instana agent, by prefixing the result of
readlinkSync with an absolute path.

See https://github.com/bazelbuild/rules_nodejs/blob/5.3.0/packages/node-patches/src/fs.ts,
line 226.

refs 89102

### DIFF
--- a/packages/collector/src/agentConnection.js
+++ b/packages/collector/src/agentConnection.js
@@ -158,7 +158,17 @@ exports.announceNodeCollector = function announceNodeCollector(cb) {
       payload.fd = String(socket._handle.fd);
 
       try {
-        payload.inode = fs.readlinkSync(pathUtil.join('/proc', String(process.pid), 'fd', payload.fd));
+        const linkPathPrefix = `${pathUtil.join('/proc', String(process.pid), 'fd')}/`;
+        const linkPath = pathUtil.join(linkPathPrefix, payload.fd);
+        payload.inode = fs.readlinkSync(linkPath);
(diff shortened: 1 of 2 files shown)


packages/collector/src/agentConnection.js
```diff
@@ -158,7 +158,17 @@ exports.announceNodeCollector = function announceNodeCollector(cb) {
       payload.fd = String(socket._handle.fd);
 
       try {
-        payload.inode = fs.readlinkSync(pathUtil.join('/proc', String(process.pid), 'fd', payload.fd));
+        const linkPathPrefix = `${pathUtil.join('/proc', String(process.pid), 'fd')}/`;
+        const linkPath = pathUtil.join(linkPathPrefix, payload.fd);
+        payload.inode = fs.readlinkSync(linkPath);
+        if (typeof payload.inode === 'string' && payload.inode.indexOf(linkPathPrefix) === 0) {
```
kirrg001, line 164: I was looking for the opposite test. Found it in /packages/collector/test/agentCommunication_test.js
But the test only runs on linux. Was wondering if you would like to cover the opposite case (return value is not a full path) in your bazel test

kirrg001 (approved):
LGTM 👍

packages/collector/src/agentConnection.js
basti1302: Yes, sure, let's do that.

Done.

PR #512, opened 2022-03-23
test: troubleshoot pino-v7 on ci

basti1302:


kirrg001:
Great investigation 🥇

PR #514, opened 2022-03-23
test: added tsoa framework to integration tests

kirrg001:
no issue

https://github.com/lukeautry/tsoa

The good thing: the normal use cases using the tsoa framework work as expected.

Background: Customer's urls (such as /api/bla/blubb and /api/du/da) were grouped together under /api/ in the ui, which is unexpected.

only a few routes are affected
these were grouped together because path_tpl was missing
only found one underyling reason:
path_tpl is missing when requests are exited too early
The python sensor is always attaching the path_tpl too, see https://github.com/instana/python-sensor/search?p=1&q=path_tpl

Please ignore the build folder. This makes it easier to review

TBD:
I personally would not fix the early exit middleware use case, because it will result in wrong path_tpl values.


basti1302 (approved):
LGTM 🎉

PR #516, opened 2022-03-24
test: speed up ci test run & reduce random failures

kirrg001:
See https://circleci.com/docs/2.0/parallelism-faster-jobs/

refs 80775
refs 93142

Goals/Description:
- reduce random failures - they are rly annoying
- speed up test run in ci -> switch from random split by filename to split by timings. Random is not good because it will put x slow tests on a single job
- we can only activate 16 parallel jobs per node version (80/5), otherwise they are waiting for one to finish (PER BUILD!)

**[Before](https://app.circleci.com/pipelines/github/instana/nodejs/3120/workflows/c76dfdd0-e732-42ba-8cd7-59e9e5a465ca):**
node-8: 14m 2s
node-16: 22m 58s
node-14: 23m 14s
node-12: 21m 26s
node-10: 21m 35s

**[After](https://app.circleci.com/pipelines/github/instana/nodejs/3189/workflows/09b3a791-9f8f-4867-8887-714cef4e396e):**
node-8: 6m 32s
node-16: 10m 3s
node-14: 8m 47s
node-12: 8m 27s
node-10: 7m 45s

The whole test runs depends on circle ci being slow/fast and our third party communication (aws etc.)
So you sometimes see they finish after 8mins and sometimes after 12min
Node 16 is the slowest run at the moment. I don't want to spend more time on this for now.

### TODO
- [x] results.xml has no file name - thats why by-timings does not work
- [x] split by timings
- [x] why are some tests not running? (shard-metrics, autoprofile)
- [x] why aws s3 broken - buckets again?
- [x] some single tests run too long
- [x] fix new ci reporter output - looks ugly -> use multi
- [x] test:mocha -> test:local
- [x] npm testci vs test
- [x] create ticket for broken collector/test/tracing/cloud/gcp tests
  - No timing found for "/home/circleci/repo/packages/collector/test/tracing/cloud/gcp/pubsub/test.js"
  - No timing found for "/home/circleci/repo/packages/collector/test/tracing/cloud/gcp/storage/test.js"
- [x] create ticket for profiling test to speed up the test (7mins locally!)
  - No timing found for "/home/circleci/repo/packages/collector/test/profiling/test.js"
- [x] cleanup

tests before: [2625](https://app.circleci.com/pipelines/github/instana/nodejs/3113/workflows/ae8a70f4-30ef-4662-a996-8203549f505e/jobs/13609/tests)
tests after: [2470](https://app.circleci.com/pipelines/github/instana/nodejs/3170/workflows/6b922e7c-92c1-4c9c-bba1-ebec5155a6bb/jobs/13754/tests)

Reasons for running less tests: 
- We run tests twice in the previous implementation e.g. otel env was wrong and always executed all tests per node version
- When the ci globbing did not have results for a package, the npm command run everything as fallback

The shorter the runtime, the less annoying it is to re-run a suite because of a random failure.

The new reporter is officially listed on their docs https://circleci.com/docs/2.0/collect-test-data/

### Summary of changes

- introduce `test:ci` - we do not need to generate xml files locally + less bugs if we have two separate commands 
- split tests which run multiple npm versions into separate files
   - putting multiple npm versions into a single test makes the test very slow 
   - it is uncomfortable to run a single npm version because you need to manually comment the version out
- fixed s3 bucket naming (see cleanup aws script!)
- adding split by timings
  - when you add a new test, circleci has no timing reference. It will run it randomly on any job and next run it will know

**Review per commit is easier than looking at the full diff**
**Merge all commits into main**


basti1302:
LGTM 👍 

One thing that we should discuss IMO: There is a good bit of code duplication now to make the distribution algorithm of CircleCI happy. But I think we can eat the cake and have it too, in this case -- I added #518 as a suggestion for a pattern that should make CircleCI happy and avoid code duplication. 

Let's have a chat tomorrow or next week.

basti1302 (approved):

PR #517, opened 2022-03-29
fix(metrics): do not report metrics from worker threads

### LINKED ISSUE #500: [Bug]: Incorrect process identification when pino v7+ transports are used
### Problem Description

When the `@instana/collector` is loaded into a Node.js application, but that application also uses pino v7+ [transports](https://getpino.io/#/docs/transports), then the Node.js process is sometimes identified as `thread-stream` instead of using the actual application name. The reproduction also includes a scenario where the same transport is used as a pino [destination](https://getpino.io/#/docs/api?id=destination) (omit the `MODE` env variable), in such cases the process is correctly identified.

**Note:** I cannot reliable reproduce the problem, sometimes the process is correctly identified even when pino transports are used. The problem may have something to do with the fact that pino transports work by using worker threads, but that is as far as my knowledge of the problem goes.

### Short, Self Contained Example

See [reproduction repository](https://github.com/jonathansamines/instana-pino-transport)

### Node.js Version

See reproduction

### package.json

```shell
See reproduction
```

### package-lock.json

```shell
See reproduction
```

basti1302:
When instrumented via NODE_OPTIONS="--require ..." (which is inherited by
worker threads), each worker thread will announce independently and act like
the main thread. This can cause issues, in particular if the worker thread is
started from a dependency in node_ module and it identifies a different
package.json as its main package.json, hence reporting a different application
name etc.

To prevent this, we disable sending metrics and snapshot data from
worker threads for now.

fixes #500

refs 88823

### DIFF
--- a/packages/collector/src/announceCycle/agentready.js
+++ b/packages/collector/src/announceCycle/agentready.js
@@ -5,6 +5,13 @@
 
 'use strict';
 
+let isMainThread = true;
+try {
+  isMainThread = require('worker_threads').isMainThread;
+} catch (err) {
+  // Worker threads are not available, so we know that this is the main thread.
+}
+
 const semver = require('semver');
 const { tracing } = require('@instana/core');
 
@@ -23,6 +30,9 @@ const TEN_MINUTES = 10 * 60 * 1000;
 
 /** @type {*} */
 let autoprofile;
+
+/** @type {*} */
+let profiler;
 
 /** @type {import('@instana/core/src/logger').GenericLogger} */
 let logger;
@@ -71,29 +81,34 @@ module.exports = exports = {
  */
 function enter(_ctx) {
   ctx = _ctx;
-  uncaught.activate();
-  metrics.activate();
+
   initializedTooLate.check();
-  transmissionCycle.activate(
-    metrics,
-    agentConnection,
-    /**
-     * @param {Array.<import('../agent/requestHandler').AnnounceRequest>} requests
-     */
-    function onSuccess(requests) {
-      requestHandler.handleRequests(requests);
-    },
-    function onError() {
-      ctx.transitionTo('unannounced');
-    }
-  );
+
+  if (isMainThread) {
+    uncaught.activate();
+    metrics.activate();
+    requestHandler.activate();
+    transmissionCycle.activate(
+      metrics,
+      agentConnection,
+      /**
+       * @param {Array.<import('../agent/requestHandler').AnnounceRequest>} requests
+       */
+      function onSuccess(requests) {
+        requestHandler.handleRequests(requests);
+      },
+      function onError() {
+        ctx.transitionTo('unannounced');
+      }
+    );
+    scheduleTracingMetrics();
+    detectEOLNodeVersion();
+  }
+
   tracing.activate();
-  requestHandler.activate();
-  scheduleTracingMetrics();
-  detectEOLNodeVersion();
 
   if (agentOpts.autoProfile && autoprofile) {
-    const profiler = autoprofile.start();
+    profiler = autoprofile.start();
     /**
      * @param {*} profiles
      * @param {(...args: *) => *} callback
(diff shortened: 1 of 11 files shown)


packages/collector/src/announceCycle/agentready.js
```diff
@@ -110,14 +124,19 @@ function enter(_ctx) {
...
+
   tracing.deactivate();
-  requestHandler.deactivate();
-  if (tracingMetricsTimeout) {
-    clearTimeout(tracingMetricsTimeout);
-    tracingMetricsTimeout = null;
+
+  if (profiler) {
+    profiler.destroy();
+    profiler = null;
```
basti1302, line 139: Destroying the profiler in `agentready#leave` is not directly related to the main intent of this PR, but I noticed that this was missing, so I added it.

packages/collector/src/logger.js
```diff
@@ -37,6 +46,7 @@ exports.init = function init(config, isReInit) {
     // we create later on.
     parentLogger = bunyan.createLogger({
       name: '@instana/collector',
+      thread: threadId,
```
basti1302, line 49: Adding the thread ID to the log object is not strictly necessary but we have the PID in there, too, and I think it might be pretty helpful when trouble shooting scenarios involving worker threads in the future.

packages/collector/src/announceCycle/agentready.js
kirrg001: My impression is that `agentready#leave` is only executed when `ctx.transitionTo('agentready');` is called, but profile will be null then, because this state was never entered. My impression is that `agentready` state can only happen once. (Autoprofile will autodestroy on process exit, which I thought is enough based on the logic. )
Maybe better to discuss offline :)

packages/collector/test/metrics/appWithWorkerThread/test.js
```diff
@@ -0,0 +1,120 @@
...
+        }
+      });
+      expect(allNames).to.contain('worker-thread-test-app');
+    });
+
+    // We have made sure that the main thread has announced itself. Now we give the worker thread a chance to announce
+    // itself and send metrics. Actually, we do _not_ want the worker thread to send metrics on its own, but to verify
+    // that this does not happen, we need to wait a bit so that the test actually would fail if the worker thread sent
+    // metrics.
+    await delay(800);
```
kirrg001, line 101: I was wondering if this can create another flaky test.
Is there a way to determine if the worker thread has collected the spans yet? (I have not looked for a solution)

packages/collector/src/announceCycle/agentready.js
```diff
@@ -71,29 +81,33 @@ module.exports = exports = {
...
-    },
-    function onError() {
-      ctx.transitionTo('unannounced');
-    }
-  );
+
+  if (isMainThread) {
+    uncaught.activate();
+    metrics.activate();
+    initializedTooLate.check();
```
kirrg001, line 88: I was wondering why we don't want to send the initialized monitoring event on a worker thread.
My impression was that a worker thread can require instana in a worker thread, which means the feature might be helpful there too.

kirrg001 (approved):
LGTM! Just left some questions :)

packages/collector/src/announceCycle/agentready.js
basti1302: agentready#leave can also be triggered when we lose the connection to the agent (for example, if the agent is stopped or crashes), see https://github.com/instana/nodejs/blob/main/packages/collector/src/announceCycle/agentready.js#L87

When that has happened, we go back to state unannounced and start the announce loop again. If the agent comes back up, we would actually execute agentready#enter a second time.

In that scenario, without an explicit call to profiler.destroy() the profiler from the first successful announce would still be active (although we no longer have a referenced to it from @instana/collector), and we would create and start a second profiler.

packages/collector/src/announceCycle/agentready.js
basti1302: Good catch! That's absolutely correct, I'll move that one out of the "if (isMainThread)".

packages/collector/test/metrics/appWithWorkerThread/test.js
basti1302: Right, the section for checking if spans have been produced needs to be wrapped in another `retry` block. Fixed.

packages/collector/src/announceCycle/agentready.js
kirrg001: Ah yeah nice. That makes sense. Thanks <3

PR #522, opened 2022-04-01
chore: improved and fixed publish-layer script

kirrg001:
- ability to send region per env
- summary of input and yield to say yes/no
- fixed `mv instana-aws-lambda-*.tgz $LAYER_WORKDIR/instana-aws-lambda.tgz`

### DIFF
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -93,9 +93,32 @@ ZIP_PREFIX=instana-nodejs-layer
 ZIP_NAME=$ZIP_PREFIX.zip
 TMP_ZIP_DIR=tmp
 
-echo Will build Lambda layer with name \"$LAYER_NAME\".
+if [[ -z $REGIONS ]]; then
+  REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
+fi
 
-REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
+if [[ -z $NO_PROMPT ]]; then
+  while true; do
+      echo "####"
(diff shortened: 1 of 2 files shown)


packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -93,9 +93,32 @@ ZIP_PREFIX=instana-nodejs-layer
 ZIP_NAME=$ZIP_PREFIX.zip
 TMP_ZIP_DIR=tmp
 
-echo Will build Lambda layer with name \"$LAYER_NAME\".
+if [[ -z $REGION ]]; then
+  REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
+else
+  REGIONS=$REGION
+fi
```
basti1302, line 100:
```suggestion
if [[ -z $REGION ]]; then
  REGIONS=$'ap-northeast-1\nap-northeast-2\nap-south-1\nap-southeast-1\nap-southeast-2\nca-central-1\neu-central-1\neu-north-1\neu-west-1\neu-west-2\neu-west-3\nsa-east-1\nus-east-1\nus-east-2\nus-west-1\nus-west-2'
fi
```

Nice addition! I think the else branch is a no-op though and can be omitted.

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -93,9 +93,32 @@ ZIP_PREFIX=instana-nodejs-layer
...
+    echo "SKIP_AWS_PUBLISH_LAYER: $SKIP_AWS_PUBLISH_LAYER"
+    echo "####"
+
+    read -p "Do you wish to continue (yes or no)? " yn
+    case $yn in
+        [Yy]* ) echo "Let's go!"; break;;
+        [Nn]* ) exit 1;;
+        * ) echo "Please answer yes or no.";;
+    esac
+done
```
basti1302, line 119: _issue:_ That's pretty nice for local use 👍    One caveat: On [Concourse](https://github.com/instana/nodejs/blob/main/packages/serverless/ci/pipeline.yaml#L310-L315), nobody will press `y`, so I think we need a parameter or env var to skip the question for usage on CI.

basti1302 (changes requested):
Looks good. There's one issue though that needs to be fixed or it will break the build job to build and publish the production layer from Concourse.

packages/aws-lambda/layer/bin/publish-layer.sh
kirrg001: Yeah good catch! I'll skip

basti1302 (approved):
LGTM

PR #527, opened 2022-04-06
feat(tracing): use new common tracing config from from agent response

basti1302:
~Note: one assertion in the tests is failing, I'll look into this tomorrow/Friday/early next week.~

With the upcoming Node.js discovery plug-in version 1.2.17, common tracing
configuration (including extra http headers and the new Kafka tracing config)
will be available under the "tracing" attribute in the agent's response.

Use the values from agentResponse.tracing over the legacy attributes for
extraHeaders. Also, apply the new Kafka tracing config (correlation
disabled/enabled, header format) to our Kafka instrumentations.

refs 93724


packages/collector/test/tracing/messaging/node-rdkafka/test.js
```diff
@@ -240,13 +241,23 @@ mochaSuiteFn('tracing/messaging/node-rdkafka', function () {
...
+          expect(span.t).to.be.a('string');
+        }
+      },
+      span => {
+        if (parent) {
+          expect(span.p).to.equal(parent.s);
+        } else {
+          expect(span.p).to.not.exist;
+        }
+      },
```
basti1302, line 260: This did not actually check for span.p to be undefined in the previous version.

packages/collector/test/tracing/messaging/node-rdkafka/test.js
```diff
@@ -291,6 +302,154 @@ mochaSuiteFn('tracing/messaging/node-rdkafka', function () {
     ]);
   }
 
+  describe('tracing enabled, header format string', function () {
```
basti1302, line 305: We did not have tests with different header formats for rdkafka before. `binary` is not used, so it is fine that that is missing, but a test for `string` was missing so I added it.

packages/collector/test/tracing/messaging/node-rdkafka/test.js
```diff
@@ -291,6 +302,154 @@ mochaSuiteFn('tracing/messaging/node-rdkafka', function () {
...
+        const httpEntry = verifyHttpRootEntry({ spans, apiPath, pid: String(producerControls.getPid()) });
+        const kafkaExit = verifyRdKafkaExit(producerControls, spans, httpEntry, null, false);
+        verifyHttpExit({ spans, parent: httpEntry, pid: String(producerControls.getPid()) });
+        const kafkaEntry = verifyRdKafkaEntry(consumerControls, spans, kafkaExit, null, false);
+        verifyHttpExit({ spans, parent: kafkaEntry, pid: String(consumerControls.getPid()) });
+      });
+    });
+  });
+
+  describe('tracing enabled, but trace correlation disabled', function () {
```
basti1302, line 379: This case had also been missing previously.

packages/core/src/tracing/constants.js
```diff
@@ -28,6 +28,11 @@ exports.kafkaTraceIdHeaderNameString = 'X_INSTANA_T';
 exports.kafkaSpanIdHeaderNameString = 'X_INSTANA_S';
 exports.kafkaTraceLevelHeaderNameString = 'X_INSTANA_L_S';
 
+exports.kafkaTraceCorrelationDefault = true;
+// Before we start phase 1 of the migration, 'binary' will be the default value. With phase 1, we will move to 'both',
+// with phase 2 it will no longer be configurable and will always use 'string'.
+exports.kafkaHeaderFormatDefault = 'binary';
```
basti1302, line 34: Extracted the default from the two instrumentation modules that are using it.

packages/core/src/tracing/index.js
```diff
@@ -23,8 +23,6 @@ let automaticTracingEnabled = false;
 let cls = null;
 /** @type {import('../util/normalizeConfig').InstanaConfig} */
 let config = null;
-/** @type {Array.<string>} */
-let extraHeaders = [];
```
basti1302, line 27: We kept a copy of the extraHeaders setting in this module but never used it for anything, so I removed it.

packages/core/src/tracing/index.js
```diff
@@ -235,10 +240,9 @@ exports.getCls = function getCls() {
 };
 
 /**
- * @param {Array.<string>} _extraHeaders
+ * @param {Array.<string>} extraHeaders
  */
-exports.setExtraHttpHeadersToCapture = function setExtraHttpHeadersToCapture(_extraHeaders) {
-  extraHeaders = _extraHeaders;
```
basti1302, line 241: This is where we wrote the obsolete extraHeaders (and never read it again).

packages/core/src/tracing/instrumentation/messaging/kafkaJs.js
```diff
@@ -15,10 +15,8 @@ logger = require('../../../logger').getLogger('tracing/kafkajs', newLogger => {
   logger = newLogger;
 });
 
-let traceCorrelationEnabled = true;
-// Before we start phase 1 of the migration, 'binary' will be the default value. With phase 1, we will move to 'both',
-// with phase 2 it will no longer be configurable and will always use 'string'.
```
basti1302, line 20: comment was moved to the constants file.

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -16,36 +16,50 @@ logger = require('../../../logger').getLogger('tracing/rdkafka', newLogger => {
...
- * Kafka migration has not officially started yet
- * The option `headerFormat` is not officially documented yet.
- *
- * To ensure customers can use rdkafka instrumentation we set the
- * default to "both" here. That means we send both header formats
- * by default.
- *
- * Node.js Sender -> Node.js Receiver works
- * Node.js Sender -> Java Receiver does not works
- */
```
basti1302, line 31: comment moved down to overrideKafkaHeaderFormat function

packages/core/test/tracing/index_test.js
```diff
@@ -7,49 +7,77 @@
...
+    setKafkaTracingConfigStubRdKafka = sinon.stub(rdKafka, 'setKafkaTracingConfig');
   });
 
-  it('deactivate instrumentation via config', () => {
-    const config = { tracing: { disabledTracers: ['grpc', 'kafka'] } };
-    normalizeConfig(config);
+  beforeEach(() => {
+    // requiring tracing/index via proxyquire gives us a module in pristine state everytime, in particular with the
+    // tracingActivated flag reset.
+    tracing = proxyquire('../../src/tracing', {});
```
basti1302, line 40: There were some issues with this test suite due to the fact that we re-used the same (stateful) instance of `require('../../src/tracing')` for all tests. In particular, the `activate` method was not executed, except for the first test in the suite.

packages/collector/test/announceCycle/unannounced_test.js
```diff
@@ -0,0 +1,187 @@
...
+const { secrets, tracing } = require('@instana/core');
+const { constants } = tracing;
+
+let unannouncedState;
+let agentResponse;
+
+const pidStoreStub = {};
+const agentOptsStub = {};
+const tracingStub = sinon.stub(tracing);
+const secretsStub = sinon.stub(secrets);
```
kirrg001, line 24: This won't work, because if you run the tests this code will be executed for **all tests**.
That means you would stub the whole core package for all tests.
The stub call needs to go into `before` or `beforeEach` - depends what you want to achieve :)

packages/collector/src/announceCycle/unannounced.js
```diff
@@ -64,43 +91,106 @@ function tryToAnnounce(ctx, retryDelay = initialRetryDelay) {
...
+      traceCorrelation:
+        kafkaTracingConfigFromAgent['trace-correlation'] != null
+          ? kafkaTracingConfigFromAgent['trace-correlation']
+          : tracingConstants.kafkaTraceCorrelationDefault,
+      headerFormat:
+        kafkaTracingConfigFromAgent['header-format'] != null
+          ? kafkaTracingConfigFromAgent['header-format']
+          : tracingConstants.kafkaHeaderFormatDefault
+    };
+    tracing.setKafkaTracingConfig(kafkaTracingConfig);
```
kirrg001, line 182: _Suggestion_:

The code to override the config from the agent for the target instrumentation is a bit hard to follow, but of course it existed before this PR :)

As far as I can see, we do not trace before the announceCycle is completed and reaches `agentReady` step, which calls `tracing.activate`.

In the `unannounced` step we try to call the core package (which I think is unneeded and too coupled) to tell the instrumentation that there is another config available, which is more important.

IMO the unannounced cycle step should write the agent config into the ctx object. And when the last step is reached we pass `tracing.activate(ctx.agentConfig)`. We then automatically pass the agent config into each instrumentation. Each instrumentation can pick the config it needs.

Advantage: 
- we do not have to iterate 3 times through all instrumentations (costs time)
- easier to read
- we don't need setExtraHttpHeadersToCapture, setKafkaTracingConfig or similar. If an instrumentation needs config from agent, it can just access the agentConfig in the activate fn

Happy to discuss and refactor afterwards.

packages/core/test/tracing/index_test.js
kirrg001: Maybe this is cased by [this line](https://github.com/instana/nodejs/pull/527/files#diff-e2d7b3fbf936c12cd8fa74f525607ea570aa81c0a5bc68de09c8ea6eeb0e95f0R23)?

kirrg001 (commented):
LGTM 👍 Left some comments :)

packages/collector/src/announceCycle/unannounced.js
basti1302: I like this idea! I'll look into that.

packages/core/test/tracing/index_test.js
basti1302: discussed offline

packages/collector/src/announceCycle/unannounced.js
basti1302: I have implemented this suggestion now with the small difference that I attached the configuration that the agent has provided to agent/opts, which already was used to store the agent uuid, so it felt somewhat natural to put the config there, too.

I know you already approved but somehow this triggered a couple more changes (deleting an obsolete test) so I would actually appreciate a re-review of the latest changes if you don't mind :-)

packages/core/src/tracing/instrumentation/protocols/http2Server.js
```diff
@@ -31,6 +31,26 @@ exports.init = function init(config) {
   extraHttpHeadersToCapture = config.tracing.http.extraHttpHeadersToCapture;
 };
 
+exports.updateConfig = function updateConfig(config) {
```
kirrg001, line 34: I'd like to talk about the `preinit` logic when you are back :)

packages/core/src/tracing/instrumentation/protocols/http2Server.js
```diff
@@ -31,6 +31,26 @@ exports.init = function init(config) {
...
+  extraHttpHeadersToCapture = config.tracing.http.extraHttpHeadersToCapture;
+};
+
+exports.activate = function activate(extraConfig) {
+  if (
+    extraConfig &&
+    extraConfig.tracing &&
+    extraConfig.tracing.http &&
```
kirrg001, line 42: Suggestion: Sad we cannot use optional chaining. But maybe `utils.get(extraConfig, 'tracing.http')` (see https://lodash.com/docs#get)

kirrg001 (approved):
🥳

packages/core/src/tracing/instrumentation/protocols/http2Server.js
basti1302: Yes, optional chaining would be great here but it is only available once we drop Node < 14.

TBH, I find the extra boiler plate okay-ish here. We can revisit the utils.get idea later, for now I would not go back and introduce a util for that here.

PR #531, opened 2022-04-22
chore: updated dev node.js version

kirrg001:
With 16.14 we are locally using the new async local storage implementation.
We would like to switch to 16.14 because of this reason


basti1302 (approved):
LGTM

PR #532, opened 2022-04-22
feat: added instrumentation for ibm db2 

kirrg001:
refs 93893

https://github.com/ibmdb/node-ibm_db
https://github.com/ibmdb/node-ibm_db/blob/master/APIDocumentation.md

- [x] discuss TODO's in codebase


.prettierignore
```diff
@@ -0,0 +1,2 @@
+*.yml
+*.yaml
```
kirrg001, line 2: `circleci/config.yaml` uses double quotes
But prettier is configured to use single quotes, see `.prettierrc`.

We either need to change all yaml files to single quotes or simply ignore all yaml files.

I use `"editor.formatOnSave": false` in my IDE, which triggers this behaviour.

.gitignore
```diff
@@ -4,4 +4,5 @@ packages/collector/test/bin/edgemicroResources/cli/edgemicro
 nodejs-license-report.csv
 test-duration-breakdown
 .nyc_output
-coverage
\ No newline at end of file
+coverage
+db2
```
kirrg001, line 8: I added db2 volume in the docker compose file.
This creates a folder in our root dir. Might be helpful for debugging later

packages/core/test/test_util/expectAtLeastOneMatching.js
```diff
@@ -15,13 +15,16 @@ const { findAllMatchingItems, reportFailure } = require('./matchingItems');
  * @type {Function}
  * @param {Array.<InstanaBaseSpan>} items
  * @param {(span: InstanaBaseSpan) => void} expectations
- * @returns {InstanaBaseSpan}
+ * @param {*} options
+ * @returns {*}
  */
-module.exports = exports = function expectAtLeastOneMatching(items, expectations) {
+module.exports = exports = function expectAtLeastOneMatching(items, expectations, options = { numberOfMatches: null }) {
```
kirrg001, line 21: For db2 I am calling `execute` twice for a prepared statement.
As we do not capture statement arguments, the span looks exactly the same.

See case "must trace prepare execute async with reusage"

packages/collector/test/tracing/database/db2/test.js
```diff
@@ -0,0 +1,1117 @@
...
+        options.error
+          ? expect(span.data.db2.error).to.contain(options.error)
+          : expect(span.data.db2.error).to.not.exist,
+      span => (options.error ? expect(span.ec).to.equal(1) : expect(span.ec).to.equal(0))
+    ]);
+  });
+};
+
+mochaSuiteFn('tracing/db2', function () {
+  const timeout = 30 * 1000;
```
basti1302, line 83: _thought:_ We often use timeouts that are different on CI and locally. I feel that with using a cloud based DB2 we might need that here as well. But if 30 seconds is good for both cases, that's fine, too.

packages/collector/test/tracing/database/db2/test.js
```diff
@@ -0,0 +1,1117 @@
...
+
+    it('must trace query promise', function () {
+      return controls
+        .sendRequest({
+          method: 'GET',
+          path: '/query-promise'
+        })
+        .then(() => {
+          return testUtils.retry(() => {
+            return verifySpans(agentControls, controls);
```
basti1302, line 135: _suggestion/issue:_ In the app under test we already return the data from the query via the HTTP response, so we have it available here (as the resolved value of the `sendRequest` promise). It would be a great addition to generally check the HTTP response. The added benefit is that we do not only verify that the produced spans are correct, but also that our instrumentation code does not change or swallow result values.  IMO we should do that in every integration test.

packages/core/test/test_util/expectAtLeastOneMatching.js
basti1302: _question:_ Can we instead use `packages/core/test/test_util/expectExactlyNMatching` for this? Maybe I'm missing something here but this use case sounds like it is already covered by `expectExactlyNMatching`.

basti1302 (commented):
Submitting a partial review here for now as I'll be in meetings for the rest of the day, I did not yet review the instrumentation code itself.

packages/collector/test/tracing/database/db2/test.js
kirrg001: Yeah db2 is tricky. It takes a longer time to bootstrap.
On CI though we can use a smaller timeout because of we use the cloud, true 👍 
Will change :)

packages/collector/test/tracing/database/db2/test.js
```diff
@@ -0,0 +1,1117 @@
...
+    })
+    .join('');
+
+  return randomStr;
+};
+
+const verifySpans = (agentControls, controls, options = {}) => {
+  return agentControls.getSpans().then(spans => {
+    if (options.expectSpans === false) {
+      expect(spans).to.be.empty;
```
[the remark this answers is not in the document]
kirrg001: Yeah good point!
Will change :)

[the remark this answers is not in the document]
kirrg001: Changed https://github.com/instana/nodejs/pull/532/commits/c1ca5b5d2d7dff1549aa6e76ee0af9313fb44281

packages/collector/test/tracing/database/db2/test.js
kirrg001: Ok will add the response body check 👍

packages/collector/test/tracing/database/db2/app.js
```diff
@@ -0,0 +1,676 @@
...
+      tries += 1;
+      return setTimeout(() => {
+        console.log('Trying again...');
+        connect();
+      }, 1000);
+    }
+
+    console.log('Successfully connected.');
+
+    conn.querySync(`drop table ${DB2_TABLE_NAME_1}`);
```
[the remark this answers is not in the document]
kirrg001: Hmm but it didn't fail on the cloud. But I can check again.

> Please add an if exists to that drop table statement.

Yeah `if exists` is the safer option 👍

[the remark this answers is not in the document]
kirrg001: Fixed 👍

packages/collector/test/tracing/database/db2/app.js
```diff
@@ -0,0 +1,676 @@
...
+  res.sendStatus(200);
+});
+
+app.delete('/db', (req, res) => {
+  console.log(`deleting database ${DB2_NAME}`);
+  ibmdb.dropDbSync(DB2_NAME, connStr);
+  console.log(`deleted database ${DB2_NAME}`);
+
+  res.sendStatus(200);
+});
```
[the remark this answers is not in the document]
kirrg001: Removed the endpoint 👍

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,539 @@
...
+    return resultPromise;
+  });
+}
+
+function instrumentExecuteHelper(ctx, originalArgs, stmtObject, httpParentSpan) {
+  const originalExecuteNonQuerySync = stmtObject.executeNonQuerySync;
+
+  stmtObject.executeNonQuerySync = function instanaExecuteNonQuerySync() {
+    return cls.ns.runAndReturn(() => {
+      cls.setCurrentSpan(httpParentSpan);
```
basti1302, line 332: _question:_ Since we cannot assume that the statement is prepared in the context of any HTTP request, we might need to rethink when and how we capture the CLS context/parent span and how we can restore it. Let's discuss that face to face.

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,539 @@
...
+}
+
+function handleTransaction(ctx, span) {
+  const originalEndTransaction = ctx.conn.endTransaction;
+  const originalEndTransactionSync = ctx.conn.endTransactionSync;
+
+  // NOTE: This is an internal fn to avoid instrumenting commit, rollback separately
+  ctx.conn.endTransaction = function instanaEndTransaction(rollback) {
+    if (rollback) {
+      span.cancel();
```
basti1302, line 514: _question:_ Should we cancel a span if a rollback happens or should that rather result in a span with an error? I tend to think the latter. The interaction with the database has actually happened, so I think this should be a span. We should check what we do for other db instrumentations.

packages/core/src/tracing/tracingUtil.js
```diff
@@ -176,6 +176,19 @@ exports.shortenDatabaseStatement = function shortenDatabaseStatement(stmt) {
...
+/**
+ * @param {string} connectionStr
+ * @returns {string}
+ */
+exports.sanitizeConnectionStr = function sanitizeConnectionStr(connectionStr) {
+  if (connectionStr == null || typeof connectionStr !== 'string') {
+    return undefined;
+  }
+
+  const replaced = connectionStr.replace(/PWD=.*?(?=;)/, 'PWD=REPLACED');
```
basti1302, line 188: _praise:_ Good catch to replace this. When we do secrets scrubbing we usually use `<redacted>` instead of `REPLACED, so we could go with that for consistency.

```suggestion
  const replaced = connectionStr.replace(/PWD=.*?(?=;)/, 'PWD=<redacted>');
```

basti1302 (changes requested):
Outstanding work, this is massive! 👍 👍 

Naturally, with such a huge instrumentation there are a few things that we might want to discuss before merging.

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,539 @@
...
+ *
+ * That's why we have to instrument `beginTransaction` and reactivate
+ * the span.
+ *
+ * We do not loose the parentSpan if the customer uses "beginTransactionSync".
+ */
+function instrumentBeginTransaction(originalFunction) {
+  return function instanaInstrumentBeginTransaction() {
+    const parentSpan = cls.getCurrentSpan();
+    const originalCallback = arguments[0];
```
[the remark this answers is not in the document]
kirrg001: Ok will add the paranoia protections 👍

packages/core/src/tracing/tracingUtil.js
kirrg001: Will change manually 👍

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,539 @@
...
+function instrumentPrepare(originalFunction) {
+  return function instanaInstrumentPrepare() {
+    const ctx = this;
+    const originalArgs = arguments;
+
+    if (skipTracing()) {
+      return originalFunction.apply(ctx, originalArgs);
+    }
+
+    const parentSpan = cls.getCurrentSpan();
```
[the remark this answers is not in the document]
kirrg001: Ok will add another test to prep the stmt on process start.
And add another `skipTracing` when `execute` is called 👍

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,539 @@
...
+
+    // CASE: There can be only one exit span, skip
+    if (constants.isExitSpan(parentSpan)) {
+      return originalFunction.apply(ctx, originalArgs);
+    }
+
+    // NOTE: prepare(stmt, cb) is the only possible usage
+    // TODO: I have not used cls.ns.bind here, because we remember the parentSpan
+    //       anyway and runAndReturn is executed per "execute" call, which creates a new context.
+    const originalCallback = originalArgs[1];
```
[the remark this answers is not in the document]
kirrg001: Same here. Paranoia protections incoming.

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,539 @@
...
+      .then(result => {
+        finishSpan(ctx, result, span);
+        return result;
+      })
+      .catch(err => {
+        span.ec = 1;
+        span.data.db2.error = tracingUtil.getErrorDetails(err);
+        finishSpan(ctx, null, span);
+        return err;
+      });
```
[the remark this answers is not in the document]
kirrg001: Ok 👍

packages/core/src/tracing/instrumentation/database/db2.js
kirrg001: Yeah true, let's trace a rollback 👍

packages/core/src/tracing/instrumentation/database/db2.js
kirrg001: I guess we can take the parenSpan from the execute call if that is doable 🤔

packages/core/src/tracing/tracingUtil.js
kirrg001: Done 👍

packages/core/src/tracing/instrumentation/database/db2.js
kirrg001: Done 👍

packages/core/src/tracing/instrumentation/database/db2.js
kirrg001: Done 👍

packages/collector/test/tracing/database/db2/app.js
```diff
@@ -0,0 +1,678 @@
...
+
+  if (error === 'prepare') {
+    stmt = '';
+  }
+
+  if (extraQuery) {
+    connection.querySync(`insert into ${DB2_TABLE_NAME_1} values (3, null, 'something')`);
+  }
+
+  connection.prepare(stmt, (err, stmtObject) => {
```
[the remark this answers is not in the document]
kirrg001: Done 👍

packages/core/src/tracing/instrumentation/database/db2.js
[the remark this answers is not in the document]
kirrg001: Done 👍

packages/core/src/tracing/instrumentation/database/db2.js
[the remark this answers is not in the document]
kirrg001: Done 👍

packages/core/src/tracing/instrumentation/database/db2.js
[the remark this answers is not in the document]
kirrg001: Done 👍

packages/core/test/test_util/expectAtLeastOneMatching.js
kirrg001: Done 👍

packages/collector/test/tracing/database/db2/app.js
```diff
@@ -0,0 +1,678 @@
...
+  let stmt = `insert into ${DB2_TABLE_NAME_1} (COLINT, COLDATETIME, COLTEXT) VALUES (?, ?, ?)`;
+  const params = [2, null, 'Adidas'];
+
+  if (error === 'execute') {
+    params[1] = 'any string';
+  }
+
+  if (error === 'prepare') {
+    stmt = '';
+  }
```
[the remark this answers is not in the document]
kirrg001: A use case to generate an error from the library to see how the span creation reacts.

> [Error: [IBM][CLI Driver] CLI0124E  Invalid argument value. SQLSTATE=HY009]

We do not trace this case right now.

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,612 @@
...
+ * Main difference of prepare and a normal query call:
+ * - you can call prepare once and re-use the statement (higher performance)
+ */
+function instrumentPrepare(originalFunction) {
+  return function instanaInstrumentPrepare() {
+    const ctx = this;
+    const originalArgs = arguments;
+
+    // CASE: we pass "true" to skip `cls.isTracing` because the prepare call can happen on process start
+    if (skipTracing(true)) {
```
kirrg001, line 131: Discussed: we can remove this

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,612 @@
...
+      return originalFunction.apply(ctx, originalArgs);
+    }
+
+    const possibleHttpParentSpan = cls.getCurrentSpan();
```
kirrg001, line 135: Discussed: 
We would like to add one more test case just to ensure we are doing everything right here:

- first http req (.prepare)
- second http req (.execute with the first http parent)

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,612 @@
...
+
+    // NOTE: prepare(stmt, cb) is the only possible usage
+    // TODO: I have not used cls.ns.bind here, because we remember the parentSpan
```
kirrg001, line 138: Discussed: We don't need it here

packages/core/src/tracing/instrumentation/database/db2.js
kirrg001: > cls.getCurrentSpan()

Change to `cls.getCurrentEntrySpan()` to be 100% sure we are remembering the right span here.

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,612 @@
...
+            span.ec = 1;
+            span.data.db2.error = tracingUtil.getErrorDetails(fetchCbErr);
+          }
+
+          return fetchCb.apply(this, arguments);
+        };
+
+        // NOTE: if the customer passes wrong arguments to the fetch fn
+        //       an error is thrown. We still want to keep the span
+        // TODO: Discuss if we want to cancel this span?
```
kirrg001, line 226: Discussed: No

kirrg001: Check again if we trace the error here correctly

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,539 @@
...
+    }
+
+    const customerCallbackIndex =
+      // eslint-disable-next-line no-nested-ternary
+      originalArgs.length === 2 && typeof originalArgs[1] === 'function'
+        ? 1
+        : originalArgs.length === 3 && typeof originalArgs[2] === 'function'
+        ? 2
+        : null;
+    const customerCallback = originalArgs[customerCallbackIndex];
```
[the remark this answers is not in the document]
kirrg001: Can change 👍

.gitignore
kirrg001, line 8: TODO: we need to be more explicit here

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,612 @@
...
+
+    // CASE: There can be only one exit span, skip
+    if (constants.isExitSpan(parentSpan)) {
+      return false;
+    }
+
+    // NOTE: remember the http parent of the first execute call,
+    //       otherwise the next execute call looses the parent
+    if (!prepareCallParentSpan && parentSpan) {
+      httpSpan = parentSpan;
```
kirrg001, line 351: TODO: check why we need this condition and explain better

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,612 @@
...
+      closeSyncCalled = true;
+      span.transmit();
+      return originalCloseSync.apply(this, arguments);
+    };
+
+    // CASE: customer forgets to call `result.closeSync`
+    //       search for result.closeSync() in:
+    //       https://github.com/ibmdb/node-ibm_db/blob/master/APIDocumentation.md
+    // TODO: discuss timeout length and the approach in general
+    setTimeout(() => {
```
kirrg001, line 541: Discussed:
`clearTimeout(t)` is missing when timeout was called
`unref` is missing e.g. when app is killed and should not block the runtime

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -0,0 +1,612 @@
...
+        if (typeof originalOnEndCallback === 'function') {
+          arguments[1] = function instanaOnEndOverride(onEndErr) {
+            if (onEndErr) {
+              span.ec = 1;
+              span.data.db2.error = tracingUtil.getErrorDetails(onEndErr) || 'Error not available.';
+            }
+
+            if (forceRollback && !onEndErr) {
+              span.ec = 1;
+              span.data.db2.error = 'Transaction was rolled back without error.';
```
kirrg001, line 574: Discussed: maybe remove for now but keep the test case

basti1302 (approved):

.gitignore
kirrg001: Done using `/db2`

packages/core/src/tracing/instrumentation/database/db2.js
kirrg001: Will remove for now till we agreed on a solution

PR #535, opened 2022-04-27
fix(tracing): cancel sqs entry span sync when no messages are available

basti1302:


packages/collector/src/logger.js
```diff
@@ -102,6 +102,9 @@ exports.getLogger = function getLogger(loggerName, reInitFn) {
   }
 
   if (reInitFn) {
+    if (registry[loggerName]) {
+      throw new Error(`Duplicate logger name: ${loggerName}.`);
+    }
```
basti1302, line 107: While working on the test for the sqs change, I noticed by chance that we have a few wrong logger names in the `getLogger('tracing/mongoose', newLogger => { ... });` calls, due to copy&paste errors. Using the same module logger name twice breaks the `instana.setLogger` call - one of the two modules that used the same logger name will not receive the new logger instance.

I know our mantra is "do not throw errors" but I am confident that this will only ever be thrown at development time when accidentally using a module logger name twice. No application under test will be able start so all tests will be red when this error condition exists.

WDYT?

packages/collector/test/test_util/CollectingLogger.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+'use strict';
+
+const { format } = require('util');
+
+/**
+ * A logger that does not emit logs anywhere but collects them. Useful for validating what has been logged in tests.
+ */
+module.exports = class CollectingLogger {
```
basti1302, line 12: This is basically a vehicle to make "Cannot start an AWS SQS entry span when another span is already active." is not logged, that is, we do not run into a situation where the previous SQS entry span is still active.

packages/collector/test/test_util/TeeLogger.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+'use strict';
+
+/**
+ * Multiplexes logs to a collection of target loggers. Think tee (Like https://man7.org/linux/man-pages/man1/tee.1.html)
+ * but for loggers.
+ */
+
+module.exports = class TeeLogger {
```
basti1302, line 12: This is used to also still see logs on the console/in the application's output while also collecting the logs for inspection in the test.

packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/receiveMessage.js
```diff
@@ -6,43 +6,74 @@
...
 const instana = require('../../../../../../')();
+
 const express = require('express');
-const port = process.env.APP_PORT || 3216;
-const agentPort = process.env.INSTANA_AGENT_PORT || 42699;
 const request = require('request-promise');
 const { sendToParent } = require('../../../../../../../core/test/test_util');
-
-const app = express();
-
```
basti1302, line 16: Re-arranged imports and variable initialization a bit, these statements have not actually been deleted, just moved.

packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/test.js
```diff
@@ -28,7 +28,7 @@ const { verifyHttpRootEntry, verifyHttpExit } = require('@instana/core/test/test
 const defaultPrefix = 'https://sqs.us-east-2.amazonaws.com/410797082306/';
 const queueUrlPrefix = process.env.SQS_QUEUE_URL_PREFIX || defaultPrefix;
 
-let queueName = 'team_nodejs';
+let queueName = 'nodejs-team';
```
basti1302, line 31: A queue with this name actually existed in our AWS account, while `team_nodejs` did not exist.

kirrg001: As long as the clean-aws.js script still works, I am fine with this change 👍

packages/collector/src/logger.js
basti1302: Basically all integration tests will fail once a module does that. We always load/require all modules at the start of the application, so no matter which integration test is executed, the error will be thrown in the application under test. Failing ~80% of our test cases should be a good enough signal to fix the problem I guess.

kirrg001: Cool 👍

kirrg001 (approved):
Sweet 👌🏻👌🏻

PR #541, opened 2022-05-10
ci(*): retry integration tests once on CI

basti1302:
Plus: 

* refactor(profiling): move global mocha hooks into dedicated file
* refactor(test): use Mocha root hook plug-ins instead of root hooks


packages/shared-metrics/test/hooks.js
```diff
@@ -0,0 +1,24 @@
...
+
+// This file defines global (root level) before and after hooks, see https://mochajs.org/#root-hook-plugins.
+// This module needs to be pre-required via Mocha's --require parameter when running the tests in this package. The npm
+// scripts for running tests take care of that.
+//
+// The globalAgent module manages an agent stub instance that can be used globally for all tests.
+
+const { startGlobalAgent, stopGlobalAgent } = require('../../collector/test/globalAgent');
+
+exports.mochaHooks = {
```
kirrg001, line 16: Cool - wasn't aware of this option.

packages/collector/test/hooks.js
```diff
@@ -0,0 +1,24 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2020
```
kirrg001, line 3: Want to change to 2022, because you touched it?
Ignore if you don't want to :)

kirrg001 (approved):
Cool!
Would like to see this in action

packages/collector/test/hooks.js
basti1302: I'm also always a bit unsure when exactly that would need to happen, but I think in this case it is more correct to leave the years as it is. In general, the copyright year would be the year when the original work has been created. True, this is a new file, but I basically renamed an existing file, so it is more a modification of the existing work from 2020. And you would not update the copyright year for any modifications of existing work. (See https://stackoverflow.com/a/2391555/2565264 for more information on this very boring topic :-) )

kirrg001: I am also not 100% sure. Ok then we agree that changing the copyright year has no use cases.

PR #542, opened 2022-05-11
fix(agent): reduce log level for sending data from debug to trace

basti1302:
When enabling debug logging, we write one log line each time something is
sent to the agent, which happens at least twice per second. These log messages
have been reduced to level trace now, so it is easier to see read debug logs
due to reduced noise. The previous behavior can still be activated
with INSTNANA_LOG_LEVEL=trace.


kirrg001 (approved):
Fine with me 👍

PR #543, opened 2022-05-15
ci(coverage): fix test for nyc coverage job

basti1302:
The app under test checked NODE_OPTIONS as the test added that for some
configurations but not for others. Turns out, nyc always adds
NODE_OPTIONS so this check needs to be a bit more sophisticated.


kirrg001 (approved):
🥳

PR #550, opened 2022-05-23
clean up .eslintrc.js structure & remove Node.js-4-compliant syntax

basti1302:


bin/.eslintrc.js
```diff
@@ -6,15 +6,6 @@
 'use strict';
 
 module.exports = {
```
kirrg001, line 8: Question: Why do we need an extra `eslintrc.js` in bin, misc folders, which does not inherit from the root file?

packages/collector/test/agentCommunication_test.js
```diff
@@ -67,8 +67,8 @@ describe('agentCommunication', function () {
         )
       ));
 
-  it('sends a Node.js EOL alert event to the agent (when applicable)', async () => {
-    return retry(async () => {
+  it('sends a Node.js EOL alert event to the agent (when applicable)', async () =>
+    retry(async () => {
```
kirrg001, line 71: Question: Is this rule enforced now? That would be not ideal. I'd like to make this optional.

kirrg001 (commented):
Great cleanup! I've left two questions

bin/.eslintrc.js
basti1302: They **all** inherit from the root eslint file (except for the test eslint configs). At some point, eslint changed the logic around inheritance a bit, now it always implicitly inherits from all eslint configs it finds when going _upwards_ in the directory hierarchy:

https://eslint.org/docs/user-guide/configuring/configuration-files#cascading-and-hierarchy

The only exception is when you explicitly tell it to `extend` from a specific path, then it will extend from that path and from everything in the directory hierarchy. 

Last but not least, with `extends` and `root: true`, it will only extend from the explicity provided `extends` path.

The reason for the config files in bin and misc are the few rules I explicitly add there:
* `'no-console': 'off'` for scripts in bin and
* `'import/no-unresolved': 'off'` for misc, because on CI we do not run `npm install` in those locations. (no-unresolved compares what is installed in node_modules with your require statements, and there is no node_modules folder if we don't run `npm install`)

packages/collector/test/agentCommunication_test.js
basti1302: Yes, that is enforced now. I don't have a strong opinion, I'll set the rule `off` again.

kirrg001: 🙏

bin/.eslintrc.js
kirrg001: Ah ok 👌🏻

kirrg001 (approved):
🌷

PR #552, opened 2022-05-24
test(google cloud storage): use random object names

basti1302:
Context: 

```
@instana/collector: Retrying test after failure: tracing/cloud/gcp/storage::must trace google cloud storage file-save (promise API) (retry 1/1)
@instana/collector: Google Cloud Storage Client (6327):	 { Error: Object 'nodejs-tracer-test-bucket/file.txt' is subject to bucket's retention policy and cannot be deleted, overwritten or archived until 2022-05-23T22:11:17.715679-07:00
@instana/collector:     at Gaxios._request (/home/circleci/repo/node_modules/gaxios/build/src/gaxios.js:127:23)
```

https://app.circleci.com/pipelines/github/instana/nodejs/3524/workflows/58b9131b-4830-4422-95fc-0da439a5e7be/jobs/15124/parallel-runs/16?invite=true#step-137-883

----

This is another attempt at making the Google Cloud Storage integration
tests more stable: Tests that performed bucket operations were already
using random bucket names, but tests for object operations were
previously all operating on the same bucket and on fixed object names.
Thus, concurrently running tests (for example the test suite for a
different Node.js version) would all operate on the same file in the
same bucket. This might lead to conflicts.

Additionally, tests sometimes failed when trying to delete an object
from the bucket, due to a retention policy. Therefore the test that sets
(and removes) a retention policy now operates on its own unique bucket.

### DIFF
--- a/packages/collector/test/tracing/cloud/gcp/storage/app.js
+++ b/packages/collector/test/tracing/cloud/gcp/storage/app.js
@@ -16,8 +16,6 @@ const path = require('path');
 const { Storage } = require('@google-cloud/storage');
 const { v4: uuid } = require('uuid');
 
-const asyncRoute = require('../../../../test_util/asyncExpressRoute');
-
 const logPrefix = `Google Cloud Storage Client (${process.pid}):\t`;
 
 const options = { projectId: process.env.GCP_PROJECT };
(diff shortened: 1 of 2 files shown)


kirrg001 (approved):
🥳

PR #553, opened 2022-05-24
chore: reverted optional dev dependencies and removed npm audit on root package

kirrg001:
I've noticed that even if the packages are installed, there is a longer check which slows down `npm install` locally


basti1302 (approved):
LGTM

kirrg001:
We removed `optionalDevDependencies` and switched back to `optionalDependencies`.
`optionalDependencies` did not work previously because `npm audit` failed. 

@basti1302 realized in the morning that we not have to execute `npm audit` on the root package. WIth this change, we can reuse `optionalDependencies`.

PR #554, opened 2022-05-25
test: split node jobs

kirrg001:
refs 97307

- we use a lot of resources of our circle CI credits, we would like to reduce them
- plus: speed up development

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -302,7 +302,7 @@ jobs:
       - <<: *nats-streaming
       - <<: *memcached
     <<: *shared
-    parallelism: 16
+    parallelism: 20
 
   "node-16":
     docker:
@@ -321,7 +321,7 @@ jobs:
       - <<: *nats-streaming
       - <<: *memcached
     <<: *shared
-    parallelism: 16
+    parallelism: 20
 
   "node-14":
     environment:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -363,7 +363,7 @@ jobs:
       - <<: *nats-streaming
       - <<: *memcached
     <<: *shared
-    parallelism: 16
+    parallelism: 20
```
kirrg001, line 366: We have the option to either run them with low parallel config or high.
My thinking was: if I want to rebuild the job before a release, I want a fast result.
In the night, I don't care. TBD

.circleci/config.yml
```diff
@@ -411,6 +411,15 @@ workflows:
     jobs:
       - "node-18"
       - "node-16"
```
kirrg001, line 413: v16 and v18 or only v16? TBD

kirrg001: Agreed v16/v18

.circleci/config.yml
basti1302: In that case, wouldn't it make sense to set a high parallelism for Node 16 ad 18, and a low one (maybe `parallelism: 1`) for the older Node versions that are only run nightly?

basti1302 (approved):
LGTM

.circleci/config.yml
kirrg001: > and a low one (maybe parallelism: 1) for the older Node versions that are only run nightly?

Yeah totally, but what about this:

> if I want to rebuild the job before a release (e.g. because it failed nightly)

PR #555, opened 2022-05-25
test: removed ts from core/test

kirrg001:
refs 97306

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,13 +14,8 @@
   "include": [
     "packages/core/src/util",
     "packages/core/src/tracing/*.js",
-    "packages/core/test/test_util/common_verifications.js",
     "packages/shared-metrics",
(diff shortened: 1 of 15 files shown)


tsconfig.json
```diff
@@ -14,13 +14,8 @@
   "include": [
     "packages/core/src/util",
     "packages/core/src/tracing/*.js",
-    "packages/core/test/test_util/common_verifications.js",
     "packages/shared-metrics",
     "packages/collector/src"
```
kirrg001, line 18: Question: What was the reason that only some public packages/exports are typed?

basti1302: I think the reason was purely that it is a lot of work to retroactively add types to existing code and Willian's plan was to do it iteratively.

basti1302 (approved):
LGTM

PR #563, opened 2022-06-21
fix: handle aws lambda timeout error handling better

kirrg001:
refs 98745

Please see the log file attached in the ticket for more context.

### DIFF
--- a/packages/serverless/src/backend_connector.js
+++ b/packages/serverless/src/backend_connector.js
@@ -183,6 +183,7 @@ function send(resourcePath, payload, finalLambdaRequest, callback) {
   let req;
   const skipWaitingForHttpResponse = !proxyAgent && !useLambdaExtension;
   const transport = getTransport();
+
   if (skipWaitingForHttpResponse) {
     // If the Lambda extension is not available to act as a proxy between the Lambda and serverless-acceptor (and
     // additionally, if no user-configured proxy is in place), we change the HTTP handling a bit to reduce the time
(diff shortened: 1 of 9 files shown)


packages/serverless/src/backend_connector.js
```diff
@@ -223,6 +224,14 @@ function send(resourcePath, payload, finalLambdaRequest, callback) {
   req.on('timeout', () => onTimeout(req, resourcePath, payload, finalLambdaRequest, callback));
 
   req.on('error', e => {
+    // CASE: we manually destroy streams, skip these errors
+    // Otherwise we will produce `Error: socket hang up` errors in the logs
+    // We already print the warning that a timeout happened
+    // https://nodejs.org/api/http.html#requestdestroyed
+    if (req.destroyed) {
```
kirrg001, line 231: When a timeout occurs when talking to the layer, we destroy the initial request, which results in the error `socket hang up` in the logs. In parallel we send the request directly. 

The error itself should be ignored, otherwise it's a confusing error and the data would be sent twice to the BE directly.

The second problem in the ticket was that the lambda kept alive 120s (the configured timeout). I doubt this is related to the fix, but I don't know for sure. Might be caused by the customer. We'll see.

basti1302 (approved):
The change for the production code looks good. Nice find!

I do not fully understand what benefit we gain from the changes to the test setup. Do we really need an extension mock that forwards data to the back end? Which cases in backend_connector can we cover with that setup, that we could not have covered with the previous setup?

basti1302:
> The change for the production code looks good. Nice find!
> 
> I do not fully understand what benefit we gain from the changes to the test setup. Do we really need an extension mock that forwards data to the back end? Which cases in backend_connector can we cover with that setup, that we could not have covered with the previous setup?

^ discussed: We need two different processes (mock extension and mock backend) to test the case where the first request goes to the extension and the second one goes to the backend.

PR #566, opened 2022-06-27
feat(aws-lambda): added support for Node v16

kirrg001:
- see https://aws.amazon.com/de/about-aws/whats-new/2022/05/aws-lambda-adds-support-node-js-16/
- `Error.js` is no longer available in /var/runtime


basti1302 (approved):
👍 👍

PR #579, opened 2022-07-19
chore: added docs for running lambda locally with a lambda emulator

kirrg001:


basti1302:
There is also https://github.com/instana/nodejs/tree/main/packages/aws-lambda/test/runtime_emulator (which I completely forgot) -- I wonder if those two approaches overlap?

kirrg001:
Oh :D that example code is in the test folder.
I'd suggest to move this into `aws-lambda/lamdas`.

(I will extend the existing README with the alternative approach)

basti1302 (approved):

PR #580, opened 2022-07-21
feat: added new package @instana/opentelemetry-sampler

kirrg001:
refs 99429

Samplers are not supported in the contrib repo of Otel.
The docs say that samplers need to be outside of their main repo.

- [x] replace NPM reference for propagator (currently a local tgz, build is red)

### DIFF
--- a/packages/opentelemetry-sampler/src/InstanaAlwaysOnSampler.js
+++ b/packages/opentelemetry-sampler/src/InstanaAlwaysOnSampler.js
@@ -1,0 +1,42 @@
+/*
+ * (c) Copyright IBM Corp. 2022
+ */
+
+'use strict';
+
+const { SamplingDecision, TraceFlags, trace } = require('@opentelemetry/api');
+
+class InstanaAlwaysOnSampler {
+  /**
+   * The Instana always on sampler can only be used in combination with the
+   * Instana propagator.
+   *
+   * The sampler will read the span context and check whether tracing is enabled based
+   * on the incoming Instana headers.
+   */
+  shouldSample(context) {
+    // NOTE: We need to use `RECORD_AND_SAMPLED`, otherwise `traceFlags` will be set to 0 (0000 0000)
+    let decision = SamplingDecision.RECORD_AND_SAMPLED;
+    try {
+      const nonRecordingSpan = trace.getSpan(context).spanContext();
+
+      // https://www.w3.org/TR/trace-context/#trace-flags
+      // eslint-disable-next-line no-bitwise
+      if ((nonRecordingSpan.traceFlags & TraceFlags.SAMPLED) !== TraceFlags.SAMPLED) {
+        decision = SamplingDecision.NOT_RECORD;
+      }
+    } catch (e) {
+      // ignore
+    }
+
+    return {
+      decision
+    };
+  }
+
+  toString() {
+    return 'InstanaAlwaysOnSampler';
+  }
+}
+
+module.exports = InstanaAlwaysOnSampler;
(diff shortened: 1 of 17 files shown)


basti1302 (approved):
LGTM

packages/opentelemetry-sampler/src/InstanaAlwaysOnSampler.js
```diff
@@ -0,0 +1,40 @@
...
+   * The sampler will read the span context and check whether tracing is enabled based
+   * on the incoming Instana headers.
+   */
+  shouldSample(context) {
+    // NOTE: We need to use `RECORD_AND_SAMPLED`, otherwise `traceFlags` will be set to 0
+    let decision = SamplingDecision.RECORD_AND_SAMPLED;
+
+    try {
+      const nonRecordingSpan = trace.getSpan(context).spanContext();
+      if (nonRecordingSpan.traceFlags === 0) {
```
basti1302, line 23: *issue(blocking):* This is not the correct way to read the flags from the context. See my change here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1077/files#diff-11002f6f5094c073f11ebb3418523fa956999dd6637cf0552e907589d9f0f475R75-R77

The W3C trace context spec describes the correct way to read/interprete the flags here: https://www.w3.org/TR/trace-context/#trace-flags:

> As this is a bit field, you cannot interpret flags by decoding the hex value and looking at the resulting number. For example, a flag 00000001 could be encoded as 01 in hex, or 09 in hex if present with the flag 00001000. A common mistake in bit fields is forgetting to mask when interpreting flags.
> 
> Here is an example of properly handling trace flags:
> 
> static final byte FLAG_SAMPLED = 1; // 00000001
> ...
> boolean sampled = (traceFlags & FLAG_SAMPLED) == FLAG_SAMPLED;

basti1302: *remark*: IIRC the plan was to always sample, unless X-INSTANA-L=0 is present. Since we read the sampled flag instead of directly reading the HTTP headers, we would also decide to not sample if any other propagator has created a context and set the sampled flag to 0. AFAICT we cannot do anything else here. So actually the original plan was flawed :-)

The docs say to only use this with the Instana propagator, so we might argue that this is a non-issue.

Anyway, I would leave the behavior here as it is (except for the bit masking change I mentioned above). Maybe we should change the name though? "AlwaysOn" does not really fit what we are doing here, changing this to just InstanaSampler might be better.

basti1302 (changes requested):
LGTM except for one blocking issue (the way we read the sampled flag).

packages/opentelemetry-sampler/src/InstanaAlwaysOnSampler.js
kirrg001: > The W3C trace context spec describes the correct way to read/interprete the flags here: https://www.w3.org/TR/trace-context/#trace-flags:

Will adapt so it is future proof 👍

kirrg001: > should change the name though? "AlwaysOn" does not really fit what we are doing here, changing this to just InstanaSampler might be better.

Not sure 🤔 Maybe we want to add more samplers in the future, then it might be better to give the sampler a more specific name. This sampler is always on except tracing is disabled. But no strong opinion here.

basti1302: I don't have a strong opinion either :-)  So use whatever you think is best. 👍

basti1302:
The propagator has been released now: https://www.npmjs.com/package/@opentelemetry/propagator-instana

basti1302:
I think this needs to be rebased? For some reason, parts of the work on the GH action appear in the diff. 😕 🤷

kirrg001:
Rebased and updated NPM package reference 👍

PR #585, opened 2022-08-05
feat(lambda): send preflight request to the AWS Lambda extension

basti1302:
Note: This PR currently contains a modified Lambda extension binary from https://github.ibm.com/instana/lambda-extension/pull/5. We need to review&merge the Lambda extension PR first, then update the binary here again, before merging this PR.

----

The Node.js Lambda in-process collector will now send a preflight
request (POST /preflight) directly at the start of the invocation.

This serves two purposes:
1) The Lambda extension can employ a much shorter timeout when HTTP
  communication between the in-process collector and the extension does
  not work. Previously, the extension would have waited until shortly
  before the Lambda timeout, potentially turning a sub-second Lambda
  invocation into a very long invocation (depending on the timeout set
  on the Lambda). Now, it can safely signal "next" to the AWS Lambda
  runtime environment when the preflight request does not come in until
  one second after the start of the invocation.
2) The in-process collector also profits from this new protocol: It can
  establish whether the Lambda extension is up and running directly at
  the start of the invocation. Previously it might have only found out
  about that when sending the bundle at the end of the invocation.

### DIFF
--- a/packages/aws-lambda/layer/Dockerfile
+++ b/packages/aws-lambda/layer/Dockerfile
@@ -1,7 +1,7 @@
 # (c) Copyright IBM Corp. 2021
 # (c) Copyright Instana Inc. and contributors 2021
 
-FROM public.ecr.aws/lambda/nodejs:14
+FROM public.ecr.aws/lambda/nodejs:16
 
 WORKDIR /opt
 COPY tmp/ .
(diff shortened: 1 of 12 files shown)


basti1302:
We decided to postpone this for now and only release #586 for now.

kirrg001 (approved):
LGTM 👍

PR #586, opened 2022-08-08
fix(lambda): increase socket and HTTP timeout for Lambda extension

basti1302:
It turns out that 100 ms is sometimes not enough, although
communication to the Lambda extension happens on localhost. If the
extension is up and runnig but communication between the Node.js
in-process collector and the extension fails at some point, the
Node.js in-process collector will stop attempting to talk to the
extension. Consequently, the extension will wait to receive the Lambda
entry span, which will never happen, causing the Lambda invocation to
run into a timeout.


basti1302:
Once https://github.ibm.com/instana/lambda-extension/pull/4 has been reviewed and merged, we could also include the new binary from that PR here (or in a separate PR).

kirrg001 (approved):
👍

PR #589, opened 2022-08-09
docs(ci): improve & document Github action for publishing a release

basti1302:


basti1302:
This is related to https://github.com/instana/nodejs/compare/6910adcc568af60c1f648abc816daff8ff0bccdb...ee4ec895b7e9e37f8bac96e83cb1ce484890b30e

CONTRIBUTING.md
```diff
@@ -175,29 +175,22 @@ Visit https://www.npmjs.com/package/@instana/${package-name}/access and set publ
...
-#### Separate lerna version And lerna publish
-
-You might want to separate the version bumping and tagging from publishing to the npm registry. This is also possible. The separate lerna publish command (see below) is also helpful if the publish did not go through successfully.
+Lerna will determine if this is going to be a major, minor or patch version from the commit comments of all commits since the last release. It will also automatically update the CHANGELOG.md files in the root of the repository and also in all individual packages, based on those commits.
 
-- Run `lerna version`.
-- Lerna will determine if this is going to be a major, minor or patch version from the commit comments of all commits since the last release. It will also update all relevant changelog files.
-- Lerna will push the commit and the tag created by `lerna version` to GitHub. Check that this has happened. Also check that the version numbers in package-lock.json have been updated as well.
-- Acquire an OTP token for 2fa.
-- `NPM_CONFIG_OTP={your token} lerna publish from-package && lerna bootstrap`
```
kirrg001, line 196: Discussion:Wondering if we would like to keep this. The GH action can still run into this error and then we need to manually fix the situation.

Exactly for this case

> That last command (`lerna publish from-package`) can also be run if the previous publish command went through for a subset of packages but not for others. Lerna will automatically figure out for which packages the latest version is not present in the registry and only publish those.

CONTRIBUTING.md
```diff
@@ -217,16 +210,6 @@ NPM_CONFIG_OTP={TOKEN} lerna publish --force-publish --conventional-graduate &&
...
-#### Rate Limited OTP
-
-If publishing the packages fails with an error like this:
-
-```
-lerna ERR! E429 Could not authenticate ${npm-user-name}: rate limited otp
-```
-
-you simply need to wait five minutes before trying again. In case some packages have already been published and others have not, refer the advice about `lerna publish from-package` in the previous section.
-
```
kirrg001, line 229: Discus: The error can still occur on the GH action I assume.

kirrg001 (approved):
👍 Great work
Left 2 questions for discussion

CONTRIBUTING.md
basti1302: Yeah, I was 50/50 on this when I deleted this. I'll revert this deletion and add some verbiage that this is can be used if an automated release goes south. Ideally, I would like to move to a setup where we never publish anything to npm from a local machine, so might want a Github action that executes this, but for now it's probably better to keep this section.

CONTRIBUTING.md
basti1302: I've not seen this issue in a long while, but we might as well keep this as long as we keep the other section.

CONTRIBUTING.md
kirrg001: Also: if the action fails, do we get a mail or similar?

basti1302: Good point. You can enable notifications for actions for your account at https://github.com/settings/notifications (in the section titled "Actions"). But enabling that might also get you notifications for other repos you contribute to. I'll add that to the docs.

I guess I better also add a Slack notification to the action. I'll take care of that today.

CONTRIBUTING.md
```diff
@@ -91,22 +91,15 @@ npm run refresh-package-lock-files
 
 Note: The intent of the procedure above is satisfy security scanners (`npm audit`, Dependabot etc.). `package-lock.json` files are never included when publishing to the npm registry. Thy are only relevant when you execute an `npm install` in _this repository_. Users of `@instana` npm packages will have their own `package-lock.json` or `yarn.lock` file for their application, those are completely independent from the `package-lock.json` files in this repository. Thus, they need to take care of updating their transitive dependencies themselves. We only can make sure that the version ranges we use for our direct dependencies allow to install a dependency tree that has no known vulnerabilities.
 
-## Release Process
-
-### When Adding A New Package
+## Adding A New Package
```
basti1302, line 94: I took the opportunity to clean up this section a bit.

CONTRIBUTING.md
```diff
@@ -167,39 +164,58 @@ addToPackageLock package-lock.json @instana/serverless false
 
 Later on, the original package-lock.json is restored when `git checkout package-lock.json;` is run from the `audit` npm script that you added before (see the package.json section above)
 
-#### IMPORTANT: Set Publishing Access For The New Package
```
basti1302, line 170: This has been moved above now (not deleted).

CONTRIBUTING.md
```diff
@@ -217,16 +235,6 @@ NPM_CONFIG_OTP={TOKEN} lerna publish --force-publish --conventional-graduate &&
 
 NOTE: With each prerelease the changelogs are getting updated. It might be wishful to delete the changelog before doing a final major release.
 
-#### Rate Limited OTP
```
basti1302, line 220: This section has moved up now.

.github/workflows/release.yaml
```diff
@@ -32,30 +37,113 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-           node-version: '18'
+           node-version: "18"
 
-      - run: |
+      - name: Disallow major releases
+        run: .github/workflows/disallow-major-release.sh
```
basti1302, line 50: As discussed, this will prevent accidental major releases.

.github/workflows/release.yaml
```diff
@@ -32,30 +37,113 @@ jobs:
...
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+      FROM_PACKAGE: ${{ inputs.fromPackage }}
       DRY_RUN: ${{ inputs.dryRun }}
       SKIP_CI_STATUS_CHECK: ${{ inputs.skipCiStatusCheck }}
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-           node-version: '18'
+           node-version: "18"
```
kirrg001, line 47: Current local development version is v16. I doubt it makes a different if we use v16 or v18. 
Wondering if we would like to use the same current version everywhere 🤔

basti1302: Right. I also don't think it makes a difference because we do not execute any tests etc. in the GH action, I just used the most recent version of Node.js that was available. But you are right, for consistency we should use 16 for now. I'll change it.

basti1302: Fixed.

PR #590, opened 2022-08-10
fix(shared-metrics): package.json not being found

kirrg001:
refs #581

- replaced deprecated usage of process.mainModule (this is not the main fix, just a replacement)
- https://nodejs.org/api/process.html#processmainmodule
- the main fix is to fallback reading the process.argv for the following case:
  - application uses ES modules and preloads a commonJS file using `--require`
  - in this preload file the Instana collector is required
  - package.json & existing fallback solutions did not work

### DIFF
--- a/packages/core/src/util/applicationUnderMonitoring.js
+++ b/packages/core/src/util/applicationUnderMonitoring.js
@@ -30,7 +30,7 @@ function isAppInstalledIntoNodeModules() {
 
 /**
  * Looks for the app's main package.json file, parses it and returns the parsed content. The search is started at
- * path.dirname(process.mainModule.filename).
+ * path.dirname(require.main.filename).
  *
  * In case the search is successful, the result will be cached for consecutive invocations.
  *
@@ -43,7 +43,7 @@ function getMainPackageJsonStartingAtMainModule(cb) {
 
 /**
  * Looks for the app's main package.json file, parses it and returns the parsed content. If the given directory is null
- * or undefined, the search will start at path.dirname(process.mainModule.filename).
+ * or undefined, the search will start at path.dirname(require.main.filename).
  *
  * In case the search is successful, the result will be cached for consecutive invocations.
  *
@@ -83,7 +83,7 @@ function getMainPackageJsonStartingAtDirectory(startDirectory, cb) {
 }
 
 /**
- * Looks for path of the app's main package.json file, starting the search at path.dirname(process.mainModule.filename).
+ * Looks for path of the app's main package.json file, starting the search at path.dirname(require.main.filename).
  *
  * In case the search is successful, the result will be cached for consecutive invocations.
  *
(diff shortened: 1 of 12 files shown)


packages/shared-metrics/src/name.js
```diff
@@ -20,24 +20,26 @@ exports.payloadPrefix = 'name';
...
       return logger.warn('Failed to determine main package json. Reason: ', err.message, err.stack);
-    } else if (!packageJson && attempts < MAX_ATTEMPTS) {
+    } else if (!packageJson && attempts < exports.MAX_ATTEMPTS) {
       logger.debug('Main package.json could not be found. Will try again later.');
-      setTimeout(exports.activate, DELAY).unref();
+      setTimeout(exports.activate, exports.DELAY).unref();
       return;
     } else if (!packageJson) {
-      if (process.mainModule) {
+      if (require.main) {
```
kirrg001, line 38: This section was not covered in our tests. Added a test. Made the attempts & delay dirtily configureable via `exports` and `reset` fn. No need to add ENV variables for this IMO

packages/core/src/util/applicationUnderMonitoring.js
```diff
@@ -113,18 +113,44 @@ function getMainPackageJsonPathStartingAtDirectory(startDirectory, cb) {
...
+      // with "--require file.cjs", neither `require.main` nor the deprecated `process.mainModule`
+      // will have a value, because ES modules use `import.meta` and soon `import.main`.
+      // See https://github.com/nodejs/modules/issues/274
+
+      // eslint-disable-next-line max-len
+      // See https://github.com/nodejs/node/blob/472edc775d683aed2d9ed39ca7cf381f3e7e3ce2/lib/internal/modules/run_main.js#L79
+      // Node.js is using `process.argv[1]` internally as main file path.
+      // Check whether a module was preloaded and use process.argv[1] as filename.
+      // @ts-ignore
+      if (process._preload_modules && process._preload_modules.length > 0) {
```
kirrg001, line 144: The fix starts here. I wasn't able to think of a different solution. But satisfied because Node.js is doing the same internally.

basti1302: One alternative I could think of is to walk through the module cache until you find a module that has import.meta.main === true. But I think that would be much worse than your current solution. 👍   Plus, the property is not even implemented yet.

packages/shared-metrics/src/name.js
```diff
@@ -20,24 +20,26 @@ exports.payloadPrefix = 'name';
 // @ts-ignore
 exports.currentPayload = undefined;
 
-const MAX_ATTEMPTS = 60;
-const DELAY = 1000;
+exports.MAX_ATTEMPTS = 10;
```
basti1302, line 23: _question:_ Any specific reason for reducing it to 10? I don't have a strong opinion at all, 60 was probably totally arbitrary to begin with and might well be not very reasonable, just asking.

packages/shared-metrics/src/name.js
```diff
@@ -49,3 +51,9 @@ exports.activate = function activate() {
     exports.currentPayload = packageJson.name;
   });
 };
+
+exports.reset = () => {
+  exports.currentPayload = undefined;
+  exports.MAX_ATTEMPTS = 60;
```
basti1302, line 57: _issue:_ This is not consistent with the change from 60 to 10 above.

basti1302 (approved):
Great research & fix 👍

packages/shared-metrics/src/name.js
kirrg001: lol It should be 60

Issue #596, opened 2022-08-17
[Bug]: Error when using Lambda Arm64 architecture with Instana Extension 


spamalot22:
### Problem Description

When setting our lambda architecture to the new "Arm64" architecture, the extension fails to initialise with an unknown error:

`EXTENSION	Name: instana-lambda-extension	State: LaunchError	Events: []	Error Type: UnknownError`

Switching back to the default "x86_64" architecture resolves the issue. 
Due to the reduced costs and AWS's intention to eventually make Arm64 the default architecture, we are looking to migrate our existing functions.

There is a write-up here regarding changes required to make functions and layers compatible with the new architecture - potentially separate versions of the layer would be required for the two settings: https://aws.amazon.com/blogs/compute/migrating-aws-lambda-functions-to-arm-based-aws-graviton2-processors/

### Short, Self Contained Example

Deploy a lambda function using the layer as usual but set the architecture to "arm64" rather than "x86_64" under the function's runtime settings.

### Node.js Version

16

### package.json

```shell
N/A
```

### package-lock.json

```shell
N/A
```

kirrg001:
Hi! Thanks for the report.

I can confirm that arm64 is not supported yet. We will update this issue as soon as possible.

kirrg001:
Hi there!

We have released the layer for arm64, but we have not released the docs yet. We will let you know as soon as the docs go live.

In the meanwhile you can use the layer version below and replace your region:

`arn:aws:lambda:{REGION}:410797082306:layer:instana-nodejs-arm64:1`

spamalot22:
Many thanks for the quick turnaround on this! We will give it a try ASAP.

kirrg001:
https://www.ibm.com/docs/en/instana-observability/current?topic=kinesis-aws-lambda-native-tracing-nodejs
Docs are live now 👍

PR #605, opened 2022-08-29
feat(aws-lambda): added support for arm64 architecture 

kirrg001:
refs https://github.com/instana/nodejs/issues/596

### DIFF
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -58,6 +58,10 @@ if [[ -z $PACKAGE_VERSION ]]; then
   PACKAGE_VERSION=latest
 fi
 
+if [[ -z $LAMBDA_ARCHITECTURE ]]; then
+  LAMBDA_ARCHITECTURE='x86_64'
+fi
+
 PACKAGE_NAMES="@instana/aws-lambda@$PACKAGE_VERSION instana-aws-lambda-auto-wrap@$PACKAGE_VERSION"
 
 # The default layer name is instana-nodejs. If you want to push experimental changes under a different layer name, you
@@ -79,10 +83,13 @@ fi
 if [[ -z $DOCKER_IMAGE_NAME ]]; then
   if [[ $BUILD_LAYER_WITH == local ]]; then
     DOCKER_IMAGE_NAME=instana-aws-lambda-nodejs-local
+  elif [[ $BUILD_LAYER_WITH == npm ]] && [[ $LAYER_NAME == instana-nodejs ]]; then
+      DOCKER_IMAGE_NAME=$CONTAINER_REGISTRY/instana/aws-lambda-nodejs
   else
-    DOCKER_IMAGE_NAME=$CONTAINER_REGISTRY/instana/aws-lambda-nodejs
+    DOCKER_IMAGE_NAME=$CONTAINER_REGISTRY/instana/aws-lambda-nodejs-experimental
   fi
 fi
+
 if [[ -n $SKIP_DOCKER_IMAGE ]]; then
   # SKIP_DOCKER_IMAGE implies SKIP_DOCKER_IMAGE_PUSH
   SKIP_DOCKER_IMAGE_PUSH=true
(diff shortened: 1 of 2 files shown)


packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -79,36 +83,50 @@ fi
 if [[ -z $DOCKER_IMAGE_NAME ]]; then
   if [[ $BUILD_LAYER_WITH == local ]]; then
     DOCKER_IMAGE_NAME=instana-aws-lambda-nodejs-local
+  elif [[ $BUILD_LAYER_WITH == npm ]] && [[ $LAYER_NAME == instana-nodejs ]]; then
+      DOCKER_IMAGE_NAME=$CONTAINER_REGISTRY/instana/aws-lambda-nodejs
   else
-    DOCKER_IMAGE_NAME=$CONTAINER_REGISTRY/instana/aws-lambda-nodejs
+    DOCKER_IMAGE_NAME=$CONTAINER_REGISTRY/instana/aws-lambda-nodejs-experimental
```
kirrg001, line 89: This prevents deploying a docker image with the name "/instana/aws-lambda-nodejs" when using an experimental layer name

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -79,36 +83,50 @@ fi
...
+echo "ZIP_NAME: $ZIP_NAME"
+echo "LAMBDA_ARCHITECTURE: $LAMBDA_ARCHITECTURE"
+echo "SKIP_DOCKER_IMAGE: $SKIP_DOCKER_IMAGE"
+echo "SKIP_DOCKER_IMAGE_PUSH: $SKIP_DOCKER_IMAGE_PUSH"
+echo "DOCKER_IMAGE_NAME: $DOCKER_IMAGE_NAME"
+echo "REGIONS: $REGIONS"
+echo "PACKAGE_VERSION: $PACKAGE_VERSION"
+echo "BUILD_LAYER_WITH: $BUILD_LAYER_WITH"
+echo "SKIP_AWS_PUBLISH_LAYER: $SKIP_AWS_PUBLISH_LAYER"
+echo "####"
```
kirrg001, line 126: Moved them up to always see the output (no matter if prompt or not)

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -193,30 +211,36 @@ if [[ $BUILD_LAYER_WITH == local ]]; then
 
   echo "Building local tar.gz for @instana/core."
   cd core
+  rm -rf instana-core-*.tgz
   npm --loglevel=warn pack
   mv instana-core-*.tgz $LAYER_WORKDIR/instana-core.tgz
 
   echo "Building local tar.gz for @instana/serverless."
   cd ../serverless
+  rm -rf instana-serverless-*.tgz
```
kirrg001, line 220: When you previously ran the script it will happen that you get two zips in the folder e.g. 
instana-serverless-2.6.2
instana-serverless-2.6.3

> mv instana-serverless-*.tgz $LAYER_WORKDIR/instana-serverless.tgz

Will then fail

packages/serverless/ci/pipeline.yaml
```diff
@@ -294,7 +294,7 @@ jobs:
         file: instana-aws-lambda-npm-package/version
         reveal: true
 
-      - task: build-and-publish-aws-lambda-image
+      - task: build-and-publish-aws-lambda-image-x86_64
```
basti1302, line 297: _nitpick:_ I think we previously mentioned that the current naming is slightly misleading (my bad). How about we  rename the job (line 283) from `aws-lambda-layer-as-container-image` to something like `aws-lambda-layer-and-container-image` to be more clear that it does both, layer and container image? And maybe rename the two tasks to `build-and-publish-aws-lambda-layer-and-image-$architecture`?

basti1302 (commented):
Looks awesome!

packages/serverless/ci/pipeline.yaml
kirrg001: Oh yeah good idea. 

> build-and-publish-aws-lambda-layer-and-image-$architecture?

👍

basti1302 (approved):

PR #608, opened 2022-09-01
fix(sqs, sns): do not add message attributes if that would violate limit

basti1302:
It turns out that SQS and SNS enforce a limit of max 10 message
attributes. If the message already has some attributes, the ones that we
add for trace correlation could lead to a violation of that limit. In
that case we now refrain from adding message attributes and log a
warning instead.

refs 103388

### DIFF
--- a/packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
+++ b/packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
@@ -135,7 +135,7 @@ function start(version) {
                 simple: withError !== 'sender'
               });
 
-              await verify(receiverControls, senderControls, response, apiPath, withError);
+              await verify({ receiverControls, senderControls, response, apiPath, withError });
               await verifyNoUnclosedSpansHaveBeenDetected(receiverControls);
             });
           });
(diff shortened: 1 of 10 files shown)


packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
```diff
@@ -239,112 +239,19 @@ function start(version) {
...
+                withError: false,
+                isBatch: true
+              });
               await verifyNoUnclosedSpansHaveBeenDetected(receiverControls);
             });
           });
         });
       });
-
-      async function verify(receiverControls, _senderControls, response, apiPath, withError, isBatch) {
```
basti1302, line 249: all the verify functions just got moved down (outside of this describe block)

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v2/sns.js
```diff
@@ -103,6 +104,12 @@ class InstanaAWSSNS extends InstanaAWSProduct {
       return;
     }
 
+    // SQS has a limit of 10 message attributes, we need to add two attributes.
+    if (Object.keys(attributes).length >= 9) {
+      logTooManyAttributesWarningOnce(attributes, 2);
```
kirrg001, line 109: *Suggestion*: Pass the logger from the actual instrumentation

e.g. `packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/sqs.js` is using the logger name `tracing/sqs/v3`

If the warning appears, there will be two different logger names for the same context.

kirrg001 (approved):

PR #609, opened 2022-09-06
feat(kafka): use kafka header format 'both' by default

basti1302:
~~DO NOT MERGE THIS YET!~~

This is the switch to phase 1 of the [Kafka header migration](https://www.ibm.com/docs/en/instana-observability/current?topic=references-kafka-header-migration). The plan is to release this in early October. 

See
https://www.ibm.com/docs/en/instana-observability/current?topic=references-kafka-header-migration
for more information.

### DIFF
--- a/packages/core/src/tracing/constants.js
+++ b/packages/core/src/tracing/constants.js
@@ -23,15 +23,22 @@ exports.kafkaLegacyTraceLevelHeaderName = 'X_INSTANA_L';
 exports.kafkaLegacyTraceLevelValueSuppressed = Buffer.from([0]);
 exports.kafkaLegacyTraceLevelValueInherit = Buffer.from([1]);
 
-// new kafka trace correlation (string values) starting approximately 2021-11
+// New kafka trace correlation (string values). Available as opt-in since 2021-10, and send out together with the legacy
+// binary headers by default starting in 2022-10. We will switch over to these headers completely (omitting the legacy
+// headers approximately in 2023-10.
 exports.kafkaTraceIdHeaderName = 'X_INSTANA_T';
 exports.kafkaSpanIdHeaderName = 'X_INSTANA_S';
 exports.kafkaTraceLevelHeaderName = 'X_INSTANA_L_S';
 
+/**
+ * @typedef {'binary' | 'string' | 'both'} KafkaTraceCorrelationFormat
+ */
+
+// With the current phase 1 of the Kafka header format migration, 'both', is the default.
+// With phase 2 (starting approximately October 2023) it will no longer be configurable and will always use 'string'.
+/** @type {KafkaTraceCorrelationFormat} */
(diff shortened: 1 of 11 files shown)


packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -340,42 +349,7 @@ function addTraceContextHeader(headers, span) {
...
-    case 'binary':
-      return addTraceContextHeaderBinary(headers, span);
-    case 'string':
-      return addTraceContextHeaderString(headers, span);
-    case 'both':
-    // fall through (both is the default)
-    default:
-      return addTraceContextHeaderBinary(addTraceContextHeaderString(headers, span), span);
-  }
-}
```
basti1302, line 353: I've removed all logic for adding binary headers from rdkafka, since they basically do not work anyway.

packages/core/src/util/normalizeConfig.js
```diff
@@ -114,9 +112,7 @@ const defaults = {
     disableW3cTraceCorrelation: false,
     kafka: {
       traceCorrelation: true,
-      // Before we start phase 3 of the migration, 'binary' will be the default value. With phase 1, we will move to
-      // 'both',  with phase 2 it will no longer be configurable and will always use 'string'.
-      headerFormat: 'binary'
+      headerFormat: constants.kafkaHeaderFormatDefault
```
basti1302, line 115: Using the default value from `tracing/constants.js`, so we do not need to define it twice. This also made it necessary to move the type definition to `tracing/constants.js`.

basti1302:
Converting back to draft so we don't accidentally merge it. Nonetheless, it can be reviewed already.

packages/core/src/tracing/constants.js
```diff
@@ -23,15 +23,20 @@ exports.kafkaLegacyTraceLevelHeaderName = 'X_INSTANA_L';
 exports.kafkaLegacyTraceLevelValueSuppressed = Buffer.from([0]);
 exports.kafkaLegacyTraceLevelValueInherit = Buffer.from([1]);
 
-// new kafka trace correlation (string values) starting approximately 2021-11
+// new kafka trace correlation (string values) starting approximately 2022-10
```
kirrg001, line 26: It says 2023 below

> (starting approximately October 2023)

basti1302: October 2022 refers to phase 1, where we start to inject both sets of headers by default. October 2023 refers to phase 2, where switch over to only send string headers.

basti1302: It is a bit ambigious. I'll make that more precise.

basti1302: Fixed, the new wording should be a bit clearer.

kirrg001 (approved):
👍

PR #612, opened 2022-09-13
feat(aws-lambda): added heartbeat

kirrg001:
refs 103353

### DIFF
--- a/packages/serverless/src/backend_connector.js
+++ b/packages/serverless/src/backend_connector.js
@@ -95,7 +95,7 @@ exports.init = function init(
   //            send data once. It can happen all the time till the Lambda handler dies!
   //            SpanBuffer sends data asap and when the handler is finished the rest is sent.
   if (useLambdaExtension) {
-    executeLambdaExtensionHeartbeatRequest();
+    scheduleLambdaExtensionHeartbeatRequest();
   }
 };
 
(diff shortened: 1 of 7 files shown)


packages/serverless/src/backend_connector.js
```diff
@@ -111,63 +115,78 @@ exports.sendSpans = function sendSpans(spans, callback) {
...
-      if (res.statusCode === 200) {
-        logger.debug('The Instana Lambda extension preflight request has succeeded.');
-      } else {
-        handlePreflightError(
-          new Error(
-            `The Instana Lambda extension preflight request has returned an unexpected status code: ${res.statusCode}`
-          )
-        );
+let heartbeatInterval;
+function executeLambdaExtensionHeartbeatRequest() {
```
basti1302, line 119: _nitpick:_ We could rename this to `scheduleLambdaExtensionHeartbeatRequest`.

packages/serverless/test/backend_connector_test.js
```diff
@@ -0,0 +1,219 @@
...
+ */
+
+'use strict';
+
+const expect = require('chai').expect;
+const sinon = require('sinon');
+const backendConnector = require('../src/backend_connector');
+const uninstrumentedHttp = require('../src/uninstrumentedHttp');
+
+const sleep = async ts => {
```
basti1302, line 12: _suggestion_: We already have a util for this, that we could use here: https://github.com/instana/nodejs/blob/main/packages/core/test/test_util/delay.js

packages/serverless/test/backend_connector_test.js
```diff
@@ -0,0 +1,219 @@
+/*
+ * (c) Copyright IBM Corp. 2022
+ */
+
+'use strict';
```
basti1302, line 5: Excellent idea to cover this with an additional unit test. 👍

basti1302 (approved):
LGTM

PR #615, opened 2022-09-19
fix(db2): redact password also from the end of the connection string

### LINKED ISSUE #614: [Bug]: Failed to sanitize DB2 connection string
### Problem Description

The sanitizeConnectionStr function regex doesn't match a connection string where PWD is the last parameter included and does not contain the semicolon. Package: tracingUtil.js
This causes string sanitize not to work.

### Short, Self Contained Example

Connection string: "DATABASE=MY_DATABASE;HOSTNAME=localhost;PORT=50000;PROTOCOL=TCPIP;UID=my_user;PWD=123456"

In this case the regex contained in the code does not match, as there is no semicolon.
**PWD=.*?(?=;)**
image

I suggest using the regex like this:
**PWD=.*?(?=;|$)**
image

### Node.js Version

Node.js v16.14.0

### package.json

```shell
{
  "name": "instana",
  "version": "1.0.0",
  "description": "",
  "main": "app.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "weslleyshavo",
  "license": "ISC",
  "dependencies": {
    "@instana/collector": "^2.7.1",
    "ibm_db": "^2.8.2"
  }
}
```

### package-lock.json

```shell
body is too long...
```

basti1302:
Improves the regex to redact the password by not expecting a subsequent semicolon character after the PWD parameter.

fixes #614

### DIFF
--- a/packages/core/src/tracing/tracingUtil.js
+++ b/packages/core/src/tracing/tracingUtil.js
@@ -185,7 +185,7 @@ exports.sanitizeConnectionStr = function sanitizeConnectionStr(connectionStr) {
     return undefined;
   }
 
-  const replaced = connectionStr.replace(/PWD=.*?(?=;)/, 'PWD=<redacted>');
+  const replaced = connectionStr.replace(/PWD\s*=\s*[^;]*/, 'PWD=<redacted>');
   return replaced;
 };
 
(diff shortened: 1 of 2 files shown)


kirrg001 (approved):
LGTM!

PR #617, opened 2022-09-19
fix(serverless): do not send x-instana-time header

basti1302:
This header is not actually required when sending data to the Instana back end.


kirrg001 (approved):
LGTM 👍

basti1302:
Currently the Instana back end will log a warning when this header is missing. The next release will no longer log this. We should keep this PR on the shelf for a while until merging it.

PR #619, opened 2022-09-20
fix(redis): suppress error for unsupported redis@4

basti1302:
Full fledged support for redis@4 will follow very soon.


basti1302:
Tested by installing redis@4 locally and verifying that the `error` log no longer appears, but the `debug` level (in case debug logging is active).

kirrg001 (approved):

PR #621, opened 2022-09-22
chore: tests, cleanup & clarity for existing redis instrumentation

kirrg001:
- Test coverage was not strong (https://output.circle-artifacts.com/output/job/0c9ea718-7097-4a14-822d-8c2e47411481/artifacts/0/coverage/lcov-report/core/src/tracing/instrumentation/database/redis.js.html)
- Clarify code sections & correct
- Added tests & proof that 0.x works/not works

V4 support is coming on top of this


packages/core/src/tracing/instrumentation/database/redis.js
```diff
@@ -137,23 +155,14 @@ function instrumentMultiExec(isAtomic, original) {
...
-    // CASE: multi = multiple executions
-    //       These executations are combined in one batch (see test).
-    //       And all of them need to same parent.
-    //       This parent has to be the incoming http server span.
-    //       If parentSpan.n is "redis", we need to use the original entry span instead. (http server)
-    //       because `cls.getCurrentSpan()` is the previous redis execution of the multi executions.
-    let parentSpan = cls.getCurrentSpan();
-    if (parentSpan.n === 'redis') {
-      parentSpan = cls.getCurrentEntrySpan();
-    }
```
kirrg001, line 149: There was a bug in the redis instrumentation. https://github.com/instana/nodejs/pull/621/files#diff-a128b813b4619f90c25af0b2da1a819592ad8f5aa182a93a1def91dbffa163deL47

Multi is defined as function, therefor `instrumentCommand` was executed for multi commands. A span was started, but never closed. Because `multi().get().set()` is not using callbacks and the callbacks are not called from the redis codebase.

We can remove the parent check, because I think we don't need this.
If there is a multi command, we capture one span with sub commands. This is what the test expects and has worked before. I have checked main against my assumption and main branch is also only creating one multi redis span with sub commands.

basti1302 (approved):
Very good improvement 👍

PR #639, opened 2022-10-12
chore(sqs, sns): remove fallback to legacy headers

basti1302:
This is the last step of a multi-year header migration from using X_INSTANA_T/X_INSTANA_S with signed long values over using X_INSTANA_ST/X_INSTANA_SS with string values back to X_INSTANA_T/X_INSTANA_S with string values.

In this final step, we remove the fallback to the legacy header names X_INSTANA_ST/X_INSTANA_SS on the consumer side. Tracers have been sending X_INSTANA_T and X_INSTANA_S with string values since approximately January 2020, so the grace period for updating has been a generous 22 months.


kirrg001 (approved):
👍

PR #643, opened 2022-10-19
feat(tracing): instrument prisma (ORM)

basti1302:


packages/core/src/util/initializedTooLateHeuristic.js
```diff
@@ -16,15 +16,19 @@ let hasBeenInitializedTooLate = false;
 
 let patterns = [
   /\/@apollo\/gateway\/dist\//,
+  /\/@aws-sdk\/smithy-client\//,
```
basti1302, line 19: The only new pattern is `/\/@prisma\/client\//`, the rest of the diff comes from restoring alphabetical order, scnr.

packages/collector/test/tracing/database/prisma/test.js
```diff
@@ -0,0 +1,212 @@
...
+
+const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : describe.skip;
+
+mochaSuiteFn('tracing/prisma', function () {
+  this.timeout(Math.max(config.getTestTimeout() * 3, 20000));
+
+  // We need to run npm install for the prisma application (which has its own package.json file) because prisma
+  // downloads a native database driver binary the current platform (Darwin, Linux, ...) as an npm postinstall hook.
+  before(async () => {
+    await executeAsync('npm install', appDir);
```
kirrg001, line 34: _Question_: Why do we not put the dependency `@prisma/client'` into our root package.json? We already have many native dependencies in their.

packages/core/src/tracing/instrumentation/database/prisma.js
```diff
@@ -0,0 +1,204 @@
...
+              if (!activeDatasource) {
+                return;
+              }
+
+              // We attach the provider and destination URL to the Prisma client instance. That way, when multiple
+              // Prisma client's are used, we do not confuse providers/database URLs
+              this.__inProvider = activeDatasource.activeProvider;
+              const dataSourceUrlObject = activeDatasource.url;
+              if (dataSourceUrlObject && dataSourceUrlObject.value) {
+                this.in_DataSourceUrl = redactPassword(activeDatasource.activeProvider, dataSourceUrlObject.value);
```
kirrg001, line 57: _Discuss_: Wondering if we should remember the values in a global array for each client, because it might be confusing if somebody is debugging and finds these custom Instana attributes on the PrismaClient class.

packages/core/src/tracing/instrumentation/database/prisma.js
```diff
@@ -0,0 +1,204 @@
...
+    const params = argsForOriginalRequest[0] || {};
+    span.data.prisma = {
+      model: params.model,
+      action: params.action,
+      provider: ctx.__inProvider,
+      url: ctx.in_DataSourceUrl
+    };
+    const requestPromise = originalRequest.apply(ctx, argsForOriginalRequest);
+    if (!requestPromise && typeof requestPromise.then !== 'function') {
+      span.cancel();
```
kirrg001, line 124: _Question_: Wondering why do we even start the span? We could call the original fn and start the span if we get a promise back. Ignore if you want to :)

kirrg001 (approved):
Great work 🥳

packages/collector/test/initEnv.js
```diff
@@ -21,6 +23,7 @@ const DEFAULT_ENV_VALUES = {
   POSTGRES_PASSWORD: 'nodepw',
   POSTGRES_DB: 'nodedb',
   POSTGRES_HOST: '127.0.0.1',
+  POSTGRES_URL: `postgresql://node:nodepw@127.0.0.1/nodedb?schema=prisma-${semver.major(process.version)}`,
```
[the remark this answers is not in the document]
basti1302: done

packages/collector/test/tracing/database/prisma/test.js
basti1302: There is a comment in line 31 that explains it. In addition to that particular reason, we also need to run `npm run generate` (that is, `prisma generate`) for each provider and Prisma is quite finicky with paths, the generate command seems to basically expect `prisma/schema.prisma` in the projects root folder.

I'll add some more comments to explain this.

basti1302: On second thought, the comment is indeed misleading. It is less about the native drivers (as you correctly pointed out), but only about `prisma generate` not working if the dependency is not installed locally here for this test app.

packages/core/src/tracing/instrumentation/database/prisma.js
```diff
@@ -0,0 +1,204 @@
...
+  if ('sqlserver' === provider.toLowerCase()) {
+    return redactPasswordFromMsSQLUrl(url);
+  }
+
+  // Parse standard datasource URI (basically everything except for MsSQL).
+  try {
+    const parsedUrl = new URL(url);
+
+    if (parsedUrl != null) {
+      parsedUrl.password = '_redacted_';
```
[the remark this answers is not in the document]
basti1302: Right, I went with `<redacted>` first but I ran into some issues because `<` and `>` are not allowed in URLs. That's why I decided to go with the safer variant `_redacted_`. We parse the URI in the back end to extract host and port for infrastructure correlation, and that will only work with a valid URI, using `<redacted>` would break that. Alternatively, we could percent-encode `<redacted>` but I think `_redacted_` is the better choice over `%3Credacted%3E`.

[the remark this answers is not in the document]
basti1302: I'll add a comment about that.

packages/core/src/tracing/instrumentation/database/prisma.js
basti1302: Good idea. 👍

basti1302: Done

packages/core/src/tracing/instrumentation/database/prisma.js
basti1302: I think that would be "less correct", strictly speaking? We want the prisma exit span to be the active span while `originalRequest.apply` is running. In practice it probably makes little difference, because we do not expect further activity that we would want to trace while that is happening. But one case where it could make a difference is our usual check that we do not start another exit span when an exit span is already active. For that to work we want the prisma exit to be active for the duration of `_request`. WDYT?

packages/collector/test/tracing/database/prisma/test.js
basti1302: I've expanded the comment for this.

PR #646, opened 2022-11-07
fix(sdk): fix memory leak when startEntrySpan is called recursively from the same context

basti1302:


packages/collector/test/tracing/sdk/entry_span_context_app.js
```diff
@@ -0,0 +1,245 @@
...
+        return createSdkSpanRecursivePromise(iterationsDone, maxIterations).then(resolve);
+      }, delayBetweenIterations);
+    });
+  }
+}
+
+async function createSdkSpanNonRecursiveAsync() {
+  await instana.sdk.async.startEntrySpan(spanName);
+  annotateWithParentContext();
+  await delay(durationOfSpans);
```
kirrg001, line 66: _question_: I was wondering if we want to add an exit span. This would guarantee that exit spans still keep the reference to it's parent entry span.

kirrg001 (approved):
Great work!

packages/core/src/tracing/clsHooked/async_local_storage_context.js
```diff
@@ -20,6 +20,8 @@ const unset = require('./unset');
 
 const CONTEXTS_SYMBOL = 'instanaClsHooked@contexts';
 
+const DEBUG_CLS = !!process.env.DEBUG_CLS;
```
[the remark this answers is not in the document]
basti1302: Good point. I wonder if we even want to keep this debug option at all. It was useful for testing and troubleshooting, but the two log statements are a bit arbitrary I think. How do you feel about removing this completely?

packages/collector/test/tracing/sdk/entry_span_context_app.js
basti1302: Various combinations of entry, exit and intermediate spans are already tested in this [test](https://github.com/instana/nodejs/blob/main/packages/collector/test/tracing/sdk/test.js) and [app](https://github.com/instana/nodejs/blob/main/packages/collector/test/tracing/sdk/app.js), I don't think we need to add this aspect here, too -- it would be redundant IMO.

Although that other test does not explicitly handly the `async` api, I'll add that there for completeness.

WDYT?

kirrg001: Perfect 👍

packages/core/src/tracing/clsHooked/async_local_storage_context.js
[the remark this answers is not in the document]
kirrg001: Good question. I was wondering about this as well, but having debug logs available in case there is a problem with a customer is quite nice. 

I would say: up to you 👍 Either remove, leave as is or optimise the usage of the env variable :D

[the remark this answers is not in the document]
basti1302: I'll remove it for now.

packages/collector/test/tracing/sdk/entry_span_context_app.js
basti1302: I'll add those tests as part of a follow up PR.

PR #650, opened 2022-11-08
test(sdk): add tests for async API

basti1302:
Technically, instana.sdk.async is only an alias for instana.sdk.promise, so the existing tests for instana.sdk.promise have covered that functionality already, but it is still good to have tests that use the typical call patterns used with code that uses async/await.

### DIFF
--- a/packages/collector/test/tracing/sdk/app.js
+++ b/packages/collector/test/tracing/sdk/app.js
@@ -18,11 +18,12 @@ const morgan = require('morgan');
 const path = require('path');
 const request = require('request-promise');
 
-const delay = require('../../../../core/test/test_util/delay');
+const { delay, getLogger } = require('../../../../core/test/test_util');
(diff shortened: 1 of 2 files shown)


packages/collector/test/tracing/sdk/app.js
```diff
@@ -166,41 +180,17 @@ app.post('/promise/create-intermediate', function createIntermediatePromise(req,
     });
 });
 
-app.post('/callback/create-overlapping-intermediates', async function createOverlappingIntermediatesCallback(req, res) {
```
basti1302, line 169: This test moved down but also changed/fixed. It did not actually test the functionality it was supposed to test (passing the span to completeIntermediateSpan). The test was still green, even when removing the extra `span2`/`span1` arguments provided to `completeIntermediateSpan` because it used the spans that were the currently active span at that point anyway.

kirrg001 (approved):

PR #653, opened 2022-11-10
feat(aws-lambda): added ES module support

kirrg001:
refs 104472

The plan:

- mark aws-lambda-auto-wrap npm package as deprecated (https://docs.npmjs.com/cli/v8/commands/npm-deprecate)
- remove publishing the package to npm in 3.x
- manually copy the commonjs handler to the layer zip in 3.x
- change the esm handler to a real ES module when AWS has fixed their layer bugs

Todo:

- [x] AWS support confirmed ES module bug in layer
- [x] add a simple test suite
- [x] create utils which both esm & cjs uses
- [x] code quality & code coverage
- [x] test docker container build
- [x] final test on AWS lambda

Docs updates coming soon

### DIFF
--- a/packages/aws-lambda-auto-wrap/esm/utils.js
+++ b/packages/aws-lambda-auto-wrap/esm/utils.js
@@ -1,0 +1,1 @@
+../src/utils.js
--- a/packages/aws-lambda-auto-wrap/test/cjs_test.js
+++ b/packages/aws-lambda-auto-wrap/test/cjs_test.js
@@ -1,0 +1,192 @@
+/*
+ * (c) Copyright IBM Corp. 2022
+ */
+
+'use strict';
+
+const sinon = require('sinon');
+const expect = require('chai').expect;
+const instana = require('@instana/aws-lambda');
+
+describe('cjs wrapper', function () {
+  before(() => {
+    const Module = require('module');
+    const originalRequire = Module.prototype.require;
+
+    Module.prototype.require = function fakeRequire(path) {
+      if (path === '/var/runtime/Errors.js') {
+        class HandlerNotFound extends Error {}
+        class MalformedHandlerName extends Error {}
+        class UserCodeSyntaxError extends Error {}
+        class ImportModuleError extends Error {}
+
+        const lambdaRuntimeErrors = {
+          HandlerNotFound,
+          MalformedHandlerName,
+          UserCodeSyntaxError,
+          ImportModuleError
+        };
+
+        return lambdaRuntimeErrors;
+      }
+
+      return originalRequire.apply(this, arguments);
+    };
+  });
+
+  beforeEach(() => {
+    sinon.stub(instana, 'wrap').callsFake(function fake(originalHandler) {
+      return function fakeWrapper() {
+        return originalHandler.apply(this, arguments);
+      };
+    });
+  });
+
+  afterEach(() => {
+    sinon.restore();
+    delete process.env.LAMBDA_TASK_ROOT;
+    delete process.env.LAMBDA_HANDLER;
+    delete process.env.LAMBDA_HANDLER;
+
+    delete require.cache[require.resolve('../src/index')];
+  });
+
+  it('should not successfully import a module which does not exist', async () => {
+    process.env.LAMBDA_TASK_ROOT = 'test/functions/doesnotexist';
+
+    try {
+      const wrapper = require('../src/index');
+      await wrapper.handler();
+    } catch (e) {
+      expect(e.message).to.contain("Error: Cannot find module 'index'");
+    }
+
+    expect(instana.wrap.called).to.be.false;
+  });
+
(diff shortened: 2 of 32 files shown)


packages/aws-lambda-auto-wrap/esm/utils.js
```diff
@@ -0,0 +1 @@
+../src/utils.js
```
basti1302, line 1: _question:_ This looks slightly weird, but probably only in the Github diff. I assume this is not actually a new `.js` file that only contains this single line, right? Is it a symlink? If so, do we need to have it as a symlink here? I guess it is about how we package/copy files in the publish-layer script, is it? Could you explain this a bit?

packages/aws-lambda-auto-wrap/test/cjs_test.js
```diff
@@ -0,0 +1,192 @@
...
+ * (c) Copyright IBM Corp. 2022
+ */
+
+'use strict';
+
+const sinon = require('sinon');
+const expect = require('chai').expect;
+const instana = require('@instana/aws-lambda');
+
+describe('cjs wrapper', function () {
```
basti1302, line 11: _praise:_ Kudos for adding tests for the original commonjs wrapper! 🙇

packages/aws-lambda-auto-wrap/test/functions/cjs-5/server
```diff
@@ -0,0 +1,9 @@
+/*
+ * (c) Copyright IBM Corp. 2022
+ */
+
+'use strict';
+
+exports.candle = async () => {
```
basti1302, line 7: 🕯️

packages/aws-lambda/lambdas/runtime_emulator/Dockerfile
```diff
@@ -12,7 +12,11 @@ COPY --from=instana-layer /opt/extensions/ /opt/extensions/
 COPY --from=instana-layer /opt/nodejs/ /opt/nodejs/
 
 # Copy function code
+# COPY esm/app.js ${LAMBDA_TASK_ROOT}
+# COPY esm/package.json ${LAMBDA_TASK_ROOT}
```
basti1302, line 16: _nitpick:_ Do we need these two commented out lines for testing esm? Do you want to add a comment here as well to explain that?

packages/aws-lambda/lambdas/runtime_emulator/build-and-run.sh
```diff
@@ -73,6 +73,7 @@ docker run \
   --env INSTANA_TIMEOUT=$instana_timeout \
   --env INSTANA_ENDPOINT_URL=$instana_endpoint_url \
   --env INSTANA_AGENT_KEY=$instana_agent_key \
+  #--env LAMBDA_HANDLER='esm/app.handler' \
```
basti1302, line 76: _nitpick:_ Do we need this line for testing esm? Do you want to add a comment here as well to explain that?

basti1302 (approved):
Looks great!

packages/aws-lambda-auto-wrap/esm/utils.js
kirrg001: Yes it is a symlink. So the esm/index.js has to require the utility and it cannot do `../src/utils`.
Because when we build the layer zip, the esm handler has no longer access to the location.
That's why we require `./utils` and create a symlink locally.

Do you have a better idea? I wanted to avoid to create a build process such as create a single JS file etc.

packages/aws-lambda/lambdas/runtime_emulator/Dockerfile
kirrg001: Yes if the developer wants to test esm, he also needs to modify the Dockerfile. 
I though the comment in line 19/20 is enough? What do you think?

packages/aws-lambda/lambdas/runtime_emulator/build-and-run.sh
kirrg001: Yes for esm. I will leave another comment :)

packages/aws-lambda/lambdas/runtime_emulator/Dockerfile
basti1302: sure

packages/aws-lambda-auto-wrap/test/cjs_test.js
kirrg001: I also ran the test against the original index.js handler from main branch. Looks good 👍

packages/aws-lambda-auto-wrap/esm/utils.js
basti1302: Thanks for the explanation, that makes sense. I don't have a better idea. We might have more options when we migrate the commonjs wrapper away from being an npm package to just copying it. Maybe we can revisit this then. Fine for now 👍

PR #654, opened 2022-11-14
test(prisma): do not disconnect manually from DB at end of test

basti1302:
We do not need this explicit disconnect step, since we are about to terminate the Prisma integration test app at that point anyway. None of the other DB integration test apps has such a manual disconnect step. Furthermore, a recent CI run failed because the `after` hook of the Prisma test suite failed, leaving the Prisma app running after the test was finished. This blocked the test apps for all subsequent tests from starting (because they all try to open the same port, 3215, and that port was still blocked by the Prisma app.)

### DIFF
--- a/packages/collector/test/tracing/database/prisma/test.js
+++ b/packages/collector/test/tracing/database/prisma/test.js
@@ -68,15 +68,6 @@ mochaSuiteFn('tracing/prisma', function () {
       });
 
       after(async () => {
-        // eslint-disable-next-line no-use-before-define
-        if (controls) {
-          // eslint-disable-next-line no-use-before-define
-          await controls.sendRequest({
-            method: 'POST',
-            path: '/disconnect'
-          });
-        }
-
         await fs.rm(schemaTargetFile, { force: true });
         await rimraf(migrationsTargetDir);
       });
--- a/packages/collector/test/tracing/database/prisma/app.js
+++ b/packages/collector/test/tracing/database/prisma/app.js
@@ -98,12 +98,6 @@ app.post('/update', async (req, res) => {
   }
 });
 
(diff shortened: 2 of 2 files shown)


packages/collector/test/tracing/database/prisma/test.js
```diff
@@ -68,15 +68,6 @@ mochaSuiteFn('tracing/prisma', function () {
       });
 
       after(async () => {
-        // eslint-disable-next-line no-use-before-define
```
kirrg001, line 71: https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/connection-management

> In most cases, you do not need to explicitly call these methods. PrismaClient automatically connects when you run your first query, creates a [connection pool](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/connection-pool), and disconnects when the Node.js process ends.


Look what I found. Prisma automatically disconnects when the process get's killed.

> None of the other DB integration test apps has such a manual disconnect step

True, but I think they should. I think disconnecting from an external resource before the process get's killed is correct. But nothing we should actively change now

kirrg001 (approved):

packages/collector/test/tracing/database/prisma/test.js
basti1302: Good find.

basti1302: About disconnecting from the DB from the apps for other integration tests: It might not matter. The app under test is killed anyway. Worst case is that the DB keeps the connection open for a while longer. Most DBs will probably close the connection on their end after a while when a client terminates without disconnecting first. And we also stop the DB containers once all tests are done.

PR #669, opened 2022-12-09
chore(aws-lambda): do not use `setTimeout` for span buffer

kirrg001:
refs 109158

We want to avoid using `setTimeout` in Faas, because the AWS Lambda runtime will execute the timeout handlers in a different lambda execution context. This is out of our control and can flood executions.

### DIFF
--- a/packages/aws-lambda/src/wrapper.js
+++ b/packages/aws-lambda/src/wrapper.js
@@ -247,6 +247,7 @@ function init(event, arnInfo, _config) {
   // instanaCore.init also normalizes the config as a side effect
   instanaCore.init(config, backendConnector, identityProvider);
 
+  spanBuffer.setIsFaaS(true);
   captureHeaders.init(config);
   metrics.init(config);
   metrics.activate();
@@ -363,6 +364,14 @@ function postHandler(entrySpan, error, result, callback) {
   }
 
   const spans = spanBuffer.getAndResetSpans();
+
+  // We want that all upcoming spans are send immediately to the BE.
+  // Span collection happens all the time and they are buffered via spanBuffer.
+  // The spanBuffer will only send out the spans early if the payload becomes too big.
+  // When the Lambda handler finishes we trigger `sendBundle`.
+  // If there is any span collected afterwards (async operations), we send them out
+  // directly and that's why we set `setTransmitImmediate` to true.
+  spanBuffer.setTransmitImmediate(true);
 
   const metricsData = metrics.gatherData();
 
--- a/packages/aws-lambda/src/wrapper.js
+++ b/packages/aws-lambda/src/wrapper.js
@@ -374,7 +374,6 @@ function postHandler(entrySpan, error, result, callback) {
   spanBuffer.setTransmitImmediate(true);
 
   const metricsData = metrics.gatherData();
-
   const metricsPayload = {
     plugins: [{ name: 'com.instana.plugin.aws.lambda', entityId: identityProvider.getEntityId(), data: metricsData }]
   };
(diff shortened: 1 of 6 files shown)


packages/aws-lambda/src/wrapper.js
```diff
@@ -232,14 +232,16 @@ function shimmedHandler(originalHandler, originalThis, originalArgs, _config) {
...
-  config = _config || {};
+  config = instanaCore.coreUtil.normalizeConfig(_config);
 
   if (config.logger) {
     logger = config.logger;
   } else if (process.env.INSTANA_DEBUG || config.level || process.env.INSTANA_LOG_LEVEL) {
     logger.setLevel(process.env.INSTANA_DEBUG ? 'debug' : config.level || process.env.INSTANA_LOG_LEVEL);
   }
 
+  config.internal.isServerless = true;
```
kirrg001, line 243: Rename `isFaas`

packages/core/src/tracing/spanBuffer.js
```diff
@@ -167,6 +183,13 @@ exports.addSpan = function (span) {
     return;
   }
 
+  // CASE: if isFaaS we no longer want to buffer spans after we have already send the bundle
+  if (isFaaS && transmitImmediate) {
```
basti1302, line 187: _question:_ Should we drop the `isFaas` part of the condition? If any code that is using `spanBuffer` sets `transmitImmediate` to true, the expectation is probably that spans are transmitted immediately, no matter which environment we are in. At the moment we only need this for FAAS, so it doesn't really matter. Still, IMO there is no reason to check `isFaas` here, checking `transmitImmediately` is enough I think. WDYT?

packages/core/test/tracing/spanBuffer_test.js
```diff
@@ -6,8 +6,9 @@
 'use strict';
 
 const { expect } = require('chai');
-
+const sinon = require('sinon');
```
basti1302, line 9: _praise:_ great stuff, testing both configurations with a unit test ❤️

basti1302 (approved):
LGTM! 🎉

packages/core/src/tracing/spanBuffer.js
```diff
@@ -120,21 +126,31 @@ exports.activate = function activate(extraConfig) {
...
+  // NOTE: We do not want to use `setTimeout` in AWS Lambda, because
+  //       the AWS runtime might execute the handler in a different lambda execution.
+  //       On AWS Lambda we wait till the handler finishes and then transmit all collected spans via
+  //       `sendBundle`. Any detected span will be sent directly to the BE.
+  if (!isFaaS) {
+    transmissionTimeoutHandle = setTimeout(transmitSpans, initialDelayBeforeSendingSpans);
+    transmissionTimeoutHandle.unref();
+    if (preActivationCleanupIntervalHandle) {
+      clearInterval(preActivationCleanupIntervalHandle);
+    }
```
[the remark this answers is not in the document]
kirrg001: 👍

packages/core/src/tracing/spanBuffer.js
kirrg001: Yes was thinking about that too 👍

packages/aws-lambda/src/wrapper.js
```diff
@@ -221,14 +221,16 @@ function shimmedHandler(originalHandler, originalThis, originalArgs, _config) {
  * Initialize the wrapper.
  */
 function init(event, arnInfo, _config) {
-  config = _config || {};
+  config = instanaCore.util.normalizeConfig(_config);
```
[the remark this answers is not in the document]
basti1302: > As far as I can see e.g. Fargate and GC doing the same

Oh, you're right. Good point. 👍 

> I think I will simplify this by adding spanBuffer.isFaaS(true). This was my initial though anyway.

I like that very much. It is not meant to be configured externally, so there is no reason to have it in the config.

PR #671, opened 2022-12-14
fix(aws-fargate): cannot read property 'cpu' of undefined

kirrg001:
refs 111063

### DIFF
--- a/packages/aws-fargate/src/metrics/task/EcsTaskProcessor.js
+++ b/packages/aws-fargate/src/metrics/task/EcsTaskProcessor.js
@@ -38,8 +38,8 @@ class EcsTaskProcessor extends DataProcessor {
       desiredStatus: metadata.DesiredStatus,
       knownStatus: metadata.KnownStatus,
       limits: {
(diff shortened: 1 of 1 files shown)


packages/aws-fargate/src/metrics/task/EcsTaskProcessor.js
```diff
@@ -38,8 +38,8 @@ class EcsTaskProcessor extends DataProcessor {
       desiredStatus: metadata.DesiredStatus,
       knownStatus: metadata.KnownStatus,
       limits: {
-        cpu: metadata.Limits.CPU,
-        memory: metadata.Limits.Memory
+        cpu: metadata.Limits ? metadata.Limits.CPU : undefined,
+        memory: metadata.Limits ? metadata.Limits.Memory : undefined
```
kirrg001, line 42: See https://github.com/instana/nodejs/blob/main/packages/aws-fargate/src/metrics/container/containerUtil.js#L32

basti1302 (approved):
👍

PR #672, opened 2022-12-22
fix(collector): missing spans when client app is ESM

kirrg001:
Added support for client app being an ESM app.

Unsupported: ESM instrumentations. We will extend the `esm-loader.mjs` asap to add this support. But this should not affect a customer right now, because there are no known big NPM modules, which have dropped the support for CJS completely (except e.g. `got`). Furthermore we do not offer an API to add custom instrumentations.


packages/collector/test/test_util/ProcessControls.js
```diff
@@ -89,7 +89,19 @@ class ProcessControls {
...
+      opts.appPath = path.join(opts.dirname, 'app.js');
+    }
+
+    if (process.env.RUN_ESM && !opts.execArgv) {
+      if (opts.dirname) {
+        const files = fs.readdirSync(opts.dirname);
+        const esmApp = files.find(f => f.indexOf('.mjs') !== -1);
+
+        if (esmApp) {
+          opts.execArgv = [`--experimental-loader=${path.join(__dirname, '..', '..', 'esm-loader.mjs')}`];
```
kirrg001, line 101: opts.execArgv = [`--require=${path.join(__dirname, '..', '..', 'load-instana.js')}`];

packages/collector/esm-loader.mjs
```diff
@@ -0,0 +1,25 @@
...
+'use strict';
+
+/**
+ * We currently only instrument CJS modules. As soon as we want
+ * to instrument ES modules (such as `got` v12), the requireHook will
+ * no longer work. Therefor we would need to wrap the target ES module
+ * with our instrumentations using the resolve & load hook.
+ */
+
+import instana from './src/index.js';
```
kirrg001, line 14: `node --experimental-loader=@instana/collector/esm-loader.mjs server.js`

packages/core/src/util/requireHook.js
```diff
@@ -54,6 +54,25 @@ exports.init = function (config) {
...
 function patchedModuleLoad(moduleName) {
+  let module = moduleName;
+
+  // CASE: when using ESM, the Node runtime passes a full path to Module._load
+  //       we try to grab the module name to being able to patch the target module
+  //       with our instrumentation
+  if (path.isAbsolute(moduleName) && moduleName.indexOf('.js') !== -1) {
+    try {
+      // node_modules/mysql/lib/index.js
+      module = moduleName.match(/node_modules\/(.*?(?=\/))/)[1];
```
basti1302, line 65: _nitpick:_ Looks like we are relying on the `[1]` access to throw a TypeError when there is no match and `match` returns `null`. Can we replace that with an actual null check on the result of `match` instead of using exceptions for flow control?

packages/core/src/util/requireHook.js
```diff
@@ -54,6 +54,25 @@ exports.init = function (config) {
  * @param {string} moduleName
  */
 function patchedModuleLoad(moduleName) {
+  let module = moduleName;
```
basti1302, line 57: _very nitpicky nitpick:_ I feel `moduleName` is still the better variable name. When I read `module` in the context of what we are doing in the require hook, I think it would refer to the actual module object that is returned by `require`. Is there a particular reason for introducing a new variable? We could reassign the `moduleName` in the code below, could we?

basti1302 (approved):
Awesome work! Looks great.


I have one small non-blocking concern about the introduction of the `app.mjs` files. They would introduce a bit of a maintenance burden. Going forward, whenever we need to change one of the apps under test, we would need to change both versions. There is also the risk of forgetting about that sometimes, so there is a small risk of `app.js` and `app.mjs` drifting apart. That said, I absolutely acknowledge the value that this new test setup brings. I also do not have better idea on how to get this level of test coverage without duplicating the files. So, to. be clear, I'm totally fine with merging these files as-is. If we later feel that the duplicated maintenance is an actual issue, we can still think of a better solution.

packages/core/src/util/requireHook.js
```diff
@@ -54,6 +54,25 @@ exports.init = function (config) {
...
+
+  // CASE: when using ESM, the Node runtime passes a full path to Module._load
+  //       we try to grab the module name to being able to patch the target module
+  //       with our instrumentation
+  if (path.isAbsolute(moduleName) && moduleName.indexOf('.js') !== -1) {
```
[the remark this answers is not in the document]
kirrg001: It was a different reason. I think I received a `.ts` file, which I wanted to ignore.

> string in their name (https://www.npmjs.com/package/gcstats.js 

Defo good point. I will try to remove the check.

packages/core/src/util/requireHook.js
kirrg001: 👍 Will use the original var

packages/core/src/util/requireHook.js
kirrg001: Sure why not

PR #685, opened 2023-01-20
chore: fixed pino tests

kirrg001:


package.json
```diff
@@ -162,6 +162,7 @@
     "pino": "^8.8.0",
     "pino-pretty": "^7.5.3",
     "pino-v7": "npm:pino@^7.8.1",
+    "pino-v6": "npm:pino@6.14.0",
```
basti1302, line 165: _remark:_ Alternatively, we could add `"pino": "pino@6.14.0"` to `packages/aws-fargate/package.json`, right? (And then do `require('pino')` the modules in `packages/aws-fargate/test/`.)

Both fine with me, no preference either way.

basti1302 (approved):
LGTM

package.json
kirrg001: Last year we decided to move all dev dependencies up to the main root package.json. 
For a better overview and to use the full potential of lerna. So not sure about this?

basti1302: I only brought it up because we have the special situation here that we need different versions in different packages; and even for different use cases, one is for testing our pino instrumentation and the other for using it as a logger in applications under test.

But nevermind, it's fine as it is now.

PR #686, opened 2023-01-23
feat(tracing, fetch): instrument native fetch

basti1302:


packages/collector/test/tracing/misc/spec_compliance/test.js
```diff
@@ -38,7 +38,7 @@ describe('spec compliance', function () {
   globalAgent.setUpCleanUpHooks();
 
   [false, true].forEach(http2 => {
-    const mochaSuiteFn = supportedVersion(process.versions.node) && !http2 ? describe : describe.skip;
+    const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : describe.skip;
```
basti1302, line 41: See commit comment:
> test(spec compliance): enable spec compliance test suite for http2 again
>
> This had been accidentally disabled in commit
> b25fe79a723f8f993e9488a8d62187c759e0cafe.

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
```diff
@@ -0,0 +1,304 @@
...
+  }
+  isActive = true;
+};
+
+exports.deactivate = function deactivate() {
+  isActive = false;
+};
+
+function instrument() {
+  global.fetch = function fetch() {
```
kirrg001, line 54: suggestion: naming `instanaFetch`

packages/core/src/tracing/instrumentation/protocols/_http.js
```diff
@@ -5,32 +5,51 @@
```
kirrg001, line 5: suggestion: can we rename this file to something like `httpUtils.js` or similar?

tsconfig.json
```diff
@@ -14,6 +14,7 @@
   "include": [
     "packages/core/src/util",
     "packages/core/src/tracing/*.js",
+    "packages/core/src/tracing/instrumentation/protocols/_http.js",
```
kirrg001, line 17: question: Is `_http` exposed as part of the collector package? What was the reason to add TS to it?

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
```diff
@@ -0,0 +1,304 @@
...
+    // eslint-disable-next-line no-unused-vars
+    let w3cTraceContext = cls.getW3cTraceContext();
+
+    const skipTracingResult = cls.skipExitTracing({ isActive, extendedResponse: true, skipParentSpanCheck: true });
+    const parentSpan = cls.getCurrentSpan() || cls.getReducedSpan();
```
kirrg001, line 59: help: I do understand the concept of reduced spans, but I currently struggle why we need this for http clients. Makes only sense to me for graphql subscriptions. Can we leave a comment?

kirrg001 (approved):
Great work <3

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
basti1302: Right, good call. Changed accordingly.

tsconfig.json
basti1302: No, it is not exposed. The reason is two-fold. I noticed that a lot of the method in that file didn't have any doc comment at all, and I wanted to fix that. There are a lot of very similar, yet slightly different methods in there -- all focussed on capturing headers from HTTP requests, tailored to very specific situations (capture from a Request object, from an object literal, merge headers etc.). It took me a while to sort out which method is doing what exactly, so I figured I should just add doc comments to them.

And while I added comments to the methods, I figured it would be best to use the JSDoc comment format, which we have in a lot of places already and which describes parameters and return values in a standardized format. Having done that, I also wanted to make sure I declared the correct types for the parameters, so that's why I added this to the typescript files. 

And the changes that were necessary in _http.js for that (except for the method comments) are rather small, just a three or four inline type declarations that's all. Added bonus: Adding types uncovered a couple of edge cases in handling the headers that we might never have noticed otherwise.

Long story short, it escalated from adding a few comments to adding types and having tsc check them.

**I can absolutely remove the declaration here if you prefer that.** In my opinion, having utility files like this type check is nice, but I do not hold that opinion strongly.

packages/core/src/tracing/instrumentation/protocols/_http.js
basti1302: Oh yes, absolutely. I agree the name is nonsense, a derelict from 2016. Since the module is only about capturing http headers, I'll rename it to `captureHttpHeadersUtil.js`.

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
basti1302: To be honest, I started the native fetch instrumentation by taking bits and pieces from the existing `httpClient` instrumentation, and that instrumentation also [falls back to the reduced span](https://github.com/instana/nodejs/blob/13c4ff3143c8ad6929d2b66414ba8420d75109a3/packages/core/src/tracing/instrumentation/protocols/httpClient.js#L150) if there is no current span.

This had been added there with [this commit](https://github.com/instana/nodejs/commit/55db775df23d0d28d342552ff7e8dfcb4051a3b6):

> feat(tracing): support tracing of deferred HTTP exits
> Also capture HTTP exits if the entry that triggered it has already
finished. 

Which means: It can happen that processing an HTTP entry has already finished when an outgoing HTTP request is made. I do not remember the concrete use case that prompted implementing it like this. I do remember that it was a real world use case from a customer, though, which got fixed by adding this fallback in `httpClient.js`.

tsconfig.json
kirrg001: 👍 Lets keep the ts types

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
basti1302: Discussed f2f, will add an additional comment that explains this a bit more.

PR #691, opened 2023-01-26
feat(collector): added support for sqs-consumer@6.2.0

kirrg001:
sqs-consumer@6.2.0 is using aws-sdk@3 under the hood.

FYI: When using the latest version of `@aws-sdk/client-sqs` to send a message to the receiver, it does not work and throws an error. Reported.

### DIFF
--- a/packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
+++ b/packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
@@ -17,7 +17,8 @@ const {
   expectAtLeastOneMatching,
   retry,
   delay,
-  stringifyItems
+  stringifyItems,
+  expectExactlyNMatching
 } = require('../../../../../../../core/test/test_util');
 const ProcessControls = require('../../../../../test_util/ProcessControls');
 const globalAgent = require('../../../../../globalAgent');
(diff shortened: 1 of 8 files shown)


packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/sqs.js
```diff
@@ -254,6 +257,13 @@ class InstanaAWSSQS extends InstanaAWSProduct {
               // _synchronously_ in the same event loop tick. See commit message for details.
               span.cancel();
             }
+
+            // NOTE: attach the async context to continue with it if sqs-consumer is used, see index.js
+            //       sqs-consumer messages are dependend on the customer's `handleMessage` function
+            data.Messages.forEach(message => {
```
kirrg001, line 263: https://github.com/bbc/sqs-consumer/blob/main/src/consumer.ts#L193

`receiveMessage` calls `this.sqs.send` where we can attach our async context. This is the only option.

https://github.com/bbc/sqs-consumer/blob/main/src/consumer.ts#L366

`executeHandler` calls the customer's `handleMessage` fn. This is the place where we need to wait for the result.

packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
```diff
@@ -188,10 +188,75 @@ function start(version) {
         });
       });
 
-      /**
-       * At the moment, SQS-Consumer does not support AWS SDK v3, although a PR exists in their repository:
-       * https://github.com/bbc/sqs-consumer/pull/252
-       */
+      // See https://github.com/bbc/sqs-consumer/issues/356
```
kirrg001, line 191: `@aws-sdk/client-sqs@3.53.0` does not work currently, reported.

basti1302 (approved):

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/sqs.js
kirrg001, line 263: check if data.messages is defined

kirrg001: works with batch? maybe check

kirrg001:
Added batch cases & support.

PR #702, opened 2023-02-10
feat(collector): nats 2.x

kirrg001:
refs 99434

- [x] headers in 2.x
- [x] test `.connect({callback: ... })`
- [x] cleanup


packages/collector/test/tracing/messaging/nats/test.js
```diff
@@ -6,320 +6,481 @@
...
+          }
+        });
+      });
 
-  describe('subscribe', function () {
-    [false, true].forEach(withError => {
-      testSubscribe.call(this, withError);
-    });
+      // NOTE: We do not track these errors yet, because there is no entry span
+      describe.skip('connect', () => {
```
kirrg001, line 415: @basti1302 Never noticed that we do not trace connection errors for instrumentations. Is that something we want to trace in the future?

basti1302: Hm, well, as you wrote: Usually this happens outside of active tracing, in the startup phase of the application, before it even starts to process incoming requests. The general rule "we don't trace an exit if there is no active entry span" still holds. So specifically for NATS I think we are fine without tracing this? 

If, on the other hand, we would instrument a driver that typically establishes a connection while processing an incoming request, we would probably want to instrument that as well.

basti1302 (approved):
LGTM! There is one thing that I think needs to be changed before merging (see comment)

packages/collector/test/tracing/messaging/nats/test.js
kirrg001: > The general rule "we don't trace an exit if there is no active entry span" still holds.

Yes. The question is: Is that something we want to trace in the future in general for all instrumentations (independent of the engine)?

packages/core/src/tracing/instrumentation/messaging/nats.js
```diff
@@ -19,39 +19,116 @@ exports.init = function init() {
...
+    const h = Headers();
+    h.append(constants.traceIdHeaderName, span.t);
+    h.append(constants.spanIdHeaderName, span.s);
+    opts.headers = h;
+  }
+}
+
+function instrumentedPublish(ctx, originalPublish, originalArgs, natsUrl, isLatest) {
   if (cls.skipExitTracing({ isActive })) {
+    addSuppressionHeaders(originalArgs, isLatest);
```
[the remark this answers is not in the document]
basti1302: The use case is skipping the tracing because the collector has not yet connected to the agent. The downstream service might also be instrumented and already connected to an agent, in this case the expected behavior is that the downstream service starts a trace with a root span. If we send X-INSTANA-L=0 downstream incorrectly, the downstream tracer would suppress instead of starting a trace.

packages/collector/test/tracing/messaging/nats/test.js
basti1302: AFAIK there are no plans to change this rule.

PR #704, opened 2023-02-16
feat(tracing): improve robustness of custom service naming

basti1302:
Previously, we only added span.data.service to _entry_ spans. The requirements around this have changed, now we add this annotation to _all_ spans, if the service name has been configured via the environment variable INSTANA_SERVICE_NAME or the in-code configuration.


kirrg001 (approved):

kirrg001:
> use INSTANA_SERVICE_NAME for all spans

Was wondering if you would like to rename the feature to reflect the benefit for the customer :)

PR #709, opened 2023-02-23
test: work around SIGTERM handling conflict between nyc and Prisma

basti1302:
Both nyc and Prisma do fancy things with signal handling. For that
reason, the Prisma test app ignores the SIGTERM signal when the tests
are run via nyc.

This change works around this issue by sending a SIGKILL one second
later, when SIGTERM does not terminate the child process (the
application under test).

### DIFF
--- a/packages/collector/test/test_util/ProcessControls.js
+++ b/packages/collector/test/test_util/ProcessControls.js
@@ -3,6 +3,8 @@
  * (c) Copyright Instana Inc. and contributors 2018
  */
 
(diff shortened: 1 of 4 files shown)


kirrg001 (approved):
LGTM
Do you want to put the GH issue reference to the big code comment?

basti1302:
> LGTM Do you want to put the GH issue reference to the big code comment?

Yes, good call, done.

PR #710, opened 2023-02-24
chore: switched to Node@18 (NPM 9)

kirrg001:
- nvm install 18.14.0
- nvm alias default 18.14.0
- nvm use default

From https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#lockfileversion
> 3: The lockfile version used by npm v7, without backwards compatibility affordances. This is used for the hidden lockfile at node_modules/.package-lock.json, and will likely be used in a future version of npm, once support for npm v6 is no longer relevant.

- updated some lock files to use version 3 already (apps)
- removed lockFileVersion: 1 in separate commit


package.json
```diff
@@ -50,7 +50,6 @@
     "test:debug": "lerna run test:debug --stream",
     "typecheck": "tsc",
     "update-deps": "lerna exec '../../node_modules/.bin/ncu --upgrade' && lerna exec 'npm i --package-lock-only' && npm run reinstall-deps",
-    "verify": "lerna run verify --stream",
```
basti1302, line 53: _question:_ What's up with this deletion?

basti1302 (approved):
Good stuff!

package.json
kirrg001: This was a mistake.

kirrg001:
FYI Node v10 build for msgpackr-extract was already failing on main
See e.g. https://app.circleci.com/pipelines/github/instana/nodejs/5003/workflows/2cd78bdf-8473-41a4-b53e-9f1a686dd479/jobs/18550

As this is an optional dependency, **it is fine that it fails** for Node v10. We 4.10.2 has dropped support for Node v10.

So why is the build now failing with Node v10?
Exit code 137 is memory issue again.

PR #715, opened 2023-02-28
feat(core): added otel integration (fs, restify, socket.io)

kirrg001:
refs 109122

- [x] finish up config - we keep a single config for now
- [x] check performance when fs is on
- [x] suppression
- [x] fix typescript
- [x] fix tests
- [x] QA
- [x] discuss @instana/utils package
- [x] check if dependabot creates a PR if the version is pinned fix
- [x] fix otel versions
- [x] create ticket to create a utility pacjage
- [x] Disable the new OTel integration code specifically on all versions below Node.js 14.
- [x] [**reported && waiting**] for Otel release because of warning of `fs.realpath.native`. They released 0.7.1.
- [x] Monkeypatch Otel ritm usage (really hard because they do not expose an object, lets ignore)
- [x] entry / exit span checking
- [x] [**reported && waiting**] wait for otel release + remove `useOpentelemetry: false` in require hook test when resolved.


packages/autoprofile/lib/samplers/async_sampler.js
```diff
@@ -7,7 +7,7 @@
 
 'use strict';
 
-const fs = require('fs');
+const { uninstrumentedFs: fs } = require('@instana/core');
```
kirrg001, line 10: TBD: Add a utility npm package `@instana/utils` to serve libs such uninstrumentedFs etc.

packages/autoprofile/precompile/build-all-addons.js
```diff
@@ -8,7 +8,6 @@
 
 const os = require('os');
 const path = require('path');
-const { copyFileSync, existsSync, mkdirSync } = require('fs');
```
basti1302, line 11: Apparently we are not running eslint on this folder. I think we should, so that we catch stuff like this directly -> https://github.com/instana/nodejs/pull/723

packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/receiveMessage.js
```diff
@@ -5,7 +5,12 @@
 
 'use strict';
 
-const instana = require('../../../../../..')();
+const instana = require('../../../../../..')({
+  tracing: {
+    // aws sdk produces some fs calls to read the config from disk on instantiation which we ignore.
+    useOpentelemetry: false
```
basti1302, line 11: _question:_ do we stil need to exclude this explicitly via config? IIRC you found a solution for not capturing fs calls outside of an ongoing tracing and the AWS SDK instantiation should happen outside of any request processing, wouldn't it?

packages/collector/test/tracing/database/db2/test.js
```diff
@@ -852,7 +852,10 @@ mochaSuiteFn('tracing/db2', function () {
         .then(() =>
           testUtils.retry(() =>
             verifySpans(agentControls, controls, {
-              spanLength: 11,
+              // 11 queries splitted from the file
+              // 5 fs operations on top (ours + from db2 internally fs-extra)
+              // https://github.com/ibmdb/node-ibm_db/blob/fb25937524d74d25917e9aa67fb4737971317986/lib/odbc.js#L916
+              spanLength: 15,
```
basti1302, line 858: _nitpick:_ `numberOfSpans` might be a better name.

packages/core/src/tracing/opentelemetry-instrumentations/index.js
```diff
@@ -0,0 +1,11 @@
+/*
+ * (c) Copyright IBM Corp. 2023
+ */
+
+'use strict';
+
+module.exports = {
+  get otelInstrumentations() {
```
basti1302, line 8: _question:_ OTel has different policies around support for older Node.js versions than we do.

https://github.com/open-telemetry/opentelemetry-js/tree/main#supported-runtimes

> Only Node.js Active or Maintenance LTS versions are supported. Previous versions of node may work, but they are not tested by OpenTelemetry and they are not guaranteed to work.

I think it might be a good idea to avoid the risk of OTel modules causing issues for customers on older Node.js versions by disabling OTel integration on our end for older versions. (probably everything below 14?)

Not sure where we would want to do that, either here or we could automatically set the config flag for OTel integration to off on older versions, or put the condition into `packages/core/src/tracing/index.js`.

packages/core/src/tracing/opentelemetry-instrumentations/restify.js
```diff
@@ -0,0 +1,27 @@
+/*
+ * (c) Copyright IBM Corp. 2023
+ */
+
+'use strict';
+
+const constants = require('../constants');
+
+module.exports.init = () => {
+  const { RestifyInstrumentation } = require('@opentelemetry/instrumentation-restify');
```
basti1302, line 10: _question:_ The test is disabled for Node.js 18, because the instrumentation does not work correctly there. Should we also skip loading the instrumentation on Node.js 18 then?

basti1302 (approved):
Great work! 🎉 🎉

packages/autoprofile/lib/samplers/async_sampler.js
basti1302: I don't have a strong opinion, I think. All the packages that would use that also either rely on `core` directly or indirectly, or are always installed together with `core`. (`autoprofiler` is never used directly, but only as a dependency of `collector`, which in turn requires `core`). But I agree, adding `core` just for this feels weird.

Let's discuss pros and cons in person at some point.

packages/core/src/tracing/opentelemetry-instrumentations/fs.js
```diff
@@ -0,0 +1,47 @@
...
+  // https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1335/files#diff-9a2f445c78d964623d07987299501cbc3101cbe0f76f9e18d2d75787601539daR428
+  // As we mix Instana & Otel instrumentations, we need to check here if
+  // we have an Instana parentSpan otherwise we loose spans.
+  // e.g. http server call is instana span && fs call exit call is otel span
+  const orig = api.trace.getSpan;
+  api.trace.getSpan = function instanaGetSpan() {
+    const parentSpan = cls.getCurrentSpan();
+
+    // CASE: we only want to trace fs calls on entry spans
+    if (!parentSpan || constants.isExitSpan()) {
```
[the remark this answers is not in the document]
kirrg001: Good catch!

packages/collector/test/tracing/database/db2/test.js
kirrg001: Will rename.

packages/core/src/tracing/opentelemetry-instrumentations/index.js
kirrg001: If the node version is unsupported, we set `automaticTracingEnabled` to false in `normalizeConfig`.
And we only enable Otel if `automaticTracingEnabled` is true. So I guess we are safe here?

packages/autoprofile/lib/samplers/async_sampler.js
kirrg001: So do we want to keep this as it is and discuss+create a ticket to add a utility package?

packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/receiveMessage.js
kirrg001: I was able to remove it for dynamodb tests, but not for sqs. Added a comment why.

packages/autoprofile/lib/samplers/async_sampler.js
basti1302: Yes, good idea, I think that would be best. We can merge this as-is and refactor later. This would also help keep the scope of this PR smaller. Or are you strongly leaning towards doing this in the scope of this PR?

packages/core/src/tracing/opentelemetry-instrumentations/index.js
basti1302: Hm, not sure that is exactly what I mean. We set have this as our expression for supported versions: `semver.satisfies(version, '^10.4.0 || ^11 || >=12.0.0');` So we basically support Node.js 10.x and above.

The OTel project on the other hand only test their code against Node.js 14 and above on their CI. 

I was thinking we might want to explicitly disable this new OTel integration for Node.js 10 and 12, and only enable it for Node.js 14 and above. Customers on Node 10 and 12 can still use our "own" auto-tracing instrumentations (which we test extensively on Node 10 and 12 on CI), but we do not load the OTel code automatically on these old Node.js versions.

The reason is mainly that by doing that we would have the stronger guarantees from the OTel test suite/CI apply to this new integration as well. 

Of course, you added a lot of tests for the new OTel integration to our test suite as well. You have a point if you say that this makes it safe to run the OTel code as part of `@instana/collector` on Node.js 10 and 12 as well.

But: At the moment we depend on the OTel-related libraries with a version range -- `"@opentelemetry/api": "^1.4.0"` etc., all are referred to with a caret (`^`) in the version.

So one possible scenario would be this: Currently everything works on Node 10, and the version of let say `@opentelemetry/instrumentation-fs` we have in our `package-lock.json` (`0.7.1`) is compatible with Node.js 10. Then `@opentelemetry/instrumentation-fs` introduces some code that breaks on Node.js 10 and releases a new version 0.8.0. Their package.json file says `engine: >= 14`, so it wouldn't even be a breaking change from their perspective. Plus, they don't run tests on Node.js 10, so they wouldn't notice. Our tests would also not go red immediately because we still have version 0.7.1 in our package-lock.json. But someone doing a fresh install of `@instana/collector` would receive `@opentelemetry/instrumentation-fs@0.8.0` and it would break their app. 

So at a minimum I think we need to do either of those two thing (or both):
* Reference all the new packages via an exact version, not a version range.
* Disable the new OTel integration code specifically on all versions below Node.js 14.

packages/autoprofile/lib/samplers/async_sampler.js
kirrg001: 👍

packages/core/src/tracing/opentelemetry-instrumentations/index.js
kirrg001: > Disable the new OTel integration code specifically on all versions below Node.js 14.

Great idea 👍 Let's do.

kirrg001:
Ok this PR stays on hold for now. We are waiting for the otel merge.

kirrg001:
Updated Otel dependencies and removed the `useOpentelemetry: false` checks. Working now 🥳

packages/core/src/tracing/opentelemetry-instrumentations/wrap.js
```diff
@@ -0,0 +1,88 @@
...
+      }
+
+      if (targetInstrumentationModule.transform) {
+        otelSpan = targetInstrumentationModule.transform(otelSpan);
+      }
+    }
+
+    if ((kind === constants.EXIT && cls.skipExitTracing()) || (kind === constants.ENTRY && cls.tracingSuppressed())) {
+      return;
+    }
```
kirrg001, line 58: @basti1302 Would love to get your input and help here.

As discussed in the tracer meeting. It was asked about entry and exit span conditions.
I have added this line to ensure we respect the rules.

I am wondering if we need anything else here, but I cannot think of any other check 🤔 
For example the http server instrumentation creates an entry span and does not care wether there is already an existing entry span or not.

FYI we create entry spans only for Socket.io ON events.

kirrg001: Discussed offline. See commits.

PR #717, opened 2023-03-02
fix(grpc-js): capture the correct destination host per grpc-js client

basti1302:
This fixes an issue where we would record the wrong destination
host/port when multiple @grpc/grpc-js clients are used by the
application.

Plus: fix(grpc-js): do not mark cancelled calls as erroneous

### DIFF
--- a/packages/core/src/tracing/instrumentation/protocols/grpcJs.js
+++ b/packages/core/src/tracing/instrumentation/protocols/grpcJs.js
@@ -114,8 +114,9 @@ function modifyArgs(name, originalArgs, span) {
 
       if (err) {
         const errorMessage = err.details || err.message;
-
-        if (errorMessage !== 'Cancelled') {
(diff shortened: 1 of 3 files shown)


packages/core/src/tracing/instrumentation/protocols/grpcJs.js
```diff
@@ -112,8 +114,9 @@ function modifyArgs(name, originalArgs, span) {
 
       if (err) {
         const errorMessage = err.details || err.message;
-
-        if (errorMessage !== 'Cancelled') {
+        if (typeof errorMessage === 'string' && errorMessage.toLowerCase().includes('cancelled')) {
```
basti1302, line 117: I noticed that the tests had all the code in place to test cancelled calls (because they are to some degree a copy of the old grpc tests), but there were actually no tests for cancelling. When adding them, I found out that we need to tweak the code a bit to correctly detect cancelled calls.

kirrg001 (approved):
LGTMÄ

PR #722, opened 2023-03-08
chore: experimental shimmer wrapper

kirrg001:
refs 117518

- [x] run legacy & esm build
- [x] add unit test

### DIFF
--- a/packages/core/src/tracing/shimmer.js
+++ b/packages/core/src/tracing/shimmer.js
@@ -17,7 +17,7 @@ exports.unwrap = shimmer.unwrap;
  * This wrapper pretends that the application of the customer dies.
  *
  * We won't catch errors which happened e.g. when the result of the original fn comes back.
- * This is fine, because we want to primarly catch sync solution.
+ * This is fine, because we want to primarly catch errors which happened before the library call.
  */
 exports.wrap = (origObject, origMethod, instrumentationWrapperMethod) => {
   shimmer.wrap(origObject, origMethod, function shimmerWrapper(originalFunction) {
@@ -46,10 +46,10 @@ exports.wrap = (origObject, origMethod, instrumentationWrapperMethod) => {
           throw err;
         }
 
+        logger.warn(`An internal error happend in the Instana Node.js collector. Please contact support. ${err.stack}`);
+
         if (originalCalled) return;
         originalCalled = true;
-
-        logger.warn(`Catched error from instrumentation: ${err.message}, ${err.stack}`);
 
         // CASE: we pretend that there was no error in our instrumentation, we only log a warning and
         //       ensure tht the customer's application does not die.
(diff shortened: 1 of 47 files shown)


packages/core/src/tracing/shimmer.js
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * This wrapper prevents that the application of the customer dies.
+ *
+ * We won't catch errors which happened e.g. when the result of the original fn comes back.
+ * This is fine, because we want to primarly catch errors which happened before the library call.
+ */
+exports.wrap = (origObject, origMethod, instrumentationWrapperMethod) => {
+  shimmer.wrap(origObject, origMethod, function shimmerWrapper(originalFunction) {
+    return function () {
```
basti1302, line 24: _nitpick:_ Do we want this to be an anonymous function or should we make it a named function as well?

packages/core/src/tracing/shimmer.js
```diff
@@ -0,0 +1,60 @@
...
+});
+
+exports.unwrap = shimmer.unwrap;
+
+/**
+ * This wrapper prevents that the application of the customer dies.
+ *
+ * We won't catch errors which happened e.g. when the result of the original fn comes back.
+ * This is fine, because we want to primarly catch errors which happened before the library call.
+ */
```
basti1302, line 21: _comment:_ While this wrapper is really super useful, it is also kind of hard to wrap your head around the order in which things are called. WDYT about adding a comment explaining how the flow of control looks?

Something like this maybe?

----

Assuming the following flow for an uninstrumented application:

_application code_ -> _original library function_

After applying instrumentation with the help of this wrapper, the flow will looks like this:

_application code_ -> _(anonymous function in line 24 which we might want to give a name, see below)_ -> `instrumentationWrapperMethod`  (e.g. `shimQuery` from our `mysql` instrumentation) -> `instanaOriginalFunctionWrapper` -> _original library function_.

This serves two purposes:
1. We can insert one central try-catch into the flow that is applied for all our instrumentations.
2. The additional named function `instanaOriginalFunctionWrapper` enables us to inspect the stack trace of errors that are thrown and thus determine whether they have been thrown in our instrumentation code or in the instrumented library.

----

I think this would make it easier to understand this wrapper.

packages/core/src/tracing/shimmer.js
kirrg001: Yes we want named fn's everywhere <3 Good catch!

packages/core/src/tracing/shimmer.js
```diff
@@ -0,0 +1,60 @@
...
+exports.wrap = (origObject, origMethod, instrumentationWrapperMethod) => {
+  shimmer.wrap(origObject, origMethod, function shimmerWrapper(originalFunction) {
+    return function () {
+      let originalCalled = false;
+      const originalThis = this;
+      const originalArgs = arguments;
+
+      function instanaOriginalFunctionWrapper() {
+        if (originalCalled) return;
```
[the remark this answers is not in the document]
kirrg001: I guess I kept it for super safety. But I think you're right, yes. There is no real use case right now. I think I would like to keep it with a comment.

packages/core/src/tracing/shimmer.js
kirrg001: Yes sure. I will add this

PR #724, opened 2023-03-09
test(aws): do not load @aws-sdk/client-dynamodb on old Node.js versions

basti1302:
The tests are loading the package @aws-sdk/client-dynamodb directly, to check whether the required tables exist and clean up in the after hook. Version 3.201.0 of @aws-sdk/client-dynamodb dropped support for Node.js
12. Thus, the latest upgrade of that package broke the test suite on Node.js 10 and 12 because we tried to require code with syntax only supported on Node.js >= 14.

Fixed by not loading that package via require on older Node.js versions and also skipping the associated test cases.

### DIFF
--- a/packages/collector/test/tracing/cloud/aws-sdk/v3/dynamodb/test_definition.js
+++ b/packages/collector/test/tracing/cloud/aws-sdk/v3/dynamodb/test_definition.js
@@ -4,8 +4,7 @@
 
 'use strict';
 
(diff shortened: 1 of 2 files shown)


basti1302:
Build looks better now: https://app.circleci.com/pipelines/github/instana/nodejs/5124/workflows/5479b03b-b1d9-42df-99b3-fd5a2b09eef2

kirrg001 (approved):

PR #739, opened 2023-03-24
feat(collector): added support for amqplib 0.10.x

kirrg001:
They changed consumers from object to map:
https://github.com/amqp-node/amqplib/commit/d483493e6825ed7d770e68f4cec60979d0feeae7


basti1302 (approved):
LGTM

packages/collector/test/tracing/messaging/amqp/test.js
```diff
@@ -24,198 +24,202 @@ let consumerControls;
...
-  consumerControls = require('./consumerControls');
+    publisherControls = require('./publisherControls');
+    consumerControls = require('./consumerControls');
 
-  ['Promises', 'Callbacks'].forEach(apiType => {
-    describe(apiType, function () {
-      registerTests.call(this, apiType);
+    ['Promises', 'Callbacks'].forEach(apiType => {
+      describe(apiType, function () {
+        registerTests.call(this, apiType, version);
```
basti1302: @kirrg001  I removed the `version` parameter here in the latest commit, there is no `versions` parameter in the function definition of `registerTests`. Both the call and the function definition are in the scope of the outermost `forEach` so there is no need to pass it along as a parameter.

PR #740, opened 2023-03-24
chore: improve announce cycle log messages

basti1302:


packages/collector/src/announceCycle/agentready.js
```diff
@@ -146,9 +146,15 @@ function sendTracingMetrics() {
   const payload = tracing._getAndResetTracingMetrics();
   agentConnection.sendTracingMetricsToAgent(payload, error => {
     if (error) {
-      logger.warn('Error received while trying to send tracing metrics to agent: %s', error.message);
+      logger.info(
```
basti1302, line 149: I downgraded this from warning to info. Tracemetrics are for our internal self-monitoring and not a direct concern for users of `@instana/collector`.

packages/collector/src/announceCycle/unannounced.js
```diff
@@ -86,19 +92,26 @@ function tryToAnnounce(ctx, retryDelay = initialRetryDelay) {
     try {
       agentResponse = JSON.parse(rawResponse);
     } catch (e) {
-      logger.warn(
-        'Failed to JSON.parse agent response. Response was %s. Will retry in %s ms',
-        rawResponse,
+      logger.error(
```
basti1302, line 95: Changed this from warn to error. This should really not happen and it would more often than not be something that will not self-heal when retrying.

packages/collector/src/announceCycle/unannounced.js
```diff
@@ -86,19 +92,26 @@ function tryToAnnounce(ctx, retryDelay = initialRetryDelay) {
...
-    logger.info('Overwriting pid for reporting purposes to: %s', pid);
-    pidStore.pid = pid;
+    if (pidStore.pid !== agentResponse.pid) {
+      logger.info(
+        'Reporting data to the Instana host agent with the PID from the root namespace (%s) instead of the ' +
+          'in-container PID (%s).',
+        agentResponse.pid,
+        pidStore.pid
+      );
+      pidStore.pid = agentResponse.pid;
```
basti1302, line 113: This could be considered a change in behavior, but in practical terms it should not. We now only overwrite the PID when the PID returned by the agent is actually different. (And also log the fact only in that case.)

packages/collector/src/announceCycle/unannounced.js
```diff
@@ -77,7 +77,13 @@ function tryToAnnounce(ctx, retryDelay = initialRetryDelay) {
   }
   agentConnection.announceNodeCollector((err, rawResponse) => {
     if (err) {
-      logger.debug('Announce attempt failed: %s. Will retry in %s ms', err.message, retryDelay);
+      logger.info(
```
basti1302, line 80: I reverted this back to info as discussed. The additional explanation should hopefully make it a bit clearer that this does not need to be reported.

kirrg001 (approved):

PR #751, opened 2023-04-17
test: added integration test otel & instana

kirrg001:
- show how otel & instana behave when they are both used

Regarding [Hapi](https://github.com/instana/nodejs/pull/751/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R77):

```
root "@hapi/hapi": "21.x.x", (21 has it's own types file)

  submodule
    @opentelemetry/instrumentation-hapi
      "@hapi/hapi": "20.1.5", 
      "@types/hapi__hapi": "20.0.9" (this will be installed on the root and then is not compatible with 21.x and throws types errors)
```

Installing "@types/hapi__hapi": "20.13" latest works as a workaround.


packages/collector/test/opentelemetry/app.js
```diff
@@ -0,0 +1,84 @@
...
+
+  const sdk = new opentelemetry.NodeSDK({
+    resource: new Resource({
+      [SemanticResourceAttributes.SERVICE_NAME]: 'my-service'
+    }),
+    spanProcessor: spanProcessor,
+    instrumentations: [nodeAutoInstrumentations]
+  });
+
+  sdk.start();
```
kirrg001, line 37: FYI: This is already the new SDK version with sync mechanism.

packages/collector/test/opentelemetry/test.js
```diff
@@ -0,0 +1,126 @@
...
+      otelSpans = [];
+    });
+
+    it('should not trace Instana spans', () =>
+      controls
+        .sendRequest({
+          method: 'GET',
+          path: '/trace'
+        })
+        .then(() =>
```
kirrg001, line 69: Add a safety delay.

packages/collector/test/opentelemetry/test.js
```diff
@@ -0,0 +1,126 @@
...
+const { retry } = require('../../../core/test/test_util');
+const ProcessControls = require('../test_util/ProcessControls');
+const portfinder = require('../test_util/portfinder');
+const globalAgent = require('../globalAgent');
+
+// NOTE: only run on the latest node version
+const mochaSuiteFn =
+  supportedVersion(process.versions.node) && semver.gte(process.versions.node, '18.0.0') ? describe : describe.skip;
+
+mochaSuiteFn('Opentelemetry usage', function () {
```
basti1302, line 23: _suggestion:_ I think it might be nice to add a comment that describes the intent of the test. Something like this maybe?

> NOTE: Using @instana/collector and the OpenTelemetry SDK in the same process is *not supported*. Thus, this test does not verify desirable behavior but simply checks what exactly happens when this unsupported setup is used. Both variants (require Instana first/OTel second and vice versa) fail, though they fail in different ways.

basti1302 (approved):

PR #756, opened 2023-04-19
ci: moved Node v16 to legacy and removed 16-esm

kirrg001:
- save resources

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ shared: &shared
   steps:
     - checkout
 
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -561,6 +537,7 @@ workflows:
         # by providing the parameter "workflow: legacy-nodejs-versions"
         - equal: [ legacy-nodejs-versions, << pipeline.parameters.workflow >> ]
     jobs:
+      - "node-16"
```
kirrg001, line 540: TBD: rename legacy-nodejs-versions to other-nodejs-versions

basti1302 (approved):

PR #759, opened 2023-04-20
feat(opentelemetry-exporter): added support for INSTANA_DEBUG env variable

kirrg001:


packages/opentelemetry-exporter/src/index.js
```diff
@@ -70,6 +70,13 @@ class InstanaExporter {
    */
   constructor({ agentKey, endpointUrl } = {}) {
     this._isShutdown = true;
+
+    // https://github.com/open-telemetry/opentelemetry-js/pull/3627
+    // NOTE: We are not using the Instana logger. We are using Otel API diag component.
+    if (process.env.INSTANA_DEBUG) {
+      process.env.OTEL_LOG_LEVEL = 'debug';
+    }
```
basti1302, line 78: _question:_ Should we also check INSTANA_LOG_LEVEL and copy that to OTEL_LOG_LEVEL? For our other components, INSTANA_DEBUG=true is usually an alias for INSTANA_LOG_LEVEL=debug.

basti1302 (approved):
Nice!

packages/opentelemetry-exporter/src/index.js
kirrg001: Sure 👍

PR #766, opened 2023-04-25
test(mysql2): skip tests on Node.js < 14

basti1302:
Despite not being advertised as a breaking change in their changelog, node-mysql2 dropped support for EOL releases by updating the lru-cache dependency to a version that requires Node.js 14.


basti1302:
New build (with fix): https://app.circleci.com/pipelines/github/instana/nodejs/5518/workflows/f6584167-4a41-4f5f-b065-5862dc3f6d22

PR #767, opened 2023-04-25
test: added SDK multiple installations case

kirrg001:
Proofs how to use the SDK with multiple installations for CJS & ESM.
Docs updates coming soon.

### DIFF
--- a/packages/collector/src/index.js
+++ b/packages/collector/src/index.js
@@ -60,7 +60,6 @@ function init(_config) {
         cacheKey => cacheKey.indexOf('collector/src/index.js') >= 0
       );
     }
-
     if (collectorIndexCacheKey) {
       return require.cache[collectorIndexCacheKey].exports;
     } else {
--- a/packages/collector/src/index.js
+++ b/packages/collector/src/index.js
@@ -60,6 +60,7 @@ function init(_config) {
         cacheKey => cacheKey.indexOf('collector/src/index.js') >= 0
       );
     }
+
     if (collectorIndexCacheKey) {
       return require.cache[collectorIndexCacheKey].exports;
     } else {
--- a/packages/collector/test/tracing/sdk/multiple_installations/cjs/src/app.js
+++ b/packages/collector/test/tracing/sdk/multiple_installations/cjs/src/app.js
@@ -26,8 +26,12 @@ const logPrefix = `CJS SDK multiple installations: (${process.pid}):\t`;
 async function createSDKSpans() {
   // NOTE: We need the delay here to ensure that the collector is fully initialized.
   //       Otherwise we will get NoopSpanHandle instances because tracing is not ready yet.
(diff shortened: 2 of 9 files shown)


.circleci/config.yml
```diff
@@ -379,6 +379,8 @@ memcached: &memcached
 
 jobs:
   "node-18":
+    environment:
+      COLLECTOR_ONLY: true
```
basti1302, line 383: Ah, so that's how the grown ups set this flag. Makes me feel stupid for having set that via https://app.circleci.com/settings/project/github/instana/nodejs/environment-variables?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2Finstana%2Fnodejs a while ago, which of course then affects all builds 🤦 

I still think this should be removed before merging ;-)

packages/collector/test/apps/agentStub.js
```diff
@@ -139,10 +139,14 @@ app.post(
 app.post(
   '/com.instana.plugin.nodejs/traces.:pid',
   checkExistenceOfKnownPid(function handleTraces(req, res) {
+    // eslint-disable-next-line no-console
+    console.log('received spans in agent', req.body);
```
basti1302, line 143: ~~This will be quite noisy. I'm -1 on adding that permanently, for the whole test suite. Can we remove this before merging?~~ (edit: after reading the whole PR, I guess you are still planning to clean up the console.logs from whole PR anyway)

packages/collector/test/apps/agentStub.js
```diff
@@ -139,10 +139,14 @@ app.post(
...
     if (rejectTraces) {
       return res.sendStatus(400);
     }
     if (!dropAllData) {
+      // eslint-disable-next-line no-console
+      console.log('push data to receive');
```
basti1302, line 149: ~~Similar thing here, -1 on this.~~ nevermind, guess this will be removed later as well

packages/core/src/tracing/cls.js
```diff
@@ -160,7 +160,11 @@ class InstanaSpan {
   }
 
   transmit() {
+    // eslint-disable-next-line no-console
+    console.log('transmit', this, this.transmitted, this.manualEndMode);
```
basti1302, line 164: Okay, I see now. The "review-and-merge" label lead me to believe that you planned to merge this PR as-is but there are still a couple more CI-troubleshooting console.logs in here, also in sdk.js, so looks like I was mistaken about that. I'll stop commenting on those :-)

basti1302 (approved):
LGTM, except for the spurious console.logs which are still in there. 👍

packages/core/src/tracing/cls.js
kirrg001: Yes correct :)

PR #771, opened 2023-04-27
fix(collector): keep EOL events open instead of recreating them

basti1302:
Issue events that are sent to the Instana agent (and thereby indirectly to the Instana backend), are discarded when they expire according to their "duration" property. To keep them open for the entire lifetime of the entity (which makes sense for the EOL events we are dealing with here), we are ought to resend them before they expire in a heartbeat- like fashion. The backend requires a unique "path" attributes on these events to be able to group them together and to know that it is supposed to keep the already created issue open instead of opening a new issue. This commit adds that "path" attribute.

Also:
- change the duration of the event from 10 minutes to 6 hours
- rename the misnamed AgentConnectionEvent type to just Event

### DIFF
--- a/packages/collector/src/announceCycle/agentready.js
+++ b/packages/collector/src/announceCycle/agentready.js
@@ -24,8 +24,9 @@ const transmissionCycle = require('../metrics/transmissionCycle');
 const uncaught = require('../uncaught');
 const { isNodeVersionEOL } = require('../util/eol');
 
-const ELEVEN_MINUTES = 11 * 60 * 1000;
-const TEN_MINUTES = 10 * 60 * 1000;
+const ONE_MINUTE = 60 * 1000;
+const EOL_EVENT_REFRESH_INTERVAL = 6 * 60 * ONE_MINUTE; // 6 hours
+const EOL_EVENT_DURATION = 6 * 60 * ONE_MINUTE + ONE_MINUTE; // 6 hours + 1 minute
 
 /** @type {*} */
 let autoprofile;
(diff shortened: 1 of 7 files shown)


packages/collector/src/announceCycle/agentready.js
```diff
@@ -211,13 +214,13 @@ function sendEOLEvent() {
 /**
  * Sends an issue event to the agent when the Node.js version has reached end of life.
  * It will work for non serverless environments where an agent is present.
- * Also, currently the serverless-acceptor does not have support for events.
+ * (At the time of writing, serverless-acceptor does not have support for events.)
```
kirrg001, line 217: Suggestion: This sounds a bit like the support for serverless environments would be added here.
How about? 

"At the time of writing, serverless-acceptor does not have support for events. See packages/serverless/src/backend_connector.js"

kirrg001:
> fix(eol): keep EOL events open instead of recreating them

Suggestion:

"fix(collector): keep EOL events open instead of recreating them"

packages/collector/src/announceCycle/agentready.js
basti1302: Not sure I follow. This is talking about the Instana **backend** service `serverless-acceptor`. Support for ingesting events would need to be added there before we can add it in any of our modules. I'll try to make that more clear.

kirrg001: Yes.
packages/serverless/src/backend_connector.js === backend service serverless-acceptor

basti1302: No. 

https://github.ibm.com/instana/backend/blob/develop/serverless-acceptor/src/main/java/com/instana/serverless/ServerlessAcceptorApp.java ===backend service serverless-acceptor

Anyway, I've changed the comment, so this should hopefully be a bit more clear now.

basti1302:
For the record, I need to keep this open for a while to make some corresponding changes in the backend as well. I want to make the backend changes before merging this.

PR #777, opened 2023-05-02
feat(w3c): support W3C trace context level 2

basti1302:
- set the flag that the trace ID has been generated randomly when we generated id
- pass on the incoming flag for the randomness of the trace ID unchanged when a traceparent header is received on incoming HTTP requests

See also: https://www.w3.org/TR/trace-context-2/, in particular section 3.2.2.5.2 "Random Trace ID Flag".

### DIFF
--- a/packages/collector/test/tracing/misc/specification_compliance/tracer_compliance_test_cases.json
+++ b/packages/collector/test/tracing/misc/specification_compliance/tracer_compliance_test_cases.json
@@ -11,7 +11,7 @@
     "X-INSTANA-T out": "$new_64_bit_trace_id",
     "X-INSTANA-S out": "$new_span_id_2",
     "X-INSTANA-L out": "1",
-    "traceparent out": "00-0000000000000000$new_64_bit_trace_id-$new_span_id_2-01",
+    "traceparent out": "00-0000000000000000$new_64_bit_trace_id-$new_span_id_2-03",
     "tracestate out": "in=$new_64_bit_trace_id;$new_span_id_2",
     "index": 1
   },
@@ -30,7 +30,7 @@
     "X-INSTANA-T out": "fa2375d711a4ca0f",
     "X-INSTANA-S out": "$new_span_id_2",
     "X-INSTANA-L out": "1",
-    "traceparent out": "00-0000000000000000fa2375d711a4ca0f-$new_span_id_2-01",
+    "traceparent out": "00-0000000000000000fa2375d711a4ca0f-$new_span_id_2-03",
     "tracestate out": "in=fa2375d711a4ca0f;$new_span_id_2",
     "index": 2
   },
@@ -50,7 +50,7 @@
     "X-INSTANA-T out": "3483a5fe1cd42e30",
     "X-INSTANA-S out": "$new_span_id_2",
     "X-INSTANA-L out": "1",
-    "traceparent out": "00-0373e5f7aca1cdef3483a5fe1cd42e30-$new_span_id_2-01",
+    "traceparent out": "00-0373e5f7aca1cdef3483a5fe1cd42e30-$new_span_id_2-03",
     "tracestate out": "in=3483a5fe1cd42e30;$new_span_id_2",
     "index": 3
   },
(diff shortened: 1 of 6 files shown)


packages/collector/test/tracing/misc/specification_compliance/tracer_compliance_test_cases.json
```diff
@@ -730,8 +730,48 @@
...
+    "exitSpan.t": "fa2375d711a4ca0f",
+    "exitSpan.p": "$new_span_id_1",
+    "exitSpan.s": "$new_span_id_2",
+    "exitSpan.data.service": "Tracer Test Suite Service",
+    "X-INSTANA-T out": "fa2375d711a4ca0f",
+    "X-INSTANA-S out": "$new_span_id_2",
+    "X-INSTANA-L out": "1",
+    "traceparent out": "00-0000000000000000fa2375d711a4ca0f-$new_span_id_2-03",
+    "tracestate out": "in=fa2375d711a4ca0f;$new_span_id_2",
+    "index": 42
```
basti1302, line 775: This two new test cases are unrelated to the change at hand, they are the result of copying over a fresh copy of the tracer test suite test case definition.

packages/core/src/tracing/w3c_trace_context/W3cTraceContext.js
```diff
@@ -6,11 +6,13 @@
...
 
 const constants = require('../constants');
+const leftPad = require('../leftPad');
 const tracingUtil = require('../tracingUtil');
 
 const VERSION00 = '00';
 const LEFT_PAD_16 = '0000000000000000';
-const SAMPLED_BITMASK = 0x00000001;
+const SAMPLED_BITMASK = 0b1;
+const RANDOM_TRACE_ID_BITMASK = 0b10;
```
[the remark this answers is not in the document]
kirrg001: Suggestion: It is not trivial to read the shortcuts 0b1 or 0b10. How about adding the [link](https://www.w3.org/TR/trace-context-2/#trace-flags) here? The link describes the rest.

packages/core/src/tracing/w3c_trace_context/W3cTraceContext.js
```diff
@@ -51,6 +55,8 @@ W3cTraceContext.fromInstanaIds = function fromInstanaIds(instanaTraceId, instana
   traceContext.traceParentTraceId = paddedTraceId;
   traceContext.traceParentParentId = instanaParentId;
   traceContext.sampled = sampled;
+  // Instana tracers always generate trace IDs randomly, so we can confidently set this flag.
+  traceContext.randomTraceId = true;
```
kirrg001, line 59: Suggestion: As this is a true or false state, I would like to suggest to call it `isRandomTraceId`.

kirrg001 (approved):
Great 👍 

Suggestion for commit:

`feat(collector): added support for w3c trace context level 2 #777`

packages/core/src/tracing/w3c_trace_context/W3cTraceContext.js
basti1302: I'm -1 on this. `sampled` is also a boolean. The names for the flags are taken directly from the specification. I think this makes sense -- you probably need to read the spec anyway to know what this code is about and if you have read it, using the same name helps making the connection between the code and the concepts discussed in the spec easier.

basti1302:
> Great 👍
> 
> Suggestion for commit:
> 
> `feat(collector): added support for w3c trace context level 2 #777`

It is not only a change for the collector though. Our AWS Lambda, Fargate, Google Cloud Run packages are affected as well.

packages/core/src/tracing/w3c_trace_context/W3cTraceContext.js
kirrg001: > The names for the flags are taken directly from the specification.

The name in the spec is "random" :) 

All good, was not important 👍

basti1302: > The name in the spec is "random" :)

Ha, right. Depends on which part of the spec you look at. There is a code example in "3.2.2.5 trace-flags" that refers to the flag simply as `random`. Section [3.2.2.5.2 Random Trace ID Flag](https://www.w3.org/TR/trace-context-2/#random-trace-id-flag) calls it "random-trace-id flag". That's what I was referring to. And yes, I _always_ need to have the last word in arguments, lolol.

PR #778, opened 2023-05-02
ci: reduce parallel executors for nightly job and make them configurable

basti1302:
The intent is to save credits on nightly jobs by not starting all containers 20 times in parallel. By having them configurable we still can get fast build times when running the job on demand, by providing the "executors" pipeline parameter.


.circleci/config.yml
```diff
@@ -397,12 +400,12 @@ jobs:
       - <<: *couchbase
       - <<: *couchbase-setup
     <<: *shared
-    parallelism: 20
-    resource_class: large
+    parallelism: 15
```
kirrg001, line 403: As discussed, we would like to run the main build a couple of times to check the time difference.
We either revert to 20 or keep 15 if the time difference is small.

kirrg001 (approved):

PR #780, opened 2023-05-08
fix(announce): remove obsolete check for Server header

basti1302:
This check actually breaks the announce procedure if a service mesh with proxies is present, as the proxy might very well be configured to replace the Instana agent's Server header with its own. The check is superfluous anyway. Instead, the collector now checks whether the HTTP status code is in the 2xx range.

### DIFF
--- a/packages/collector/src/announceCycle/agentHostLookup.js
+++ b/packages/collector/src/announceCycle/agentHostLookup.js
@@ -27,7 +27,6 @@ logger = require('../logger').getLogger('announceCycle/agentHostLookup', newLogg
 // The host differs, when the collector is running inside a Docker container and the
 // agent is running on the host.
 
-const expectedServerHeader = 'Instana Agent';
 const requestTimeout = 5000;
 const retryTimeoutMillis = 60 * 1000;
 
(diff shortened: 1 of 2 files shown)


packages/collector/src/announceCycle/agentHostLookup.js
```diff
@@ -125,14 +124,13 @@ function checkHost(host, cb) {
         timeout: requestTimeout
       },
       res => {
-        if (res.headers.server === expectedServerHeader) {
+        if (res.statusCode >= 200 && res.statusCode < 300) {
```
kirrg001, line 127: Question: Will you add proper 404 error handling in our agent announce cycle (see fn `announceNodeCollector`) as part of this PR or as part of another PR?

basti1302: What kind of error handling are you looking for? [Currently](https://github.com/instana/nodejs/blob/remove-server-header-check/packages/collector/src/agentConnection.js#L129-L132) we abort the attempt and retry. This is the correct behavior IMO since this usually means that the agent was not able to start the sensor instance yet.

kirrg001: So I was thinking that if the host is not the agent, we want to tell that to the customer. 
But I guess we cannot differentiate this case from the case that the agent is not yet ready to accept connections.

Both cases would return a 404 status code and print:

> Establishing the connection to the Instana host agent has failed: %s. This usually means that the Instana
    host agent is not yet ready to accept connections....

Which is not true for the first case that the provided port is not even the agent.
I don't have an idea how to detect that.

PR #786, opened 2023-05-11
feat(aws-lambda): added support for payload version format 2.0

kirrg001:
refs #34347


packages/aws-lambda/src/capture_headers.js
```diff
@@ -11,28 +11,32 @@ exports.init = function init(config) {
   extraHttpHeadersToCapture = config.tracing.http.extraHttpHeadersToCapture;
 };
 
-exports.captureHeaders = function captureHeaders(event) {
+// https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
+exports.captureHeaders = function captureHeaders(data) {
```
kirrg001, line 15: FYI: This is either the lambda response or the lambda event.

packages/aws-lambda/src/triggers.js
```diff
@@ -62,24 +62,50 @@ exports.enrichSpanWithTriggerData = function enrichSpanWithTriggerData(event, co
 };
 
 function isApiGatewayProxyTrigger(event) {
-  // Note: An application load balancer event also has event.path and event.httpMethod but it does not have
+  // NOTE: An application load balancer event also has event.path and event.httpMethod but it does not have
   // event.resource.
-  return event.resource != null && event.path != null && event.httpMethod != null;
+  return (
+    (event.resource != null && event.path != null && event.httpMethod != null) ||
+    (event.rawPath && event.version === '2.0')
```
kirrg001, line 69: FYI: no other trigger has this format. See https://docs.aws.amazon.com/lambda/latest/dg/lambda-services.html

packages/aws-lambda/test/integration_test/test_definition.js
```diff
@@ -1039,7 +1067,7 @@ function registerTests(handlerDefinitionPath) {
       trigger: 'api-gateway-proxy',
       instanaEndpointUrl: backendBaseUrl,
       instanaAgentKey,
-      serverTiming: 'string'
+      serverTiming: true
```
kirrg001, line 1070: FYI: Removed the array case, because we already test adding `response.multiValueHeaders`. No need to test that twice.

kirrg001:
FYI When creating a Gateway REST API with lambda proxy there is no option to define the payload format. It is currently always 1.0. When creating a AWS Gateway HTTP API you can choose between 2.0 and 1.0 [2.0 is default].

packages/aws-lambda/test/lambdas/async.js
```diff
@@ -20,33 +20,37 @@ require('../../src/metrics/rootDir').root = require('path').resolve(__dirname, '
...
-  multiValueHeaders: {
-    'X-Response-Header-2': ['response', 'header', 'value 2'],
-    'X-Response-Header-4': ['response', 'header', 'value 4']
-  },
-  body: {
-    message: 'Stan says hi!'
+const handler = async event => {
+  console.log('in actual handler');
+
+  if (event.version === '1.0') {
```
kirrg001, line 28: FYI: We could refactor and optimize the lambda handler types. But ignore that for now.

packages/aws-lambda/test/integration_test/test_definition.js
```diff
@@ -1064,40 +1092,6 @@ function registerTests(handlerDefinitionPath) {
     });
   });
 
-  describe('API Gateway/EUM - Server-Timing array header already present', function () {
```
basti1302, line 1067: _issue(blocking):_ I disagree with this deletion, I think we should restore this test. 

> FYI: Removed the array case, because we already test adding response.multiValueHeaders. No need to test that twice.

I think those are two different things. The "normal" test for `response.multiValueHeaders` we only test capturing headers that the Lambda handler sets on its own, that is, in particular this code: https://github.com/instana/nodejs/blob/feat/lambda-payload2.0/packages/aws-lambda/src/capture_headers.js#L34-L43

For the Server-Timing header it is different. We manipulate this header in the response deliberately to inject the trace ID for website monitoring there. And this needs to be handled differently depending on whether the lambda handler already sets this header or not, and whether it has been set by the Lambda handler as a multi value header or a single header.

This test was for this part of the code: https://github.com/instana/nodejs/blob/feat/lambda-payload2.0/packages/aws-lambda/src/process_result.js#L71-L74

basti1302 (approved):

PR #788, opened 2023-05-15
fix(db2): ensure span is correctly processed as an IBM DB2 span

basti1302:


packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -15,7 +15,7 @@ let isActive = false;
 
 const CLOSE_TIMEOUT_IN_MS = process.env.DB2_CLOSE_TIMEOUT_IN_MS || 1000 * 30;
 
-exports.spanName = 'db2';
+exports.spanName = 'ibmdb2';
```
basti1302, line 18: This is the actual fix.

PR #794, opened 2023-05-23
refactor(collector): remove dependency on sbin and awk in containers

basti1302:
Previously, the executables `sbin` and `awk` were used to determine the default gateway IP. This is now done by parsing a file in the `/proc` file system, without assuming the existence of any particular executables.

refs 127632

----

~~TODO: I still need to test this in some real world scenarios.~~ Tested ✅ 

See also: step (3.) in https://github.ibm.com/instana/technical-documentation/tree/master/tracing/specification#phase-1--host-lookup


packages/collector/src/announceCycle/agentHostLookup.js
```diff
@@ -166,6 +164,51 @@ function checkHost(host, cb) {
...
+  // We inspect the payload, checking whether it is JSON and has the "version" property, to verify that we have
+  // actually connected to an Instana host agent.
+
+  /** @type {Array.<Buffer>} */
+  const chunks = [];
+  res.on('data', chunk => {
+    chunks.push(chunk);
+  });
+  res.on('end', () => {
+    const responsePayload = String(Buffer.concat(chunks));
```
kirrg001, line 184: Question: Which encoding is this? Wondering why we do not use `res.setEncoding('utf-8')`. Then we concat the received chunk strings and parse the JSON.

packages/collector/src/announceCycle/agentHostLookup.js
```diff
@@ -166,6 +164,51 @@ function checkHost(host, cb) {
...
+      cb(
+        new Error(
+          `The attempt to connect to the Instana host agent on ${host}:${agentOpts.port} has failed, the response ` +
+            `cannot be parsed as JSON. The party listening on ${host}:${agentOpts.port} does not seem to be an ` +
+            `Instana host agent. Full response: ${responsePayload}`
+        )
+      );
+      return;
+    }
+    if (responseJson.version !== undefined) {
```
kirrg001, line 198: Suggestion: Add a comment saying that this check is a "non official/professional/temporarly way" to figure out if the host might be the agent.

kirrg001 (approved):

packages/collector/src/announceCycle/agentHostLookup.js
```diff
@@ -124,17 +121,18 @@ function checkHost(host, cb) {
         timeout: requestTimeout
       },
       res => {
-        if (res.statusCode >= 200 && res.statusCode < 300) {
-          cb(null);
+        if (res.statusCode >= 200 && res.statusCode <= 300) {
```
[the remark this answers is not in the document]
basti1302: Ooops, I screwed up there, I'll revert this change. Thanks for spotting this! (While working on that I changed that to the opposide condition at some point `if (res.statusCode < 200 || res.statusCode >= 300) { // early exit ... ` and apparently messed up when changing it back.)

packages/collector/src/announceCycle/agentHostLookup.js
basti1302: It is utf-8. I wonder which is cheaper, collecting Buffer objects, concatenating them, and then converting to a string or converting each chunk and concatenating it. It probably does not matter at all. I'll go with your suggestion, that's also how the examples on https://nodejs.org/api/http.html are written. (If we found an agent it is only one chunk anyway, so it matters even less).

kirrg001: Also wondering what is cheaper.

packages/collector/src/announceCycle/agentHostLookup.js
basti1302: It is quite "official" though: https://github.ibm.com/instana/technical-documentation/blob/master/tracing/specification/README.md#phase-1--host-lookup. I'll add a link to that part of the spec here instead.

basti1302: ... that is, I'll add it to the existing comment at the start of the method: https://github.com/instana/nodejs/pull/794/files#diff-44f44550231a95107349e86cfa335b336d7205479f2422abb87cc6386d4b882bR175-R176

packages/collector/src/announceCycle/agentHostLookup.js
```diff
@@ -124,17 +121,18 @@ function checkHost(host, cb) {
...
+          handleResponse(host, res, cb);
         } else {
+          // We are not interested in the body of a non-2xx HTTP response, but we still need to consume the response.
```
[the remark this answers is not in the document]
basti1302: Changed. 👍

PR #802, opened 2023-06-13
test: fix bunyan, q, DB2 & OpenTracing tests for Node.js 10 & 12

basti1302:
The OTel integration is disabled for Node.js 10 and 12, the expectations
in the tests that have file system access need to accomodate for that.

### DIFF
--- a/packages/collector/test/test_util/isOTelIntegrationEnabled.js
+++ b/packages/collector/test/test_util/isOTelIntegrationEnabled.js
@@ -1,0 +1,13 @@
+/*
+ * (c) Copyright IBM Corp. 2023
+ */
+
+'use strict';
+
+const semver = require('semver');
+
+const minimumNodeJsVersionForOTelIntegration =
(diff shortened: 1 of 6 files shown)


basti1302:
https://app.circleci.com/pipelines/github/instana/nodejs/5935/workflows/c2a9e108-a4ba-4532-9434-b822b443ce5f

packages/collector/test/test_util/isOTelIntegrationEnabled.js
```diff
@@ -0,0 +1,14 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2016
```
kirrg001, line 3: 🌚

kirrg001 (approved):
Thank you! Alternatively we could also disable the integration for these tests.

basti1302:
> Thank you! Alternatively we could also disable the integration for these tests.

That would also be fine with me. You updated the tests in in your PR to check for these additional spans, so I went with that for now. If you want to reconsider that and disable the integration, we can do that as well, I do not have a strong opinion here.

PR #803, opened 2023-06-13
feat(sdk): add method to mark the current span as an error

basti1302:
see also https://github.ibm.com/instana/docs/pull/3907

### DIFF
--- a/packages/collector/test/tracing/api/test.js
+++ b/packages/collector/test/tracing/api/test.js
@@ -30,7 +30,6 @@ mochaSuiteFn('tracing/api', function () {
       const now = Date.now();
       return controls
         .sendRequest({
-          method: 'GET',
           path: '/span/active'
         })
         .then(response => {
@@ -55,7 +54,6 @@ mochaSuiteFn('tracing/api', function () {
     it('must annotate a nested value (path given as flat string)', () =>
       controls
         .sendRequest({
-          method: 'GET',
           path: '/span/annotate-path-flat-string'
         })
         .then(response => {
@@ -74,7 +72,6 @@ mochaSuiteFn('tracing/api', function () {
     it('must annotate a nested value (path given as array)', () =>
       controls
         .sendRequest({
-          method: 'GET',
           path: '/span/annotate-path-array'
         })
         .then(response => {
@@ -89,11 +86,53 @@ mochaSuiteFn('tracing/api', function () {
           expect(span.handleConstructorName).to.equal('SpanHandle');
         }));
 
+    it('must mark the current span as erroneous', () =>
+      controls
+        .sendRequest({
+          path: '/span/mark-as-erroneous'
+        })
+        .then(response => {
+          const span = response.span;
+          expect(span).to.exist;
+          expect(span.name).to.equal('node.http.server');
+          expect(span.errorCount).to.equal(1);
+          expect(span.data.http.error).to.equal(
+            'This call has been marked as erroneous via the Instana Node.js SDK, no error message has been supplied.'
+          );
+        }));
+
+    it('must mark the current span as erroneous with a custom error message', () =>
+      controls
+        .sendRequest({
+          path: '/span/mark-as-erroneous-custom-message'
+        })
+        .then(response => {
+          const span = response.span;
+          expect(span).to.exist;
+          expect(span.name).to.equal('node.http.server');
+          expect(span.errorCount).to.equal(1);
+          expect(span.data.http.error).to.not.exist;
+          expect(span.data.custom.path.error).to.equal('custom error message');
+        }));
+
+    it('must mark the current span as non-erroneous', () =>
+      controls
+        .sendRequest({
+          path: '/span/mark-as-non-erroneous'
+        })
+        .then(response => {
+          const span = response.span;
+          expect(span).to.exist;
+          expect(span.name).to.equal('node.http.server');
+          expect(span.errorCount).to.equal(0);
+          expect(span.name).to.equal('node.http.server');
+          expect(span.data.http.error).to.not.exist;
+        }));
+
     it('must manually end the currently active span', () => {
       const now = Date.now();
       return controls
         .sendRequest({
-          method: 'GET',
           path: '/span/manuallyended'
         })
         .then(response => {
@@ -126,7 +165,6 @@ mochaSuiteFn('tracing/api', function () {
     it('must provide a noop span handle', () =>
       controls
         .sendRequest({
-          method: 'GET',
           path: '/span/active'
         })
         .then(response => {
@@ -148,7 +186,6 @@ mochaSuiteFn('tracing/api', function () {
     it('must do nothing when trying to manually end the currently active span', () =>
       controls
         .sendRequest({
-          method: 'GET',
           path: '/span/manuallyended'
         })
         .then(response => {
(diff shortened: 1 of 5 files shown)


packages/collector/test/tracing/api/test.js
```diff
@@ -89,11 +86,34 @@ mochaSuiteFn('tracing/api', function () {
           expect(span.handleConstructorName).to.equal('SpanHandle');
         }));
 
+    it('must mark a the current span as an error', () =>
```
kirrg001, line 89:
```suggestion
    it('must mark the current span as an error', () =>
```

packages/collector/test/tracing/api/test.js
```diff
@@ -89,11 +86,34 @@ mochaSuiteFn('tracing/api', function () {
...
+          path: '/span/mark-as-error'
+        })
+        .then(response => {
+          const span = response.span;
+          expect(span).to.exist;
+          expect(span.name).to.equal('node.http.server');
+          expect(span.errorCount).to.equal(1);
+        }));
+
+    it('must mark a the current span as not an error', () =>
```
kirrg001, line 101:
```suggestion
    it('must mark the current span not as an error', () =>
```

packages/core/src/tracing/spanHandle.js
```diff
@@ -194,6 +203,22 @@ function _annotateWithArray(target, path, value) {
...
 
+/**
+ * Marks the span as an error (that is, it sets the error count for the span to 1). You can optionally provide a boolean
+ * attribute: if the provided value is true, the span will be marked as an error (just as if no argument had been
+ * provided), if it is false, the span will be marked as not being an error (that is, the error count will be set to 0).
+ * If no argument is provided, the default behavior is to mark the span as an error.
+ *
+ * @param {boolean?} error whether to mark the span as an error
+ */
+SpanHandle.prototype.markAsError = function markAsError(error = true) {
```
kirrg001, line 214: TBD: This change provides a helper function for customers to easily add an error count (instead of directly using `span.ec = 1`). But what about the **error message**? Wondering if we want to extend the fn design. Without an error message this feels a little incomplete?

kirrg001 (commented):
Added a comment for discussion.

packages/core/src/tracing/spanHandle.js
basti1302: > instead of directly using span.ec = 1

To clarify, this would only work like this:
```
const span = instana.currentSpan();
// get a reference to the internal span object
const internalSpan = span.span;
internalSpan..ec = 1;
```

So it is not so much about making it easier, but more about having a clear separation between our public API and our internal data structure. If we recommend to users to work on `instana.currentSpan().span` directly, we can no longer change our internal format easily in the future. Of course, users can still do that, but then they use internals and if that breaks, it is not our responsibilty.

Back to the error message suggestion: The error message is a normal annotation in span.data. It also is expected in different paths, depending on the type of span (that is, `span.data.http.error` vs. `span.data.mysql.error` vs `span.data.lambda.error` etc.) 

So from my perspective there are two reasons against adding this to the API here:

* Users can already use `span.annotate` to add the error message (if they know what kind of span they are dealing with and where the error message needs to go).
* On the API level here, it would be somewhat cumbersome for ourselves to figure out whether we would need to write `span.data.http.error` or `span.data.mysql.error` or `span.data.lambda.error` or whatever.

kirrg001: > Users can already use span.annotate to add the error message

Yes true but that is complicated, don't you think?

> On the API level here, it would be somewhat cumbersome for ourselves to figure out whether we would need to write

I thought we usually have a single key in `span.data`? Is that not correct?

We could add something like this:

```
markAsError(true, 'error message')
```

```
function markAsError(bool, msg, path) {
   if (msg) {
     const path = path || tryToGetPath(this.span.data)
     
     if (path && path.length) _annotateWithString(this.span.data, path, msg);
     else logger.warn('Cannot set error message on span. Please annotate the error message using the annotate function.');
   }
   
   ...
 }
 
 function tryToGetPath(data) {
   const keys = Object.keys(data)
   if (keys.length && keys.length === 1) return keys[0]
   return false;
}
```

basti1302: Sold! I thought about this a bit more, and I think you are right. I'll start working on this and update the PR.

packages/core/src/tracing/spanHandle.js
```diff
@@ -304,6 +419,8 @@ NoopSpanHandle.prototype.setCorrelationType = function setCorrelationType() {};
 
 NoopSpanHandle.prototype.annotate = function annotate() {};
 
+NoopSpanHandle.prototype.markAsError = function markAsError() {};
```
kirrg001, line 422: I guess this must be renamed too?

packages/core/src/tracing/spanHandle.js
```diff
@@ -194,6 +209,106 @@ function _annotateWithArray(target, path, value) {
...
+};
+
+/**
+ * Marks the span as being not an error (that is, it sets the error count for the span to 0). This is useful if the span
+ * has been marked erroneous previously (either by autotracing or via span.markAsErroneous) and that earlier decision
+ * needs to be reverted.
+ * @param {string|Array.<string>} errorMessagePath the annotation path where the error message has been written to
+ *   earlier; there is usually no need to provide this argument as this will be handled automatically
+ */
+SpanHandle.prototype.markAsNonErroneous = function markAsNonErroneous(errorMessagePath) {
```
kirrg001, line 236: I like this idea of being able to mark a span as non error 👍

packages/collector/test/tracing/api/test.js
```diff
@@ -89,11 +86,51 @@ mochaSuiteFn('tracing/api', function () {
...
+    it('must mark the current span as non-erroneous', () =>
+      controls
+        .sendRequest({
+          path: '/span/mark-as-non-erroneous'
+        })
+        .then(response => {
+          const span = response.span;
+          expect(span).to.exist;
+          expect(span.name).to.equal('node.http.server');
+          expect(span.data.http.error).to.not.exist;
```
kirrg001, line 127: Q: Wondering if we also should assert for ec 0?

kirrg001 (changes requested):
Great work 👍 

There is one issue which I think should be fixed.

packages/core/src/tracing/spanHandle.js
basti1302: Oh, very good catch. Fixed and I added a test to catch missing methods on NoopSpanHandle automatically in the future, so this should not happen again.

packages/collector/test/tracing/api/test.js
basti1302: Definitely. I accidentally dropped these assertions in the second commit. No idea why. Thanks for spotting this.

packages/core/test/tracing/spanHandle_test.js
```diff
@@ -0,0 +1,33 @@
...
+  const NoopSpanHandle = spanHandle._NoopSpanHandle;
+
+  it('NoOpSpanHandle should have the same methods as SpanHandle', () => {
+    verifyMethods(SpanHandle, NoopSpanHandle);
+  });
+
+  it('SpanHandle should have the same methods as NoOpSpanHandle', () => {
+    verifyMethods(NoopSpanHandle, SpanHandle);
+  });
+
```
basti1302, line 21: FYI: I plan to add a couple more unit tests for `annotate` and `findAndAnnotateErrorMessage` here as well later today (so please don't merge yet :-) )

kirrg001 (approved):

kirrg001:
Approved.

> so please don't merge yet

👍

PR #805, opened 2023-06-15
feat: added node v20 support

kirrg001:
- node-gyp 5.1.1 (which is installed by lerna) does no longer build with Node v20
- ESM & Node v20 support needs to wait, see https://github.com/nodejs/help/issues/4190
  - we wait till the issue is resolved and then add ESM support
- we had to bump `nan` for gcstats, see https://github.com/nodejs/nan/commit/865921e00c8e3c74ef6f835ac819fcea580c3ed1
- node-rdkafka (https://github.com/Blizzard/node-rdkafka/issues/1025)
- grpc no longer builds with Node v20
- sequilize does not work with Node v20, But their alpha release added support. Can update later


kirrg001:
Node v20 worker threads: somehow we get `false` for the main thread inside the test env. Investigating

.circleci/config.yml
```diff
@@ -563,6 +585,7 @@ workflows:
         - not:
           equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
     jobs:
+      - "node-20"
```
kirrg001, line 588: TBD: run node-20 via the nightly build or not

PR #807, opened 2023-06-19
chore: fixed coverage

kirrg001:
`nyc` has an inbuilt feature to use `.map` files instead of `.js` files if they exist.
The Otel assets had a reference to a `.map` file, which did not exist.


basti1302:
Merged via https://github.com/instana/nodejs/commit/bf4c2a23e701719f3ca4c7c383c474d0ca4ad25b

PR #811, opened 2023-06-20
fix(sdk): do not overwrite span.ec after it has been set via the SDK

kirrg001:
…neous

### DIFF
--- a/packages/collector/test/tracing/api/test.js
+++ b/packages/collector/test/tracing/api/test.js
@@ -11,6 +11,7 @@ const supportedVersion = require('@instana/core').tracing.supportedVersion;
 const config = require('../../../../core/test/config');
 const ProcessControls = require('../../test_util/ProcessControls');
 const globalAgent = require('../../globalAgent');
+const { retry, expectExactlyOneMatching } = require('../../../../core/test/test_util');
 
 const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : describe.skip;
 
@@ -18,7 +19,7 @@ mochaSuiteFn('tracing/api', function () {
   this.timeout(config.getTestTimeout());
 
   globalAgent.setUpCleanUpHooks();
-
+  const agentControls = globalAgent.instance;
   describe('when tracing is enabled', () => {
     const controls = new ProcessControls({
       dirname: __dirname,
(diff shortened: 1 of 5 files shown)


packages/collector/test/tracing/api/test.js
```diff
@@ -86,18 +87,19 @@ mochaSuiteFn('tracing/api', function () {
...
-        .then(response => {
-          const span = response.span;
-          expect(span).to.exist;
-          expect(span.name).to.equal('node.http.server');
-          expect(span.errorCount).to.equal(1);
-          expect(span.data.http.error).to.equal(
-            'This call has been marked as erroneous via the Instana Node.js SDK, no error message has been supplied.'
+        .then(() => {
+          return retry(() =>
+            agentControls.getSpans().then(spans => {
```
kirrg001, line 97: TBD: The test was not correct. We have compared the response span data. This data is not the same data as the transmitted data. I think this has a bigger underlying issue. We keep the span as instance in the instrumentation.
The span we ask with `cls` is not the same instance as far as I can see.

packages/collector/test/tracing/api/app.js
```diff
@@ -55,7 +55,14 @@ app.get('/span/annotate-path-array', (req, res) => {
 
 app.get('/span/mark-as-erroneous', (req, res) => {
   const span = instana.currentSpan();
+
+  // TODO: problem http server has a span variable in process
+  //       this is a different instance as asking the cls for the currentSpan
+  //       usually you end the span manually via the SDK .end(), but in this case we do not
   span.markAsErroneous();
+  span.disableAutoEnd();
+  span.end();
```
kirrg001, line 64: Just added these lines temporary.

packages/collector/test/tracing/api/test.js
basti1302: Gnaah, good find. The test is indeed unsuitable :-//  

> The span we ask with cls is not the same instance as far as I can see.

No, I'm pretty sure it is the same instance. I think it is a bit simpler: We basically write `span.ec` at the end of every instrumentation, before we close the span. Like [here](https://github.com/instana/nodejs/blob/main/packages/core/src/tracing/instrumentation/protocols/httpServer.js#L186-L190). 

If I remove line 189 (span.ec = ...), the test passes, even without disableAutoEnd etc. 

We probably could to do something similar as we do with `span.data.http.path_tpl` (see freezePathTemplate method in spanHandle.js): Once `span.ec` has been set explicitly via the SDK, we need to remember that fact and make span.ec read-only.

basti1302: And the reason why the disableAutoEnd/end combo fixes the test is because `span.end()` directly tansmits the span, in this case without overwriting `span.ec`.

packages/core/src/tracing/cls.js
```diff
@@ -101,7 +103,32 @@ class InstanaSpan {
...
+        }
+      },
+      enumerable: true
+    });
+    Object.defineProperty(this, 'ecHasBeenSetManually', {
+      value: false,
+      writable: true,
+      enumerable: false
+    });
+
```
basti1302, line 131: Using three properties here (the internal value `_ec`, the enumerable value `ec` plus `ecHasBeenSetManually` feels a bit clunky. But I currently do not have any idea how to do this more elegantly.

basti1302 (approved):

basti1302 (approved):

basti1302:
@kirrg001 I know you cannot formally approve this since it is your PR, but I'd like a review nonetheless, pretty please :-)

packages/core/src/tracing/instrumentation/protocols/httpServer.js
```diff
@@ -170,6 +170,13 @@ function shimEmit(realEmit) {
       });
 
       function finishSpan() {
+        if (span.transmitted) {
+          // We listen to multiple events like aborted, finish and close. In some scenarios, finishSpan will be called
+          // multiple times. However, if the span has already been transmitted to the agent, we do not need to do
+          // anything here, it has been taken care of by an earlier invocation of finishSpan.
+          return;
+        }
```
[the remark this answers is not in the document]
kirrg001: Great 👍 I remember we did something similar a while ago in our agent connection.

kirrg001 (commented):
👍

PR #815, opened 2023-06-27
test(otel): skip flaky Instana+OTel test

basti1302:
This test is flaky on CI. It basically serves to demonstrate that with this unsupported setup (combining OTel and Instana, initializing OTel first), we might receive duplicated spans, one from OTel, one from Instana. However, sometimes the OTel span is missing. That might be due to a timing issue. Anyways, since this test mostly exist to document an unsupported setup, we skip it on CI.

See also:
https://app.circleci.com/insights/github/instana/nodejs/workflows/build/tests?branch=main, section "Flaky tests".


packages/collector/test/opentelemetry/test.js
```diff
@@ -93,7 +93,11 @@ mochaSuiteFn('Opentelemetry usage', function () {
         ));
   });
 
-  describe('otel first', function () {
+  // This test is flaky on CI. It basically serves to demonstrate that with this unsupported setup (combining OTel and
+  // Instana, initializing OTel first), we might receive duplicated spans, one from OTel, one from Instana. However,
+  // sometimes the OTel span is missing. That might be due to a timing issue. Anyways, since this test mostly exist to
+  // document an unsupported setup, we skip it on CI.
+  describe.skip('otel first', function () {
```
kirrg001, line 100: TBD: Do we want to add the check `isCI()`?

basti1302: Good call, done.

PR #816, opened 2023-06-29
test: replace requests to instana.com

basti1302:
instana.com now responds with a redirect to ibm.com, which changed which spans are produced and broke some tests.


kirrg001 (approved):
Thanks so much

PR #821, opened 2023-07-11
build: moved the ibm_db to optional dependency

aryamohanan:
npm installation was blocked due to ibm_db issue with M1. Raised an [issue](https://github.com/ibmdb/node-ibm_db/issues/935), Once its resolved will correct the packages.


kirrg001 (approved):

kirrg001:
We will revert this change as soon as we've found a solution for the M1 architecture.

PR #822, opened 2023-07-11
chore: fix docker compose m1

aryamohanan:
- [x] mysql not working fully (we need to use a cloud solution for M1 I think - needs to be discussed)
- [x] check if the updates work on Kate's laptop too

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -127,8 +127,8 @@ services:
       SCHEMA_REGISTRY_HOST_NAME: schema-registry
 
   mysql:
-    image: mysql:8.0.23
-    platform: linux/x86_64
+    image: mysql:8.0.1
+    platform: linux/amd64
     ports:
       - 3306:3306
     environment:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -123,7 +127,8 @@ services:
       SCHEMA_REGISTRY_HOST_NAME: schema-registry
 
   mysql:
-    image: mysql:8.0.1
+    image: mysql:8.0.23
```
kirrg001, line 130: We need to revert that. This won't work with the latest mysql client anymore.

docker-compose.yml
```diff
@@ -123,7 +127,8 @@ services:
...
+    platform: linux/x86_64
```
kirrg001, line 131: platform: linux/amd64

kirrg001 (approved):
Looks good to me 👍 

This will fix the services which complained that they are not compatible with M1. Docker is using the arch from the host as default, which is M1 arch.

PR #826, opened 2023-07-18
feat: new trigger function URL added for aws lambda

aryamohanan:
ref 94000

- [x] double check if we should add the aws triggers to our internal tech docs
- [x] finish adding unit test

### DIFF
--- a/packages/aws-lambda/src/triggers.js
+++ b/packages/aws-lambda/src/triggers.js
@@ -54,7 +54,7 @@ exports.enrichSpanWithTriggerData = function enrichSpanWithTriggerData(event, co
     span.data.lambda.trigger = 'aws:lambda.invoke';
     return;
   } else if (isFunctionURLTrigger(event)) {
-    span.data.lambda.trigger = 'aws:lambda.functionurl';
+    span.data.lambda.trigger = 'aws:lambda.function.url';
     extractFunctionUrlEvent(event, span);
     return;
   }
--- a/packages/aws-lambda/src/triggers.js
+++ b/packages/aws-lambda/src/triggers.js
@@ -52,6 +52,10 @@ exports.enrichSpanWithTriggerData = function enrichSpanWithTriggerData(event, co
   } else if (isInvokeFunction(context)) {
     span.data.lambda.trigger = 'aws:lambda.invoke';
     return;
+  } else if (isFunctionURLTrigger(context)) {
+    span.data.lambda.trigger = 'aws:lambda.functionurl';
+    extractFunctionUrlEvent(event, span);
+    return;
   }
 
   // When an API Gateway is used without the "Use Lambda Proxy" setting, the body from the HTTP request is forwarded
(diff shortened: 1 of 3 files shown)


packages/aws-lambda/src/triggers.js
```diff
@@ -69,9 +73,22 @@ function isApiGatewayProxyTrigger(event) {
...
+  );
+}
+
+function isFunctionURLTrigger(event) {
+  // NOTE: Function URL -> Lambda sets routeKey to $default as a placeholder.
+  // NOTE: Function URL -> Currently support version 2.0.
+  // NOTE: Function URL -> Domain name follows the format https://<url-id>.lambda-url.<region>.on.aws
+  return (
+    event.routeKey === '$default' && event.version === '2.0' &&
+    (event.requestContext.domainName && event.requestContext.domainName.includes('lambda-url'))
```
basti1302, line 91: _suggestion:_ We might be a bit more specific here than `domainName.includes('lambda-url')`. In theory, other URLs might contain "lambda-url" as a substring. Maybe a regex match against the documented pattern `https://<url-id>.lambda-url.<region>.on.aws`? (And if we change it here, the same change would then be needed in line 81.)

aryamohanan: added the regex pattern

packages/aws-lambda/src/triggers.js
```diff
@@ -52,6 +52,10 @@ exports.enrichSpanWithTriggerData = function enrichSpanWithTriggerData(event, co
   } else if (isInvokeFunction(context)) {
     span.data.lambda.trigger = 'aws:lambda.invoke';
     return;
+  } else if (isFunctionURLTrigger(context)) {
```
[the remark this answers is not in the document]
aryamohanan: done 👍

packages/aws-lambda/src/triggers.js
```diff
@@ -387,3 +404,16 @@ function readSqsStringMessageAttribute(messageAttributes, key) {
...
 }
+
+function extractFunctionUrlEvent(event, span) {
+  const requestCtx = event.requestContext || {};
+  const requestCtxHttp = requestCtx.http || {};
+
+  span.data.http = {
+    method: requestCtxHttp.method,
+    url: requestCtxHttp.path,
+    path_tpl: event.rawPath,
```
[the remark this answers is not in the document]
aryamohanan: removed the span.data.http.path_tpl annotation

packages/aws-lambda/src/triggers.js
```diff
@@ -52,6 +53,10 @@ exports.enrichSpanWithTriggerData = function enrichSpanWithTriggerData(event, co
   } else if (isInvokeFunction(context)) {
     span.data.lambda.trigger = 'aws:lambda.invoke';
     return;
+  } else if (isFunctionURLTrigger(event)) {
+    span.data.lambda.trigger = 'aws:lambda.functionurl';
```
basti1302, line 57: Sorry, I have one more tiny nitpick (which I should have brought up yesterday 🙈 ). The other triggers use `.` as a word separator. (`load.balancer`, `cloudwatch.logs` etc.)

For consistency, we could do the same here as well and go with `aws:lambda.function.url`. What do you think?

aryamohanan: That sounds like an excellent catch 🥇

packages/aws-lambda/test/integration_test/test_definition.js
```diff
@@ -1595,6 +1595,38 @@ function registerTests(handlerDefinitionPath) {
           ])
         ));
   });
+describe('triggered by aws lambda function url', function () {
```
basti1302, line 1598: The test looks 👍 👍 

One nitpick: The formatting looks off. I would recommend to configure your IDE/editor so that it automatically runs `prettier` when you save a file.

Something like [this](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) maybe for Visual Studio Code (not sure what you are using).

You can also run `npm run prettier` to format all source code files - that might be a good idea because probably one or two other files that you have touched in this PR also need to be reformatted slightly.

basti1302:
I believe this is almost ready to be merged (except for the source code formatting, see inline comment). [CI](https://app.circleci.com/pipelines/github/instana/nodejs?branch=aws-lambda-function-url) also looks good.

@kirrg001, WDYT? I added some review comments already which Arya already fixed. Do you also want to take a quick peek?

packages/aws-lambda/src/triggers.js
```diff
@@ -69,9 +74,22 @@ function isApiGatewayProxyTrigger(event) {
   // NOTE: Gateway Rest API with lambda proxy -> always format 1.0
   // NOTE: Gateway HTTP API with lambda proxy -> default is 2.0, but can be configured on creation
   // NOTE: Gateway REST/HTTP API without lambda proxy -> format cannot be interpreted
+  // NOTE: Function URL also support version 2 and contains rawPath, Gateway HTTP API does not follow the format
```
kirrg001, line 77: > Function URL also support version 2

Question: Is there a way to configure Function URL with format version 1.0? Or does Function Url only support 2.0?

kirrg001 (approved):
Great work <3

basti1302 (approved):

PR #828, opened 2023-07-21
fix: stack.slice not a function

kirrg001:
Reported by customer.

### DIFF
--- a/packages/core/src/util/stackTrace.js
+++ b/packages/core/src/util/stackTrace.js
@@ -37,11 +37,17 @@ exports.captureStackTrace = function captureStackTrace(length, referenceFunction
     // drawback is that it might show too much, but that's better than not having any stack trace at all.
     Error.captureStackTrace(stackTraceTarget);
   }
-  const stack = stackTraceTarget.stack;
+
+  let stack = stackTraceTarget.stack;
+
+  if (!Array.isArray(stack)) {
+    stack = [];
+  }
+
   Error.stackTraceLimit = originalLimit;
   Error.prepareStackTrace = originalPrepareStackTrace;
 
-  if (drop > 0) {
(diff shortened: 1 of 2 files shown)


packages/core/src/util/stackTrace.js
```diff
@@ -37,11 +37,17 @@ exports.captureStackTrace = function captureStackTrace(length, referenceFunction
...
+
+  if (!Array.isArray(stack)) {
+    stack = [];
+  }
+
   Error.stackTraceLimit = originalLimit;
   Error.prepareStackTrace = originalPrepareStackTrace;
 
-  if (drop > 0) {
+  if (drop > 0 && stack.length >= drop) {
```
kirrg001, line 50: IMO there were two bugs.

1. stack not being a string (the actual use case is not 100% known yet)
2. stack.length >= drop behaviour

basti1302, line 50: _praise:_ good catch, that way we at least get some stack trace if `drop > stack.length` accidentally. (I don't think this should ever happen, but anyway.)

basti1302 (approved):

packages/core/src/util/stackTrace.js
basti1302: I wonder when (2) would have happened in practice (without the new stack = [] fix)

aryamohanan (commented):
Code looks good to me 👍

packages/core/src/util/stackTrace.js
kirrg001: Yeah 👍

kirrg001: I am not sure tbh, I just looked for potential bugs in this fn :)

PR #832, opened 2023-07-25
feat(aws-lambda): added support for missing regions

kirrg001:
refs 129560

### DIFF
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -108,6 +108,8 @@ fi
 ZIP_NAME=$ZIP_PREFIX.zip
 TMP_ZIP_DIR=tmp
 
+echo "step 1/9: Fetching AWS regions"
+
 # us-gov-* only available to US government agencies, U.S. government etc.
 # cn-* (china regions) completely disconnected from normal AWS account. 
 SKIPPED_REGIONS=$'cn-north-1\ncn-northwest-1\nus-gov-east-1\nus-gov-west-1'
@@ -163,15 +165,6 @@ fi
 
 if [[ -z $AWS_ACCESS_KEY_ID ]] || [[ -z $AWS_SECRET_ACCESS_KEY ]]; then
   echo Warning: AWS_ACCESS_KEY_ID or AWS_SECRET_ACCESS_KEY are not set. This might be okay if you have set up AWS authentication via other means. If not, the AWS cli commands to publish the layer will fail.
-fi
-
-echo "step 1/9: AWS regions"
-
-if [[ -z $SKIP_AWS_PUBLISH_LAYER ]]; then
-  echo Will publish to regions:
-  echo "$REGIONS"
-else
-  echo Publishing to AWS will be skipped.
 fi
 
 echo "step 2/9: Prepare build environment"
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -108,7 +108,13 @@ fi
 ZIP_NAME=$ZIP_PREFIX.zip
 TMP_ZIP_DIR=tmp
 
-echo "step 1/9: Fetching AWS regions"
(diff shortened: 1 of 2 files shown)


packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -160,16 +165,13 @@ if [[ -z $AWS_ACCESS_KEY_ID ]] || [[ -z $AWS_SECRET_ACCESS_KEY ]]; then
   echo Warning: AWS_ACCESS_KEY_ID or AWS_SECRET_ACCESS_KEY are not set. This might be okay if you have set up AWS authentication via other means. If not, the AWS cli commands to publish the layer will fail.
 fi
 
-echo "step 1/9: fetching AWS regions (skipping, using fixed list of regions for now)"
-
-# Actually, this should give us the regions where the Lambda service is provided:
-# REGIONS=$(aws ssm get-parameters-by-path --path /aws/service/global-infrastructure/services/lambda/regions --output text --query "Parameters[].Value" | tr '\t' '\n')
-# But for some reason, publishing to all of these regions does not work. In particular, the
-# following regions either require special authorization/subscription status or don't support Lambdas: ap-east-1 me-south-1 ap-northeast-3
+echo "step 1/9: AWS regions"
```
basti1302, line 168: _suggestion:_ Let's move this `echo` above the `aws ssm` call (and maybe print "fetching AWS regions" again as before), to give an indication of what is currently happening.

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -108,8 +108,12 @@ fi
 ZIP_NAME=$ZIP_PREFIX.zip
 TMP_ZIP_DIR=tmp
 
+# us-gov-* only available to US government agencies, U.S. government etc.
+# cn-* (china regions) completely disconnected from normal AWS account. 
+SKIPPED_REGIONS=$'cn-north-1\ncn-northwest-1\nus-gov-east-1\nus-gov-west-1'
```
basti1302, line 113: 👍

basti1302 (approved):
LGTM. Thanks for also updating the report-layer-version script as well 👍

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -160,16 +165,13 @@ if [[ -z $AWS_ACCESS_KEY_ID ]] || [[ -z $AWS_SECRET_ACCESS_KEY ]]; then
...
+echo "step 1/9: AWS regions"
 
 if [[ -z $SKIP_AWS_PUBLISH_LAYER ]]; then
   echo Will publish to regions:
   echo "$REGIONS"
+else
+  echo Publishing to AWS will be skipped.
 fi
 
 echo "step 2/9: Prepare build enviornment"
```
aryamohanan, line 177: A small typo, `enviornment`, I know it's not part of this PR, but it's a minor change.

packages/aws-lambda/layer/bin/publish-layer.sh
kirrg001: Sure 👍

kirrg001: The problem with this suggestion is that we are fetching the regions before we print the confirm prompt.
The step printing should come after the confirm prompt. 

How about we just remove this step?

aryamohanan (approved):

packages/aws-lambda/layer/bin/publish-layer.sh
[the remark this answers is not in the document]
kirrg001: Discussed in slack :)

packages/aws-lambda/layer/bin/publish-layer.sh
basti1302: Is there a specific reason why we don't want to print any messages before the prompt? I think it is fine to 
* print "step 1: I am fetching all the regions now", 
* then print all the env vars and settings and 
* then ask via the prompt whether it's fine to continue.

I'm mostly thinking about the CI job here anyway, where the prompt is off. If the CI job fails, I want to see immediately which step failed.

kirrg001: Ok why not

PR #837, opened 2023-07-28
feat: added support for the latest  aws sdk v3 changes

aryamohanan:
ref: [issue](https://github.com/aws/aws-sdk-js-v3/pull/4873)
All aws sdk v3 services went through  refactoring of smithy client.


package.json
```diff
@@ -59,8 +59,8 @@
     "@apollo/gateway": "0.50.2",
     "@aws-sdk/client-dynamodb": "3.284.0",
     "@aws-sdk/client-dynamodb2": "npm:@aws-sdk/client-dynamodb@3.20.0",
-    "@aws-sdk/client-s3": "3.321.1",
-    "@aws-sdk/client-s3-2": "npm:@aws-sdk/client-s3@3.21.0",
```
aryamohanan, line 63: We decided to drop this version to avoid testing against three versions.

basti1302 (approved):
Awesome! Thanks for catching this and fixing it so quickly!

PR #838, opened 2023-07-31
feat: aws sdk v3 kinesis instrumentation

aryamohanan:
ref: [133974](https://instana.kanbanize.com/ctrl_board/56/cards/133974/details/)


package.json
```diff
@@ -61,6 +61,7 @@
     "@aws-sdk/client-dynamodb2": "npm:@aws-sdk/client-dynamodb@3.20.0",
     "@aws-sdk/client-s3": "3.379.1",
     "@aws-sdk/client-s3-2": "npm:@aws-sdk/client-s3@3.321.1",
+    "@aws-sdk/client-kinesis": "^3.363.0",
```
kirrg001, line 64: Latest version is 3.379.1. I would suggest to test against this version :)

packages/collector/test/tracing/cloud/aws-sdk/v3/kinesis/test_definition.js
```diff
@@ -0,0 +1,239 @@
...
+  'putRecord',
+  'listStreams',
+  'getRecords',
+  'getShardIterator'
+];
+
+const getNextCallMethod = require('@instana/core/test/test_util/circular_list').getCircularList(requestMethods);
+function start(version) {
+  let mochaSuiteFn;
+  if (!supportedVersion(process.versions.node)) {
```
kirrg001, line 55: Question: I think we need to skip the test on Node v14?
See https://github.com/aws/aws-sdk-js-v3/blob/v3.379.1/clients/client-kinesis/package.json#L74

You skipped it already in the cleanup function. I am wondering why we need to check this in the cleanup function at all?

packages/collector/test/tracing/cloud/aws-sdk/v3/kinesis/util.js
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+'use strict';
+
+const semver = require('semver');
+const minNodeJsVersion = '14.0.0';
+const { KinesisClient, DescribeStreamCommand, DeleteStreamCommand } = require('@aws-sdk/client-kinesis');
+
+if (semver.lt(process.versions.node, minNodeJsVersion)) {
+  exports.cleanup = async function () {
```
kirrg001, line 12: I am still wondering why we need to check for the Node version in the cleanup script?
I know this was copied from s3, but it looks wrong. I might be wrong :)

packages/collector/test/tracing/cloud/aws-sdk/v3/kinesis/app.js
```diff
@@ -0,0 +1,216 @@
...
+      return new Promise((resolve, reject) => {
+        kinesisSendWithCB(command, (err, result) => {
+          if (err) {
+            reject(err);
+          } else {
+            resolve(result);
+          }
+        });
+      });
+    case 'async-v2':
```
kirrg001, line 164: Suggestion: rename all of the v2 strings in this file to v3.

aryamohanan: This case is to support v2 compatible style in v3.

kirrg001: Ah ok then nvm :)

packages/collector/test/tracing/cloud/aws-sdk/v3/kinesis/test_definition.js
aryamohanan: The missing node version check has been added. Concerning the node version check in the cleanup method, all other similar instruments have it, which is why I included it. We can remove it if it is no longer required 🙂

kirrg001: Yeah it would be cool if you could double check that, because the Node version check in the cleanup fn is confusing IMO.

aryamohanan: Removed the check from cleanup function.

packages/collector/test/tracing/cloud/aws-sdk/v3/kinesis/util.js
aryamohanan: Removed the version check

kirrg001: Great :)

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/kinesis.js
```diff
@@ -0,0 +1,127 @@
...
+    if (cls.skipExitTracing()) {
+      return originalSend.apply(ctx, smithySendArgs);
+    }
+
+    const command = smithySendArgs[0];
+    return cls.ns.runAndReturn(() => {
+      const self = this;
+      const span = cls.startSpan(this.spanName, EXIT);
+      span.ts = Date.now();
+      span.stack = getStackTrace(this.instrumentedSend, 1);
```
kirrg001, line 55: Bug: `this.instrumentedSend` is not defined.

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/kinesis.js
```diff
@@ -0,0 +1,127 @@
...
+      span.data[this.spanName] = this.buildSpanData(command.constructor.name, command.input);
+      this.captureRegion(ctx, span);
```
kirrg001, line 57: bug: There is no region for kinesis.

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/kinesis.js
```diff
@@ -0,0 +1,127 @@
...
+
+      if (typeof smithySendArgs[1] === 'function') {
```
kirrg001, line 59: TBD: I think the callback position can vary?

See https://github.com/awslabs/smithy-typescript/blob/main/packages/smithy-client/src/client.ts#L36

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/kinesis.js
```diff
@@ -0,0 +1,127 @@
...
+        const _callback = smithySendArgs[1];
+
+        smithySendArgs[1] = cls.ns.bind(function (err) {
+          if (err) {
+            _callback.apply(this, arguments);
+            self.finishSpan(err, span);
+          } else {
+            _callback.apply(this, arguments);
+            self.finishSpan(null, span);
```
kirrg001, line 68: nitpick: I think it doesn't matter much (and I know other services use the same logic), but we usually first send the span and then the customer's fn is called.

packages/collector/test/tracing/cloud/aws-sdk/v3/kinesis/test_definition.js
```diff
@@ -0,0 +1,239 @@
...
+    describe('tracing disabled', () => {
+      this.timeout(config.getTestTimeout() * 2);
+
+      const appControls = new ProcessControls({
+        appPath: path.join(__dirname, 'app'),
+        useGlobalAgent: true,
+        tracingEnabled: false,
+        env: {
+          AWS_KINESIS_STREAM_NAME: streamName,
+          AWS_SDK_CLIENT_REQUIRE: version
```
kirrg001, line 160: question: Is `AWS_SDK_CLIENT_REQUIRE` used? 🤔

kirrg001 (changes requested):
Great work 🥳 Left some comments

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/kinesis.js
```diff
@@ -0,0 +1,127 @@
...
+'use strict';
+
+const cls = require('../../../../cls');
+const { EXIT } = require('../../../../constants');
+const { getStackTrace } = require('../../../../tracingUtil');
+const { InstanaAWSProduct } = require('./instana_aws_product');
+
+const SPAN_NAME = 'kinesis';
+
+const operationsInfo = {
```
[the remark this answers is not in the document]
kirrg001: **Discussed**: We want to collect all commands. We do no longer want to use a lookup `operationsInfo` object. We can remove this object and lower case `command.constructor.name` and remove the `Command` postfix.

e.g. `DeleteStreamCommand` will be `deleteStream`

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/kinesis.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+    if (params.ShardIteratorType) {
+      spanData.shardType = params.ShardIteratorType;
+    }
+    if (params.StartingSequenceNumber) {
+      spanData.startSequenceNumber = params.StartingSequenceNumber;
+    }
+    if (params.ShardId) {
+      spanData.shard = params.ShardId;
+    }
```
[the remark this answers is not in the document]
kirrg001: **Discussed**: The UI client has a couple of more span attributes, which the BE does not process. It is fine as it is :)

packages/collector/test/tracing/cloud/aws-sdk/v3/kinesis/test_definition.js
aryamohanan: Removed

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/kinesis.js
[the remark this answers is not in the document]
aryamohanan: Updated the code to support all operations.

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/index.js
```diff
@@ -81,14 +82,25 @@ function shimSmithySend(originalSend) {
     const command = smithySendArgs[0];
 
     if (isActive) {
-      const awsProduct = operationMap[command.constructor.name];
-
-      if (awsProduct) {
-        return awsProduct.instrumentedSmithySend(self, originalSend, smithySendArgs);
+      if (self.config.serviceId) {
```
kirrg001, line 85: nitpik: Wondering if `self.config` could be undefined for any older v3 version

aryamohanan: Done

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/kinesis.js
aryamohanan: done 👍

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/index.js
```diff
@@ -81,14 +82,23 @@ function shimSmithySend(originalSend) {
     const command = smithySendArgs[0];
 
     if (isActive) {
-      const awsProduct = operationMap[command.constructor.name];
+      const serviceId = self.config && self.config.serviceId;
+      const awsProduct = awsProducts.find(aws => aws.spanName === (serviceId ? serviceId.toLowerCase() : ''));
```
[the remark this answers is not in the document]
aryamohanan: done 💯

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/instana_aws_product.js
```diff
@@ -24,11 +24,11 @@ class InstanaAWSProduct {
    */
   getOperations() {
     const operationMap = {};
-
-    this.operations.forEach(op => {
-      operationMap[op] = this;
-    });
-
+    if (this.operations && this.operations.length) {
```
kirrg001, line 27: suggestion: in line 16 do

`this.operations = operations || []`

or do

```constructor(spanName, operations: []) {```

Then you don't need this check at all.

**Review by kirrg001 — APPROVED** [review]
IMO I think this is good to go now :)

File: packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/instana_aws_product.js (hunk 14 above)
Comment (aryamohanan) [reply packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/instana_aws_product.js]: Thanks, updated the code with the suggestion

File: packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/index.js [hunk 15]
```diff
@@ -81,14 +82,23 @@ function shimSmithySend(originalSend) {
...
-      const awsProduct = operationMap[command.constructor.name];
+      const serviceId = self.config && self.config.serviceId;
+      const awsProduct = serviceId && awsProducts.find(aws => aws.spanName === serviceId.toLowerCase());
 
-      if (awsProduct) {
+      if (awsProduct && awsProduct.supportsOperation(command.constructor.name)) {
         return awsProduct.instrumentedSmithySend(self, originalSend, smithySendArgs);
+      } else {
+        // This code will be removed when we release support for version 3x.
+        // We are keeping it here for now to avoid any unknown problems.
```
basti1302, line 92: _question:_ I do not understand this comment. Does the "support for 3.x" mean "we will remove this when we release @instana/core@3.x? Or does the 3.x refer to the version of the AWS SDK or the kinesis client? 

AFAICT, we can only remove this, when we have migrated all other AWS SDK v3 instrumentations to this new pattern, right?

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/index.js
```diff
@@ -81,14 +82,23 @@ function shimSmithySend(originalSend) {
...
+        handleLegacy(originalSend, smithySendArgs);
       }
-
-      return originalSend.apply(self, smithySendArgs);
     }
     return originalSend.apply(self, smithySendArgs);
```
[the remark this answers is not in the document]
aryamohanan: Great! I'll retain the existing code as-is in the 'else' section to prevent potential future bugs. Thank you all for bringing this issue to my attention 💯

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/index.js
aryamohanan: Absolutely! Our current plan is to remove the 'else' section within version 3.x. We're optimistic that by that time, we'll have successfully transitioned all other AWS SDK v3 instrumentations to this new approach. [ref](https://instana.kanbanize.com/ctrl_board/56/cards/96212/details/). @kirrg001

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/index.js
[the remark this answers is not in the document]
basti1302: Good call. I think that is easier to read. 👍

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/index.js
```diff
@@ -81,13 +82,21 @@ function shimSmithySend(originalSend) {
...
-      const awsProduct = operationMap[command.constructor.name];
+      const serviceId = self.config && self.config.serviceId;
+      let awsProduct = serviceId && awsProducts.find(aws => aws.spanName === serviceId.toLowerCase());
 
-      if (awsProduct) {
+      if (awsProduct && awsProduct.supportsOperation(command.constructor.name)) {
         return awsProduct.instrumentedSmithySend(self, originalSend, smithySendArgs);
+      } else {
+        // This code will be removed in version 3.x release.
+        // We are keeping it here for now to avoid any unknown problems.
```
basti1302, line 92: Following up from the earlier discussion on this comment.
> Absolutely! Our current plan is to remove the 'else' section within version @instana/core@3.x.
> We're optimistic that by that time, we'll have successfully transitioned all other AWS SDK v3
> instrumentations to this new approach. [ref](https://instana.kanbanize.com/ctrl_board/56/cards/96212/details/).

That makes sense. Great that you have noted this down as a major version refactoring in the Kanbanize card.

_nitpick:_ How about we make this a bit more explicit here in the comment?

```suggestion
        // This code can be removed once all AWS SDK v3 instrumentations have been refactored to use the new approach
        // introduced in https://github.com/instana/nodejs/pull/838 for kinesis. That is: Do not use an explicit
        // operationsInfo/operationsMap map that restricts the traced operations to a subset of possible operations, but
        // instead allow _all_ operations to be traced, using the operation name from `command.constructor.name` for
        // span.data.$spanName.op. We plan to finish this refactoring before or with the next major release (3.x) of the
        // @instana packages.
```

I know, this is much more verbose, but maybe our future selves will find it easier to understand.

basti1302 (approved):

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/index.js
aryamohanan: Incorporated the suggestion 💯

basti1302:
@aryamohanan I second Kathi, let's merge this 👍 👍 🎉

PR #840, opened 2023-08-02
fix: reduced noisy log warnings when active entry could not be found

kirrg001:
I noticed when running tests for suppressed tests, that we are seeing:

> Cannot start an exit span as this requires an active entry (or intermediate) span as parent

This is not right. When a trace is suppressed, then we do not create an entry span.
Therefor the check to skip the exit span won't have a parent span.

Found two things which we had to optimize:

1. `options` need to be merged. Bug.
2. Reorder suppress and parentSpan check. IMO this is okay.

### DIFF
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -546,13 +546,16 @@ function call(fn) {
  * @param {Object.<string, *>} options
  */
 function skipExitTracing(options) {
-  const opts = Object.assign({}, options, {
-    isActive: true,
-    extendedResponse: false,
-    skipParentSpanCheck: false,
-    log: true,
-    skipIsTracing: false
-  });
+  const opts = Object.assign(
+    {
+      isActive: true,
+      extendedResponse: false,
+      skipParentSpanCheck: false,
+      log: true,
+      skipIsTracing: false
+    },
+    options
+  );
 
   const parentSpan = getCurrentSpan();
   const suppressed = tracingSuppressed();
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -545,20 +545,28 @@ function call(fn) {
  *
  * @param {Object.<string, *>} options
  */
-function skipExitTracing(
-  options = {
+function skipExitTracing(options) {
+  const opts = Object.assign({}, options, {
     isActive: true,
     extendedResponse: false,
     skipParentSpanCheck: false,
     log: true,
     skipIsTracing: false
-  }
-) {
+  });
+
   const parentSpan = getCurrentSpan();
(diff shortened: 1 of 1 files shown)


packages/core/src/tracing/cls.js
```diff
@@ -545,20 +545,28 @@ function call(fn) {
...
+  const opts = Object.assign({}, options, {
     isActive: true,
     extendedResponse: false,
     skipParentSpanCheck: false,
     log: true,
     skipIsTracing: false
-  }
-) {
+  });
+
```
basti1302, line 556: _issue(blocking):_ Oh right, the previous code was definitely broken. Good catch. However, this also does not work correctly, I think? See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign?retiredLocale=de#merging_objects_with_same_properties ->

> The properties are overwritten by other objects that have the same properties later in the parameters order.

I think we need to put the defaults first and the actual incoming parameters last, so they overwrite the defaults. I think what you want is this:

```
function skipExitTracing(options) {
  const opts = Object.assign(
    {
      isActive: true,
      extendedResponse: false,
      skipParentSpanCheck: false,
      log: true,
      skipIsTracing: false
    },
    options
  );
  // console.table(opts);
  ...
}
```

or shorter, without an extra identifier, like this:

```
function skipExitTracing(options) {
  options = Object.assign(
    {
      isActive: true,
      extendedResponse: false,
      skipParentSpanCheck: false,
      log: true,
      skipIsTracing: false
    },
    options
  );
  // console.table(options);
   ...
}
```

packages/core/src/tracing/cls.js
```diff
@@ -545,20 +545,28 @@ function call(fn) {
...
   const isExitSpanResult = isExitSpan(parentSpan);
-  if (!options.skipParentSpanCheck && (!parentSpan || isExitSpanResult)) {
-    if (options.log) {
+
+  // CASE: first ask for suppressed, because if we skip the entry span, we won't have a parentSpan
+  //       on the exit span, which would create noisy log messages.
+  if (suppressed) {
+    if (opts.extendedResponse) return { skip: true, suppressed, isExitSpan: isExitSpanResult };
+    return true;
+  }
```
basti1302, line 566: _praise:_ Yes, that makes total sense. Again, great catch. 🎣

basti1302 (changes requested):

packages/core/src/tracing/cls.js
kirrg001: Ah ja good catch 👍 I thought it was the other way around :)

basti1302 (approved):

PR #846, opened 2023-08-17
test: added test case for testing asynchronous pg calls

aryamohanan:
ref: [136255](https://instana.kanbanize.com/ctrl_board/56/cards/136255/details/)


kirrg001 (approved):
Thank you :)

PR #848, opened 2023-08-18
chore(dependencies): replace recursive copy with built in method

jimmywarting:
NodeJS have `cp` built in so there is no need to have a dependency for it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -136,7 +136,6 @@
         "proxyquire": "2.1.3",
         "q": "1.5.1",
         "read-pkg-up": "^7.0.1",
-        "recursive-copy": "^2.0.14",
         "redis": "^4.3.1",
         "redis-v0": "npm:redis@^0.12.1",
         "redis-v3": "npm:redis@^3.1.2",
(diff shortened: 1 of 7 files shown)


basti1302:
I'm sorry, but we cannot merge this. [fs.cp](https://nodejs.org/dist/latest-v20.x/docs/api/fs.html#fscpsrc-dest-options-callback) was added in Node.js v16.7.0 (see the History for that API).

image

We currently support Node.js 10 - 20. This change would break all customer using Node.js 10, 12, and 14 (plus early versions of 16). Even with the next major release of our packages, where we might drop support for e.g. Node.js 10, we will probably not drop support for Node.js 12, 14, and 16.

basti1302:
And yes, I saw the comment you added to a very old PR, I just didn't have time to respond there yet. I didn't expect you to create a PR for that one day later :-)   Sorry about the time you have put into this PR. Maybe open an issue before creating the PR next time to discuss whether what you want to do is actually feasible and has a chance of being accepted?

jimmywarting:
That's alright. then you could close this or have it in the backlog for later when you do decide to drop support for older nodejs version. 
I did not know that you supported really old nodejs version all the way down to v10. 
10 new major releases have been created since then, we are now at v20+ 😝

jimmywarting:
i had a look at https://npmgraph.js.org/?q=%40instana%2Fshared-metrics and i tough there where so many sub dependencies that is getting downloaded in vain.
every (sub)dependency chain is a potential security risk so i'm only trying to reduce the footprint

kind of have a "package-phobia" for things 🤪
think you should target current LTS, but understand if you want to keep support for older versions.

basti1302:
> 10 new major releases have been created since then, we are now at v20+ 😝
> think you should target current LTS, but understand if you want to keep support for older versions.

Believe me, I understand perfectly well how outdated Node.js 10 is.

You would need to discuss that with our customer base. Let's just say that running production applications on very old runtimes and never upgrading to a newer runtime version is far from unheard of. We need to support what our customers use, that's a basic reality of being a commercial APM product ;-)

Issue #849, opened 2023-08-21
[Bug]: version upgrade from 2.29.0 to 2.30.0 causing log flood.


Vhndaree:
### Problem Description

I upgrade from 2.29.0 to 2.30.0 and after that 
```
Cannot start an exit span as this requires an active entry (or intermediate) span as parent. Currently there is no span active at all
```

is being logged as INFO a lot of time, which causes issues in service performance

### Short, Self Contained Example

Usage
```
if (agentPort && agentHost) {
    require('@instana/collector')();
}
```

### Node.js Version

^18.14.0

### package.json

```shell
{
    "name": "service",
    "version": "0.1.0",
    "private": true,
    "description": "",
    "main": "",
    "scripts": {
        "build": "tsc"
    },
    "engines": {
        "node": "^18.14.0"
    },
    "license": "UNLICENSED",
    "dependencies": {
        "@instana/collector": "^2.30.0"
    }
}
```

kirrg001:
Could you please share your package.json with me? Thank you

Vhndaree:
````
{
    "name": "service",
    "version": "0.1.0",
    "private": true,
    "description": "",
    "main": "",
    "scripts": {
        "build": "tsc",
        "build:debug": "tsc --sourceMap",
        "start": "node --max-http-header-size=32768 dist/cmd/service/main.js",
        "lint": "eslint . --ext .ts",
        "lint:fix": "eslint . --ext .ts --fix",
        "format": "prettier '**/*.ts' --write",
        "test:all": "npm run test"
    },
    "engines": {
        "node": "^18.14.0"
    },
    "keywords": [],
    "license": "UNLICENSED",
    "dependencies": {
        "@instana/collector": "^2.29.0",
        "@wireapp/lru-cache": "^3.8.1",
        "agentkeepalive": "^4.5.0",
        "ajv": "^8.12.0",
        "avsc": "^5.7.7",
        "axios": "^0.27.2",
        "body-parser": "^1.20.2",
        "chokidar": "^3.5.3",
        "cookie-parser": "^1.4.6",
        "dns-cache": "^2.0.0",
        "ejs": "^3.1.9",
        "express": "^4.18.2",
        "express-promise-router": "^4.1.1",
        "html-entities": "^2.4.0",
        "ioredis": "^5.3.2",
        "jsonwebtoken": "^9.0.1",
        "kafkajs": "^2.2.4",
        "method-override": "^3.0.0",
        "murmurhash": "^2.0.1",
[...]

PR #850, opened 2023-08-23
fix: reduced log flooding when collector is not rdy

kirrg001:
refs #849

This is a draft. This is not fully designed or final. We might want to use EventEmitter. TBD


kirrg001:
Discussed: We want to use `isActive` attribute of the instrumentations.

basti1302:
> https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-collector-configuration#disabling-individual-tracers

For the record, we discussed this independently. We can only disable all of the AWS v3 instrumentations at once currently, more or less. The available strings for disabling would be:
* `aws-sdk/v2/index`,
* `aws-sdk/v3/index`,
* `aws-sdk/v2/sdk`,
* `aws-sdk/v2/sqs`,
* `gcp/pubsub`, and
* `gcp/storage`

We agreed to not change this in the scope of this PR. We will consider refactoring the mechanism for disabling individual instrumentation plug-ins with the next major update (3.0.0).

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/index.js
```diff
@@ -74,27 +75,24 @@ function shimSmithySend(originalSend) {
...
+    } else {
+      // This code can be removed once all AWS SDK v3 instrumentations have been refactored to use the new approach
+      // introduced in https://github.com/instana/nodejs/pull/838 for kinesis. That is: Do not use an explicit
+      // operationsInfo/operationsMap map that restricts the traced operations to a subset of possible operations, but
+      // instead allow _all_ operations to be traced, using the operation name from `command.constructor.name` for
+      // span.data.$spanName.op. We plan to finish this refactoring before or with the next major release (3.x) of the
+      // @instana packages.
+      awsProduct = operationMap[smithySendArgs[0].constructor.name];
+      if (awsProduct) {
+        return awsProduct.instrumentedSmithySend(self, isActive, originalSend, smithySendArgs);
```
basti1302, line 92: _praise:_ I think this is actually better now, because it is more consistent. Each individual awsProduct now checks its `isActive` flag and either calls the original send method or the instrumented one. And we get rid of one `originalSend.apply(self, smithySendArgs);` here in index.js.

I like! :)

basti1302 (approved):
LGTM 👍

PR #851, opened 2023-08-23
fix: resolved missing dynamodb spans

kirrg001:
refs 136351

Problem:
One library is using `@aws-sdk/smithy-client` and the other is already using `@smithy/smithy-client`.
If one is loaded earlier, the other one won't create spans.

@aws-sdk/client-dynamodb is currently using **3.188.0 on Lambda.**
This version still uses the **old smithy** `@aws-sdk/smithy-client`.

But `@aws-sdk/lib-dynamodb` uses already the new smithy `@smithy/smithy-client`. That's why the customer didn't see any spans.


aryamohanan (approved):
Looks good to me

PR #853, opened 2023-08-25
fix(dynamodb): resolved all operation names

kirrg001:
e.g. lib-dynamodb is using PutCommand instead of PutItemCommand

This is an additional fix for the previous dynamodb fix. It wasn't working for the lib-dynamodb commands, because `operationInfo.op` crashed.


aryamohanan (approved):
LGTM

PR #858, opened 2023-08-31
feat: added support for batch write dynamodb

aryamohanan:
ref: [137369](https://instana.kanbanize.com/ctrl_board/56/cards/137369/details/)


basti1302 (approved):
Awesome, thanks for the quick fix!

PR #860, opened 2023-09-04
feat(aws): added support for sns v3 

kirrg001:
refs 135931

I am seeing an error in the localstack logs on CI and locally:

> 2023-09-04T14:26:21.391 ERROR --- [-functhread3] hypercorn.error            : Error in ASGI Framework

But it looks like we can ignore it.

- [x] add refactoring ticket for sqs


packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/receiver.js
```diff
@@ -38,10 +38,16 @@ const agentPort = process.env.INSTANA_AGENT_PORT || 42699;
...
-
-const sqs = new awsSdk3.SQSClient({ region: awsRegion });
-const sqsv2 = new awsSdk3.SQS({ region: awsRegion });
-
+let sqs;
+let sqsv2;
+
+if (process.env.AWS_ENDPOINT) {
+  sqs = new awsSdk3.SQSClient({ region: awsRegion, endpoint: process.env.AWS_ENDPOINT });
+  sqsv2 = new awsSdk3.SQS({ region: awsRegion, endpoint: process.env.AWS_ENDPOINT });
```
kirrg001, line 46: FYI: This is really nice, because we proofed that localstack is working with SNS & SQS.

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -107,7 +107,10 @@ function shouldBeBypassed(parentSpan, options) {
   // Same regex used by AWS SDK at /lib/services/sqs.js
   const hostMatchesSQS = evaluateHeaderValue(hostInfo, header => header.match(/^sqs\.(?:.+?)\./) !== null);
 
-  if (parentSpan && parentSpan.n === 'sqs' && isAWSNodeJSHeader && hostMatchesSQS) {
+  // 'user-agent': 'aws-sdk-js/3.329.0 os/darwin/22.5.0 lang/js md/nodejs/18.14.2 api/sqs/3.329.0'
+  const agentMatchesSQS = evaluateHeaderValue(userAgent, header => header.toLowerCase().indexOf('api/sqs') > -1);
```
kirrg001, line 111: FYI: @basti1302 Setting the hostname does not change the http hostname header. I solved it like this now.

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/sqs.js
```diff
@@ -263,14 +265,14 @@ class InstanaAWSSQS extends InstanaAWSProduct {
...
               const messages = data.Messages;
-
               let tracingAttributes = readTracingAttributes(messages[0].MessageAttributes);
               if (!hasTracingAttributes(tracingAttributes)) {
                 tracingAttributes = readTracingAttributesFromSns(messages[0].Body);
               }
               if (tracingAttributes.level === '0') {
                 cls.setTracingLevel('0');
                 setImmediate(() => span.cancel());
+                return data;
```
kirrg001, line 275: TBD: IMO there was a return missing here. Why would we call the logic below for a cancelled span.

basti1302: 👍 👍 

Yes, I think returning here is better. Looking at this, I think if we add both `span.cancel` and `this.finishSpan` to the next tick of the event loop via setImmediate, I am not even sure if the behavior would be deterministic. If we happen to call span.cancel first, everything is fine and the `this.finishSpan` should be ignored I think. If it happens the other way around, we would create a span even though it should have been supressed.

I think the callback case in line 236 probably has the same potential issue. There, we cannot simply return, though, we would have to call `_callback.apply(self, arguments);` and then we can return.

WDYT?

packages/core/src/tracing/instrumentation/protocols/httpClient.js
basti1302: Looks good. 👍

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/sqs.js
kirrg001: https://github.com/instana/nodejs/pull/860/commits/5c5543af89fbccc44b615a4f7a06ccdfcc2cac04

packages/collector/test/tracing/cloud/aws-sdk/v3/sns/test_definition.js
```diff
@@ -0,0 +1,305 @@
...
+  PublishCommand: {
+    Message: 'STRING_VALUE',
+    TargetArn: 'STRING_VALUE',
+    Subject: 'STRING_VALUE'
+  }
+};
+
+/*
+fs
+  .readdirSync(`${path.dirname(require.resolve('@aws-sdk/client-sns'))}/commands`)
```
aryamohanan, line 38: Suggestion: Can we remove this commented code if it is no longer needed?

packages/collector/test/tracing/cloud/aws-sdk/v3/sns/test_definition.js
```diff
@@ -0,0 +1,305 @@
...
+  .map(cmd => {
+    return cmd.match(/(.*).js/)[1];
+  });
+*/
+
+function start(version) {
+  let mochaSuiteFn;
+  this.timeout(config.getTestTimeout() * 4);
+
+  // https://github.com/aws/aws-sdk-js-v3/blob/v3.391.0/clients/client-sns/package.json#L70
```
aryamohanan, line 52: 💯

packages/core/src/tracing/tracingUtil.js
```diff
@@ -254,3 +254,24 @@ exports.requireModuleFromApplicationUnderMonitoringSafely = function requireModu
 ) {
   return require(path.join(basePath, ...relativePath));
 };
+
+exports.findCallback = (/** @type {string | any[]} */ originalArgs) => {
```
aryamohanan, line 258: 🥇

packages/collector/test/tracing/cloud/aws-sdk/v3/sns/test_definition.js
kirrg001: Yeah sure we can do. I wasn't sure tbh

package.json
```diff
@@ -62,6 +62,7 @@
     "@aws-sdk/client-kinesis": "^3.379.1",
     "@aws-sdk/client-s3": "3.379.1",
     "@aws-sdk/client-s3-2": "npm:@aws-sdk/client-s3@3.321.1",
+    "@aws-sdk/client-sns": "^3.388.0",
```
kirrg001, line 65: Latest is 3.405.0
Will update before merge

aryamohanan (approved):

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/sqs.js
kirrg001: @aryamohanan Did you review https://github.com/instana/nodejs/commit/5c5543af89fbccc44b615a4f7a06ccdfcc2cac04? That would be helpful :)

PR #866, opened 2023-09-12
fix: disabling aws sdk instrumentation

kirrg001:
refs 139073

This is

a) not documented https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-collector-configuration#disabling-individual-tracers
b) customers would need to use `aws/sdk/v2/index`, which is really not a good design

This is a first fix to this. 
In the next major release we could drop the support for the old syntax. The major ticket contains it already.
This PR ensures that we still support the old notation, because it could be that some customers are using this.

Another addition could be: add an optional function like `handleDisablingInstrumentation`  and each instrumentation can handle it on it's own if wished e.g. aws sdk could disable only certain instrumentations. Right now users can only disable the while library.

### DIFF
--- a/packages/core/src/tracing/index.js
+++ b/packages/core/src/tracing/index.js
@@ -125,6 +125,20 @@ exports.util = tracingUtil;
 exports.esmSupportedVersion = esmSupportedVersion;
 
 /**
+ * @param {import('../util/normalizeConfig').InstanaConfig} cfg
+ * @param {string} instrumentationKey
+ */
+const isInstrumentationDisabled = (cfg, instrumentationKey) => {
+  const extractedInstrumentationName = instrumentationKey.match(/.\/instrumentation\/[^/]*\/(.*)/)[1];
+
+  return (
+    cfg.tracing.disabledTracers.includes(extractedInstrumentationName.toLowerCase()) ||
+    (instrumentationModules[instrumentationKey].instrumentationName &&
+      cfg.tracing.disabledTracers.includes(instrumentationModules[instrumentationKey].instrumentationName))
+  );
+};
+
+/**
  * @param {Array.<InstanaInstrumentedModule>} _additionalInstrumentationModules
  */
 exports.registerAdditionalInstrumentations = function registerAdditionalInstrumentations(
@@ -184,14 +198,7 @@ function initInstrumenations(_config) {
     instrumentations.forEach(instrumentationKey => {
       instrumentationModules[instrumentationKey] = require(instrumentationKey);
 
-      const extractedInstrumentationName = instrumentationKey.match(/.\/instrumentation\/[^/]*\/(.*)/)[1];
-
-      const isInstrumentationDisabled =
-        _config.tracing.disabledTracers.includes(extractedInstrumentationName.toLowerCase()) ||
-        (instrumentationModules[instrumentationKey].instrumentationName &&
-          _config.tracing.disabledTracers.includes(instrumentationModules[instrumentationKey].instrumentationName));
-
-      if (!isInstrumentationDisabled) {
+      if (!isInstrumentationDisabled(_config, instrumentationKey)) {
         instrumentationModules[instrumentationKey].init(_config);
       }
 
@@ -223,14 +230,7 @@ exports.activate = function activate(extraConfig = {}) {
 
     if (automaticTracingEnabled) {
       instrumentations.forEach(instrumentationKey => {
-        const extractedInstrumentationName = /.\/instrumentation\/[^/]*\/(.*)/.exec(instrumentationKey)[1];
-
-        const isDisabled =
-          config.tracing.disabledTracers.includes(extractedInstrumentationName.toLowerCase()) ||
-          (instrumentationModules[instrumentationKey].instrumentationName &&
-            config.tracing.disabledTracers.includes(instrumentationModules[instrumentationKey].instrumentationName));
-
-        if (!isDisabled) {
+        if (!isInstrumentationDisabled(config, instrumentationKey)) {
           instrumentationModules[instrumentationKey].activate(extraConfig);
         }
       });
--- a/packages/core/src/tracing/index.js
+++ b/packages/core/src/tracing/index.js
@@ -41,8 +41,10 @@ let processIdentityProvider = null;
 const instrumentations = [
   './instrumentation/cloud/aws-sdk/v2/index',
   './instrumentation/cloud/aws-sdk/v3/index',
+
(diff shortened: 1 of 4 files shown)


packages/core/src/tracing/index.js
```diff
@@ -41,8 +41,10 @@ let processIdentityProvider = null;
 const instrumentations = [
   './instrumentation/cloud/aws-sdk/v2/index',
   './instrumentation/cloud/aws-sdk/v3/index',
+
```
aryamohanan, line 44: suggestion: can we avoid the empty lines in line no: 44 and 47?

aryamohanan (approved):

packages/core/src/tracing/index.js
kirrg001: 👍

packages/core/test/tracing/index_test.js
```diff
@@ -67,21 +78,27 @@ mochaSuiteFn('[UNIT] tracing/index', function () {
...
-    initAndActivate({ tracing: { disabledTracers: ['grpc', 'kafkajs'] } });
+    initAndActivate({ tracing: { disabledTracers: ['grpc', 'kafkajs', 'aws-sdk/v2/index'] } });
     expect(activateStubGrpc).to.not.have.been.called;
     expect(activateStubGrpcJs).to.have.been.called;
     expect(activateStubKafkaJs).to.not.have.been.called;
     expect(activateStubRdKafka).to.have.been.called;
     expect(initStubGrpc).not.to.have.been.called;
+
+    expect(initAwsSdkv2).not.to.have.been.called;
+    expect(activateAwsSdkv2).not.to.have.been.called;
```
basti1302, line 89: _suggestion:_ For some of the disabled instrumentations, we only check whether their activate has or has not been called, for others we check this for init as well as activate. Then, we also have a separate test case 'skip init step when instrumentation disabled' that checks all the init methods. 

Maybe we should merge this all into this test case to make it a bit more consistent? Something like this:

```
it('deactivate instrumentation via config', () => {
  initAndActivate({ tracing: { disabledTracers: ['grpc', 'kafkajs', 'aws-sdk/v2/index'] } });
  // grpc instrumentation  has been disabled, make sure neither its init nor its activate method are called
  expect(initStubGrpc).not.to.have.been.called;
  expect(activateStubGrpc).to.not.have.been.called;
  
  // grpcJs instrumentation has not been disabled, make sure its init and activate are called
  expect(initStubGrpcJs).to.have.been.called;
  expect(activateStubGrpcJs).to.have.been.called;
  
  // kafkajs has been disabled...
  expect(initStubKafkaJs).to.not.have.been.called;
  expect(activateStubKafkaJs).to.not.have.been.called;
      
  // ...but rdkafka has not been disabled
  expect(initStubRdKafka).to.have.been.called;
  expect(activateStubRdKafka).to.have.been.called;
  
  // aws-sdk/v2 has been disabled (via aws-sdk/v2/index)
  expect(initAwsSdkv2).not.to.have.been.called;
  expect(activateAwsSdkv2).not.to.have.been.called;
});
```

And then remove the separate test case 'skip init step when instrumentation disabled'.

Of course, this is not directly related to the changes in this PR.

And since you probably want to specifically test for `'aws-sdk/v2/index'` as well as `'aws-sdk/v2'`, we could have two separate test cases, one with the string with `/index` and one for the string without `/index`.

packages/core/test/tracing/index_test.js
```diff
@@ -67,21 +78,27 @@ mochaSuiteFn('[UNIT] tracing/index', function () {
...
 
   it('deactivate instrumentation via env', () => {
     process.env.INSTANA_DISABLED_TRACERS = 'grpc';
     initAndActivate({});
+
     expect(activateStubGrpc).to.not.have.been.called;
     expect(activateStubGrpcJs).to.have.been.called;
     expect(activateStubKafkaJs).to.have.been.called;
     expect(activateStubRdKafka).to.have.been.called;
+    expect(activateAwsSdkv2).to.have.been.called;
```
basti1302, line 100: _thought:_ We could use the same pairs of init and activate checks as in the test case 'deactivate instrumentation via config' here. On the other hand, the test setup only deactivates `grpc`, so to be honest, I do not remember why we check for all the other instrumentations here as well 🤔   I think just having expectations for `grpc` and maybe `grpcjs` for their respective init and activate methods would be enough?

packages/core/test/tracing/index_test.js
```diff
@@ -100,9 +117,10 @@ mochaSuiteFn('[UNIT] tracing/index', function () {
   });
 
   it('skip init step when instrumentation disabled', () => {
-    initAndActivate({ tracing: { disabledTracers: ['grpcjs', 'kafkajs'] } });
+    initAndActivate({ tracing: { disabledTracers: ['grpcjs', 'kafkajs', 'aws-sdk/v2'] } });
     expect(initStubKafkaJs).not.to.have.been.called;
     expect(initStubGrpcJs).not.to.have.been.called;
+    expect(initAwsSdkv2).to.not.have.been.called;
     expect(initStubGrpc).to.have.been.called;
   });
```
basti1302, line 125: _though:_ We could merge this into 'deactivate instrumentation via config' instead of having a separate test case, see above.

packages/core/src/tracing/index.js
```diff
@@ -182,8 +183,14 @@ function initInstrumenations(_config) {
...
-      const instrumentationName = instrumentationKey.match(/.\/instrumentation\/[^/]*\/(.*)/)[1];
-      const isInstrumentationDisabled = _config.tracing.disabledTracers.includes(instrumentationName.toLowerCase());
+
+      const extractedInstrumentationName = instrumentationKey.match(/.\/instrumentation\/[^/]*\/(.*)/)[1];
+
+      const isInstrumentationDisabled =
+        _config.tracing.disabledTracers.includes(extractedInstrumentationName.toLowerCase()) ||
+        (instrumentationModules[instrumentationKey].instrumentationName &&
+          _config.tracing.disabledTracers.includes(instrumentationModules[instrumentationKey].instrumentationName));
+
```
basti1302, line 193: _suggestion:_ We should have probable done that before, but how about extracting this part into a function that checks whether an instrumentation is disabled? It looks like we have the exact same snippet twice, once in `initInstrumenations` and once in `activate`.

basti1302 (approved):
Looks good! 👍 👍 

I left some rambling comments (as always :-) ) which are mostly a matter of taste and all completely optional.

PR #876, opened 2023-09-27
Migrate to npm workspaces and lerna v7

kirrg001:
refs 104038

Finally updating lernav3 to lerna v7, which includes a complete replacement of how dependencies are managed. They are now managed via **npm workspaces**.

https://lerna.js.org/docs/introduction

**Instead of having separate lock files for each package, npm workspaces keep the dependency tree in a single lock file for all the packages within a workspace that helps streamline the dependency management in a large javascript monorepo setup.**

**Lerna no longer takes care of dependency management / dependency linking.**

**Lerna publish and version is still the same.** (https://lerna.js.org/docs/features/version-and-publish)

- [x] fix CI caching lock file
- [x] fix tests
- [x] optimise docs
- [x] ~bring back bookshelf~


CONTRIBUTING.md
```diff
@@ -288,7 +230,7 @@ This assumes that you are on a branch where the versions in all package.json fil
 Doing a stable release after a prerelease requires you to use:
 
 ```
-NPM_CONFIG_OTP={TOKEN} lerna publish --force-publish --conventional-graduate && lerna bootstrap
+NPM_CONFIG_OTP={TOKEN} lerna publish --force-publish --conventional-graduate
```
kirrg001, line 233: TBD: Do we need do add `npm i` instead? 🤔

package.json
```diff
@@ -31,28 +34,24 @@
   ],
   "scripts": {
     "audit": "lerna run audit --stream",
-    "bootstrap": "lerna bootstrap",
-    "bump": "lerna publish",
     "commit": "cz",
     "doctoc": "doctoc README.md packages/collector/README.md",
     "lint": "node misc/is_eol || (eslint .*.js *.js && npm run typecheck && lerna run lint --stream && eslint misc && eslint dockerfile-examples && eslint bin)",
-    "postinstall": "husky install && npm run bootstrap",
+    "postinstall": "husky install && lerna run node_modules:exists",
```
kirrg001, line 40: FYI: npm workspaces is very optimized to install the dependencies across the project. For some packages we do not have dependencies which will be installed in the target folder. That's why it can happen that a package does not have a `node_modules` folder. But we need a `node_modules` folder for our core/metrics package. The metrics package [determines](https://github.com/instana/nodejs/blob/v2.33.1/packages/core/src/util/applicationUnderMonitoring.js#L217) the app package.json based on the existence of the `node_modules` folder. I am wondering if should remove this check.

package.json
```diff
@@ -31,28 +34,24 @@
...
     "open-coverage": "open coverage/lcov-report/index.html",
     "prettier": "prettier --write 'misc/**/*.js' 'dockerfile-examples/**/*.js' 'bin/**/*.js' && lerna run prettier --stream",
-    "refresh-package-lock-files": "lerna exec 'rm -rf node_modules' && npm install",
     "reinstall-deps": "lerna clean -y && rm -rf node_modules && npm install",
     "test": "lerna run test --stream",
     "test:ci": "lerna run test:ci --stream",
     "test:ci:collector": "lerna run test:ci  --stream --scope=@instana/collector",
     "test:debug": "lerna run test:debug --stream",
     "typecheck": "tsc",
-    "update-deps": "lerna exec '../../node_modules/.bin/ncu --upgrade' && lerna exec 'npm i --package-lock-only' && npm run reinstall-deps",
```
kirrg001, line 52: TBD: I never used this command and I don't see a use case for it. I can re-add the cmd.

package.json
```diff
@@ -31,28 +34,23 @@
...
     "verify": "lerna run verify --stream",
-    "depcheck": "lerna exec 'npx depcheck --skip-missing --ignores nan,no-code2,agent-base,@types/*' --stream",
-    "version": "lerna clean --yes && lerna bootstrap --no-ci --ignore-scripts -- --package-lock-only --no-audit && git add packages/*/package-lock.json"
```
kirrg001, line 55: TBD: No idea what the use case is for this command.

package.json
```diff
@@ -103,10 +101,12 @@
     "apollo-server-express": "^2.24.1",
     "apollo-server-express-v3": "npm:apollo-server-express@3.6.2",
     "async": "^3.2.2",
+    "semver": "7.5.4",
```
kirrg001, line 104: FYI: I had to add the semver version here because NPM was complaining because a subdependency relied on an older semver version which was then stored under `node_modules/semver` (main semver version) and this version was vulnerable. `npm audit` failed.

package.json
```diff
@@ -195,7 +196,7 @@
     "restify": "^8.6.1",
     "rimraf": "^3.0.2",
     "sequelize": "^6.17.0",
-    "sinon": "^11.1.1",
+    "sinon": "^15.2.0",
```
kirrg001, line 199: FYI: Had to update sinon too beause typescript failed.

packages/autoprofile/package.json
```diff
@@ -24,7 +24,8 @@
     "url": "git+https://github.com/instana/nodejs.git"
   },
   "scripts": {
-    "audit": "npm audit --production",
+    "audit": "npm audit ----omit=dev",
```
kirrg001, line 27: FYI: `--production` is deprecated.

packages/aws-fargate/package.json
```diff
@@ -68,9 +69,7 @@
     "@instana/metrics-util": "2.33.1",
     "@instana/serverless": "2.33.1"
   },
-  "devDependencies": {
-    "no-code2": "2.0.0"
```
kirrg001, line 72: FYI: There is no need to keep `node-code2` anymore. We've added it in the past because lerna had a bug with updating the package lock file correctly.

packages/aws-fargate/src/index.js
```diff
@@ -42,7 +41,6 @@ try {
     //
     // eslint-disable-next-line
     module.exports = exports = require('/instana/legacy-1x/node_modules/@instana/aws-fargate/src/index');
-    return;
```
kirrg001, line 45: FYI: typescript complained

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -30,8 +28,7 @@ describe('retry loading native addons', function () {
   const metricAddonsTestConfigs = [
     {
       name: 'event-loop-stats',
-      nodeModulesPath: sharedMetricsNodeModules,
-      nativeModulePath: path.join(sharedMetricsNodeModules, 'event-loop-stats'),
+      nativeModulePath: require.resolve('event-loop-stats'),
```
kirrg001, line 31: FYI: npm workspaces installs event-loop-stats on the root node_modules folder.

packages/collector/test/tracing/frameworks/bookshelf/test.js
```diff
@@ -15,7 +15,14 @@ const globalAgent = require('../../../globalAgent');
...
 
-mochaSuiteFn('frameworks/bookshelf', function () {
+// Bookshelf wasn't updated since 2020.
+// Waiting for https://github.com/bookshelf/bookshelf/pull/2125
+// We currently can't install bookshelf, because
+// it requires knex<0.22 and knex<0.22 requires mssqlv6
+// if we install mssql@6 into the collector workspace, typeorm will complain
+// that it needs mssql@9. If we install typeorm into the root, typeorm
+// will complaint that it needs mongodb@5.
+mochaSuiteFn.skip('frameworks/bookshelf', function () {
```
kirrg001, line 25: FYI: ^

kirrg001: I am not sure how to deal with that in the future.

packages/core/package.json
```diff
@@ -126,9 +126,9 @@
   "dependencies": {
     "@opentelemetry/api": "1.4.1",
     "@opentelemetry/context-async-hooks": "1.9.1",
-    "@opentelemetry/instrumentation-fs": "0.7.3",
-    "@opentelemetry/instrumentation-restify": "0.32.3",
-    "@opentelemetry/instrumentation-socket.io": "0.33.3",
+    "@opentelemetry/instrumentation-fs": "0.8.1",
+    "@opentelemetry/instrumentation-restify": "0.34.0",
+    "@opentelemetry/instrumentation-socket.io": "0.34.1",
```
kirrg001, line 131: FYI: updated these packages because they used a vulnerable semver version.

packages/core/src/tracing/instrumentation/protocols/captureHttpHeadersUtil.js
```diff
@@ -223,7 +223,7 @@ exports.mergeExtraHeadersFromNormalizedObjectLiteral = function mergeExtraHeader
  *
  * @param {Object.<string, string> | undefined} headersAlreadyCapturedIfAny an object literal representing the headers
  * that have already been captured earlier
- * @param {Headers} headers a Fetch API Headers object
+ * @param {import('node-fetch').Headers} headers a Fetch API Headers object
```
kirrg001, line 226: FYI: typescheck complained

package.json
```diff
@@ -103,10 +101,12 @@
     "apollo-server-express": "^2.24.1",
     "apollo-server-express-v3": "npm:apollo-server-express@3.6.2",
     "async": "^3.2.2",
+    "semver": "7.5.4",
     "aws-sdk": "2.1450.0",
     "basic-auth-parser": "0.0.2",
     "bluebird": "^3.7.2",
     "body-parser": "^1.19.0",
+    "bson": "^4.7.2",
```
kirrg001, line 109: FYI: Had to install bson 4 to fix typescript

>node_modules/@types/mongodb/index.d.ts:58:74 - error TS2724: '"bson"' has no exported member named 'ObjectID'. Did you mean 'ObjectId'?

mongodb v6 uses bson v6
@types/mongodb is deprecated but still used by mongoose-v5@npm:mongoose@5.13.17

That's why we need to pin bson to v4 on the root

kirrg001:
> Lerna publish and version is still the same. (https://lerna.js.org/docs/features/version-and-publish)

I have tested one/two commands manually. I think the release is not affected.

CONTRIBUTING.md
basti1302: Uhh, not sure. The `lerna bootstrap` was there because `lerna publish` left the local repository in a weird state where the leaf packages would have their node_modules folder removed (if I remember correctly). The additional `lerna bootstrap` fixed that then. This could be totally different after upgrading from lerna 2 to 7. Suggestion: Let's not add `npm i` now and let's see when we do the next major release. Also, we wanted to create a CI job for that anyway, and then it doesn't matter in which state the local folder is after the release.

packages/aws-fargate/package.json
basti1302: I think we also had this as a workaround to make sure there is a node_modules folder in each package. But maybe I remember that wrong, not sure. If it was because of that, the new solution (`node_modules:exist`) is better anyway.

packages/aws-fargate/package.json
```diff
@@ -68,9 +69,7 @@
...
-  },
+  "devDependencies": {},
```
basti1302, line 72: _nitpick:_ We could remove this section entirely, right? (Also in all the other packages where this is empty now.)

packages/collector/test/tracing/frameworks/bookshelf/test.js
basti1302: Me neither but IMO it's fine if we need to let go of this test. We do not have specific instrumentation for bookshelf anyway, and the test is mostly just for making sure our database driver level instrumentation also works when used via bookshelf.

basti1302 (approved):
This is awesome! 😍

CONTRIBUTING.md
```diff
@@ -6,11 +6,13 @@ You should use the Node.js version defined in [.nvmrc](https://github.com/instan
 
 Python3 (< 3.11) is required, otherwise the db2 package won't build, see https://github.com/nodejs/node-gyp/issues/2219.
 
+`npm install npx -g` to execute npm package binaries.
```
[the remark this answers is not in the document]
kirrg001: You are right 👍

CONTRIBUTING.md
```diff
@@ -65,73 +67,62 @@ Thank you for your interest in the Instana Node.js project!
...
-Run `lerna add --scope=@instana/$package $dependencyName@$version && npm run refresh-package-lock-files`. For example, when you want to update the `semver` dependency in `@instana/core` to version 7.5.3, run `lerna add --scope=@instana/core semver@7.5.3 && npm run refresh-package-lock-files`.
+`npm uninstall ${dependency-name}`
+`npm uninstall -D ${dependency-name}`
+`npm uninstall ${dependency-name} -w packages/collector`
+`npm uninstall -D ${dependency-name} -w packages/collector`
 
-You can manage dev dependencies in the same way, that is `lerna add --dev --scope=@instana/autoprofile semver@7.5.3 && npm run refresh-package-lock-files`.
+### Updating A Single Version In A `package.json` File
 
-Do *not* run `npm install ${dependency-name}` in the directory of a package (like, in `packages/core`) to update a dependency. This will mess up the `package-lock.json` file in that package. (If you accidentally did this, you can do a `git checkout` of the `package.json` file of the package and then run `npm run refresh-package-lock-files` to fix the lock file and the content of `node_modules`.)
```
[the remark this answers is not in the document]
kirrg001: Ja its equivalent :) That is super smart from npm workspaces

package.json
kirrg001: Good point 👍 Will cleanup

packages/aws-fargate/package.json
kirrg001: Ok let's remove it 👍

packages/collector/test/tracing/frameworks/bookshelf/test.js
kirrg001: Ok I will remove the whole test for now 👍

PR #880, opened 2023-10-05
Fixed lerna for older node versions

kirrg001:
Trying to fix the build


kirrg001:
Closing because I cannot find a workaround for Node 10 & 12.
See #883

Issue #885, opened 2023-10-09
[Bug]: "Cannot start an exit span" log messages in conjunction with web socket


chromey:
### Problem Description

A recent update of this package caused two of our NestJS services to be flooded with log statements like this:
```
Cannot start an exit span as this requires an active entry (or intermediate) span as parent. Currently there is no span active at all
```
basically, similar to #867

We have many Nest services that don't exhibit this problem. What sets the two affected ones apart is that they both use a [websocket gateway](https://docs.nestjs.com/websockets/gateways), which is why I believe this causes the problem.

I could not add the package-lock.json as it was too long for this form. Let me know if you need specific version details on certain packages.

`@instana/collector` is 2.33.1
`@nestjs/websockets` is 10.2.6

Thanks!

### Node.js Version

Node.js 20.x

### package.json

```shell
{
	"name": "xxx",
	"version": "0.0.9",
	"license": "MIT",
	"scripts": {
		"start": "nx serve",
		"build": "nx build",
		"test": "nx test",
		"prepare": "is-docker || (husky install && nx run-many --target env-decrypt --all)",
		"mongo-rs": "run-rs -v 5.0.14 --host 127.0.0.1 --portStart 45000 --dbpath mongodb-data"
	},
	"private": true,
	"devDependencies": {
		"@compodoc/compodoc": "^1.1.15",
		"@golevelup/ts-jest": "^0.4.0",
		"@jest-mock/express": "^2.0.1",
		"@nestjs/schematics": "10.0.2",
		"@nestjs/testing": "10.2.6",
		"@nx-tools/nx-container": "^4.0.2",
		"@nx-tools/nx-docker": "^3.0.1",
		"@nx/devkit": "16.9.1",
		"@nx/eslint-plugin": "16.9.1",
		"@nx/jest": "16.9.1",
		"@nx/js": "16.9.1",
		"@nx/linter": "16.9.1",
		"@nx/nest": "16.9.1",
		"@nx/node": "16.9.1",
		"@nx/plugin": "16.9.1",
		"@nx/webpack": "16.9.1",
		"@nx/workspace": "16.9.1",
		"@swc-node/register": "~1.6.7",
		"@swc/cli": "~0.1.62",
		"@swc/core": "1.3.90",
		"@swc/helpers": "0.5.2",
		"@twittwer/compodoc": "^1.8.0",
		"@types/amqp-connection-manager": "^3.4.1",
		"@types/boom": "^7.3.2",
[...]

PR #891, opened 2023-10-10
fix: reduce log warnings when there is no entry span

kirrg001:
refs #885

We decided to remove the log warning for now completely.
Alternative solutions: add a delay mechanisms or only log once. But it's too complicated for the outcome we get from the warning.


aryamohanan (approved):
LGTM

PR #897, opened 2023-10-20
fix(rdkafka): only log warning about header format once

athirasomanath:
As Discussed, made the changes for logging in the activate function.


basti1302:
@athirasomanath The changes look good from my perspective. 👍 

The PR is still in draft mode. I think it's ready for review, right? Please mark it as ready for review then :-) 

Please also rebase on `main`. 

Note: We generate our CHANGELOG file directly from the commit messages, so it is good practice to write them in style that is relevant for people using the package. If you want to merge the PR later with rebase&merge, please squash the two commits into one and provide a more extensive commit message suitable for the changelog. Maybe something like "fix(rdkafka): only log warning about header format once"? Alternatively, you can also merge the PR later with "squash & merge" and provide a proper message then.

basti1302 (approved):

basti1302 (approved):

PR #904, opened 2023-10-25
docs: update support/ticket system links

basti1302:
We still mentioned https://support.instana.com in a few places, but that domain does not exist anymore. Updated to https://www.ibm.com/mysupport, which is the external link for CSP.


aryamohanan (approved):
🥇

PR #906, opened 2023-10-27
fix: adds version information for all NPM packages

athirasomanath:
This change adds a version information for any NPM packages: aws-lambda, fargate and Google Cloud Run

### DIFF
--- a/packages/google-cloud-run/src/index.js
+++ b/packages/google-cloud-run/src/index.js
@@ -4,10 +4,12 @@
 
 'use strict';
 
+const path = require('path');
 // MAINTENANCE NOTE: All code in this file needs to be compatible with all Node.js versions >= 6.0.0.
 
 try {
   const nodeJsVersion = process.version;
+  console.log('@instana/google-cloud-run module version:', require(path.join(__dirname, '..', 'package.json')).version);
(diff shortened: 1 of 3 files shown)


basti1302 (approved):
LGTM

packages/google-cloud-run/src/index.js
```diff
@@ -4,10 +4,13 @@
 
 'use strict';
 
+const path = require('path');
 // MAINTENANCE NOTE: All code in this file needs to be compatible with all Node.js versions >= 6.0.0.
 
 try {
   const nodeJsVersion = process.version;
+  // eslint-disable-next-line no-console
+  console.log('@instana/google-cloud-run module version:', require(path.join(__dirname, '..', 'package.json')).version);
```
basti1302, line 13: _nitpick:_ this should be above line 11, I think. The `const nodeJsVersion = process.version;` and the `if (typeof nodeJsVersion !== 'string') {` belong together and now this sits right in the middle between them.

basti1302 (approved):

packages/google-cloud-run/src/index.js
athirasomanath: Resolved the comment

PR #913, opened 2023-11-03
 feat(google cloud storage): added support for google cloud storage v7

athirasomanath:


packages/core/src/util/initializedTooLateHeuristic.js
```diff
@@ -20,7 +20,7 @@ let patterns = [
   /\/@elastic\/elasticsearch\/index.js/,
   /\/@google-cloud\/pubsub\/build\/src\/publisher\/index.js/,
   /\/@google-cloud\/pubsub\/build\/src\/subscriber.js/,
-  /\/@google-cloud\/storage\/build\/src\/index.js/,
+  /\/@google-cloud\/storage\/build\/cjs\/src\/index.js/,
```
basti1302, line 23: Maybe we should keep both entries, the old one and the new path, so that the detection works for older as well as for newer package versions?

athirasomanath: Pushed the changes as per review.

package.json
```diff
@@ -73,7 +73,7 @@
     "@elastic/elasticsearch-v7.17.0": "npm:@elastic/elasticsearch@7.17.0",
     "@elastic/elasticsearch-v7.9.0": "npm:@elastic/elasticsearch@7.9.0",
     "@google-cloud/pubsub": "4.0.2",
-    "@google-cloud/storage": "^6.9.4",
+    "@google-cloud/storage": "^7.5.0",
```
basti1302, line 76: _issue:_ IIRC version 7 dropped support for Node.js 12, making Node.js 14 the minimum required version. That means we need to update for which versions we execute the test. Currently we skip the test when the current Node.js version is < 12. See https://github.com/instana/nodejs/blob/main/packages/collector/test/tracing/cloud/gcp/storage/test.js#L56-L64

basti1302 (changes requested):

basti1302 (approved):

package.json
```diff
@@ -147,7 +147,7 @@
     "koa": "^2.13.4",
     "koa-bodyparser": "^4.3.0",
     "koa-morgan": "^1.0.1",
-    "koa-router": "12.0.0",
+    "koa-router": "12.0.1",
```
kirrg001, line 150: This shouldn't be here. Please update the pull request with main again 👍

kirrg001 (approved):
LGTM

Please see https://github.com/instana/nodejs/pull/913/files#r1386235569
The PR needs to be rebased against main.

If you merge the pull request, please squash the commits into one.

package.json
basti1302: This change is already on [main](https://github.com/instana/nodejs/blob/main/package.json#L150) I think, but you (Kathi) are right it shouldn't show up as a diff. I think this PR needs to be rebased on main? Maybe you (Athira) cherry-picked the commit from main? 🤔

athirasomanath: rebased to main and pushed

PR #918, opened 2023-11-07
feat(fastify): added support for fastify v4- redacted

athirasomanath:


package.json
```diff
@@ -133,6 +133,7 @@
     "fastify": "1.14.6",
     "fastify2": "npm:fastify@2.15.3",
     "fastify3": "npm:fastify@3.21.5",
+    "fastify4": "npm:fastify@4.24.3",
```
kirrg001, line 136: If there is no good reason to test against four major versions, could you please drop the last two major versions?
And: could you please use the v4 as latest version?

packages/collector/test/tracing/frameworks/fastify/test.js
```diff
@@ -20,7 +20,7 @@ const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : descri
 mochaSuiteFn('tracing/fastify', function () {
   this.timeout(config.getTestTimeout());
 
-  ['1.14.6', '2.15.3', '3.21.5'].forEach(version => {
+  ['1.14.6', '2.15.3', '3.21.5', '4.24.3'].forEach(version => {
```
kirrg001, line 23: Suggestion: We got rid of this specific version pattern in the past. Could you please adapt it for fastify?
See example https://github.com/instana/nodejs/blob/v2.34.1/packages/collector/test/tracing/database/mongodb/test.js#L23

kirrg001 (changes requested):

packages/core/src/tracing/instrumentation/frameworks/fastify.js
```diff
@@ -58,9 +58,8 @@ function instrument(build) {
 
     app.addHook('onRequest', function onRequest(request, reply, done) {
       try {
-        // NOTE: v1 uses _context https://github.com/fastify/fastify/blob/1.x/fastify.js#L276
-        //       v2/v3 uses context https://github.com/fastify/fastify/blob/2.x/test/handler-context.test.js#L41
-        const url = reply._context ? reply._context.config.url : reply.context.config.url;
+        // NOTE: v4/v3 uses context https://github.com/fastify/fastify/blob/2.x/test/handler-context.test.js#L41
+        const url = reply.context.config.url;
```
kirrg001, line 62: bug: we still support v1. we only drop testing against v1

kirrg001 (changes requested):

package.json
```diff
@@ -130,9 +130,8 @@
     "express": "^4.17.1",
     "express-pino-logger": "^6.0.0",
     "express-winston": "^4.2.0",
-    "fastify": "1.14.6",
-    "fastify2": "npm:fastify@2.15.3",
-    "fastify3": "npm:fastify@3.21.5",
+    "fastify": "4.24.3",
```
kirrg001, line 133: Can we add the caret please? I don't remember why the 1.14 version wasn't making use of the caret 🤔 Maybe it was a mistake

PR #922, opened 2023-11-09
fix(couchbase): added missing bucket type and name for bucket.query

kirrg001:
no issue

Reported internally.


aryamohanan (approved):
LGTM

PR #923, opened 2023-11-09
chore(package.json): added the missing carets in package.json

athirasomanath:
This PR is really regarding the dev dependencies. To maintain consistency across devdependencies, I have added caret for all missing packages. 

refs: [120539](https://instana.kanbanize.com/ctrl_board/56/cards/120539/details/)

(Suggesting and open for further discussion regarding the dependencies. Although it needs digging into the package history to make a decision to keep the dependency versions with or without caret. )


kirrg001:
FYI there is an existing issue which you can refer to: 120539

basti1302 (approved):
LGTM, thanks!

kirrg001 (approved):
👍 

We can mark the consistency subtask as done, but leave the ticket open regarding this idea. Move back to backlog.

> using a caret ^ for development dependencies does not give any transparency for us which latest minor version we are testing against, because only the lock file get's updated and not the package.json

athirasomanath:
Reverted the caret for "@elastic/elasticsearch-v7.9.0" as it is failing for node v10

aryamohanan (approved):

PR #931, opened 2023-11-15
feat(sqs-consumer): support for sqs-consumer v7 -test branch

athirasomanath:
V7 has made node v18.X as minimum required version. Removed the testing for V6, since the instrumentation however remain the same and V7 still support aws sdk v3

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,7 +201,6 @@
     "socket.io": "^4.6.0",
     "socket.io-client": "^4.6.0",
     "sqs-consumer": "5.7.0",
-    "sqs-consumer-v6": "npm:sqs-consumer@^6.2.0",
     "sqs-consumer-v7": "npm:sqs-consumer@^7.4.0",
     "superagent": "^6.1.0",
     "tsoa": "^3.14.1",
--- a/package.json
+++ b/package.json
@@ -202,6 +202,7 @@
     "socket.io-client": "^4.6.0",
     "sqs-consumer": "5.7.0",
     "sqs-consumer-v6": "npm:sqs-consumer@^6.2.0",
+    "sqs-consumer-v7": "npm:sqs-consumer@^7.4.0",
     "superagent": "^6.1.0",
     "tsoa": "^3.14.1",
     "typescript": "^5.0.4",
--- a/packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
+++ b/packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
@@ -586,9 +586,9 @@ function start(version) {
       if (withError !== 'publisher') {
         const sqsEntry = verifySQSEntry({ receiverControls, spans, parent: sqsExit, withError, isBatch });
         /**
-       * When receiving messages in batch, SQS may not send all messages in one batch and we cannot guarantee that all 
-       * messages will be in the batch. The expectation can vary for the no:of http exit spans sent per message  received.
-       * So we are expectign atleast one matching for http exit spans.
+       * When receiving messages in batch, SQS may not send all messages in one batch and we cannot guarantee that all
+       * messages will be in the batch. The expectation can vary for the no:of http exit spans sent per message
+       * received. So we are expectign atleast one matching for http exit spans.
        */
 
         if (isSQSConsumer) {
--- a/packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
+++ b/packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
@@ -45,7 +45,7 @@ function start(version) {
     mochaSuiteFn = describe;
   }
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -201,7 +201,7 @@
     "socket.io": "^4.6.0",
     "socket.io-client": "^4.6.0",
     "sqs-consumer": "5.7.0",
-    "sqs-consumer-v6": "npm:sqs-consumer@^6.2.0",
+    "sqs-consumer-v7": "npm:sqs-consumer@^7.4.0",
```
kirrg001, line 204: Question: We are testing aws-sdk v2 with the `sqs-consumer` package (see the test in aws-sdk v2 sqs/sqs-consumer). I thought sqs-consumer v7 is using aws-sdk v3 under the hood.

kirrg001: Ignore the question. I thought we are only installing one sqs-consumer version now.
The packages needed to be switched though 👍

packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
```diff
@@ -184,7 +183,7 @@ function start(version) {
       });
 
       // See https://github.com/bbc/sqs-consumer/issues/356
-      if (version !== '@aws-sdk/client-sqs' && semver.gte(process.versions.node, '14.0.0')) {
+      if (version !== '@aws-sdk/client-sqs' && semver.gte(process.versions.node, '18.0.0')) {
```
kirrg001, line 186: Question: Is the issue still present? Can we remove the condition and check?

packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/test_definition.js
```diff
@@ -586,14 +585,19 @@ function start(version) {
...
+       */
 
         if (isSQSConsumer) {
           verifyHttpExit({
             spans,
             parent: sqsEntry,
             pid: String(receiverControls.getPid()),
             testMethod: (exitSpans, tests) => {
-              return expectExactlyNMatching(exitSpans, 4, tests);
+              return expectAtLeastOneMatching(exitSpans, tests);
```
kirrg001, line 600: Question: I don't understand this change. Can you please explain? :)

athirasomanath: We are verifying the http exit spans for matching cases. In case of batch message, we are sending 4 messages as a batch and often it is received as 2 batches with 2 messages each ( SQS may not send all messages in one batch ). For each message received we are sending an http exit span. In such cases, for the first 2 messages in the first batch, we send 2 http exit spans with same  parent id. But for the second batch of 2 messages, the span id differs and hence is the parent id for the http exit spans. But we were expecting for 4 matching spans which returns false

kirrg001: Okay and why is the behavior "suddenly" different?

athirasomanath: I noticed it as flaky.Only occurring in local so far. Often passing in local too. Unsure, if the polling time from local has something to do with it

kirrg001: Maybe there is a way to make it more stable?

athirasomanath: I looked up for an attribute to set a min number of messages in sqs level. But it is unavailable. So i changed in the expectation level. Any suggestions/data points I might have overlooked? @kirrg001

kirrg001: Can't we wait till we have four exit spans and then check expectExactlyNMatching(exitSpans, 4, tests)?

package.json
```diff
@@ -200,8 +200,8 @@
     "sinon-chai": "^3.7.0",
     "socket.io": "^4.6.0",
     "socket.io-client": "^4.6.0",
-    "sqs-consumer": "5.7.0",
-    "sqs-consumer-v6": "npm:sqs-consumer@^6.2.0",
+    "sqs-consumer": "^7.4.0",
+    "sqs-consumer-v5": "npm:sqs-consumer@^5.7.0",
```
kirrg001, line 204: Question: Why do we test against v7 and v5? 🤔

kirrg001: I don't mind testing against v5 and v7 if there is a reason 👍

athirasomanath: v5 is using aws sdk v2 and sqs v7 is using aws sdk v3

kirrg001: Okay thx 👌 

I am wondering if we should drop support for aws sdk v2.

> NOTE: The AWS SDK for JavaScript (v2) will be put into maintenance mode in 2023.

PR #933, opened 2023-11-15
(v3) chore: dropped node 10/12

kirrg001:
BREAKING CHANGE
Dropped Node v10 & v12 support.


packages/aws-lambda/test/infra-mock/index.js
```diff
@@ -35,7 +35,7 @@ const legacySensorMode = process.env.LEGACY_SENSOR != null;
 const anotherLambda = process.env.ANOTHER != null;
 const onlyLatest = process.env.ONLY_LATEST;
 
-const name = process.env.LAMBDA_FUNCTION_NAME || 'wrapped_async_8_10';
+const name = process.env.LAMBDA_FUNCTION_NAME || 'wrapped-async';
```
kirrg001, line 38: FYI: `wrapped_async_8_10` did no longer exist.

packages/core/src/tracing/instrumentation/messaging/amqp.js
```diff
@@ -153,7 +153,6 @@ function instrumentedDispatchMessage(ctx, originalDispatchMessage, originalArgs)
   const consumerTag = fields.consumerTag;
   let consumer;
 
-  // 0.10.x
   // https://github.com/amqp-node/amqplib/commit/d483493e6825ed7d770e68f4cec60979d0feeae7
```
kirrg001, line 156: NOTE: I was not able to remove the Node v10 check. The comment was wrong.

packages/aws-fargate/src/preactivate.js
```diff
@@ -11,7 +11,8 @@ if (isNodeJsTooOld()) {
   // eslint-disable-next-line no-console
   console.error(
     `The package @instana/aws-fargate requires at least Node.js ${minimumNodeJsVersion} but this process is ` +
-      `running on Node.js ${process.version}. This Fargate container will not be monitored by Instana.`
+      `running on Node.js ${process.version}. This Fargate container will not be monitored by Instana.` +
+      'See https://www.ibm.com/docs/en/instana-observability/current?topic=agents-aws-fargate#versioning.'
```
basti1302, line 15: It is a good idea to link there 👍  Maybe we should also add a table to that section of the docs that explains which image version can be used for which Node.js version?

packages/aws-lambda/lambdas/README.md
```diff
@@ -28,7 +28,7 @@ The AWS Lambda demo is comprised of all Lambda functions in this folder prefixed
 
 ### Initial Setup
```
basti1302, line 29: This is out of scope for this PR, but to be honest, I think we should delete everything about the Lambda demo. It has been moved over to the Instana demo repository ages ago and we will never need to deploy the Lambda demo from the sources in this repo, ever again.

packages/aws-lambda/lambdas/plain-ec2-app/setup.log
```diff
@@ -4,7 +4,7 @@ change:
   mode = APM
```
basti1302, line 4: Also out of scope, but the plain-ec2-app can be deleted as well, as it was only relevant for the demo.

packages/core/src/tracing/instrumentation/messaging/amqp.js
basti1302: I don't understand this comment. I think the `0.10.x` comment referred to the version number of the amqplib package, not to the Node.js version?

packages/aws-fargate/images/instana-aws-fargate/Dockerfile-npm
```diff
@@ -1,23 +1,6 @@
-FROM node:12.22.7-buster AS instana-aws-fargate-build-nodejs
+FROM node:18-buster AS instana-aws-fargate-build-nodejs
```
basti1302, line 1: _thought:_ The tag `18-buster` will be updated everytime a new 18.x version comes out. That is, currently it refers to the same image as the `18.18.2-buster` tag. When Node.js 18.18.3 or 18.19.0 is released, it will point to that newer image. That means the outcome of our image build depends on the point in time when the build is executed. To have a reproducible build that does not depend on the time of execution, we might want to use a full version tag like 18.18.2-buster here. WDYT?

The same argument could be made for the Google cloud run image.

basti1302:
I found two more small things that we could remove/update:

* we can remove https://github.com/instana/nodejs/blob/drop/v10v12/packages/collector/test/tracing/protocols/graphql/test_definition.js#L26-L30 as well
* not directly related, but we could update https://github.com/instana/nodejs/blob/drop/v10v12/packages/collector/src/util/eol.js#L17 to  `return satisfies(process.versions.node, '<18 || 19');` (and update the related test, https://github.com/instana/nodejs/blob/drop/v10v12/packages/collector/test/util/isNodeVersionEOL_test.js)

packages/core/src/tracing/instrumentation/messaging/amqp.js
kirrg001: hahahah yes you are right :D

native-dep-packs/rebuild-precompiled-addons.sh
```diff
@@ -11,9 +11,7 @@ cd `dirname $BASH_SOURCE`
 # Maintenance Note:
 # This should be kept in sync with packages/autoprofile/precompile/build-all-addons.js-> ABI_VERSIONS.
 declare -A ABI_VERSIONS=( \
-  ["64"]="10.24.1" \
-  ["72"]="12.22.1" \
-  ["83"]="14.17.0" \
+  ["83"]="14.21.3" \
   ["88"]="15.14.0" \
```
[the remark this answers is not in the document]
kirrg001: Let's do it. I was wondering about this anyway. Perfect 🦖

misc/is_eol.js
```diff
@@ -7,10 +7,10 @@
 'use strict';
 
 /**
- * Exits with exit 1 if running on Node.js >= 12.x.y, otherwise with exit code 0.
- * That is, for EOL versions (10, 8, 6, ...), this script will pass and for non-EOL versions it will fail.
+ * See https://github.com/nodejs/Release
+ * Exits with exit 1 if running on Node.js > latest EOL version, otherwise with exit code 0.
  */
-if (parseInt(/v(\d+)\./.exec(process.version)[1], 10) >= 12) {
+if (parseInt(/v(\d+)\./.exec(process.version)[1], 10) > 16) {
```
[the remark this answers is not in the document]
kirrg001: Ok 👍

native-dep-packs/rebuild-precompiled-addons.sh
```diff
@@ -11,9 +11,7 @@ cd `dirname $BASH_SOURCE`
...
   ["93"]="16.3.0" \
   ["108"]="18.0.0" \
```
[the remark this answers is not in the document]
kirrg001: Yes will do. I was thinking about this while refactoring and then forgot 👍

packages/aws-lambda/lambdas/README.md
kirrg001: I can remove the lambda demo as part of the v3 branch 👍

native-dep-packs/rebuild-precompiled-addons.sh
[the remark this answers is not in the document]
basti1302: Yeah, not entirely sure either. My guess is that at the time we added this, Node 14 was the most recent LTS and Node 15 was already available with a new ABI? Not sure.

packages/aws-fargate/src/preactivate.js
kirrg001: Yes we will update the docs as part of the v3 release 👍

packages/aws-fargate/images/instana-aws-fargate/Dockerfile-npm
kirrg001: Yeah I know, I thought. it doesn't really matter 🤔 Will concourse cache docker images at all? Or bootstrap a new fresh instance on each build?

dockerfile-examples/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3'
 
 services:
   node:
-    image: node-12.18.3-instana
+    image: node-16.13.0-instana
```
aryamohanan, line 5: Qs: any particular reason to use this version here?

aryamohanan (approved):
Great work 💯

packages/aws-fargate/images/instana-aws-fargate/Dockerfile-npm
kirrg001: I will change it for the production use case.

packages/aws-fargate/images/test-images/utils
```diff
@@ -19,16 +19,14 @@ function normalizeArgs {
...
-    NODEJS_VERSION=12.16.3
-  elif [[ $nodeJsVersion = 10 ]]; then
-    NODEJS_VERSION=10.20.1
-  elif [[ $nodeJsVersion = 8 ]]; then
-    NODEJS_VERSION=8.17.0
-  else
-    NODEJS_VERSION=$nodeJsVersion
+  if [[ -z "${nodeJsVersion-}" ]]; then
+    NODEJS_VERSION=18
+  elif [[ $nodeJsVersion -lt 14 ]]; then
```
aryamohanan, line 24: I think you missed the else part 
```suggestion
  else
    NODEJS_VERSION=$nodeJsVersion
```

packages/google-cloud-run/images/test-images/utils
```diff
@@ -18,14 +18,12 @@ function normalizeArgs {
...
-  elif [[ $nodeJsVersion = 10 ]]; then
-    NODEJS_VERSION=10.20.1
-  elif [[ $nodeJsVersion = 8 ]]; then
-    NODEJS_VERSION=8.17.0
-  else
-    NODEJS_VERSION=$nodeJsVersion
+  if [[ -z "${nodeJsVersion-}" ]]; then
+    NODEJS_VERSION=18
+    LAMBDA_BASE_IMAGE=public.ecr.aws/lambda/nodejs:18
+  elif [[ $nodeJsVersion -lt 14 ]]; then
```
aryamohanan, line 24: Missed the else part
```suggestion
  else
    NODEJS_VERSION=$nodeJsVersion
```

kirrg001: You are right. Thank you :)

PR #938, opened 2023-11-17
fix: added missing prebuild for gcstats Node v20

kirrg001:
[ci skip]

I noticed that this prebuild was missing.
User facing change? Yeah I think so.


basti1302:
> User facing change? Yeah I think so.
+1

basti1302 (approved):

aryamohanan (approved):

PR #941, opened 2023-11-18
feat(sqs-consumer): added v7 support in sqs-consumer

athirasomanath:
V7 has made node v18.X as minimum required version. Removed the testing for V6, since the instrumentation however remain the same and V7 still support aws sdk v3. keeping the testing for v5 as is it using aws sdk v2


kirrg001 (approved):
👍

PR #944, opened 2023-11-22
(v3) dropped EOL libraries

kirrg001:
- We cannot drop [@apollo/federation](https://www.npmjs.com/package/@apollo/federation) yet, because we have not implemented the new package. See card on board.
- Cannot move graphql v16 to latest, because apollo-express-server requires graphql v15. Installation will fail.
- Deprecated kafka-node. No releases since 4y, but high usage still on NPM trends.

I have added `BREAKING CHANGE` to the commit messages. Please review commit per commit.
Therefor these commits will be appear in the changelog automatically.


aryamohanan (approved):
Great work 💯

package.json
```diff
@@ -135,7 +133,6 @@
     "got": "^11.8.5",
     "graphql": "^15.8.0",
     "graphql-subscriptions": "^2.0.0",
-    "graphql-v14": "npm:graphql@14.7.0",
     "graphql-v16": "npm:graphql@16.6.0",
```
basti1302, line 136: _question:_ Not really related to your PR but for most libs I think we use the unqualified package name for the most recent version, for some reason for graphql we use `graphql` for 15 and graphql-v16 for 16. I wonder if we would want to switch that around. But not really important I guess.

basti1302 (approved):
🎉

package.json
kirrg001: Yeah a little inconsistent. I want to update some libs versions on Monday. Will take a look too

PR #945, opened 2023-11-23
fix(fetch): fix header handling for native fetch for Node.js >= 20.10.0

basti1302:
refs 146305


basti1302:
The release to fix the fetch header handling has been released with 20.10.0 yesterday. There is no fix for the 18.x release line available yet.

kirrg001 (approved):

PR #884, opened 2023-10-09
v3

kirrg001:
- [ ] Drop Node 10 & Node 12
- [ ] Lerna v7 update
- [ ] Lerna v7 Node 14 fix

PR #948, opened 2023-11-27
(v3) more package updates

kirrg001:
Please review the PR by going through each commit.


package.json
```diff
@@ -56,13 +56,13 @@
     "@apollo/federation": "0.38.1",
     "@apollo/gateway": "0.50.2",
     "@aws-sdk/client-dynamodb": "3.415.0",
-    "@aws-sdk/client-dynamodb2": "npm:@aws-sdk/client-dynamodb@3.20.0",
+    "@aws-sdk/client-dynamodb-v3": "npm:@aws-sdk/client-dynamodb@3.20.0",
     "@aws-sdk/client-kinesis": "^3.379.1",
     "@aws-sdk/client-s3": "3.379.1",
-    "@aws-sdk/client-s3-2": "npm:@aws-sdk/client-s3@3.321.1",
+    "@aws-sdk/client-s3-v3": "npm:@aws-sdk/client-s3@3.321.1",
```
basti1302, line 62: _nitpick:_ The other version is also v3 (3.x.x) though. Just different minor versions of 3.x.

basti1302 (approved):

basti1302:
@kirrg001 

image

I think the commit message style you have used for v3 might become a bit of an issue with respect to changelog generation? Or at least, it is different than what we used for v2. 

This was a commit message with a breaking change for v2:
image

And this is how that looked in the changelog:

image

Differences:
* For v2 we used the format "BREAKING CHANGE: ..." with a colon and then the description of the change in the same line, for v3 it looks like we mostly use BREAKING CHANGE without colon, and then a newline. Not entirely sure if that is relevant, but https://www.conventionalcommits.org/en/v1.0.0/ reads as if the colon needs to be there?
* It looks like some (most? all?) commit messages only list a some reasoning _why_ we dropped support for something in the BREAKING CHANGE footer. If I am not mistaken, the content of the BREAKING CHANGE footer will be put in a separate section in the changelog, without any reference to the full commit message.

So the last part would mean we generate this list of breaking changes for the 3.0.0 changelog:
* ...
* v1 is from 2019
* ...

But the "v1 is from 2019" would not have any indication that it relates to fastify I think?

package.json
kirrg001: See https://github.com/instana/nodejs/commit/3936a9da70752102e90cf3721e63bd83dce92151
This was the bigger smithy client refactoring from aws sdk.

basti1302: Sorry, I didn't mean to imply we should only test one particular v3 version. I was talking about the label we use for that additional version in the package.json file. Since both are v3, labeling one of the with the suffix "-v3" is a bit misleading/confusing. But whatever... it is only a label that we use internally.

kirrg001:
> TL;DR: I think this particular commit message should have been

Good catch 👍 I remember that we also did the mistake in the beginning of v2. I will rewrite the breaking change messages before merging v3 👍

package.json
kirrg001: Ah okay. Any better idea? This label follows our pattern of `-vN`.

kirrg001:
Added a task to #884 

>  Rewrite breaking changes messages

PR #954, opened 2023-12-04
feat: added azure blob instrumentation

athirasomanath:


packages/core/src/tracing/instrumentation/cloud/azure/blob.js
```diff
@@ -0,0 +1,107 @@
...
+const shimmer = require('../../../shimmer');
+
+const requireHook = require('../../../../util/requireHook');
+const tracingUtil = require('../../../tracingUtil');
+const constants = require('../../../constants');
+const cls = require('../../../cls');
+
+let isActive = false;
+
+exports.spanName = 'az_storage';
```
kirrg001, line 17: Discussed: We want to use `az.storage`.

PR #971, opened 2023-12-18
fix: updated code formats

kirrg001:
`npm run prettier`


aryamohanan (approved):
Thanks for fixing this ❤️

PR #974, opened 2023-12-18
fix: ignored opentelemetry spans which are not traces by us

kirrg001:
- install any opentelemetry instrumentation inside the client / customer appliation + use our Instana collector
- it will add Otel spans for this instrumentation although we do not support it as part of our integration
- found this issue in https://github.com/instana/nodejs/pull/967

### DIFF
--- a/packages/core/src/tracing/opentelemetry-instrumentations/wrap.js
+++ b/packages/core/src/tracing/opentelemetry-instrumentations/wrap.js
@@ -53,6 +53,10 @@ module.exports.init = (_config, cls) => {
       if (targetInstrumentationModule.transform) {
         otelSpan = targetInstrumentationModule.transform(otelSpan);
       }
+    } else {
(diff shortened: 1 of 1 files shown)


kirrg001:
See https://github.com/instana/nodejs/pull/967/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519R8415

kirrg001:
Azure SDK always installs a package called azure tracing.
https://www.npmjs.com/package/@azure/core-tracing
And our Otel integration needs to ignore all spans which were not supported by our set of Otel instrumentations.

We should also check in #967 if the azure tracing package is somehow colliding with our Instana tracing.
@athirasomanath Could you please add a TODO to your PR? Thanks

basti1302 (approved):

aryamohanan (approved):

PR #975, opened 2023-12-18
feat(aws-lambda): added support for Node.js 20 AWS Lambda runtime

aryamohanan:
Node.js 20.x runtime now available in AWS Lambda, [reference](**url**)


kirrg001 (approved):
👍 

Commit could be merged as

> feat(aws-lambda): added support for Node.js v20 runtime #975

Please add ticket reference to the commit.

PR #989, opened 2024-01-05
chore: updated the contributors list

aryamohanan:


CONTRIBUTING.md
```diff
@@ -59,8 +59,6 @@ This is an open source project, and we appreciate your help!
 
 In order to clarify the intellectual property license granted with contributions from any person or entity, a Contributor License Agreement ("CLA") must be on file that has been signed by each contributor, indicating agreement to the license terms below. This license is for your protection as a contributor as well as the protection of Instana and its customers; it does not change your rights to use your own contributions for any other purpose.
 
-Please print, fill out, and sign the [contributor license agreement](https://github.com/instana/nodejs/raw/main/misc/instana-nodejs-cla-individual.pdf). Once completed, please scan the document as a PDF file and email to the following email address: bastian.krol@instana.com.
```
aryamohanan, line 62: I don't think it's necessary to have this on file any longer because Bastian has left the company, but I'm sure we need to update this part in the future.

kirrg001: There is a slack thread which currently discusses this. IMO we should remove this and replace it in a separate PR.

aryamohanan: I removed the chnage fro the PR

kirrg001 (approved):

PR #1004, opened 2024-01-16
fix: deprecated request module

aryamohanan:
- request module deprecated and replacing request with node-fetch

The request module has been officially deprecated. Consequently, we are substituting the request module with the node-fetch module, a lightweight and widely used package. Additionally, Node.js has a native fetch module available from version 18 onwards. Therefore, once we cease support for Node.js versions 14 and 16, we can seamlessly transition to using the fetch module from node-fetch.

### DIFF
--- a/packages/collector/test/tracing/misc/require_hook/app.js
+++ b/packages/collector/test/tracing/misc/require_hook/app.js
@@ -27,11 +27,17 @@ app.get('/', (req, res) => {
   res.sendStatus(200);
 });
 
-app.get('/multipleRequestsWithStealthyRequire', async (req, res) => {
-  await makeSuperagentRequest('https://example.com/1');
-  await makeSuperagentRequest('https://example.com/2');
+app.get('/multipleRequireWithStealthyRequire', async (req, res) => {
+  // Wrap the require calls with stealthyRequire to avoid caching
+  const firstInstanceOfGot = stealthyRequire(require.cache, () => require('got'));
+  const secondInstanceOfGot = stealthyRequire(require.cache, () => require('got'));
 
-  res.sendStatus(200);
+  // Check if the two instances of 'got' are different
+  if (firstInstanceOfGot !== secondInstanceOfGot) {
+    res.sendStatus(200);
+  } else {
+    res.sendStatus(500);
+  }
 });
 
 app.listen(port, () => {
@@ -42,13 +48,4 @@ function log() {
   const args = Array.prototype.slice.call(arguments);
   args[0] = logPrefix + args[0];
   console.log.apply(console, args);
-}
-async function makeSuperagentRequest(url) {
-  const superagentFresh = stealthyRequire(require.cache, () => require('superagent'));
-  try {
-    const response = await superagentFresh.get(url);
-    log(`response ${response.body}`);
-  } catch (error) {
-    log(`error ${error}`);
-  }
 }
--- a/packages/collector/test/tracing/misc/require_hook/app.js
+++ b/packages/collector/test/tracing/misc/require_hook/app.js
@@ -12,6 +12,7 @@ require('../../../..')();
 const bodyParser = require('body-parser');
 const express = require('express');
 const morgan = require('morgan');
+const fetch = require('node-fetch');
 const port = require('../../../test_util/app-port')();
 const app = express();
 const logPrefix = `requireHook App (${process.pid}):\t`;
@@ -26,11 +27,16 @@ app.get('/', (req, res) => {
   res.sendStatus(200);
 });
 
-app.get('/requireRequestPromiseMultipleTimes', (req, res) => {
-  require('request');
-  require('request-promise'); // executes stealthy-require
-  require('request-promise-native'); // executes stealthy-require
-  res.sendStatus(200);
+app.get('/requireRequestPromiseMultipleTimes', async (req, res) => {
+  try {
+    await fetch('https://example.com/test1');
+    await fetch('https://example.com/test2');
+    await fetch('https://example.com/test3');
+    res.sendStatus(200);
+  } catch (error) {
+    console.error(error);
+    res.sendStatus(500);
+  }
 });
 
 app.listen(port, () => {
--- a/packages/collector/test/tracing/misc/require_hook/app.js
+++ b/packages/collector/test/tracing/misc/require_hook/app.js
@@ -32,7 +32,11 @@ app.get('/multipleRequireWithStealthyRequire', async (req, res) => {
   const firstInstanceOfGot = stealthyRequire(require.cache, () => require('got'));
   const secondInstanceOfGot = stealthyRequire(require.cache, () => require('got'));
 
-  // Check if the two instances of 'got' are different
+  // Verify that the two instances of the 'got' module are different. In this scenario, we emulate the behavior of
+  // stealthy-require, ensuring that each time we load the module, it is a fresh instance. This precaution is taken
+  // to prevent require caching, and to guarantee the loading of a new module each time, as discussed and resolved
+  // in the following PR: https://github.com/instana/nodejs/pull/71
+
   if (firstInstanceOfGot !== secondInstanceOfGot) {
     res.sendStatus(200);
   } else {
--- a/packages/collector/test/tracing/misc/require_hook/app.js
+++ b/packages/collector/test/tracing/misc/require_hook/app.js
@@ -12,10 +12,10 @@ require('../../../..')();
 const bodyParser = require('body-parser');
 const express = require('express');
 const morgan = require('morgan');
-const fetch = require('node-fetch');
 const port = require('../../../test_util/app-port')();
 const app = express();
 const logPrefix = `requireHook App (${process.pid}):\t`;
+const stealthyRequire = require('stealthy-require');
 
 if (process.env.WITH_STDOUT) {
   app.use(morgan(`${logPrefix}:method :url :status`));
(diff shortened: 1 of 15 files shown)


packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -19,17 +19,24 @@ const {
...
 exports.init = function init(config) {
   instrument(coreHttpModule, false);
   instrument(coreHttpsModule, true);
-
   extraHttpHeadersToCapture = config.tracing.http.extraHttpHeadersToCapture;
+  requireHook.onModuleLoad('request', logDeprecatedWarning);
 };
-
+function logDeprecatedWarning() {
+  logger.warn('The support for request has been deprecated.');
```
kirrg001, line 38: Suggestion: One unified message format for deprecation.

> logger.warn('kafka-node is deprecated. The support will be removed in the next major release.');
> logger.warn('The support for request has been deprecated.');
>   logger.warn(
    'The support for @apollo-federation has been deprecated, following its official end-of-life in September 2023.'
  );

packages/collector/test/tracing/misc/require_hook/app.js
```diff
@@ -26,11 +27,16 @@ app.get('/', (req, res) => {
   res.sendStatus(200);
 });
 
-app.get('/requireRequestPromiseMultipleTimes', (req, res) => {
-  require('request');
-  require('request-promise'); // executes stealthy-require
-  require('request-promise-native'); // executes stealthy-require
```
kirrg001, line 32: _issue_: This is a very special test use case. See https://github.com/instana/nodejs/pull/71/files
I doubt your code simulates the underlying problem with stealthy-require, because node-fetch doesn't make use of the require cache. Either we remove the whole test or we need to find a way to simulate the problem on our selfs (preferred solution). I'd suggest to use stealthy-require directly in the test. You just need to choose a library you want to test the behaviour with.

packages/collector/test/tracing/protocols/http/proxy/expressProxy.js
```diff
@@ -40,38 +39,17 @@ app.use((req, res) => {
       url = `http://localhost:${process.env.UPSTREAM_PORT}/proxy-call${req.url}`;
     }
 
-    if (req.query.httpLib === 'node-fetch') {
```
kirrg001, line 43: `httpLib` can be removed from packages/collector/test/tracing/protocols/http/proxy/expressProxyControls.js#77

packages/collector/test/tracing/protocols/http/proxy/test.js
```diff
@@ -187,15 +186,14 @@ mochaSuiteFn('http with proxy', function () {
...
+                span => expect(span.n).to.equal('node.http.server'),
                 span => expect(span.error).to.not.exist,
                 span => expect(span.ec).to.equal(1),
                 span => expect(span.f.e).to.equal(String(expressProxyControls.getPid())),
                 span => expect(span.f.h).to.equal('agent-stub-uuid'),
                 span => expect(span.async).to.not.exist,
-                span => expect(span.data.http.error).to.be.a('string'),
                 span => expect(span.data.http.method).to.equal('POST'),
-                span => expect(span.data.http.url).to.equal('http://10.123.456.555:49162/foobar')
+                span => expect(span.data.http.url).to.equal('/callNonExistingTarget')
```
kirrg001, line 196: q: Why?

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -19,17 +19,24 @@ const {
...
 
 let extraHttpHeadersToCapture;
 let isActive = false;
 
 exports.init = function init(config) {
   instrument(coreHttpModule, false);
   instrument(coreHttpsModule, true);
-
   extraHttpHeadersToCapture = config.tracing.http.extraHttpHeadersToCapture;
+  requireHook.onModuleLoad('request', logDeprecatedWarning);
```
kirrg001, line 35: Suggestion: Maybe we should ensure that we only print the warning once? The `request` module is used in lot's of sub dependencies.

packages/core/test/util/initializedTooLateHeuristic_test.js
```diff
@@ -73,7 +73,7 @@ describe('[UNIT] util.initializedTooLateHeurstic', function () {
       }
 
       require.cache[resolvedPath] = {};
-      const exclude = ['bluebird', 'bunyan', 'winston'];
+      const exclude = ['bluebird', 'bunyan', 'winston', 'request'];
```
kirrg001, line 76: Q: Why?

kirrg001 (changes requested):
Found some issues.

packages/collector/test/tracing/protocols/http/proxy/test.js
```diff
@@ -187,15 +186,14 @@ mochaSuiteFn('http with proxy', function () {
           retry(() =>
             agentControls.getSpans().then(spans => {
               expectAtLeastOneMatching(spans, [
-                span => expect(span.n).to.equal('node.http.client'),
+                span => expect(span.n).to.equal('node.http.server'),
```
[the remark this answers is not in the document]
aryamohanan: The handling of invalid URLs differs between `request` and `node-fetch`. `request` throws an `ENOTFOUND` error, indicating its failure to locate the specified hostname. This results in the creation of a parent entry span 'node.http.server' and a corresponding exit span 'node.http.client'. On the other hand, `node-fetch` throws a `TypeError` with the code `ERR_INVALID_URL`. Consequently, when using `node-fetch`, only the entry span is reported, and there is no corresponding exit span.  From my understanding this is how it works, For a similar case, please see, https://github.com/instana/nodejs/pull/1004/files#diff-a51bdbe8abc67e6baf4a9b0c234c9bddba0e6d22b724b45d8fe8fd5d899eaaa0R202.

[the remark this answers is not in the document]
kirrg001: Thanks! What is the difference between both test cases?
Why do we not check against the error message? Why did you remove the line?

packages/core/test/util/initializedTooLateHeuristic_test.js
aryamohanan: I invoked `requireHook` on the `onModuleLoad` function for the `request` module to include the deprecated warning. As a result, this module has already been added to the list. Since this list isused in the current test case, it was necessary to exclude the `request` module from the list; otherwise, the test would fail.

packages/collector/test/tracing/protocols/http/proxy/test.js
kirrg001: Ignore this comment. See other [conversation](https://github.com/instana/nodejs/pull/1004#discussion_r1455367171).

packages/core/test/util/initializedTooLateHeuristic_test.js
aryamohanan: I mean here https://github.com/instana/nodejs/pull/1004/files#diff-93014d2f82f6d0b77f9edd3762ab8828fa57ff97edf9dd7a0ca114323df081eb

packages/collector/test/tracing/protocols/http/proxy/test.js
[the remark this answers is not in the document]
aryamohanan: For node-fetch error property is not there.

packages/core/test/util/initializedTooLateHeuristic_test.js
kirrg001: Discussed offline: please add the request module to the patterns of the utility 👍

packages/core/src/tracing/instrumentation/protocols/httpClient.js
aryamohanan: Would it be possible to consider incorporating the following log messages for deprecation warnings?

[Deprecation Warning]: The support for request has been deprecated.
[Deprecation Warning]: The support for @apollo-federation has been deprecated.

packages/collector/test/tracing/protocols/http/proxy/test.js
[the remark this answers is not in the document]
aryamohanan: For the current test cases, no error message captured for httpServer spans.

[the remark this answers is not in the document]
kirrg001: Okay then let's ignore the error message 👍

packages/collector/test/tracing/misc/require_hook/app.js
aryamohanan: update, could please check, I used superagent libarary to test

kirrg001: If you comment out https://github.com/instana/nodejs/pull/71/files#diff-107040a34b1332891e461280fb3a19011b4a0dbbbb5e8cca84e657ad28768769R44, the test should fail. Does it?

packages/core/src/tracing/instrumentation/protocols/httpClient.js
kirrg001: Two more things:
- maybe we can add the link to the target npm package?
- What about "The support will be removed in the next major release"?

packages/core/src/tracing/instrumentation/protocols/httpClient.js
kirrg001: In app.js:

> require('request') -> uncached
> require('request') -> cached

This should trigger the warning twice.

packages/collector/test/tracing/misc/require_hook/app.js
kirrg001: Looks great 👍

packages/core/src/tracing/instrumentation/protocols/httpClient.js
aryamohanan: Verified and it logged once.

packages/core/src/tracing/instrumentation/protocols/httpClient.js
aryamohanan: >  logger.warn('[Deprecation Warning] The support for "request" is deprecated and will be removed in the next major release. Please consider migrating to an appropriate package.');
> 
> logger.warn('[Deprecation Warning] The support for "@apollo-federation" is deprecated and will be removed in the next major release. Please consider migrating to an appropriate package.');
> 
> logger.warn('[Deprecation Warning] The "kafka-node" library is deprecated and will be removed in the next major release. Please consider migrating to an appropriate package.');

could you please check these updates

kirrg001 (approved):
👍

PR #1006, opened 2024-01-17
test: experimental parallel run

kirrg001:


kirrg001:
Closing. This is not an acceptable approach. Too slow & too error prone. We should aim for sub tests tasks.

Such as
- `npm run test:ci:databases`
- `npm run test:ci:frameworks`

Then we can run these tasks in parallel.
We can even skip specific groups when nothing was modified.

kirrg001:
Furthermore: [reporters do not work](https://github.com/michaelleeallen/mocha-junit-reporter/issues/140) with parallel modus.

PR #1016, opened 2024-01-23
fix: deprecated request-promise-native

aryamohanan:
The `request-promise-native` module is now deprecated as it relies on the deprecated `request` package. Please refer to the following link for more information: https://github.com/request/request/issues/3142

In light of this deprecation, we are discontinuing the use of the `request-promise-native` module and adopting the `node-fetch` module as an alternative. `node-fetch` is a lightweight and widely utilized package. It is important to note that Node.js has introduced a native fetch module starting from version 18. Consequently, once support for Node.js versions 14 and 16 is discontinued, we will seamlessly transition to utilizing the fetch module from `node-fetch`.

What's pending?

- [x] [PR](https://github.com/instana/nodejs/pull/1004)  needs to be merged first.

### DIFF
--- a/packages/collector/test/tracing/database/couchbase/app.mjs
+++ b/packages/collector/test/tracing/database/couchbase/app.mjs
@@ -10,7 +10,7 @@ import express from 'express';
 import { v1 } from 'uuid';
 import morgan from 'morgan';
 import requestPromise from 'request-promise';
-const fetch = require('node-fetch');
+import fetch from 'node-fetch';
 import portFactory from '../../../test_util/app-port.js';
 import testUtil from '../../../../../core/test/test_util/index.js';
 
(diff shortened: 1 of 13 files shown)


packages/collector/test/tracing/database/couchbase/app.mjs
```diff
@@ -10,7 +10,7 @@ import express from 'express';
 import { v1 } from 'uuid';
 import morgan from 'morgan';
 import requestPromise from 'request-promise';
-import request from 'request-promise-native';
+const fetch = require('node-fetch');
```
kirrg001, line 13: issue: This needs to be an import statement.

packages/collector/test/tracing/database/pg_native/app.mjs
```diff
@@ -10,7 +10,7 @@ const agentPort = process.env.INSTANA_AGENT_PORT;
 import Client from 'pg-native';
 import express from 'express';
 import morgan from 'morgan';
-import request from 'request-promise-native';
+const fetch = require('node-fetch');
```
kirrg001, line 13: And here :)

packages/collector/test/tracing/database/couchbase/app.mjs
aryamohanan: resolved

PR #1017, opened 2024-01-23
fix: depreacted request-promise module

aryamohanan:
The `request-promise` module is now deprecated as it relies on the deprecated request package. Please refer to the following link for more information: https://github.com/request/request/issues/3142

In light of this deprecation, we are discontinuing the use of the `request-promise` module and adopting the node-fetch module as an alternative. `node-fetch` is a lightweight and widely utilized package. It is important to note that Node.js has introduced a native fetch module starting from version 18. Consequently, once support for Node.js versions 14 and 16 is discontinued, we will seamlessly transition to utilizing the fetch module from `node-fetch`.

what's pending?

- [x]  Fix all unit test cases other than collector package
- [x] Fix all unit test cases for collector package
- [x] uninstall request-promise

Notable Changes

- request-promise response body is automatically parsed based on content type. With node-fetch, we need to manually parse the response according to the usage.

- request-promise  automatically converts basic data types like objects and arrays to JSON-formatted strings by default, but node-fetch not.

- node-fetch does not automatically serialize the body in the same way as request-promise, so according to the response header content-type, manually converted to the desired response type.

-  To address the certificate authorization issue with node-fetch, process.env.NODE_TLS_REJECT_UNAUTHORIZED was set to '0' in tests.

The changes are only in test files.

### DIFF
--- a/.tracer-support-matrix/configuration.yaml
+++ b/.tracer-support-matrix/configuration.yaml
@@ -322,20 +322,6 @@ additionalLibraries:
   link: https://nodejs.org/docs/latest/api/http2.html
   packageNameForLatestAvailableVersion: node
 
-- name: request
-  category: http
-  supportedSinceVersion: v1.10.0
-  support:
-    lowerBound: 2.74.0
-    upperBound: 2.88.2
-
-- name: request-promise
-  category: http
-  supportedSinceVersion: v1.10.0
-  support:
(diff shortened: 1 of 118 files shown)


kirrg001 (approved):
Great work!
As discussed, just moving the tls check to the mocharc 👍

PR #1026, opened 2024-02-01
build: resolved the failing unit test case for google-cloud/storage

aryamohanan:
It appears that the issue with the google-cloud/storage test failing for node versions 21.2.0 and above has been rectified.(https://github.com/googleapis/nodejs-storage/issues/2368).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,13 @@ Troubleshooting `pg_config: command not found`: The tests in this package depend
 
 Install the [`aws-cli`](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html) to publish AWS layers from local.
 
+### Upgrade npm for Node.js v14
+
+If you are using Node.js v14 and would like to take advantage of the "overrides" feature in package.json, you need to upgrade npm manually since Node.js v14 ships with npm v6, and "overrides" were introduced in npm v8.3. Upgrade npm manually to the latest version using the following command:
+```
+npm install npm@latest
+```
+
 ## Executing Tests Locally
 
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -209,7 +209,8 @@
     "pg-native": "^3.0.1"
   },
   "overrides": {
-    "nx": "15.8.9"
+    "nx": "15.8.9",
+    "stream-shift": "1.0.2"
```
kirrg001, line 213: We need to add a hint for Node v14 somewhere.
Node v14 automatically ships with NPM v6. But NPM v6 does not use NPM v8. Overrides came available in Node v8.3.

For CircleCI we already set the npm version for Node v14 to v8.
But locally we do not have a strategy yet.

kirrg001: Maybe we can simply document to run `nvm install-latest-npm` when switching the node version with nvm?
I could not see a way to automate this e.g. with a hook from nvm.

aryamohanan: That makes sense. I think it's better to put this information in our contributing.md since it's related to the local development environment issue. I'll include a short note about it.

kirrg001: Yes 👍

aryamohanan: I have added a short note.

CONTRIBUTING.md
```diff
@@ -18,6 +18,13 @@ Troubleshooting `pg_config: command not found`: The tests in this package depend
 
 Install the [`aws-cli`](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html) to publish AWS layers from local.
 
+### Upgrade npm for Node.js v14
+
+If you are using Node.js v14 and would like to take advantage of the "overrides" feature in package.json, you need to upgrade npm manually since Node.js v14 ships with npm v6, and `overrides` were introduced in npm v8.3. Upgrade npm manually to the latest version using the following command:
+```
+npm install npm@latest
```
kirrg001, line 25: Q: Does this work? I thought we need to use `npm install npm@8.x.x` or `nvm install-latest-npm`?

kirrg001: @aryamohanan

kirrg001: Perfect. Thanks for testing

kirrg001 (approved):

PR #1050, opened 2024-02-21
test: added fix for timeout issues with azure sql db

aryamohanan:


packages/collector/test/tracing/opentelemetry/test.js
```diff
@@ -25,8 +25,8 @@ const {
 const ProcessControls = require('../../test_util/ProcessControls');
 const globalAgent = require('../../globalAgent');
 const DELAY_TIMEOUT_IN_MS = 500;
-const testTimeout = Math.max(50000, config.getTestTimeout());
-const retryTime = 30000;
+const testTimeout = Math.max(70000, config.getTestTimeout());
+const retryTime = 40000;
```
kirrg001, line 29: IMO I think you can use retryTime (which is I think the frequency) to 10s
And then also pass "until": needs to be slower than the testTimeout to see errors.

kirrg001:
Feel free to just merge it whenever you think the issue is resolved.

kirrg001 (approved):

PR #1056, opened 2024-02-23
chore: improved docker compose usage

kirrg001:
Starting all containers eats too much resources and makes debugging harder.

refs 155406


aryamohanan (approved):
Awesome work 🎉

PR #1069, opened 2024-03-12
fix: included the matrix parameters in url

aryamohanan:
The truncation of variables after the semicolon (;) was made a few years ago in response to a reported [issue](https://github.com/instana/nodejs/issues/327). The purpose of this truncation is to address cases where credential  are present in the URL. So as per the request from the customer we are about to add the matrix parameters back in the url.

### DIFF
--- a/packages/collector/test/tracing/protocols/http/client/clientApp.js
+++ b/packages/collector/test/tracing/protocols/http/client/clientApp.js
@@ -287,6 +287,13 @@ app.post('/upload-s3', (req, res) => {
     }
   });
 });
+app.get('/matrix-params/:params', (req, res) => {
+  try {
+    res.sendStatus(200);
+  } catch (error) {
+    res.sendStatus(500);
+  }
+});
 
 function createUrl(req, urlPath) {
   const pathWithQuery = req.query.withQuery ? `${urlPath}?q1=some&pass=verysecret&q2=value` : urlPath;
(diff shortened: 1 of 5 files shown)


packages/core/src/util/url.js
```diff
@@ -15,23 +15,24 @@ const secrets = require('../secrets');
 exports.sanitizeUrl = function sanitizeUrl(urlString) {
   let normalizedUrl;
   try {
-    // This currently uses the legacy URL API. As soon as we drop support for Node.js 6 we should move to the
-    // WHATWG URL API (https://nodejs.org/api/url.html#url_the_whatwg_url_api).
-    const p = url.parse(urlString);
-    if (p.protocol == null && p.host == null && p.pathname == null) {
+    const parsedUrl = new URL(urlString);
```
aryamohanan, line 18: Since The signature '(urlString: string): UrlWithStringQuery' of 'url.parse' is deprecated, USING https://nodejs.org/api/url.html#the-whatwg-url-api

packages/collector/test/tracing/protocols/http/client/clientApp.js
```diff
@@ -287,6 +287,13 @@ app.post('/upload-s3', (req, res) => {
     }
   });
 });
+app.get('/matrix-params/:params', (req, res) => {
+  try {
+    res.sendStatus(200);
+  } catch (error) {
+    res.sendStatus(500);
+  }
```
kirrg001, line 295: 🤔

PR #1070, opened 2024-03-12
fix(collector): gracefully shutdown if core module process is not ava…

kirrg001:
…ilable

refs 155995

- we can add this check later for the serverless packages


aryamohanan (approved):
🎉

PR #1071, opened 2024-03-12
chore: fixed sendRequest function in ProcessControls

kirrg001:
refs https://github.com/instana/nodejs/pull/1017

We saw so many random failures on CircleCI and Tekton because `waitUntilServerIsUp` no longer waited for the app to be ready. The function `sendRequest` was refactored wrong.

### DIFF
--- a/packages/collector/test/test_util/ProcessControls.js
+++ b/packages/collector/test/test_util/ProcessControls.js
@@ -169,7 +169,8 @@ class ProcessControls {
         () =>
           this.sendRequest({
             method: 'GET',
-            suppressTracing: true
+            suppressTracing: true,
+            checkStatusCode: true
           }),
         retryTime,
         until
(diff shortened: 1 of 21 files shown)


packages/collector/test/tracing/frameworks/sequelize/app.js
```diff
@@ -81,7 +81,7 @@ app.get('/raw', async (req, res) => {
     bind: ['parapeter']
   });
 
-  res.json();
```
kirrg001, line 84: This generates invalid json body.

packages/collector/test/test_util/ProcessControls.js
```diff
@@ -248,33 +252,42 @@ class ProcessControls {
...
-          response = await result.text();
-        } else {
-          response = {
-            body: await result.text(),
-            status: result.status,
-            headers: result.headers
-          };
-        }
 
-        return response;
```
kirrg001, line 271: The function simply returned "Internal Server Error" if the status code of the app was 500.

packages/collector/test/test_util/ProcessControls.js
```diff
@@ -248,33 +256,44 @@ class ProcessControls {
...
+        response = await result.json();
+      } else if (contentType && (contentType.includes('text/html') || contentType.includes('text/plain'))) {
+        response = await result.text();
+      } else {
+        // CASE: Some tests do not use express and then the header is missing
+        //       Some tests use `res.send`, which is not a JSON response.
+        response = await result.text();
+      }
+
+      if (checkStatusCode) {
```
kirrg001, line 283: This is a temporary workaround because I do not have the time to go through all error tests.
The correct behaviour is here to throw an error by default. If the test case throws a 500 from the app, this helper function needs to forward the error. Otherwise we could make false assumptions in the target test case. Could you please create a ticket? @aryamohanan

packages/collector/test/test_util/ProcessControls.js
```diff
@@ -169,7 +169,8 @@ class ProcessControls {
         () =>
           this.sendRequest({
             method: 'GET',
-            suppressTracing: true
+            suppressTracing: true,
+            checkStatusCode: true
```
kirrg001, line 173: This is the fix for respecting the response from the app for the health endpoint `/`.

packages/collector/test/tracing/misc/specification_compliance/test.js
```diff
@@ -148,13 +148,17 @@ describe('spec compliance', function () {
           const request = {
             path: fullPath,
             headers,
-            resolveWithFullResponse: true
+            resolveWithFullResponse: true,
+            extendedResponse: true
```
kirrg001, line 152: You can ask the utility to return an extended payload.

aryamohanan (approved):

PR #1075, opened 2024-03-14
feat: added support for sqs-consumer v8, v9

kirrg001:
refs 158386


kirrg001:
v8 is already broken

kirrg001:
The underlying problem is https://github.com/instana/nodejs/pull/1076

kirrg001:
@aryamohanan We can ship that on Monday as well.

aryamohanan (approved):

PR #1076, opened 2024-03-15
fix(sqs): spans not being captured caused by @aws-sdk/client-sqs@3.481.0

kirrg001:
[ci all-tests]

refs 158386


kirrg001:
We are loosing the context for one specific use case: sending via callback and receiving via v2 style.

kirrg001:
3.478.0 works
3.481.0 does not work

aryamohanan (approved):

PR #1084, opened 2024-03-21
test: fixed the failing testcase for w3c_trace_context

aryamohanan:
If an invalid URL is found, such as when simply a path is provided without a host part, this scenario is taken into consideration.
Invalid URLs, such as /start?key=value, result in an error and the same input string. Handled this in the fix
Earlier the url.parse was used, so we deprecate that and use the new base parameter of the URL constructor , Related PR https://github.com/instana/nodejs/pull/1069

### DIFF
--- a/packages/core/src/util/url.js
+++ b/packages/core/src/util/url.js
@@ -19,6 +19,7 @@ exports.sanitizeUrl = function sanitizeUrl(urlString) {
     const url = new URL(urlString);
 
     // If URL has no protocol, host, or path, return the original URL.
+    // TODO: This case need adjustment for complete sanitization of the URL.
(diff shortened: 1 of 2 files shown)


packages/core/src/util/url.js
```diff
@@ -15,6 +15,9 @@ const secrets = require('../secrets');
 exports.sanitizeUrl = function sanitizeUrl(urlString) {
   let normalizedUrl;
   try {
+    if (!URL.canParse(urlString)) {
```
aryamohanan, line 18: If an invalid URL is encountered, for example, when only a path is provided without a host portion, this scenario is accounted for.

kirrg001 (approved):
LGTM
Thanks for creating the follow up card 👍

PR #1094, opened 2024-03-26
docs: tekton

kirrg001:


packages/shared-metrics/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@instana/shared-metrics",
-  "version": "3.2.0",
```
kirrg001, line 3: That was my mistake form #1036

PR #1103, opened 2024-04-08
ci: added currency report

kirrg001:
refs [INSTA-1016](https://jsw.ibm.com/browse/INSTA-1016)


.tekton/tasks/upload-currency-report.yaml
```diff
@@ -0,0 +1,54 @@
...
+      env:
+        - name: GH_ENTERPRISE_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: $(params.continuous-delivery-context-secret)
+              key: "GH_ENTERPRISE_TOKEN"
+      script: |
+        #!/bin/bash
+
+        if [ "$(params.type)" != "main" ]; then
```
kirrg001, line 28: FYI: We upload the report only for the main pipeline. Nightly and weekly jobs are excluded. I wanted to avoid adding a new pipeline only for the report. We can change that in the future if we want

PR #1112, opened 2024-04-10
[Currency Bot] Bumped tedious from 15.1.3 to 18.1.0

kirrg001:
Tada!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -157,7 +157,7 @@
         "sqs-consumer-v5": "npm:sqs-consumer@^5.7.0",
         "stealthy-require": "1.1.1",
         "superagent": "^8.1.2",
-        "tedious": "^15.1.3",
+        "tedious": "^18.1.0",
         "tsoa": "^6.0.1",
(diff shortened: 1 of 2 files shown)


kirrg001:
@aryamohanan Do we have a card for that to increase the support at Otel?

aryamohanan:
> @aryamohanan Do we have a card for that to increase the support at Otel?

Created one INSTA-1561, OpenTelemetry instrumentation for tedious only covers tedious version 15, so we can close this PR for now and consider update later.

kirrg001:
> Created one INSTA-1561, OpenTelemetry instrumentation for tedious only covers tedious version 15, so we can close this PR for now and consider update later.

Thanks. I added the label "Currency" to 1561.
Could you close this PR and update the `currencies.json` to avoid further PR's for Tedious for now? You can push that straight to main. There is a flag `ignoreUpdates`.

aryamohanan:
Added the package to the ignore list for currency. Closing the PR for now.

PR #1113, opened 2024-04-12
test: fixed how we test prisma

kirrg001:
refs INSTA-743

Goal: Move prisma to the root dependencies. Simplify the prisma test.


aryamohanan (approved):
👏

PR #1117, opened 2024-04-15
build: removed ibm_db from devdependencies

aryamohanan:
Previously, we had made the ibm_db dependency to optional in package.json, because it is not compactable  with M1 arch, [see](https://github.com/instana/nodejs/pull/821). The currency bot added it back to devdependencies, it causes issues in M1, so added fix in currency bot https://github.com/instana/nodejs/pull/1115 and also removing from package.json


kirrg001 (approved):
Thanks for the cleanup!

PR #1130, opened 2024-04-25
[Currency Bot] Bumped pino from 8.21.0 to 9.0.0

kirrg001:
Tada!


aryamohanan:
Checking for pino release note for v9 https://github.com/pinojs/pino/releases, can't find any, but new version available athttps://www.npmjs.com/package/pino/v/9.0.0?activeTab=versions

kirrg001:
We are using pino for our tests. v9 dropped support for < 18. 
Let's try to just use v9 and see if the tests are breaking, otherwise we need an improvement to read the latest package version not from `pino`, we would need to read it from `pino-v9`.

kirrg001:
@aryamohanan RFR

package.json
```diff
@@ -188,11 +188,11 @@
     "node-fetch": "^2.7.0",
     "node-nats-streaming": "^0.3.2",
     "pg": "^8.11.5",
-    "pino": "^8.21.0",
+    "pino": "^9.0.0",
     "pino-http": "^9.0.0",
     "pino-pretty": "^10.0.0",
-    "pino-v6": "npm:pino@^6.13.3",
     "pino-v7": "npm:pino@^7.8.1",
```
aryamohanan, line 194: Should we retain pino version 7, two version checks are sufficient?

aryamohanan (approved):

package.json
kirrg001: We can do that, but then I also have to remove pino-pretty. We kept it for testing different transports, but I guess its not important. I will try

kirrg001: Done

PR #1144, opened 2024-05-07
fix(azure-container-services): gracefully shutdown if node version is unsupported

kirrg001:
A return was missing. This is a bug.

Furthermore: At the moment all serverless packages **return** without a default export. Consistency change.
Its worth raising an issue to test if we have to export a default interface in case a customers is doing a manual instrumentation. We do not want to crash the customer's application.


aryamohanan (approved):

kirrg001:
Raised [INSTA-3511](https://jsw.ibm.com/browse/INSTA-3511)

PR #1063, opened 2024-03-04
fix: resolved esm loader issue for manual instrumentation node v18.19.0 and above

aryamohanan:
Starting from version 20.6, ESM loaders are [off-threaded](https://github.com/nodejs/node/pull/44710), loaded in separate thread, whereas previously we were loading our Instana collector in loader, so the change broke existing implementation. To resolve this, we used the `--import` flag, replacing the deprecated `--experimental-loader `and allowing the collector to load within the main thread.

Note that --import is supported only in Node.js versions 18.19 and later. Thus, we maintain both loading styles for compatibility across different Node.js versions.

To enable experimental ESM support, use the following command:

- For Node.js versions greater than or equal to 18.19:

    `node --import  /path/to/instana/node_modules/@instana/collector/esm-register.mjs entry-point`

- For Node.js versions less than 18.19:

    `node --experimental-loader /path/to/instana/node_modules/@instana/collector/esm-loader.mjs entry-point`

We're actively working on implementing native ESM support, targeted for resolution in INSTA-807.

**References**:

[Move ESM loaders off-thread](https://github.com/nodejs/node/pull/44710)

https://github.com/nodejs/node/issues/47880

Notes:

- Node 18.19 is also affected due to back-porting.
- `--import` flag is still on experimental phase.

**Tasks**

- [x]  loaders are off threaded!
- [x] experimental-loader flag deprecated
- [x] support all Node.js versions above 18.19
- [x]  run all unit testcases

Post Release

- [ ] Update the build version 18.18 to the latest  18.x.x(18.20.1)
- [ ] Consider changing the development version to v20 from 18
- [ ]  Automated instrumentation update

PR #1152, opened 2024-05-16
chore: switched development version to v20

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-983

Now that we have [resolved the issue with ESM ](https://github.com/instana/nodejs/pull/1063)and Node v18.18/v18.19, we can switch to Node v20.
Node v20 is LTS.

- [x] switch pipelines to use v20
- [x] manually verify that the docker images build
- [x] fixed slack notifications for ci.instana.io
- [x] add global "node-version" in Tekton which is always used when the pipeline itself did not overwrite it (Tekton automatically sends the correct version)


packages/azure-container-services/package.json
```diff
@@ -61,6 +61,9 @@
   "bugs": {
     "url": "https://github.com/instana/nodejs/issues"
   },
+  "engines": {
+    "node": ">=14.0.0"
+  },
```
kirrg001, line 66: @aryamohanan Engine definition was missing for this package. Do you have any idea why? Is Node v14 correct for azure container services? Thanks

aryamohanan: If I recall correctly, the reason is that when we added the support azure-app-services, the supported Node.js versions were 16, 18, and 20. You can refer to the comment here: [GitHub](https://github.com/instana/nodejs/blob/main/packages/azure-container-services/src/index.js#L9). Additionally, we can verify the supported versions in the official documentation: [Microsoft documentation](https://learn.microsoft.com/en-us/azure/app-service/configure-language-nodejs?pivots=platform-linux).

kirrg001: I saw that the [test](https://github.com/instana/nodejs/blob/v3.8.1/packages/azure-container-services/test/using_api/test.js) is not skipped for < 16. Thats why I got confused :) 
Should I use "node": ">=14.0.0" or "node": ">=16.0.0"?

aryamohanan: We are not calling the actual Azure app service in the tests, same as GCP.

kirrg001: Then I keep 

> "node": ">=14.0.0"

packages/serverless/ci/pipeline.yaml
```diff
@@ -108,7 +108,7 @@ resources:
     type: slack-notification
     icon: slack
     source:
-      url: ((tracer-community-slack-hook))
+      url: ((team-nodejs-slack-hook))
```
kirrg001, line 111: Added a new nodejs slack webhook.

kirrg001:
I have already started to test the images. PR is RFR.

packages/serverless/ci/testing/pipeline-debug.yaml
```diff
@@ -0,0 +1,73 @@
...
+      username: ((instanacd-github-api-token))
+      password: x-oauth-basic
+
+
+  - name: instana-aws-fargate-npm-package
+    type: npm-resource
+    icon: cube-outline
+    source:
+      scope: instana
+      package: aws-fargate
```
aryamohanan, line 20: Is this specific to aws-fargate? As its for local testing we can name it generic?

packages/aws-fargate/images/inspector/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM node:18-alpine
+ARG NODEJS_VERSION
```
aryamohanan, line 1: qs: From where this value is set?

packages/serverless/ci/testing/pipeline-debug.yaml
kirrg001: This resource downloads the fargate npm package. The testing pipeline is just a minified version of the real serverless pipeline. We can't rename this.

packages/aws-fargate/images/inspector/Dockerfile
kirrg001: https://github.com/instana/nodejs/pull/1152/files/15ed7d909f067d2e6e6cb507d3fd7f9ca256b9e5#diff-02f198aef71805a77d9836533174d3e061195b4560547217b8fcad4978a8bbbb

aryamohanan (approved):

PR #1160, opened 2024-05-22
feat: added support for node-fetch v3

aryamohanan:
Upgraded our support for the [node-fetch](https://www.npmjs.com/package/node-fetch) module to version 3.

In our instrumentation setup, while native ESM support is still pending and the requireHook isn't compatible with native ESM, the 'node-fetch' module transitions to a pure ESM module from version 3 onwards. This transition remains functional due to 'node-fetch' being instrumented alongside the 'http' module, which works [without](https://github.com/instana/nodejs/blob/v3.8.1/packages/core/src/tracing/instrumentation/protocols/httpClient.js#L32) the requireHook. Notably, this differs from our other instrumentations, which require 'http' and 'https' at the top. Importantly, native ESM libraries importing core Node modules do not trigger Module._load and thus do not utilize the requireHook. However, our requireHook is triggered when an ESM library imports a CommonJS package.

As Node.js [native fetch API](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch) only supports from Node.js v18 we used node-fetch for majority of test cases, in this PR we are added a check in the test cases to use native fetch if exists else cjs version of native-fetch module, which is v2.


tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "target": "ES6",
     "allowJs": true,
     "checkJs": true,
-    "lib": ["esnext.asynciterable"],
+    "lib": ["esnext.asynciterable","DOM"],
```
aryamohanan, line 7: There is an open issue with the latest version of `node-fetch` related to type definitions. You can track the discussion and updates on this issue at node-fetch/node-fetch#1617. AS a temp fix added DOMthe tsconfig

packages/aws-fargate/images/inspector/index.js
```diff
@@ -11,8 +11,6 @@ const app = new http.Server();
 const port = 3000;
 
 const { promises: fs } = require('fs');
-const fetch = require('node-fetch');
```
aryamohanan, line 14: These are used for the manual testing, for that we can use the global fetch, instead of node-fetch

packages/collector/test/tracing/frameworks/node-fetch/app.mjs
```diff
@@ -0,0 +1,51 @@
...
+  process.disconnect();
+  process.exit(0);
+});
+
+import express from 'express';
+import morgan from 'morgan';
+import bodyParser from 'body-parser';
+import getAppPort from '../../../test_util/app-port.js';
+const port = getAppPort();
+import fetch from 'node-fetch';
```
aryamohanan, line 20: I have add a testcase for node-fetch, we only have native fetch testcase, which under protocol folder.

aryamohanan:
> Could you please extend the requireHook comment and add node-fetch? Could you please update currencies.json?

Extented https://github.com/instana/nodejs/pull/1160/files#diff-3a1dfba9ca7ff13e28abf38067dba2d7f4132142b09915f2ee83b41ad0c0250d

> Could you please extend the requireHook comment and add node-fetch? Could you please update currencies.json?
Sure

tsconfig.json
kirrg001: Okay 👍 Can you please add a comment? Thanks!

packages/collector/test/tracing/frameworks/node-fetch/app.js
```diff
@@ -0,0 +1,52 @@
+/*
```
kirrg001, line 1: Q: Putting this here. node-fetch is not a framework.  How about putting it to "misc"?

aryamohanan: Moved to misc

kirrg001 (approved):

PR #1180, opened 2024-06-12
Revert "chore(aws-lambda): removed timeout detection"

kirrg001:
Reverts instana/nodejs#668

Brining this back but we make it optional.
Customer need this feature for debugging purposes.
We will also update the docs as well to make it clear


kirrg001:
Closing because of branch name. Re-opening.

PR #1184, opened 2024-06-14
feat: added support for tedious v16 and v17

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-1561

An underlying npm workspaces bug made us believe that we are testing against tedious v17.
See comment.


kirrg001:
Draft PR still. Needs investigation because tedious is a sub dependency of typeorm and otel has a bug as it looks like because it loads the wrong tedious version

kirrg001:
Trying to switch to yarn to get rid of the annoying installation problems with npm.

kirrg001:
> Trying to switch to yarn to get rid of the annoying installation problems with npm.

I will do that after the merge, because we need to ship this update.

package.json
```diff
@@ -94,6 +94,7 @@
     "@grpc/proto-loader": "^0.7.13",
     "@hapi/hapi": "^21.3.10",
     "@ibm/tekton-lint": "^1.0.0-beta.9",
+    "@opentelemetry/instrumentation": "0.50.0",
```
kirrg001, line 97: This is a workaround for now.
Hopefully we can move to yarn asap.

PR #1190, opened 2024-06-19
feat: added support for mssql v11

instanacd:
refs https://jsw.ibm.com/browse/INSTA-8736

### DIFF
--- a/package.json
+++ b/package.json
@@ -183,7 +183,7 @@
     "mongoose-v7": "npm:mongoose@^7.6.10",
     "morgan": "^1.10.0",
     "mssql": "^11.0.0",
-    "mssql-v9": "npm:mssql@^9.0.1",
+    "mssql-v10": "npm:mssql@^10.0.4",
     "mysql": "^2.18.1",
     "mysql2": "^3.10.1",
     "nats": "^2.26.0",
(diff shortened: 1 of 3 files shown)


aryamohanan (approved):

package.json
```diff
@@ -182,8 +182,8 @@
     "mongoose-v6": "npm:mongoose@^6.12.7",
     "mongoose-v7": "npm:mongoose@^7.6.10",
     "morgan": "^1.10.0",
-    "mssql": "^10.0.3",
-    "mssql-v9": "npm:mssql@^9.0.1",
```
kirrg001, line 186: Suggestion: Only run v11 tests.
If there is no "real change" between major versions, there is no need to run both versions.

kirrg001: I just noticed that the adapted test is anyway wrong. You are just iterating over the versions but you are not using both versions (mock missing).

kirrg001 (changes requested):
See second comment

package.json
kirrg001: My bad. The logic existed already 👍

aryamohanan:
> LGTM
> 
> Please consider my suggestion: https://github.com/instana/nodejs/pull/1190/files#r1647083061
> 
> Otherwise, feel free to merge

Considering that we still support node versions 16 and 14, and the latest MSSQL doesn't support these versions, I intend to retain the tests for versions 10 and 11.

PR #1210, opened 2024-07-03

janritter:
Previously this resulted in a high file descriptor usage in Lambda until the function reached the hard limit

This is a fix for the IBM customer support ticket: TS016453828
More details about the detection of the problem and steps to reproduce are documented in the ticket.

**A short summary of the problem:** 
Old requests were not correctly cleaned up which was mainly a problem for Lambda functions with a lot of reuse (warm Lambdas), since the number of file descriptors continued increasing over time until the Lambda hard limit was reached and the function crashed.

While this fixes the problem described above, I have the feeling the root rause which led to this problem becoming visible might be related to failing heartbeat checks and the actual extension binary which I couldn't check further.

**Testing**
| Node version of the Lambda     | File descriptor problem resolved      |  Traces received in Instana
| ------------- | ------------- | ------- |
| Node 16.x | ✅  | ✅
| Node 18.x | ✅  |✅
| Node 20.x | ✅  |✅

Successor of #1208 (using - instead of / in the branch name)

### DIFF
--- a/packages/serverless/src/backend_connector.js
+++ b/packages/serverless/src/backend_connector.js
@@ -308,6 +308,9 @@ function send(resourcePath, payload, finalLambdaRequest, callback) {
       if (finalLambdaRequest) {
         req.removeAllListeners();
         req.on('error', () => {});
+        // At this point we already received a response from the server, but since we removed all listeners we have to
+        // manually clean up the request.
+        req.destroy();
       }
 
       handleCallback();
--- a/packages/serverless/test/backend_connector_test.js
+++ b/packages/serverless/test/backend_connector_test.js
@@ -43,7 +43,8 @@ describe('[UNIT] backend connector', () => {
         on: onStub,
         setTimeout: sinon.stub(),
         end: sinon.stub(),
-        removeAllListeners: sinon.stub()
+        removeAllListeners: sinon.stub(),
+        destroy: sinon.stub()
       });
     });
 
--- a/packages/serverless/test/backend_connector_test.js
+++ b/packages/serverless/test/backend_connector_test.js
@@ -32,19 +32,21 @@ describe('[UNIT] backend connector', () => {
     this.timeout(config.getTestTimeout());
 
     let onStub;
+    let destroyStub;
 
     beforeEach(() => {
       sinon.spy(global, 'setInterval');
       sinon.spy(global, 'clearInterval');
 
       onStub = sinon.stub();
+      destroyStub = sinon.stub();
 
       sinon.stub(uninstrumentedHttp.http, 'request').returns({
         on: onStub,
         setTimeout: sinon.stub(),
         end: sinon.stub(),
         removeAllListeners: sinon.stub(),
-        destroy: sinon.stub()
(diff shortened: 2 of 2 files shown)


aryamohanan:
@janritter Thank you once again for creating the PR. It appears that the tests are failing; I will look into it as soon as possible.

aryamohanan:
Please note that the commit message should include a sign-off statement, as outlined in our [contributing guidelines](https://github.com/instana/nodejs/blob/main/CONTRIBUTING.md#how-to-contribute).

janritter:
@aryamohanan sorry, I missed the sign-off part.
I pushed an updated version of the commit which is signed-off. I also included the test addition from your commit, I hope that's ok.

packages/serverless/src/backend_connector.js
```diff
@@ -308,6 +308,9 @@ function send(resourcePath, payload, finalLambdaRequest, callback) {
       if (finalLambdaRequest) {
         req.removeAllListeners();
         req.on('error', () => {});
+        // At this point we already received a response from the server, but since we removed all listeners we have to
+        // manually clean up the request.
+        req.destroy();
```
aryamohanan, line 313: Changes in the backend connector necessitate a unit test. Could you please add a test for the bug and its corresponding solution?

janritter: I had a look and tried different options, but then realized that the problem can't be reproduced with the test setup here. Since no real requests are send, but a sinon stub is used, we don't see any unclosed / leaking sockets.

Now knowing that the req must be destroyed I can check hat req.destroy actually gets called

aryamohanan: Thank you for adding the case. We'll consider adding the integration test later.

aryamohanan (approved):

aryamohanan:
I have created another PR at https://github.com/instana/nodejs/pull/1217 and merged it into the main branch. There were two issues encountered with this PR:

- Verification was missing (the commit appeared as unverified).
- The commit length exceeded our CI's limit, causing it to fail.

janritter:
@aryamohanan thanks, the verification thing is strange, especially since it's marked as verified in GitHub. For the commit length, the included commitlint seems to use a different length config.

Do you already have an estimate when the fix will be available as a Lambda Layer?

aryamohanan:
@janritter We are considering releasing it on Monday. Does that work for you?

janritter:
@aryamohanan Can you let me know once the new version is available with the layer version numbers for eu-central-1 (I think the docs with the versions are usually updated a bit delayed)

aryamohanan:
@janritter We will release the new version at the earliest opportunity and will let you know as soon as it is available. Thank you for your patience.

aryamohanan:
Hello @janritter, we've rolled out a new version containing the fix. You can find more details at https://github.com/instana/nodejs/releases. The layer version for eu-central-1 is now 193 (arn:aws:lambda:eu-central-1:410797082306:layer:instana-nodejs:193). Thanks for your work on the fix, and please reach out if you encounter any issues.

PR #1222, opened 2024-07-08
test: updated couchbase tests to run on node.js v18 and above

aryamohanan:
Discovered that the latest Couchbase silently dropped support for v16. Their official documentation states:

> The Couchbase Node.js Client will run on any [supported LTS version of Node.js](https://github.com/nodejs/Release).

They are using [node-addon-api](https://github.com/nodejs/node-addon-api ) as a dependency, and in couchbase version 4.4.0, it was updated to [8.0.0](https://github.com/nodejs/node-addon-api/releases/tag/v8.0.0), which dropped support for v16. This change causes our v16 tests to fail.


kirrg001:
https://github.com/couchbase/couchnode/commit/ab478ed2019501cdbb94992fb3aee6f15c83b6ef
https://docs.couchbase.com/nodejs-sdk/current/project-docs/compatibility.html#node-js-version-compatibility

PR #1227, opened 2024-07-09
chore: clarified finalLambdaRequest variable and updated comments

kirrg001:
refs https://github.com/instana/nodejs/commit/82c37f43a0cea18f53d87dc208d6a3dfda7c094c

### DIFF
--- a/packages/serverless/src/backend_connector.js
+++ b/packages/serverless/src/backend_connector.js
@@ -107,6 +107,15 @@ exports.setLogger = function setLogger(_logger) {
   logger = _logger;
 };
 
+/**
+ *
+ * "finalLambdaRequest":
+ * When using AWS Lambda, we send metrics and spans together
+ * using the function "sendBundle". The variable was invented to indicate
+ * that this is the last request to be sent before the AWS Lambda runtime might freeze the process.
+ * Currently, there is exactly one request to send all the data and
+ * the variable is always true.
+ */
 exports.sendBundle = function sendBundle(bundle, finalLambdaRequest, callback) {
   send('/bundle', bundle, finalLambdaRequest, callback);
 };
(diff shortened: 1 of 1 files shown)


packages/serverless/src/backend_connector.js
```diff
@@ -405,7 +417,7 @@ function send(resourcePath, payload, finalLambdaRequest, callback) {
         // Finally, abort the request because from our end we are no longer interested in the response and we also do
         // not want to let pending IO actions linger in the event loop. This will also call request.destoy and
         // req.socket.destroy() internally.
-        req.abort();
+        destroyRequest(req);
```
kirrg001, line 420: req.abort is deprecated. Re-using the helper function.

packages/serverless/src/backend_connector.js
```diff
@@ -302,15 +311,20 @@ function send(resourcePath, payload, finalLambdaRequest, callback) {
...
         req.removeAllListeners();
         req.on('error', () => {});
-        // At this point we already received a response from the server, but since we removed all listeners we have to
-        // manually clean up the request.
-        req.destroy();
+
+        // Finally, abort the request because from our end we are no longer interested in the response and we also do
+        // not want to let pending IO actions linger in the event loop. This will also call request.destoy and
+        // req.socket.destroy() internally.
+        destroyRequest(req);
```
kirrg001, line 327: Re-using the helper function.

packages/serverless/src/backend_connector.js
```diff
@@ -107,6 +107,15 @@ exports.setLogger = function setLogger(_logger) {
   logger = _logger;
 };
 
+/**
+ *
+ * "finalLambdaRequest":
```
kirrg001, line 112: Please double check if my conclusion is correct. It looks to me the variable is at the moment mis-named, but if we have a proper comment it might be less confusing.

aryamohanan (approved):

packages/serverless/src/backend_connector.js
kirrg001: @aryamohanan Did you make the same conclusion?

aryamohanan: In my opinion, your comments are accurate and clarify the situation well.

Issue #1231, opened 2024-07-10
[Bug]: event-loop-stats addon in musl 115 contains only null bytes


Llois41:
### Problem Description

In https://github.com/instana/nodejs/tree/main/packages/shared-metrics/addons/linux/x64/musl/115 the `[event-loop-stats.tar.gz](https://github.com/instana/nodejs/blob/main/packages/shared-metrics/addons/linux/x64/musl/115/event-loop-stats.tar.gz)` file only contains null bytes which results in a archive size of only 29 bytes. Comparing this to the similar archives in the other musl versions (which are 20+ kilobytes of size) this seems broken.

This results in the startup error which looks like this: `Unpacking the precompiled build for event-loop-stats (linux/x64/musl/ABI 115) failed. Error: TAR_BAD_ARCHIVE: Unrecognized archive format\n`. This then leads to event-loop-stats not being available during the runtime.

### Node.js Version

20.15.1

### package.json

```shell
irrelevant
```

### package-lock.json

```shell
irrelevant
```

aryamohanan:
Thank you for the report. We will take a look asap.

kirrg001:
@Llois41 

We have successfully released [3.14.3](https://github.com/instana/nodejs/releases/tag/v3.14.3). 

The release was manually tested in a Docker container on linux x64 musl Node v20. We couldn't see any errors. 

Let use know if you are still encountering any problems.

PR #1232, opened 2024-07-10
fix(shared-metrics): recreated prebuild for linux x64 musl abi 115

kirrg001:
refs #1231

See https://github.com/instana/nodejs/issues/1231#issuecomment-2220839505.

I double checked the other prebuilds, looks good to me.
I could not figure out why this prebuild was not generated correctly.


aryamohanan (approved):
LGTM 💯

PR #1242, opened 2024-07-23
test: fixed the timeout issue while connectiong the azure db that resulted in tedious test to fail

abhilash-sivan:
refs INSTA-7718

### DIFF
--- a/packages/collector/test/tracing/opentelemetry/tedious-app.js
+++ b/packages/collector/test/tracing/opentelemetry/tedious-app.js
@@ -52,14 +52,14 @@ const config = {
   },
   options: {
     database: azureConfig?.AZURE_SQL_DATABASE || process.env.AZURE_SQL_DATABASE,
-    connectTimeout: 20000
+    connectTimeout: 30000
   }
 };
 
 let connected = false;
 let connection;
 
-const retryDelay = 1000;
+const retryDelay = 30000;
 const maxRetries = 2;
 let currentRetry = 0;
 
--- a/packages/collector/test/tracing/opentelemetry/test.js
+++ b/packages/collector/test/tracing/opentelemetry/test.js
@@ -497,18 +497,14 @@ mochaSuiteFn('opentelemetry/instrumentations', function () {
         ));
   });
 
-  const runTedious = describe;
-  runTedious('tedious', function () {
+  describe('tedious', function () {
     describe('opentelemetry is enabled', function () {
       globalAgent.setUpCleanUpHooks();
       const agentControls = globalAgent.instance;
       let controls;
 
-      /* This line sets the timeout duration for a test case to 2 minutes.
-       * This is added because we are having some issue with the free tier of azure database which goes from
-       * `online` state to `paused` state after some time of inactivity (seems to be 30 minutes).
-       * So the initial connection req on the paused DB needs time for it to get back online.
-      */
+    // We need to increase the waiting timeout here for the initial azure connection,
+    // because it can take up to 1-2 minutes till azure replies if the db is in paused state
       this.timeout(1000 * 60 * 2);
 
       before(async () => {
@@ -520,7 +516,6 @@ mochaSuiteFn('opentelemetry/instrumentations', function () {
           }
         });
 
-        // Waiting time added for the agent connection
         await controls.startAndWaitForAgentConnection(5000, Date.now() + 1000 * 60 * 2);
       });
 
--- a/packages/collector/test/tracing/opentelemetry/test.js
+++ b/packages/collector/test/tracing/opentelemetry/test.js
@@ -8,7 +8,6 @@ const expect = require('chai').expect;
 const path = require('path');
 const supportedVersion = require('@instana/core').tracing.supportedVersion;
 const constants = require('@instana/core').tracing.constants;
-const { isCI } = require('@instana/core/test/test_util');
 const config = require('../../../../core/test/config');
 const portfinder = require('../../test_util/portfinder');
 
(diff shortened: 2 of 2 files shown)


packages/collector/test/tracing/opentelemetry/test.js
```diff
@@ -516,7 +520,8 @@ mochaSuiteFn('opentelemetry/instrumentations', function () {
           }
         });
 
-        await controls.startAndWaitForAgentConnection();
+        // Waiting time added for the agent connection
+        await controls.startAndWaitForAgentConnection(5000, Date.now() + 1000 * 60 * 2);
```
kirrg001, line 524: Putting it here.

Q: How is the solution in the PR working if the tedious library does not fire the `connection.on('connect', err => {` event? 🤔

abhilash-sivan: My understanding is that the library is firing the event but not handling the response when the database status is paused. It was waiting for the connection timeout. But our tests were failing before that because the test cases were not waiting.

kirrg001: I'd suggest lets wait exactly 24h and not running the tests. And then manually run the tests locally.
If it passes one the first run, then we are good 👍

abhilash-sivan: Okay sure. I will check and post the results by tomorrow 👍

abhilash-sivan: @kirrg001 The tests are running perfectly even after 24h

kirrg001: Perfect 👍

packages/collector/test/tracing/opentelemetry/tedious-app.js
```diff
@@ -52,14 +52,14 @@ const config = {
...
-    connectTimeout: 20000
+    connectTimeout: 30000
   }
 };
 
 let connected = false;
 let connection;
 
-const retryDelay = 1000;
+const retryDelay = 30000;
```
kirrg001, line 62: Q: I don't understand this change 🤔 

The retry mechanism helps to retry connecting to Azure. Why would we wait 30s on each retry?
From my understanding the initial run does not even trigger the retry logic.

Does this change help to fix the problem? I thought the `connectionTimeout` and modifying `await controls.startAndWaitForAgentConnection` solved the problem.

packages/collector/test/tracing/opentelemetry/test.js
```diff
@@ -498,15 +497,20 @@ mochaSuiteFn('opentelemetry/instrumentations', function () {
         ));
   });
 
-  // Skipping the tests on CI for now, because there is a existing timeout issue while connectiong the azure sql,
-  // see github issue https://github.com/tediousjs/tedious/issues/1277
-  const runTedious = isCI() ? describe.skip : describe;
+  const runTedious = describe;
   runTedious('tedious', function () {
```
kirrg001, line 501:
```suggestion
  describe('tedious', function () {
```

Suggestion: There is no need to save mocha's `describe` into a new variable called `runTedious` anymore.

packages/collector/test/tracing/opentelemetry/test.js
```diff
@@ -498,15 +497,20 @@ mochaSuiteFn('opentelemetry/instrumentations', function () {
         ));
   });
 
-  // Skipping the tests on CI for now, because there is a existing timeout issue while connectiong the azure sql,
-  // see github issue https://github.com/tediousjs/tedious/issues/1277
-  const runTedious = isCI() ? describe.skip : describe;
+  const runTedious = describe;
```
kirrg001, line 500:
```suggestion

```

Related to https://github.com/instana/nodejs/pull/1242/files#r1691246162

packages/collector/test/tracing/opentelemetry/tedious-app.js
abhilash-sivan: This has to be updated because we have to give some time for the DB to get online. Since we are retrying only 2 times, this 30-second delay won't be an issue. 
This might work with a 20 or 25 decond delay. but as a safer I put it for 30 seconds

packages/collector/test/tracing/opentelemetry/test.js
```diff
@@ -516,7 +520,8 @@ mochaSuiteFn('opentelemetry/instrumentations', function () {
           }
         });
 
-        await controls.startAndWaitForAgentConnection();
+        // Waiting time added for the agent connection
```
kirrg001, line 523: Suggestion: This comment describes "what" you have "done" while coding. IMO doesn't add any benefit for the team when working again on this code.

```suggestion
```

packages/collector/test/tracing/opentelemetry/tedious-app.js
abhilash-sivan: The initial connection request triggers the retry after the connectionTimeout mentioned. 
Even the connection timeout can also be kept as 20 sec, but increased for safety.

packages/collector/test/tracing/opentelemetry/test.js
kirrg001: Alternatively you could say: We need to increase the waiting timeout here for the initial azure connection, because it can take up to 1-2 minutes til Azure replies.

packages/collector/test/tracing/opentelemetry/test.js
abhilash-sivan: Sure. Will update

packages/collector/test/tracing/opentelemetry/tedious-app.js
kirrg001: But you told me that the retry mechanmisn is not even triggered? 🤔 
And I haven't seen the logs as well:

```sh
Connected to the database
[ProcessControls] server is up.
[ProcessControls] started
```

> setTimeout(connectWithRetry, retryDelay);

Is called after

> connection.on('connect', err => {

Thats why I am confused by this change.

kirrg001: Ah okay 👍 Thanks for explaining.

kirrg001 (approved):
LGTM

PR #1243, opened 2024-07-24
chore: removed the currency branch if no new commits

aryamohanan:
Currently, the currency-bot branch is created even when there are no version updates. This results in automatic updates failing because the branch already exists.

### DIFF
--- a/bin/currency/update-currencies.js
+++ b/bin/currency/update-currencies.js
@@ -22,6 +22,17 @@ console.log(`MAJOR_UPDATES_MODE: ${MAJOR_UPDATES_MODE}`);
 console.log(`BRANCH: ${BRANCH}`);
 console.log(`SKIP_PUSH: ${SKIP_PUSH}`);
 
+// eslint-disable-next-line no-shadow
+const hasCommits = (branch, cwd) => {
+  try {
+    const result = execSync(`git log main..${branch} --pretty=format:"%h"`, { cwd }).toString().trim();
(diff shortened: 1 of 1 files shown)


kirrg001:
That is so cool

bin/currency/update-currencies.js
```diff
@@ -124,11 +142,14 @@ currencies.forEach(currency => {
 
     if (!SKIP_PUSH) {
       execSync(`git push origin ${branchName} --no-verify`, { cwd });
-      execSync(
-        // eslint-disable-next-line max-len
-        `gh pr create --base main --head ${branchName} --title "[Currency Bot] Bumped ${currency.name} from ${installedVersion} to ${latestVersion}" --body "Tada!"`,
-        { cwd }
-      );
+
+      if (removeBranchIfNoCommits(branchName)) {
```
kirrg001, line 146: Suggestion: IMO there is no need to push the remote branch if there are no commits.

How about?

> if (hasCommitCount())
>   execSync(`git push origin ${branchName} --no-verify`, { cwd });
>   ...
> else
>   delete local branch

aryamohanan: Good suggestion ❤️

kirrg001 (approved):
LGTM
I did not test the `hasCommits` function. But if you have tested it, then it should be fine 👍

aryamohanan:
Locally tested the changes.

PR #1265, opened 2024-08-09
fix: removed recursive-copy module from the production dependency

aryamohanan:
The module [recursive-copy](https://www.npmjs.com/package/recursive-copy) currently has [inflight](https://www.npmjs.com/package/inflight)(deprecated) as a subdependency, which triggers following npm warnings during the installation of our collector. 

> npm WARN deprecated inflight@1.0.6: This module is not supported and has memory leaks. Avoid using it. Consider `lru-cache` for a more robust and tested solution for coalescing async requests by key value.

To resolve this issue, we are removing recursive-copy from our production dependencies and will instead use [fs-extra](https://github.com/jprichardson/node-fs-extra), which is a drop-in replacement for the native fs module. All methods in fs are included in fs-extra.

While I generally prefer to avoid external libraries and additional dependencies, and would prefer to copy directories recursively using native fs methods, the fs.cp function was not available in earlier versions of Node.js 16.

Note: [fs.cp](https://nodejs.org/api/fs.html#fscpsrc-dest-options-callback) and [fsPromises.cp](https://nodejs.org/api/fs.html#fspromisesmkdirpath-options), which can copy directories, are available in Node v16.7.0. However, since we also support Node v14, we will not use these methods at this time. Instead, we’ll implement a custom solution using fs-extra methods to ensure compatibility with our existing Node.js versions. In the next major version we can consider to replace it with native fs.cp method. The syntax is very similar.

Reasons to choose [fs-extra](https://github.com/jprichardson/node-fs-extra):

- The methods closely resemble those of the native `fs` module, making migration straightforward.
- With the last commit being 5 months ago.
- Favorable comparison on npm trends: [fs-extra vs ncp vs recursive-copy](https://npmtrends.com/fs-extra-vs-ncp-vs-recursive-copy).
- Compatible with Node.js v14.
- Alternatives like [ncp](https://www.npmjs.com/package/ncp) and [wrench](https://www.npmjs.com/package/wrench) are no longer maintained. Thus, [fs-extra](https://www.npmjs.com/package/fs-extra) is likely the best option. The Developer of Wrench directs users to use fs-extra as he has deprecated his library.

refs : INSTA-12493


packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -184,22 +184,9 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
...
-          },
-          // @ts-ignore
-          cpErr => {
-            if (cpErr) {
-              logger.warn(`Copying the precompiled build for ${opts.nativeModuleName} ${label} failed.`, cpErr);
-              callback(false);
-              return;
-            }
-
+        fse
```
aryamohanan, line 187: In the replace-copy method, we pass the options `overwrite: true` (to allow overwriting of destination files) and `dot: true` (to enable copying of files that begin with a dot). Similarly, in `fs-extra`, the default behavior sets `overwrite` to `true` and also supports copying files that begin with a dot(added a test to verify this functionality). For reference, see the documentation at [fs-extra copy method](https://github.com/jprichardson/node-fs-extra/blob/master/docs/copy.md).

packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -211,10 +198,14 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
             // it will remember and not try to load anything from that path again (a `false` will be
             // put into the cache for that cache key). Instead, we force a new path, by adding precompiled
             // to the module path and use the absolute path to the module to load it.
+            logger.debug(`Copying the precompiled build for ${opts.nativeModuleName} ${label} successful.`);
```
aryamohanan, line 201: Added a debug log here, it might be helpful for the troubleshooting.

kirrg001:
> Note: [fs.cp](https://nodejs.org/api/fs.html#fscpsrc-dest-options-callback) and [fsPromises.cp](https://nodejs.org/api/fs.html#fspromisesmkdirpath-options), which can copy directories, are available in Node v16.7.0. However, since we also support Node v14

refs https://github.com/instana/nodejs/pull/848

Maybe its worth adding a task to the v4 release to replace fs-extra with the native fs.cp?

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -133,4 +154,20 @@ function runCopyPrecompiledForNativeAddonTest(agentControls, opts) {
         ]).then(opts.check)
       ));
   });
+
+  describe('validate precompiled binaries', () => {
+    it('should successfully copy the precompiled binaries', async () => {
```
kirrg001, line 159: Q: Did you run the extended test on main? 

A good approach is to create a PR to only exend the test and then merge into main. And then open another PR and just replace the library.

packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -211,10 +198,14 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
             // it will remember and not try to load anything from that path again (a `false` will be
             // put into the cache for that cache key). Instead, we force a new path, by adding precompiled
             // to the module path and use the absolute path to the module to load it.
+            logger.debug(`Successfully copied the precompiled build for ${opts.nativeModuleName} ${label}.`);
```
kirrg001, line 201: S: I was wondering if its good to have a debug log for "copying from X to Y" as well in line 186?

kirrg001 (changes requested):
Thanks for taking care of this. Added some comments 👍

packages/collector/test/nativeModuleRetry/test.js
aryamohanan: Created a [PR](https://github.com/instana/nodejs/pull/1276)

aryamohanan: PR to extending the precompiled binaries merged, and rebased the PR.

packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -183,23 +183,14 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
...
-          // @ts-ignore
-          cpErr => {
-            if (cpErr) {
-              logger.warn(`Copying the precompiled build for ${opts.nativeModuleName} ${label} failed.`, cpErr);
-              callback(false);
-              return;
-            }
-
+        fse
+          .copy(sourceDir, targetDir, { dereference: true })
```
kirrg001, line 192: Q: Why do we need `dereference: true`?

> dereference (boolean): dereference symlinks, default is false.

aryamohanan: I encountered an error while running the tests in local:

> Error: Cannot copy '/usr/bin/python3' to a subdirectory of itself

Following the advice in the comment [here](https://github.com/chegele/AutoGitUpdate/issues/6#issuecomment-948901575), I have used this option.

I will verify by removing the option to see if the issue is specific to my local environment.

kirrg001: Ok! Thx

kirrg001: The reason to **not** use `dereference: true ` is that `recursive-copy` does [not](https://github.com/timkendrick/recursive-copy/blob/2.0.14/lib/copy.js#L280) dereference symlinks. It copies the symlink itself, not the file or directory it points to.

aryamohanan: Yes, setting `dereference: true` is not necessary in our case. The issue was specific to my local machine.

aryamohanan: Thanks for catching this ❤️

kirrg001 (approved):
Commit suggestion

> fix(shared-metrics): removed recursive-copy dependency #1265
> refs ...
> 
> - resolves the following warning: 
>.   "npm WARN deprecated inflight@1.0.6: This module is not supported and has memory leaks. Avoid using it. Consider lru-cache for a more robust and tested solution for coalescing async requests by key value."

packages/shared-metrics/src/util/nativeModuleRetry.js
kirrg001: MUA

PR #1268, opened 2024-08-12
fix: exposed installation method in debug mode

abhilash-sivan:
JIRA card: [INSTA-778](https://jsw.ibm.com/browse/INSTA-778)

- [x] test in local with --experimental-loader flag: `node --experimental-loader ./node_modules/@instana/collector/esm-loader.mjs ./index-esm.mjs`
- [x] test in local with --import flag: `node --import ./node_modules/@instana/collector/esm-register.mjs ./index-esm.mjs`
- [x] test in local with --require flag: `node --require ./node_modules/@instana/collector/src/immediate ./index.js`
- [x] test in local with no flags: `node ./index.js`


abhilash-sivan:
commit messge to main:
```
feat: exposed installation method in debug mode
refs INSTA-778
```

abhilash-sivan:
For the normal commonJS instrumentation using `require('@instana/collector')();` in the code level currently considered as default and the debug log will print like 

> The app isnstrumented instana using: no additional flags

aryamohanan:
> commit messge to main:
> 
> ```
> feat: exposed installation method in debug mode
> refs INSTA-778
> ```

For me, this isn't a feature but rather a fix, which is useful for troubleshooting. It doesn’t add any new functionality for the customer.

packages/core/src/util/esm.js
```diff
@@ -66,3 +66,31 @@ exports.isESMApp = function isESMApp() {
 
   return isESM;
 };
+
+exports.tracerInstrumentationInfo = function tracerInstrumentationInfo() {
```
aryamohanan, line 70: This is only applicable for the collector package? what about the serverless packages?

packages/core/src/util/esm.js
```diff
@@ -66,3 +66,31 @@ exports.isESMApp = function isESMApp() {
...
+  const instrumentationMap = {
```
aryamohanan, line 71: This logic should not be here, this file specific to esm.

packages/core/src/tracing/index.js
```diff
@@ -197,6 +198,10 @@ exports.init = function init(_config, downstreamConnection, _processIdentityProv
   automaticTracingEnabled = config.tracing.automaticTracingEnabled;
 
   if (tracingEnabled) {
+    const loader = tracerInstrumentationInfo();
```
aryamohanan, line 201: IMO: This should be called only in debug mode. Could you explore the options to do that?

packages/core/src/util/esm.js
abhilash-sivan: Okay will update.

packages/core/src/tracing/index.js
abhilash-sivan: If there is an issue reported and we want to check the logs, then IMO it might be helpful to know the instrumentation details. 
But for that, we can do `console.info` instead of `console.debug`!
Whats your take on that?

packages/core/src/util/esm.js
abhilash-sivan: > Instead of
> 
> includes('@instana/collector')
> 
> can check for
> 
> esm-register.mjs

this won't work btw


> The serverless whole is another scope.

> 

Yeah, I agree.

packages/core/src/tracing/index.js
aryamohanan: IMO this should only be logged if debug mode is enabled.

packages/core/src/util/clientInstrumentationCheck.js
```diff
@@ -0,0 +1,33 @@
...
+
+exports.tracerInstrumentationInfo = function tracerInstrumentationInfo() {
+    const instrumentationMap = {
+      usingExperimentalLoaderFlag: '--experimental-loader flag',
+      usingImport: '--import flag',
+      usingRequire: '--require flag',
+      noFlags: 'no additional flags'
+    };
+
+    const usingExperimentalLoaderFlag =
```
aryamohanan, line 15: Qs: Why isn't there a check for `@instana/collector` included here? I noticed that you added checks for other flags.

aryamohanan:
I'm done with the review. Added one more comment.

packages/core/src/util/clientInstrumentationCheck.js
abhilash-sivan: Thanks for catching. That was not added in the checklist.
I will update this and add to checklist

packages/core/src/util/esm.js
aryamohanan: I believe we need to check the flag for all our packages. Checking `@instana/collector` would only apply to the collector package. We already have a similar check in place for detecting ESM, as seen [here](https://github.com/instana/nodejs/blob/main/packages/core/src/util/esm.js#L51). I'm thinking of either extending or referencing this function . This approach would help us avoid maintaining flags in multiple functions. What do you think?

abhilash-sivan: That would be a good idea. Initially, I designed the function like this but then later decided to check for a single package for checking serverless and collector specifically.
Re-using the existing function will be a good idea, even I thought of this. But since the outputs and checks differ in the `function isESMApp()` and `function tracerInstrumentationInfo()`, need to decide how can we extend this. I will get back to you after that. Please share if you have any better solutions in mind

abhilash-sivan: Implemented a new logic!

packages/core/src/tracing/index.js
```diff
@@ -179,6 +180,13 @@ exports.preInit = function preInit(preliminaryConfig) {
  * @param {CollectorPIDStore} _processIdentityProvider
  */
 exports.init = function init(_config, downstreamConnection, _processIdentityProvider) {
+  if (process.env.INSTANA_DEBUG || process.env.INSTANA_LOG_LEVEL === 'debug') {
+    const method = getPackageInstallation();
+
+    // eslint-disable-next-line no-console
+    console.debug(`The App has instrumented instana using: ${method}`);
```
kirrg001, line 187: issue: I am getting

> The App has instrumented instana using: no additional flags

When doing

> $ INSTANA_DEBUG=true node --require ~/dev/instana/nodejs-2/packages/collector/src/immediate.js app.js

kirrg001 (changes requested):

packages/core/src/tracing/index.js
abhilash-sivan: I am getting correctly!! but why🤔 
Screenshot 2024-08-22 at 3 30 33 PM

kirrg001: The path is different. Maybe we only cover specific use cases?

abhilash-sivan: The reason is that you are referring to the packages directly. 
The logic behind the comparison is checking for the `@instana` keyword in the arguments (as the clients are using our npm package).
So in a client app perspective it should work if they are including our package in the application start arguments

packages/core/src/tracing/index.js
```diff
@@ -179,6 +180,13 @@ exports.preInit = function preInit(preliminaryConfig) {
  * @param {CollectorPIDStore} _processIdentityProvider
  */
 exports.init = function init(_config, downstreamConnection, _processIdentityProvider) {
+  if (process.env.INSTANA_DEBUG || process.env.INSTANA_LOG_LEVEL === 'debug') {
```
kirrg001, line 183: I like this check a lot 👍 

S: It would be cool if you could raise a second PR to add a utility for that. We use this check in many places e.g.

https://github.com/search?q=repo%3Ainstana%2Fnodejs%20%22if%20(process.env.INSTANA_DEBUG%20%7C%7C%20process.env.INSTANA_LOG_LEVEL%22&type=code

packages/core/src/tracing/index.js
kirrg001: We could simplify the whole function by printing all values if `--require`, `--import` or `--experimental-loader` is used.

e.g. customer uses `node --import X --import X app.js`

**Advantages:**
- less complicated logic, less checks
- covers the use case that if a customer uses multiple import values to dig down potential problems -> conflicts
- this is only debug information
- covers more use cases (such as the case I showed)

The output would be:

```
console.debug(`The App is using the following preload flags: ${preloadFlags}`);
```

packages/core/src/tracing/index.js
abhilash-sivan: Yeah sure! 👍

packages/core/src/tracing/index.js
abhilash-sivan: Yeah perfect 👍

packages/core/src/util/getPreloadFlags.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+'use strict';
+
+exports.getPreloadFlags = function getPreloadFlags() {
+  if (
+    process.env.NODE_OPTIONS?.includes('--require')
+    || process.env.NODE_OPTIONS?.includes('--import')
+    || process.env.NODE_OPTIONS?.includes('--experimental-loader')) {
+      return process.env.NODE_OPTIONS;
```
kirrg001, line 12: NODE_OPTIONS can contain [multiple information.](https://nodejs.org/api/cli.html#node_optionsoptions).

Possible update:

```js
const flags = ['--require', '--import', '--experimental-loader'];
  
  if (process.env.NODE_OPTIONS) {
    const relevantFlags = process.env.NODE_OPTIONS.split(' ').filter(option =>
      flags.some(flag => option.includes(flag))
    );
    if (relevantFlags.length > 0) {
      return relevantFlags.join(' ');
    }
  }
  
  if (process.execArgv.length > 0) {
    const result = process.execArgv.find(arg =>
      flags.some(flag => arg.includes(flag))
    );
    return result || 'noFlags';
  } else {
    return 'noFlags';
  }
  ````

Comment (abhilash-sivan) [reply packages/core/src/util/getPreloadFlags.js:12]: > NODE_OPTIONS can contain [multiple information.](https://nodejs.org/api/cli.html#node_optionsoptions).
>

Thanks for sharing.  I will update the same according to this suggestion 👍

Comment (abhilash-sivan) [reply packages/core/src/util/getPreloadFlags.js:12]: since Array.prototype.some() function returns a boolean, we might have to make some changes in the returning string if the whole argument list is required in the debug logs. 
But as you pointed out, the arguments and NODE_OPTIONS can contain more information that might not be relevant to us in the debug log.
Let me think of other options.

Comment (kirrg001) [reply packages/core/src/util/getPreloadFlags.js:12]: Any solution works :)

Comment (abhilash-sivan) [reply packages/core/src/util/getPreloadFlags.js]: Instead of printing out the whole, now logging only the flags (--import, --require, etc)
Also added a check for `@instana` keyword which will be sufficient to verify that one/some of instana packages are passed in the argument list or as NODE_OPTIONS

File: packages/core/src/util/getPreloadFlags.js [hunk 8]
```diff
@@ -0,0 +1,34 @@
+/*
+ * (c) Copyright IBM Corp. 2024
+ */
+
+'use strict';
+
+exports.getPreloadFlags = function getPreloadFlags() {
```
kirrg001, line 7: issue: I think it would have been better to add a quick unit test. The function now prints:

> $ INSTANA_DEBUG=true node --require '@instana/collector/src/immediate.js' ./dummy-app/src/index.js
> The App is using the following preload flags: --require

I expected to see:

> The App is using the following preload flags: --require @instana/collector/src/immediate.js

We need the type + the value.

And as discussed:

> const instanaKeyword = '@instana';

We don't really need that? We can just print all --require and all --import flags etc. That will help debugging conflicts with other preload flags. Or why did you re-add this again?

kirrg001 (changes requested):

packages/core/src/util/getPreloadFlags.js
abhilash-sivan: > We don't really need that? We can just print all --require and all --import flags etc. 
> 
I thought of filtering our packages alone. No other reasons. But as you mentioned we can print all the preloaded packages because that can help debug the conflicts. I will update the  code as per the suggestion

kirrg001 (approved):

abhilash-sivan:
LGTM!

PR #1269, opened 2024-08-12
build: moved superagent to optional dependencies

aryamohanan:
The most recent version of Superagent v10 fails to install with Node 16.
related issue https://github.com/ladjs/superagent/issues/1814

### DIFF
--- a/package.json
+++ b/package.json
@@ -225,7 +225,6 @@
     "square-calc": "3.2.1",
     "square-calc-v2": "npm:square-calc@2.4.0",
     "stealthy-require": "1.1.1",
-    "superagent": "^10.0.0",
     "tedious": "^17.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -242,7 +241,8 @@
     "kafka-avro": "^3.2.0",
     "node-rdkafka": "^3.1.0",
     "pg-native": "^3.1.0",
-    "prisma": "^5.18.0"
+    "prisma": "^5.18.0",
+    "superagent": "^10.0.0"
```
kirrg001, line 245: I guess you need to skip the superagent test < Node v18?

aryamohanan: Yes, skipped the tests

kirrg001 (approved):

PR #1271, opened 2024-08-12
chore: fix version of sonarqube-scanner to v3.5 

abhilash-sivan:
fix version of sonarqube-scanner to v3.5  as they introduced br…aking changes in v4
 - they removed support for node v16 in sonarqube-scanner v4
 - our test pipeline was failing with some authentication error

refs I[NSTA-12663](https://jsw.ibm.com/browse/INSTA-12663)

### DIFF
--- a/.tekton/tasks/tasks-results.yaml
+++ b/.tekton/tasks/tasks-results.yaml
@@ -26,16 +26,6 @@ spec:
             secretKeyRef:
               name: $(params.continuous-delivery-context-secret)
               key: "SONAR_TOKEN"          
-        - name: SONAR_PWD
-          valueFrom:
-            secretKeyRef:
-              name: $(params.continuous-delivery-context-secret)
-              key: "SONAR_PWD"          
-        - name: SONAR_LOGIN
-          valueFrom:
-            secretKeyRef:
-              name: $(params.continuous-delivery-context-secret)
-              key: "SONAR_LOGIN"          
       script: |
         #!/bin/bash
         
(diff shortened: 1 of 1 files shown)


.tekton/tasks/tasks-results.yaml
```diff
@@ -55,7 +45,7 @@ spec:
           npx eslint packages/ -f json > eslint-report.json || true
 
           echo "Install sonarcube scanner..."
-          npm install -g sonarqube-scanner
+          npm install -g sonarqube-scanner@3.5
```
aryamohanan, line 48: Were you able to figure out what was causing the issues with v4? I believe we're using v20 for coverage.

kirrg001 (commented):
Yeah they [silently](https://github.com/SonarSource/sonar-scanner-npm/pull/119/files#diff-aa09ea4be3c9607f4bad376ab62723493b90fb3054156bb0c8cb2ff88686ed5cL6) dropped Node 16.

But why is the authentication suddenly broken?

We can remove SONAR_PWD an SONAR_LOGIN from our secure manager 👍

.tekton/tasks/tasks-results.yaml
abhilash-sivan: No, I wasn't able to do that. 
We can consider this as an immediate fix and can find out the issue with node20 later. Only then we can update the version to the latest

kirrg001 (approved):
> We can consider this as an immediate fix and can find out the issue

Oki 👍

.tekton/tasks/tasks-results.yaml
abhilash-sivan: Maybe we can create a ticket for the version update

aryamohanan (approved):

PR #1275, opened 2024-08-14
ci: reduced connection timeout in  aws lambda layer publish

abhilash-sivan:


kirrg001 (approved):

kirrg001:
Increasing the timeout did not work because we were running into a signature error from AWS CLI.

> An error occurred (InvalidSignatureException) when calling the PublishLayerVersion operation: Signature expired: 20240813T133841Z is now earlier than 20240813T140553Z (20240813T141053Z - 5 min.)

PR #1276, opened 2024-08-14
test: extended the  tests for native module retry

aryamohanan:
Extended the test to check the copying of precompiled binaries, related https://github.com/instana/nodejs/pull/1265
Added comments to describe the problems with the existing test setup.

### DIFF
--- a/packages/collector/test/nativeModuleRetry/test.js
+++ b/packages/collector/test/nativeModuleRetry/test.js
@@ -6,9 +6,10 @@
 'use strict';
 
 const { expect } = require('chai');
-const fs = require('node:fs/promises');
+const fsPromise = require('node:fs/promises');
 const os = require('os');
 const path = require('path');
+const fs = require('fs');
 
 const config = require('../../../core/test/config');
 const { retry } = require('../../../core/test/test_util');
@@ -67,11 +68,11 @@ function runCopyPrecompiledForNativeAddonTest(agentControls, opts) {
 
   const rename = async (oldPath, newPath) => {
     try {
-      await fs.rename(oldPath, newPath);
+      await fsPromise.rename(oldPath, newPath);
     } catch (err) {
-      if (err.code === 'EXDEV') {
+      if (err.code === 'EXDEV' || err.code === 'ENOTEMPTY') {
         await copyFolder(oldPath, newPath);
-        await fs.rmdir(oldPath, { recursive: true });
+        await fsPromise.rm(oldPath, { recursive: true });
       } else {
         throw err;
       }
@@ -79,8 +80,8 @@ function runCopyPrecompiledForNativeAddonTest(agentControls, opts) {
   };
 
   async function copyFolder(src, dest) {
-    await fs.mkdir(dest, { recursive: true });
-    const entries = await fs.readdir(src, { withFileTypes: true });
+    await fsPromise.mkdir(dest, { recursive: true });
+    const entries = await fsPromise.readdir(src, { withFileTypes: true });
 
     const copyPromises = entries.map(async entry => {
       const srcPath = `${src}/${entry.name}`;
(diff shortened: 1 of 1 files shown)


packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -6,15 +6,49 @@
...
+ *
+ * With `npm workspaces`, dependencies are installed in the root `node_modules` folder instead of individual
+ * package `node_modules` folders. This can complicate managing native modules and their precompiled binaries,
+ * as the fallback mechanism must copy to the correct path. This test verifying that precompiled binaries are copied
+ * and loaded correctly to the @instana/shared-metrics/node_modules folder.
+ *
+ * The test operates as follows:
+ *
+ * 1. **Setup**:
+ *      - Temporarily moves the installed native module to a backup directory to simulate its absence.
```
kirrg001, line 36: from the .... root node_modules? And why do we do that?

aryamohanan: npm workspaces installs event-loop-stats on the root node_modules folder.

kirrg001: Yes. Why do we have to move the root modules dependencies if we operate on shared metrics node_modules?

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -6,15 +6,49 @@
...
+ *
+ * 2. **Test Cases**:
```
kirrg001, line 38: S: No need to describe the test case. We only want to describe how the test is designed.

aryamohanan: updated the description

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -6,15 +6,48 @@
...
 const config = require('../../../core/test/config');
 const { retry } = require('../../../core/test/test_util');
 const ProcessControls = require('../test_util/ProcessControls');
 const globalAgent = require('../globalAgent');
 
+/**
+ * Test suite for verifying the fallback mechanism for loading native add-ons.
+ *
+ * This test ensures that if a native add-on (such as event-loop-stats or gcstats.js) cannot be loaded initially,
+ * the system correctly falls back to using precompiled versions of these add-ons. This ensures the application remains
```
kirrg001, line 23:
```suggestion
 * the system correctly falls back to using precompiled versions of these add-ons. This ensures the application
```

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -6,15 +6,48 @@
...
+ * functions as expected even if the native module is missing or cannot be compiled. The function `attemptRequire`,
+ * defined here: https://github.com/instana/nodejs/blob/main/packages/shared-metrics/src/util/nativeModuleRetry.js#L111
```
kirrg001, line 25: S. Never use main in a github link. Main can change.

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -6,15 +6,48 @@
...
+ * is used for the initial attempt to load the native add-on module. It tries to require the native module from the
+ * root. If this attempt fails, the fallback mechanism is triggered.
```
kirrg001, line 27:
```suggestion
 * root for this test, because we use npm workspaces and the dependencies are hoisted to the root. If this attempt fails, the fallback mechanism is triggered.
```

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -6,15 +6,48 @@
...
+ *
+ * **Issue with npm Workspaces**:
```
kirrg001, line 29:
```suggestion
 * **Design problem of the test**:
```

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -6,15 +6,48 @@
...
+ *
+ * With `npm workspaces`, dependencies are installed in the root `node_modules` folder instead of individual
+ * package `node_modules` folders. This can complicate managing native modules and their precompiled binaries,
```
kirrg001, line 32:
```suggestion
 * package `node_modules` folders. This can complicate managing the installed native modules and their copied precompiled binaries,
```

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -6,15 +6,48 @@
...
+ * as the fallback mechanism must copy to the correct path. This test verifying that precompiled binaries are copied
+ * and loaded correctly to the @instana/shared-metrics/node_modules folder.
```
kirrg001, line 34:
```suggestion
 * and loaded correctly to the @instana/shared-metrics/node_modules folder. 
 
 The test is **mis-designed**, because the test should not use the local dev environment to check the presence of precompiled binaries. Instead it should create a tmp folder.
```

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -6,15 +6,48 @@
...
+ *
+ * 1. **Setup**:
+ *    - The installed native module is temporarily moved to a backup directory to simulate its absence. As the system
+ *      first tries to require the native module from the root, so we relocate the modules.
+ *
+ * 2. **Fallback Mechanism**:
+ *    - The test checks if the fallback mechanism correctly copies the precompiled binaries to the appropriate directory
+ *      when the native module is not found.
+ *
+ * Note: Currently, we are not addressing the npm workspace issues with prebuilds because there is a
```
kirrg001, line 46:
```suggestion
 * Note: Currently, we are not addressing the wrong test design because there will be a
```

packages/collector/test/nativeModuleRetry/test.js
```diff
@@ -6,15 +6,48 @@
...
+ * potential migration to a tool like prebuildify in the future, which might handle these issues differently.
```
kirrg001, line 47:
```suggestion
 * migration to a tool called prebuildify in the near future, which will change the whole feature.
```

packages/collector/test/nativeModuleRetry/test.js
aryamohanan: removed the link.

kirrg001 (approved):

PR #1286, opened 2024-08-23
test(lambda): enabled span batching for big calls

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-10320

https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-collector-configuration#automatic-aggregation-of-short-exit-calls

### DIFF
--- a/packages/aws-lambda/test/multiple_data/test.js
+++ b/packages/aws-lambda/test/multiple_data/test.js
@@ -94,7 +94,7 @@ describe('multiple data lambda handler', function () {
     });
   });
 
-  describe('with 100 iterations', function () {
+  describe('[batching disabled] with 100 iterations', function () {
     this.timeout(config.getTestTimeout() * 2);
     let control;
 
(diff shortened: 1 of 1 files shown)


packages/aws-lambda/test/multiple_data/test.js
```diff
@@ -152,6 +160,87 @@ describe('multiple data lambda handler', function () {
...
+
+    before(async () => {
+      control = new Control({
+        faasRuntimePath: path.join(__dirname, '../runtime_mock'),
+        handlerDefinitionPath: path.join(__dirname, './lambda'),
+        startBackend: true,
+        env: {
+          INSTANA_AGENT_KEY: instanaAgentKey,
+          WITH_CONFIG: 'true',
+          INSTANA_NUMBER_OF_ITERATIONS: 100,
```
aryamohanan, line 175: Suggestion: We can consider adding this env variables in our public doc.

packages/aws-lambda/test/multiple_data/test.js
```diff
@@ -152,6 +160,87 @@ describe('multiple data lambda handler', function () {
...
+        .then(() => {
+          return retry(
+            () => {
+              return Promise.all([
+                //
+                control.getSpans(),
+                control.getRawBundles(),
+                control.getRawSpanArrays()
+              ]).then(([spans, rawBundles, rawSpanArrays]) => {
+                // 1 X bundle requestat the end of the lambda fn
```
aryamohanan, line 213: nit pick: typo in comment

aryamohanan (approved):
LGTM

packages/aws-lambda/test/multiple_data/test.js
kirrg001: The comment is on the wrong line I think?

INSTANA_SPANBATCHING_ENABLED is documented
https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-collector-configuration#automatic-aggregation-of-short-exit-calls

But the docs are not so good.

I think we need to re-work the feature first and then document only whats needed.
Because documenting INSTANA_DEV_MIN_DELAY_BEFORE_SENDING_SPANS is not right.

PR #1292, opened 2024-08-28
feat: added support for ioredis cluster

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-9475

### DIFF
--- a/packages/collector/test/tracing/database/ioredis/connect-via/cluster.js
+++ b/packages/collector/test/tracing/database/ioredis/connect-via/cluster.js
@@ -1,0 +1,51 @@
+/*
+ * (c) Copyright IBM Corp. 2024
+ */
+
+'use strict';
+
+// NOTE: We run the tests locally and on CI against azure redis cluster.
+// NOTE: We cannot run redis cluster on Tekton https://github.com/bitnami/charts/issues/28894
+// NOTE: We cannot use a docker based redis cluster at the moment!
+//       See https://github.com/redis/node-redis/issues/2815.
+//       These commands are useful as soon as we switch to a docker based cluster.
+//       node bin/start-test-containers.js --redis-node-0 --redis-node-1 --redis-node-2
+//       docker exec -it 2aaaac7b9112 redis-cli -p 6379 cluster info
+module.exports = async function connect(ioredis, log) {
+  if (!process.env.AZURE_REDIS_CLUSTER || !process.env.AZURE_REDIS_CLUSTER_PWD) {
+    log(
+      'Please set the environment variables AZURE_REDIS_CLUSTER and AZURE_REDIS_CLUSTER_PWD ' +
+        'to connect to the cloud redis cluster.'
+    );
+
+    process.exit(1);
+  }
+
+  const hostAndPort = process.env.AZURE_REDIS_CLUSTER.split(':');
+  const cluster = new ioredis.Cluster(
+    [
+      {
+        host: hostAndPort[0],
+        port: hostAndPort[1]
+      }
+    ],
+    {
+      redisOptions: {
+        tls: true,
+        password: process.env.AZURE_REDIS_CLUSTER_PWD,
+        connectTimeout: 10000
+      },
+      retryDelayOnFailover: 1000,
+      maxRetriesPerRequest: 10
+    }
+  );
+
+  log(`Connecting to cluster host: ${hostAndPort[0]}, port: ${hostAndPort[1]}.`);
+
+  return new Promise(resolve => {
+    cluster.on('ready', () => {
+      log('Connected to cluster.');
+      resolve({ connection: cluster });
+    });
+  });
+};
--- a/packages/collector/test/tracing/database/ioredis/test.js
+++ b/packages/collector/test/tracing/database/ioredis/test.js
@@ -24,14 +24,23 @@ const globalAgent = require('../../../globalAgent');
 const mochaSuiteFn =
   semver.gte(process.versions.node, '14.0.0') && supportedVersion(process.versions.node) ? describe : describe.skip;
 
-// Please run this command on the root folder to start the redis instance:
+function checkConnection(span, setupType) {
+  if (setupType === 'cluster') {
+    // Internal IP of azure cluster.
+    expect(span.data.redis.connection).to.equal(process.env.AZURE_REDIS_CLUSTER);
+  } else {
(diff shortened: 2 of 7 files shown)


packages/collector/test/tracing/database/ioredis/test.js
```diff
@@ -24,596 +24,633 @@ const globalAgent = require('../../../globalAgent');
...
+  }
+}
+
+// Please run this command on the root folder to start the single redis instance (default):
+// node bin/start-test-containers.js --redis
+//
+// Please set the environment variables to run the tests against azure redis cluster:
+//    export AZURE_REDIS_CLUSTER=team-nodejs-redis-cluster-tekton.redis.cache.windows.net:6380
+//    export AZURE_REDIS_CLUSTER_PWD=
+['default', 'cluster'].forEach(setupType => {
```
kirrg001, line 41: The test kinda remained the same. Only:

- wrapped the tests to run all tests with a cluster
- added the check connection check
- added span length assertion, which discovered a bug (explained on the instrumentation)

packages/core/src/tracing/instrumentation/database/ioredis.js
```diff
@@ -29,10 +29,37 @@ exports.init = function init() {
...
+  Object.defineProperty(ioredis, 'Cluster', {
+    value: InstanaClusterClass,
+    writable: false,
+    configurable: false,
+    enumerable: true
+  });
+
+  // We cannot register `ioredis.Cluster.prototype.sendCommand` otherwise we produce two spans for the same command.
+  // Because `ioredis.prototype.sendCommand` is triggerd when using a cluster, but `client.isCluster` is false.
+  // Thats why we override the Cluster Class and extract the connection string.
```
kirrg001, line 56: Please read this comment. Makes the changes more clear.

packages/core/src/tracing/instrumentation/database/ioredis.js
```diff
@@ -49,18 +76,19 @@ function instrumentSendCommand(original) {
...
+    // ".multi()" commands could actually be recorded as multiple spans, but we ONLY want to record ONE
+    // batched span considering that a multi call represents a transaction.
+    // The same is true for pipeline calls, but they have a slightly different semantic.
     if (
       parentSpan.n === exports.spanName &&
-      (parentSpan.data.redis.command === 'multi' || parentSpan.data.redis.command === 'pipeline') &&
-      // the multi call is handled in instrumentMultiCommand but since multi is also send to Redis it will also
-      // trigger instrumentSendCommand, which is why we filter it out.
-      command.name !== 'multi'
+      (parentSpan.data.redis.command === 'multi' || parentSpan.data.redis.command === 'pipeline')
```
kirrg001, line 84: Here was a bug IMO:

For multi, we only want to capture one span. As the original comment describes it:

> multi commands could actually be recorded as multiple spans, but we only want to record one
   batched span considering that a multi call represents a transaction.
   
But we recorded 3 spans, not one. Because the span length assertion was missing, we did not know this.

I removed

> && command.name !== 'multi'

Which was wrong.

We want to NOT create a span when the parent span is a redis span (the multi/pipeline execution) and the command is MULTI or PIPELINE.
And then we skip recording sub commands if command name is multi or pipeline (line 87)

packages/core/src/tracing/instrumentation/database/ioredis.js
```diff
@@ -71,7 +99,7 @@ function instrumentSendCommand(original) {
       const span = cls.startSpan(exports.spanName, constants.EXIT);
       span.stack = tracingUtil.getStackTrace(wrappedInternalSendCommand);
       span.data.redis = {
-        connection: `${client.options.host}:${client.options.port}`,
+        connection: clusterConnectionString || `${client.options.host}:${client.options.port}`,
```
kirrg001, line 102: This is a little bit "hacky" because if a customer uses both single client and cluster connection, the connection string could be wrong. But I would ignore the edge case.

abhilash-sivan: That is a likely, unlikely situation. We can ignore it for now.

abhilash-sivan (approved):

packages/core/src/tracing/instrumentation/database/ioredis.js
aryamohanan: I agree as well. If an issue arises in the future, we can address it then. We know where to make the fix. 😄

packages/collector/test/tracing/database/ioredis/connect-via/cluster.js
```diff
@@ -0,0 +1,51 @@
...
+
+// NOTE: We run the tests locally and on CI against azure redis cluster.
+// NOTE: We cannot run redis cluster on Tekton https://github.com/bitnami/charts/issues/28894
+// NOTE: We cannot use a docker based redis cluster at the moment!
+//       See https://github.com/redis/node-redis/issues/2815.
+//       These commands are useful as soon as we switch to a docker based cluster.
+//       node bin/start-test-containers.js --redis-node-0 --redis-node-1 --redis-node-2
+//       docker exec -it 2aaaac7b9112 redis-cli -p 6379 cluster info
+module.exports = async function connect(ioredis, log) {
+  if (!process.env.AZURE_REDIS_CLUSTER || !process.env.AZURE_REDIS_CLUSTER_PWD) {
```
aryamohanan, line 15: While running the tests locally, I initially encountered the following issue:

> [ioredis] Unhandled error event: ClusterAllFailedError: Failed to refresh slots cache.

The issue seems to have resolved itself after some time. It might be related to the cluster since we use the same cluster for both Redis and ioredis testing. Running tests in parallel could potentially cause conflicts. We should consider alternative solutions, such as using namespaces, to prevent this issue in the future.

packages/core/src/tracing/instrumentation/database/ioredis.js
aryamohanan: Its really good we added the span count check in our tests 👍

aryamohanan (approved):
LGTM

packages/collector/test/tracing/database/ioredis/connect-via/cluster.js
kirrg001: https://jsw.ibm.com/browse/INSTA-14042

PR #1293, opened 2024-08-28
chore: excluded misc folder from sonar cube report

kirrg001:
I saw this file in our report, which reduces our test coverage:

> misc/eslint-plugin/lib/rules/no-unsafe-require.js

I don't know what the difference is between:

> sonar.exclusions=...
> sonar.coverage.exclusions=...

I was too lazy to investigate. Feel free to do it!


abhilash-sivan:
`sonar.exclusions`: Excludes files from all types of analyses.
`sonar.coverage.exclusions`: Excludes files only from code coverage analysis, but still includes them in other analyses.

detailed explanation: https://brainly.com/question/43500185

aryamohanan (approved):

PR #1297, opened 2024-08-29
feat: added optional root exit spans

abhilash-sivan:
Allow root exit span in all environments
- [x] feature implementation
- [x] unit tests
- [x] integration tests
- [x] internal doc updation
- [x] ~public doc updation~ Public documentation will be part of the v4 ticket.
- [x] db2
- [x] graphql (nothing todo)
- [x] redis
- [x] bull
- [x] amqp
- [x] ioredis
- [x] Create ticket for ioredis/redis cluster allowRootExitSpan test (https://jsw.ibm.com/browse/INSTA-15876)
- [x] make build stable @abhilash-sivan 
- [x] change branch to v4 (before merging)

refs [INSTA-793](https://jsw.ibm.com/browse/INSTA-793)


packages/core/src/tracing/cls.js
```diff
@@ -48,6 +50,7 @@ function init(config, _processIdentityProvider) {
     serviceName = config.serviceName;
   }
   processIdentityProvider = _processIdentityProvider;
+  allowRootExitSpan = process.env.INSTANA_ALLOW_ROOT_EXIT_SPAN === '1' || config?.tracing?.allowRootExitSpan;
```
abhilash-sivan, line 53: This variable is introduced here because the config is already imported in here as a param. So there is no need to import it again

packages/core/src/tracing/cls.js
```diff
@@ -565,6 +568,11 @@ function skipExitTracing(options) {
     return true;
   }
 
+  if (allowRootExitSpan) {
```
abhilash-sivan, line 571: Only this check is required as the suppresed check is already covered in the above line. If tracing is suppressed then there is no point in sending spans according to `allowRootExitSpan`

packages/core/src/tracing/cls.js
kirrg001, line 53: issue: If we normalize the configuration options, there is no need to check for the env variable here again.

kirrg001 (changes requested):
Please add one integration test for the SDK.
e.g. `packages/collector/test/tracing/sdk/test.js`

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -186,12 +186,7 @@ function instrument(coreModule, forceHttps) {
     // packages/core/src/tracing/clsHooked/unset.js#storeReducedSpan.
     const parentSpan = cls.getCurrentSpan() || cls.getReducedSpan();
 
-    if (
-      skipTracingResult.skip ||
-      !parentSpan ||
-      constants.isExitSpan(parentSpan) ||
-      shouldBeBypassed(parentSpan, options)
-    ) {
+    if (skipTracingResult.skip || shouldBeBypassed(parentSpan, options)) {
```
kirrg001, line 189:
```
const resp = cls.skipExitTracing({
  checkReducedSpan: true,
  extendedResponse: true,
  skipIsTracing: true
})
...
if (skipTracingResult.skip || shouldBeBypassed(resp.parentSpan, options)) {
```

abhilash-sivan: ```shouldBeBypassed(resp.parentSpan, options)``` this check is specific to httpClient alone. IMO we can keep this in here itself. no need to move to cls

packages/core/src/tracing/instrumentation/messaging/bull.js
```diff
@@ -45,6 +45,7 @@ function shimJobCreate(originalJobCreate) {
       skipIsTracing
     });
 
+    // TODO: -> check and remove skipTracingResult.isExitSpan check for allowRootExitSpan
```
abhilash-sivan, line 48:
```
if (
      skipTracingResult.skip ||
      skipTracingResult.isExitSpan ||
      (!parentSpan && !repeatableJob) ||
      repeatableJobIsSuppressed
    ) {}
    ```
The check skipTracingResult.isExitSpan can be removed. need to verify

File: packages/collector/test/tracing/sdk/opt_in/test.js [hunk 5]
```diff
@@ -0,0 +1,78 @@
...
+
+const supportedVersion = require('@instana/core').tracing.supportedVersion;
+const config = require('../../../../../core/test/config');
+const ProcessControls = require('../../../test_util/ProcessControls');
+const globalAgent = require('../../../globalAgent');
+const { retry, delay } = require('../../../../../core/test/test_util');
+
+const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : describe.skip;
+
+mochaSuiteFn('tracing/sdk/opt_in', function () {
```
kirrg001, line 18: suggestions: 

- folder name "sdk/rootExitSpans"
- app names: app_default.js, app_with_entry.js
- I assume you wanted to proof that both cases still work?

packages/collector/test/tracing/protocols/http/client/test.js
```diff
@@ -93,6 +93,56 @@ mochaSuiteFn('tracing/http client', function () {
       });
     });
   });
+
+  describe('OPT-IN APP CASE 1', function () {
```
kirrg001, line 97: suggestion: Allow Root Exist Spans Case 1

packages/collector/test/tracing/protocols/http/client/optInApp.js
```diff
@@ -0,0 +1,44 @@
...
+process.on('SIGTERM', () => {
+  process.disconnect();
+  process.exit(0);
+});
+
+require('../../../../..')();
+const { delay } = require('../../../../../../core/test/test_util');
+const nodeFetch = require('node-fetch-v2');
+
+const main = async () => {
```
kirrg001, line 17: qs: What is the difference between this test and the sdk/opt_in/without_sdk.js?

packages/core/src/tracing/cls.js
```diff
@@ -549,31 +552,50 @@ function skipExitTracing(options) {
...
+    if (opts.extendedResponse) return { skip: true, suppressed, isExitSpan: isExitSpanResult, parentSpan };
     return true;
   }
 
+  if (allowRootExitSpan) {
+    if (opts.extendedResponse) return { skip: false, suppressed, isExitSpan: isExitSpanResult, parentSpan };
+    else return false;
+  }
+
+  if (opts.checkReducedSpan && (!parentSpan || isExitSpan(parentSpan))) {
```
kirrg001, line 584: suggestion: These conditions become quite complex. Could you please describe the cases + why? Then it will be easier to review 👍

packages/core/src/tracing/opentelemetry-instrumentations/wrap.js
```diff
@@ -70,6 +70,7 @@ module.exports.init = (_config, cls) => {
       return;
     }
 
+    // just for checking I added INTERMEDIATE SPAN here, I am not sure of the functionality
```
kirrg001, line 73: ?

packages/core/src/util/normalizeConfig.js
```diff
@@ -220,6 +222,7 @@ function normalizeTracingConfig(config) {
   normalizeSpanBatchingEnabled(config);
   normalizeDisableW3cTraceCorrelation(config);
   normalizeTracingKafka(config);
+  normalizeAllowExitRootSpan(config);
```
kirrg001, line 225: suggestion: normalizeAllowRootExitSpan

packages/core/src/tracing/instrumentation/protocols/http2Client.js
```diff
@@ -64,13 +64,14 @@ function instrumentClientHttp2Session(clientHttp2Session) {
   const originalRequest = clientHttp2Session.request;
```
kirrg001, line 64: Putting it here.

issue: there is another usage of getReducedSpan

> const parentSpan = cls.getCurrentSpan(true) || cls.getReducedSpan(true);

In graphql instrumentation.

I assume this won't work if a customer sets "INSTANA_ALLOW_ROOT_EXIT_SPAN=true` and we ignore it in this place.

packages/core/src/tracing/cls.js
```diff
@@ -549,31 +552,50 @@ function skipExitTracing(options) {
...
+    if (opts.extendedResponse) return { skip: true, suppressed, isExitSpan: isExitSpanResult, parentSpan };
     else return true;
   }
 
   const skipIsActive = opts.isActive === false;
   const skipIsTracing = !opts.skipIsTracing ? !isTracing() : false;
   const skip = skipIsActive || skipIsTracing;
-  if (opts.extendedResponse) return { skip, suppressed, isExitSpan: isExitSpanResult };
+
+  if (opts.extendedResponse) return { skip, suppressed, isExitSpan: isExitSpanResult, parentSpan };
```
kirrg001, line 598: Putting it here.

In bull.js we can use the returned parentSpan.

>     const parentSpan = cls.getCurrentSpan();
>    const skipTracingResult = cls.skipExitTracing({

kirrg001: Same for ampq.js

>   const skipTracingResult = cls.skipExitTracing({ isActive, extendedResponse: true, skipParentSpanCheck: true });
>  const parentSpan = cls.getCurrentSpan();
>  const isExitSpan = skipTracingResult.isExitSpan;

kirrg001: As well as: redis.js, ioredis.js, not sure about db2 🤔

packages/collector/test/tracing/sdk/opt_in/test.js
```diff
@@ -0,0 +1,78 @@
...
+    after(async () => {
+      await agentControls.stop();
+    });
+
+    it('should collect spans including sdk wrap', async () => {
+      await delay(100);
+
+      await retry(async () => {
+        const spans = await globalAgent.instance.getSpans();
+        expect(spans.length).to.equal(2);
```
kirrg001, line 48: issue: Please add span asserts.

packages/core/src/tracing/cls.js
```diff
@@ -549,31 +552,50 @@ function skipExitTracing(options) {
       isActive: true,
       extendedResponse: false,
       skipParentSpanCheck: false,
-      skipIsTracing: false
+      skipIsTracing: false,
+      checkReducedSpan: false
```
aryamohanan, line 556: Clarify what `checkReducedSpan` does, as it’s listed in the code but not documented in the comments.

packages/core/src/tracing/cls.js
```diff
@@ -549,31 +552,50 @@ function skipExitTracing(options) {
...
     },
     options
   );
 
-  const parentSpan = getCurrentSpan();
+  let parentSpan = getCurrentSpan();
```
aryamohanan, line 561:
```suggestion
  let parentSpan = getCurrentSpan();
  // If there is no active entry span, we fall back to the reduced span of the most recent entry span.
  // See comment in packages/core/src/tracing/clsHooked/unset.js#storeReducedSpan.
if (opts.checkReducedSpan && !parentSpan) {
  parentSpan = getReducedSpan();
}
```
The getCurrentSpan() is called twice in the condition. To avoid duplicate function calls, you could cache its result before checking for the reduced span:

packages/collector/test/tracing/sdk/opt_in/test.js
abhilash-sivan: done

packages/collector/test/tracing/protocols/http/client/optInApp.js
abhilash-sivan: Both are similar but here the environment variable is used to enable root exit spans whereas in sdk/opt_in/without_sdk.js, the configuration variable is used.

packages/core/src/tracing/opentelemetry-instrumentations/wrap.js
abhilash-sivan: removed!

packages/core/src/tracing/cls.js
abhilash-sivan: redis and ioredis does not use the extendedResponse attribute and thus the result from skipExitTracing function will be a boolean. So can we leave it ?

In db2, intead of falling back to reducedSpan     `const parentSpan = cls.getCurrentSpan() || prepareCallParentSpan;
` prepareCallParentSpan is used. So the reducedSpan logic may not work here

packages/collector/test/tracing/protocols/http/client/optInApp.js
```diff
@@ -0,0 +1,44 @@
...
+
+// NOTE: c8 bug https://github.com/bcoe/c8/issues/166
+process.on('SIGTERM', () => {
+  process.disconnect();
+  process.exit(0);
+});
+
+require('../../../../..')();
+const { delay } = require('../../../../../../core/test/test_util');
+const nodeFetch = require('node-fetch-v2');
```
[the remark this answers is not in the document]
abhilash-sivan: Native fetch support is only from node v18. I will update this

packages/core/test/tracing/cls_test.js
```diff
@@ -218,6 +218,35 @@ describe('tracing/cls', () => {
...
+          allowRootExitSpan: true
+        }
+      });
+
+      cls.startSpan('Vito-Corleone', constants.EXIT);
+      expect(cls.skipExitTracing()).to.equal(false);
+    });
+  });
+
+  it('must skip an exit span without a parent span', () => {
```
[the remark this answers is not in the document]
abhilash-sivan: This is the default condition. However in the test configured it using 
```
tracing: {
          allowRootExitSpan: false
        }
```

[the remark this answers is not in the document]
abhilash-sivan: > it should skip and omit an exit span if there is no parent span.

> 

Exactly

packages/core/src/tracing/instrumentation/protocols/http2Client.js
abhilash-sivan: In graphql,
const activeContext = storage.getStore() || (fallbackToSharedContext ? this.sharedContext : undefined);

In the code, since ```parentSpan = cls.getCurrentSpan(true) || cls.getReducedSpan(true);``` true is paased along, we might be using shared context if nothing is found in the current.
  

>    * Retrieves a value by key from the current CLS context (or a parent context), assuming the key/value pair has
>    * been set earlier in this context.
>    *
>    * When `fallbackToSharedContext` is set to true, the caller tries to access a parent context,
>    * which got remembered in a previous instrumentation.

This is somewhat different from other instrumentations as we were looking only at the current CLS context only.

packages/collector/test/tracing/sdk/opt_in/test.js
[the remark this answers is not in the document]
abhilash-sivan: done

packages/core/src/tracing/cls.js
kirrg001: There is a list of attributes described here:
https://github.com/instana/nodejs/pull/1297/files

Please extend this list.

packages/core/src/tracing/instrumentation/protocols/http2Client.js
kirrg001: Rewrite to be something like:
>   // CASE 3: Third use case is to disable entry spans. If parentSpan is null.

packages/core/src/tracing/cls.js
kirrg001: Three cases:
1. no parent span
2. entry span
3. exitspan

Idea: add this as an integration test.
app_mixed.js

worker.js
```
instana = {allowNoEntrySpan:true}

// trace 1
async ioredis.send 

// trace 2
async instana.async.sdk.startEntry
async ioredis.send
```

packages/core/test/tracing/cls_test.js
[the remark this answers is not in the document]
kirrg001: Can be ignored.

packages/collector/test/tracing/messaging/bull/test.js
```diff
@@ -53,6 +54,44 @@ mochaSuiteFn('tracing/messaging/bull', function () {
...
+      await agentControls.clearReceivedTraceData();
+    });
+
+    it('must trace', async function () {
+      await retry(async () => {
+        const spans = await agentControls.getSpans();
+
+        // TODO: all other bull tests also produce a huge number of spans
+        //       https://jsw.ibm.com/browse/INSTA-15029
+        expect(spans.length).to.be.eql(6);
```
abhilash-sivan, line 84: After the fix, now we only have 1 exit span. That should be our actual expectation here!

packages/collector/test/tracing/messaging/bull/test.js
```diff
@@ -70,17 +69,14 @@ mochaSuiteFn.only('tracing/messaging/bull', function () {
       await retry(async () => {
         const spans = await agentControls.getSpans();
 
-        // 1 x bull send?
-
         // TODO: all other bull tests also produce a huge number of spans
```
kirrg001, line 72: @abhilash-sivan Comment can be moved to e.g. line 232? 

> TODO: https://jsw.ibm.com/browse/INSTA-15029

packages/collector/test/tracing/messaging/bull/test.js
```diff
@@ -169,12 +165,6 @@ mochaSuiteFn.only('tracing/messaging/bull', function () {
               response,
               apiPath,
               testId,
-              // 1 x node.http.server 1
-              // 1 x bull receive
-              // 1 x otel (?)
-              // 1 x node.http.client (?)
-              // 1 x redis
-              // 1 x bull sender
```
kirrg001: @abhilash-sivan Could we keep the comment please? I would love to know why there is 1 otel and one node.http.client span?

packages/core/src/tracing/instrumentation/database/ioredis.js
```diff
@@ -65,10 +65,10 @@ function instrumentSendCommand(original) {
       const parentSpanSubCommands = (parentSpan.data.redis.subCommands = parentSpan.data.redis.subCommands || []);
       parentSpanSubCommands.push(command.name);
     } else if (
-      (parentSpan && constants.isExitSpan(parentSpan)) ||
+      (!skipExitTracingResult.allowRootExitSpan && parentSpan && constants.isExitSpan(parentSpan)) ||
```
kirrg001, line 68: @abhilash-sivan Could you please add a comment **why**?

packages/core/src/tracing/instrumentation/database/ioredis.js
```diff
@@ -127,10 +127,15 @@ function instrumentPipelineCommand(original) {
...
   return function wrappedInternalMultiOrPipelineCommand() {
     const client = this;
-    const parentSpan = cls.getCurrentSpan();
+    const skipExitTracingResult = cls.skipExitTracing({ isActive, skipParentSpanCheck: true, extendedResponse: true });
+    const parentSpan = skipExitTracingResult.parentSpan;
 
     // NOTE: multiple redis transaction can have a parent ioredis call
-    if (cls.skipExitTracing({ isActive, skipParentSpanCheck: true }) || constants.isExitSpan(parentSpan)) {
+    if (
+      skipExitTracingResult.skip ||
```
kirrg001, line 135: @abhilash-sivan Same here :) I would love to the an explanation

packages/collector/test/tracing/messaging/bull/test.js
abhilash-sivan: yes. done

packages/collector/test/tracing/messaging/bull/test.js
abhilash-sivan: Yeah. Added back the comment.
I think this will be resolved with https://jsw.ibm.com/browse/INSTA-15029

packages/core/src/tracing/instrumentation/database/ioredis.js
abhilash-sivan: Added 👍

packages/core/src/tracing/instrumentation/database/ioredis.js
abhilash-sivan: Added 👍

packages/collector/test/tracing/messaging/bull/test.js
kirrg001: Ok agree 👍

packages/collector/test/tracing/messaging/bull/test.js
```diff
@@ -194,6 +240,8 @@ mochaSuiteFn('tracing/messaging/bull', function () {
           const withError = true;
           const urlWithParams = withError ? `${apiPath}&withError=true` : apiPath;
 
+          // TODO: all other bull tests also produce a huge number of spans
```
kirrg001, line 243: Why did you move it here? 🤔 I suggested 232 (231 now). 

Pls reword as suggested:

> TODO: Investigate no. of spans https://jsw.ibm.com/browse/INSTA-15029

abhilash-sivan: Updated to the mentioned line number.
We are having this issue for many assertions. But one comment is enough I guess.

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
```diff
@@ -80,19 +80,20 @@ function instrument() {
...
 
-    const skipTracingResult = cls.skipExitTracing({ isActive, extendedResponse: true, skipParentSpanCheck: true });
-
-    // If there is no active entry span, we fall back to the reduced span of the most recent entry span. See comment in
-    // packages/core/src/tracing/clsHooked/unset.js#storeReducedSpan.
-    const parentSpan = cls.getCurrentSpan() || cls.getReducedSpan();
+    const skipTracingResult = cls.skipExitTracing({
+      isActive,
+      extendedResponse: true,
+      skipParentSpanCheck: true,
```
kirrg001, line 86: issue: This looks wrong.

Before: we check the parent manually. Now: no one checks the parent.
Please break it with TDD.

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -179,19 +179,11 @@ function instrument(coreModule, forceHttps) {
       isActive,
       extendedResponse: true,
       skipParentSpanCheck: true,
```
kirrg001, line 181: issue: Same here. I might be wrong, but it looks not correct.

packages/core/src/tracing/instrumentation/protocols/http2Client.js
```diff
@@ -64,13 +64,14 @@ function instrumentClientHttp2Session(clientHttp2Session) {
...
     let w3cTraceContext = cls.getW3cTraceContext();
-    const skipTracingResult = cls.skipExitTracing({ isActive, extendedResponse: true, skipParentSpanCheck: true });
-
-    // If there is no active entry span, we fall back to the reduced span of the most recent entry span. See comment in
-    // packages/core/src/tracing/clsHooked/unset.js#storeReducedSpan.
-    const parentSpan = cls.getCurrentSpan() || cls.getReducedSpan();
+    const skipTracingResult = cls.skipExitTracing({
+      isActive,
+      extendedResponse: true,
+      skipParentSpanCheck: true,
```
kirrg001, line 70: issue: Same here. I might be wrong, but it looks not correct. Before: we check the parent manually. Now: nothing.

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
abhilash-sivan: https://github.com/instana/nodejs/blob/bb263a8c4ca5186bfda7153b1442d4d0b5a62d64/packages/core/src/tracing/cls.js#L598

I have moved the basic check for parentSpan to `cls.skipExitTracing` function.
So there is no need to check in again in the instrumentation.

packages/collector/test/tracing/protocols/http/client/rootExitSpanApp.js
```diff
@@ -0,0 +1,42 @@
+/*
```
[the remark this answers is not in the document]
abhilash-sivan: done

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
```diff
@@ -96,15 +95,19 @@ function instrument() {
...
+    ) {
       if (skipTracingResult.suppressed) {
         injectSuppressionHeader(originalArgs, w3cTraceContext);
       }
       return originalFetch.apply(originalThis, originalArgs);
     }
 
     return cls.ns.runAndReturn(() => {
-      const span = cls.startSpan('node.http.client', constants.EXIT, parentSpan.t, parentSpan.s);
+      const span = cls.startSpan('node.http.client', constants.EXIT, parentSpan?.t, parentSpan?.s);
```
kirrg001, line 110: LGTM 👍 
Could you pls add a comment:

// NOTE: We need to check for parentSpan existence, because of `allowExitRootSpan` being true

kirrg001: @abhilash-sivan

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
```diff
@@ -96,15 +95,19 @@ function instrument() {
       originalArgs[i] = arguments[i];
     }
 
-    if (skipTracingResult.skip) {
+    if (
+      skipTracingResult.skip ||
+      (!skipTracingResult.allowRootExitSpan && !parentSpan) ||
+      (!skipTracingResult.allowRootExitSpan && skipTracingResult.isExitSpan)
+    ) {
```
kirrg001, line 102: I can't follow this change. Too much in one commit without any explanation.

kirrg001: Discussed: Revert and create separate commit with proper explanation

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -199,7 +201,8 @@ function instrument(coreModule, forceHttps) {
     }
 
     cls.ns.run(() => {
-      const span = cls.startSpan('node.http.client', constants.EXIT, parentSpan.t, parentSpan.s);
+      // NOTE: added type-safety in case of parentSpan is undefined or null
```
kirrg001, line 204: > // NOTE: added type-safety in case of parentSpan is undefined or null

// NOTE: added type-safety in case of `allowExitRootSpan` being set to true

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
abhilash-sivan: Added a comment for the current change

packages/core/src/tracing/instrumentation/protocols/nativeFetch.js
abhilash-sivan: done

packages/collector/test/tracing/database/ioredis/test.js
```diff
@@ -42,8 +43,56 @@ function checkConnection(span, setupType) {
 //    export AZURE_REDIS_CLUSTER=team-nodejs-redis-cluster-tekton.redis.cache.windows.net:6380
 //    export AZURE_REDIS_CLUSTER_PWD=
 ['default', 'cluster'].forEach(setupType => {
+  if (setupType !== 'cluster') {
```
aryamohanan, line 46: Why the tests skip case for cluster?

packages/collector/test/tracing/database/ioredis/test.js
```diff
@@ -42,8 +43,56 @@ function checkConnection(span, setupType) {
...
+      globalAgent.setUpCleanUpHooks();
+      const agentControls = globalAgent.instance;
+      let controls;
+
+      before(async () => {
+        controls = new ProcessControls({
+          useGlobalAgent: true,
+          appPath: path.join(__dirname, 'allowRootExitSpanApp'),
+          env: {
+            REDIS_CLUSTER: setupType === 'cluster'
```
aryamohanan, line 59: 🤔

packages/collector/test/tracing/messaging/bull/test.js
```diff
@@ -128,6 +162,12 @@ mochaSuiteFn('tracing/messaging/bull', function () {
               response,
               apiPath,
               testId,
+              // 1 x node.http.server 1
```
aryamohanan, line 165: ❤️

packages/collector/test/tracing/sdk/rootExitSpans/test.js
```diff
@@ -0,0 +1,115 @@
...
+const semver = require('semver');
+
+const supportedVersion = require('@instana/core').tracing.supportedVersion;
+const config = require('@instana/core/test/config');
+const ProcessControls = require('../../../test_util/ProcessControls');
+const globalAgent = require('../../../globalAgent');
+const { retry, delay, expectExactlyOneMatching } = require('@instana/core/test/test_util');
+const constants = require('@instana/core').tracing.constants;
+
+const mochaSuiteFn =
```
aryamohanan, line 18: QS: We are still supporting Node.js 14 and 16. Why are we choosing native fetch here? We've already tested native fetch with this feature in other tests. Can we consider using `got` instead of fetch?

packages/core/src/tracing/cls.js
```diff
@@ -549,32 +556,79 @@ function skipExitTracing(options) {
...
+
+    // We need to remember if a reduced span was used, because for reduced spans
+    // we do NOT trace anymore. The async context got already closed.
+    if (parentSpan) {
+      isReducedSpan = true;
+    }
+  }
+
   const suppressed = tracingSuppressed();
   const isExitSpanResult = isExitSpan(parentSpan);
```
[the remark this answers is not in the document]
abhilash-sivan: Yeah that is a good suggestion 👍

packages/core/src/tracing/instrumentation/database/db2.js
```diff
@@ -48,12 +48,31 @@ function instrument(db2) {
...
+
+  // CASE: `cls.skipExitTracing` decided to skip, respect that
+  if (skipExitResult.skip) return true;
+
+  // CASE: the target parent function wants us to ignore cls.isTracing
+  //       because e.g. we don't know if we trace (parent got lost)
+  if (ignoreClsTracing) return false;
+
+  // CASE: We ignore if tracing or not for `allowRootExitSpan`. See cls file.
+  if (skipExitResult.allowRootExitSpan) return false;
```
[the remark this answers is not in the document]
abhilash-sivan: @kirrg001 

Maybe we can remove this. Please confirm if there is any other specific case for this

packages/collector/test/tracing/database/ioredis/test.js
abhilash-sivan: @kirrg001 
could you please comment on this ?

packages/collector/test/tracing/sdk/rootExitSpans/test.js
abhilash-sivan: let me check that :)

packages/core/src/tracing/cls.js
```diff
@@ -549,32 +556,79 @@ function skipExitTracing(options) {
...
     return true;
   }
 
+  // DESC: If `allowRootExitSpan` is true, then we have to ignore if there is a parent or not.
+  // NOTE: The feature completely ignores the state of `isTracing`, because
+  //       every exit span would be a separate trace. `isTracing` is always false,
+  //       because we don't have a parent span. The http server span also does not check of `isTracing`,
+  //       because it's the root span.
+  // CASE: Instrumentations can disable the `allowRootExitSpan` feature e.g. loggers.
+  if (!opts.skipAllowRootExitSpanPresence && allowRootExitSpan) {
```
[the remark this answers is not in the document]
abhilash-sivan: Using the same variable name `allowRootExitSpan` will make the instrumentation code more difficult to read as we want the 2 contexts as different.

1 - ROOT EXIT SPAN ENABLED
2-  ROOT EXIT SPAN DISABLED 

We can pass an additional variable to just bypass the checking like above. The check seems like ok to me because we don't want to interfere with the workflow of current instrumentation. So clubbing this check `if (!opts.skipAllowRootExitSpanPresence && allowRootExitSpan) { }` seems relevant to me.

maybe we can change the variable name to a more simple one.

cc: @kirrg001

packages/collector/test/tracing/sdk/rootExitSpans/test.js
abhilash-sivan: > Warning: This package is native [ESM](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) and no longer provides a CommonJS export. If your project uses CommonJS, you will have to [convert to ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c). Please don't open issues for questions regarding CommonJS / ESM.

> 


> You probably want [Ky](https://github.com/sindresorhus/ky) instead, by the same people. It's smaller, works in the browser too, and is more stable since it's built upon [Fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).

packages/core/src/tracing/instrumentation/database/db2.js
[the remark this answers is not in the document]
kirrg001: > // CASE: We ignore if tracing or not for `allowRootExitSpan`. See cls file.

IMO we need to keep this, because this function `skipTracing` skips checking for `isTracing`.
And for `allowRootExitSpan` we need to ignore this check.

packages/collector/test/tracing/database/ioredis/test.js
kirrg001: Because the `allowRootExitSpanApp` has no implementation for cluster. 
@abhilash-sivan Please add a comment

packages/collector/test/tracing/sdk/rootExitSpans/test.js
kirrg001: As long as we support 14/16, we should not use native fetch in our test env.
Please use the same library we use for all other tests.

Furthermore: Please rename the folder from `rootExitSpans` to `allowRootExitSpans`. Thanks

packages/collector/test/tracing/database/ioredis/test.js
kirrg001: Can be removed. @abhilash-sivan Please just set to false.

packages/core/src/tracing/cls.js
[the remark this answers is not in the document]
kirrg001: The config `allowRootExitSpan` comes from the tracer configuration level.
Theoretically we could do what you have suggested

> If we need to disable allowRootExitSpan for a particular instrumentation, couldn’t we just pass allowRootExitSpan: false

But I doubt that the readability is not good as Abhilash said

> same variable name allowRootExitSpan will make the instrumentation code more difficult to read

**We want to skip the "root exit span behavior".** We do not want to set the variable to false.

packages/collector/test/tracing/database/ioredis/test.js
abhilash-sivan: working on it

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -178,20 +178,12 @@ function instrument(coreModule, forceHttps) {
     const skipTracingResult = cls.skipExitTracing({
       isActive,
       extendedResponse: true,
-      skipParentSpanCheck: true,
-      skipIsTracing: true
+      checkReducedSpan: true
```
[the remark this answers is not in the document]
abhilash-sivan: This has been replaced by an additional flag introduced `checkReducedSpan` and moved all the logic including reduced span and `skipParentSpanCheck` to cls.skipExitTracing function;

packages/collector/test/tracing/database/ioredis/test.js
kirrg001: @abhilash-sivan We only need a comment.

abhilash-sivan: I will create a new PR to implement the allowRootExitSpanApp in cluster mode.

kirrg001: No. Please track this as a TODO on the PR for now. We will raise a card.

kirrg001 (approved):
Last steps:
- Resolve all comments
- Make the build on CI stable
- Change the base branch to v4 (last step)

We already have a TODO for the public docs on the v4 ticket.

packages/collector/test/tracing/sdk/rootExitSpans/test.js
kirrg001: @abhilash-sivan Resolved?

abhilash-sivan: Resolved with `node-fetch-v2` as in other tests.
Renaming is also completed.

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -178,20 +178,12 @@ function instrument(coreModule, forceHttps) {
     const skipTracingResult = cls.skipExitTracing({
       isActive,
       extendedResponse: true,
-      skipParentSpanCheck: true,
```
abhilash-sivan, line 181: Removing skipParentSpanCheck & skipIsTracing and added a new additional flag `checkReducedSpan` and moved all the logic including reduced span and skipParentSpanCheck to cls.skipExitTracing function;

abhilash-sivan:
Random pipeline failure occurred in:
`packages/collector/test/tracing/sdk/allowRootExitSpans/test.js:47:33`
`packages/collector/test/tracing/protocols/http/native_fetch/test.js:542:33`

While testing locally, there were no issues.

Re-running the pipeline for now

aryamohanan (approved):
Kudos for the great work team! 🎉

PR #1306, opened 2024-09-03
chore: enabled updates for currencies packages

kirrg001:
Deprecated packages should still get updates IMO

**@azure/storage-blob**
Dont know why ignoreUpdates was set to true.

**kafka-avro**
Not even deprecated yet. Comes in v4.

**@apollo/federation / request / request-promise**
Fully npm deprecated. Still, there is IMO no reason to set ignoreUpdates to true.

**Deprecation means, we will drop support in our tracer in the future.
Deprecation does not mean, we do not care about updates.**

We probably need a new flag in the future: removed/dropped support. Then these packages are getting "ignoreUpdates: true".

### DIFF
--- a/currencies.json
+++ b/currencies.json
@@ -30,7 +30,7 @@
     "latestVersion": "",
     "cloudNative": false,
     "isBeta": false,
-    "ignoreUpdates": true,
+    "ignoreUpdates": false,
     "note": "",
     "deprecated": true,
     "core": false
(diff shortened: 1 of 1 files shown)


abhilash-sivan:
Note:
To date, even though the `ignoreUpdates` flag was set to TRUE, it was skipping the update for major releases only. It was updated in [here](https://github.com/instana/nodejs/commit/a2d75f3d8b054eaef1027c6b35f237ed524a726f) such that it will enable/disable all updates according to `ignoreUpdates`.

So for the packages updated in this PR, we may not find updates for all packages in tomorrow's currency update as it might have been already taken care of.

abhilash-sivan (approved):

PR #1312, opened 2024-09-05
chore: support policy updated for timers in currencies.json

aryamohanan:
Timers is a core module, update the support policy to 0 day.


kirrg001 (approved):
Good catch!

PR #1313, opened 2024-09-05
test: added proof test for parallel intermediates

kirrg001:
refs https://bigblue.aha.io/ideas/ideas/INSTANA-I-2600

We could possibly change the SDK interface to support `startIntermediateSpan(...parentSpan (optional)...)`


aryamohanan (approved):
LGTM

Issue #1316, opened 2024-09-06
[Bug]: Logging Sending spans via span buffer message without debug level


roberto-nascimento:
### Problem Description

Hello, we are using instana with fargate, we´ve noticed that even with instana env INSTANA_LOG_LEVEL=warn, and no env INSTANA_DEBUG is set, we are getting this debug log. only this debug log is appearing. It is ingesting a lot of cloudwatch log that we do not need. 

The strange thing is that on your code thia log is using debug logger.

logger.debug(`Sending spans via span buffer (no. of spans: ${spansToSend.length})`);

2024-09-05T21:19:08.479-03:00 Sending spans via span buffer (no. of spans: 5)
2024-09-05T21:19:07.477-03:00 Sending spans via span buffer (no. of spans: 20)
2024-09-05T21:19:06.476-03:00 Sending spans via span buffer (no. of spans: 20)

### Short, Self Contained Example

We are only using docker layer, no npm collector:

# INSTALL INSTANA AGENT
COPY --from=icr.io/instana/aws-fargate-nodejs:latest /instana /instana
RUN /instana/setup.sh
and starting it with in NODE_OPITIONS.

### Node.js Version

NodeJs 20

### package.json

```shell
no npm package
```

### package-lock.json

```shell
no npm package
```

abhilash-sivan:
Hi @roberto-nascimento ,
We will take a close look and get back ASAP

kirrg001:
We have successfully [released 3.18.0.](https://github.com/instana/nodejs/releases/tag/v3.18.0)

Could you please test the new version and let us know if the problem is resolved?
Thank you so much!

roberto-nascimento:
Hello, thanks, we will deploy our app here to check right now.

roberto-nascimento:
Done, no debug logs, thanks guys , you were awesome!!

PR #1317, opened 2024-09-06
fix(serverless): resolved printing debug logs by default

aryamohanan:
refs https://github.com/instana/nodejs/issues/1316

### DIFF
--- a/packages/aws-fargate/src/activate.js
+++ b/packages/aws-fargate/src/activate.js
@@ -21,8 +21,10 @@ const config = normalizeConfig({});
 config.logger = logger;
 
 function init() {
-  if (process.env.INSTANA_DEBUG || process.env.INSTANA_LOG_LEVEL) {
-    logger.setLevel(process.env.INSTANA_DEBUG === 'true' ? 'debug' : process.env.INSTANA_LOG_LEVEL);
+  const logLevel = process.env.INSTANA_DEBUG === 'true' ? 'debug' : process.env.INSTANA_LOG_LEVEL;
+
+  if (logLevel) {
+    logger.setLevel(logLevel);
   }
 
   instanaCore.preInit();
--- a/packages/aws-fargate/src/activate.js
+++ b/packages/aws-fargate/src/activate.js
@@ -18,10 +18,11 @@ const { normalizeConfig } = coreUtil;
 let logger = consoleLogger;
 
 const config = normalizeConfig({});
+config.logger = logger;
 
 function init() {
   if (process.env.INSTANA_DEBUG || process.env.INSTANA_LOG_LEVEL) {
-    logger.setLevel(process.env.INSTANA_DEBUG ? 'debug' : process.env.INSTANA_LOG_LEVEL);
+    logger.setLevel(process.env.INSTANA_DEBUG === 'true' ? 'debug' : process.env.INSTANA_LOG_LEVEL);
   }
 
   instanaCore.preInit();
--- a/packages/aws-fargate/src/activate.js
+++ b/packages/aws-fargate/src/activate.js
@@ -21,10 +21,9 @@ const config = normalizeConfig({});
 config.logger = logger;
 
 function init() {
-  const logLevel = process.env.INSTANA_DEBUG === 'true' ? 'debug' : process.env.INSTANA_LOG_LEVEL;
-
-  if (logLevel) {
-    logger.setLevel(logLevel);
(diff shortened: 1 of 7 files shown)


packages/core/src/logger.js
```diff
@@ -59,8 +53,11 @@ exports.init = function init(config = {}) {
     // A custom logger which is neither bunyan nor pino has been provided via config. We use it as is.
     parentLogger = config.logger;
   } else {
-    // No custom logger has been provided via config, we create a new minimal logger.
-    parentLogger = consoleLogger;
+    // No custom logger has been provided via config, we create a new minimal logger with pino.
+    // @ts-ignore
+    parentLogger = pino({
```
kirrg001, line 58: There is indeed a problem with falling back to a "minimal" logger. 
A soon as you require the core module, all required files call `getLogger` on the top of the file. Thats why we have added this fallback. This needs to be refactored away, because it does not respect any log level. I'd suggest to raise an issue for that. But IMO this is not the cause of the problem. 

IMO we are seeing debug logs by default from **any** serverless packages because we do not forward their logger!

See https://github.com/instana/nodejs/blob/v3.17.1/packages/azure-container-services/src/activate.js

We define which logger we want to use for the pkg (logger from serverless pkg).
We normalize the config (which I also do think is very wrong!, but lets ignore that. Please add a comment.) and then we do not forward the logger in the config:

https://github.com/instana/nodejs/blob/v3.17.1/packages/core/src/index.js#L69
https://github.com/instana/nodejs/blob/v3.17.1/packages/core/src/logger.js#L43

The **shortest quick fix** (without having too much to refactor) is IMO for every serverless pkg:

```
 const config = normalizeConfig({});
 config.logger = logger;
```

packages/serverless-collector/src/activate.js
```diff
@@ -15,10 +15,11 @@ const { normalizeConfig } = coreUtil;
 let logger = consoleLogger;
 
 const config = normalizeConfig({});
+config.logger = logger;
 
 function init() {
   if (process.env.INSTANA_DEBUG || process.env.INSTANA_LOG_LEVEL) {
-    logger.setLevel(process.env.INSTANA_DEBUG ? 'debug' : process.env.INSTANA_LOG_LEVEL);
+    logger.setLevel(process.env.INSTANA_DEBUG === 'true' ? 'debug' : process.env.INSTANA_LOG_LEVEL);
```
kirrg001, line 22: We need to do the string boolean check in line 21.
Otherwise you can call `setLevel(null)` if INSTANA_DEBUG is not 'true'.

kirrg001 (approved):
Pre-approving. My comment needs to be fixed before merging.

Commit suggestion

> fix(serverless): resolved printing debug logs by default
> refs https://github.com/instana/nodejs/issues/1316

packages/aws-fargate/src/activate.js
```diff
@@ -18,10 +18,13 @@ const { normalizeConfig } = coreUtil;
 let logger = consoleLogger;
 
 const config = normalizeConfig({});
+config.logger = logger;
 
 function init() {
-  if (process.env.INSTANA_DEBUG || process.env.INSTANA_LOG_LEVEL) {
-    logger.setLevel(process.env.INSTANA_DEBUG ? 'debug' : process.env.INSTANA_LOG_LEVEL);
+  const logLevel = process.env.INSTANA_DEBUG === 'true' ? 'debug' : process.env.INSTANA_LOG_LEVEL;
```
abhilash-sivan, line 24: In my understanding, we don't have to set the INSTANA_DEBUG to 'true' or anything. We can just set the value as say `test` || `1` or any "TRUTHY" values for the debugger to be enabled.

Find the [public doc here](https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-collector-configuration#log-level-configuration)

> You can also configure the log level by setting the environment variable INSTANA_LOG_LEVEL to either debug, info, warn or error. Finally, setting INSTANA_DEBUG to any non-empty string will set the log level to debug.
> 

This applies to all the similar changes.

abhilash-sivan (changes requested):

kirrg001 (approved):

abhilash-sivan (approved):

packages/aws-fargate/src/activate.js
```diff
@@ -18,8 +18,10 @@ const { normalizeConfig } = coreUtil;
 let logger = consoleLogger;
 
 const config = normalizeConfig({});
+config.logger = logger;
```
kirrg001, line 21: I will add another comment here after merge. Dont wanne wait again

PR #1319, opened 2024-09-09
chore: logged Kafka header deprecation warning only if Kafka module was used

aryamohanan:
Now the kafka warning is logged always, instead it should log only if the module is required.

### DIFF
--- a/packages/core/src/tracing/instrumentation/messaging/rdkafka.js
+++ b/packages/core/src/tracing/instrumentation/messaging/rdkafka.js
@@ -13,7 +13,6 @@ const cls = require('../../cls');
 const shimmer = require('../../shimmer');
 const { getFunctionArguments } = require('../../../util/function_arguments');
 let traceCorrelationEnabled = constants.kafkaTraceCorrelationDefault;
-let configHeader = null;
 
 let logger;
 logger = require('../../../logger').getLogger('tracing/rdkafka', newLogger => {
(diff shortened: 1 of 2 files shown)


packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -21,6 +21,7 @@ logger = require('../../../logger').getLogger('tracing/rdkafka', newLogger => {
 });
 
 let isActive = false;
+let logKafkaHeaderWarning = false;
```
aryamohanan, line 24: The flag that is turned on when the module is required, as we need to check the header format from the agent, which becomes available at a later stage.

kirrg001: 🤔 I am not able to follow this comment.

Why do we need put the `logWarningForKafkaHeaderFormat` into e.g. `instrumentConsumer`?

aryamohanan: The logKafkaHeaderWarning flag controls whether or not the warning will be logged. Before calling logWarningForKafkaHeaderFormat, we check if the flag is enabled. The flag is turn on within functions like instrumentConsumer, instrumentConsumerAsStream, and instrumentProducer because these functions trigger when the rdkafka module is loaded in different usecases. To ensure all scenarios are covered, the flag is incorporated into these methods. This flag will be removed in version 4 as part of the header migration phase-2.

kirrg001: Ok I understand now. Thanks
Can't you just add another registration `hook.onModuleLoad('node-rdkafka', logWarning);`?

aryamohanan: That is possible, but in method we have two warning messages for rdkafka, this [warning](https://github.com/instana/nodejs/pull/1319/files#diff-24bbb2057190b9840c5c35b03ff3849adf0501722464d3ba476378bf0dfd8fd4L73), we are checking the header values, and the values coming from agent is not available at this point.

kirrg001: Why do we need to warnings? 🤔 I'd suggest to just put them as one warning. Two large warnings is way too much.

kirrg001 (approved):
LGTM

Can be merged as "chore", because we already have the other commit as "fix".

Issue #1327, opened 2024-09-12
ioredis clash


a-reuss:
### Problem Description

In our application, we use ioredis as short term database. Since the update of instana to version 3.18.0 where there were changes to the interaction between instana collector and redis used as instana internal data store, our traces cannot be correctly related to traces recorded in the rest of the system. 

### Short, Self Contained Example

The reason is that although the necessary header ist correctly transferred through all services involved, the trace ids between our subsystem and the rest of the system do not match anymore. Is it possible that there is a misconfiguration for the ioredis connection instana uses. The most likely candidate for this is the newly introduced code block from line 40 in packages/core/src/tracing/instrumentation/database/ioredis.js
 
Text
clusterConnectionString = this.startupNodes.map(node => `${node.host}:${node.port}`).join(',');
 
 
Additionally, we see issues with parent spans that might come from the change in line 84 in packages/core/src/tracing/instrumentation/database/ioredis.js where 'multi' commands are now treated differently to before. We see error messages of the kind
Cannot start an intermediate span (Foo.bar()) as this requires an active entry (or intermediate) span as parent. But the currently active span is an exit span:
that we did not see before.
 
We do not know whether the 2 issues are linked to each other at all, or whether they are independent.

### Node.js Version

20.x

### package.json

```shell
{
  "devDependencies": {
    "@aws-sdk/client-ssm": "3.624.0",
    "@cucumber/cucumber": "10.8.0",
    "@eslint/eslintrc": "3.1.0",
    "@eslint/js": "9.8.0",
    "@types/async-retry": "1.4.8",
    "@types/http-cache-semantics": "4.0.4",
    "@types/jest": "29.5.12",
    "@types/js-yaml": "4.0.9",
    "@types/jsonwebtoken": "9.0.6",
    "@types/koa": "2.15.0",
    "@types/koa__router": "12.0.4",
    "@types/koa-bodyparser": "4.3.12",
    "@types/koa-etag": "3.0.3",
    "@types/koa-mount": "4.0.5",
    "@types/koa-router": "7.4.8",
    "@types/koa-static": "4.0.4",
    "@types/luxon": "3.4.2",
    "@types/mustache": "4.2.5",
    "@types/node": "20.14.14",
    "@types/proxy-addr": "2.0.3",
    "@types/selenium-webdriver": "4.1.24",
    "@types/semver": "7.5.8",
    "@types/traverse": "0.6.37",
    "@types/uuid": "10.0.0",
[...]

PR #1334, opened 2024-09-17
fix(ioredis): repaired cluster functionality

kirrg001:
refs #1327
refs https://jsw.ibm.com/browse/INSTA-14540
refs https://jsw.ibm.com/browse/INSTA-14538

Its either caused by the connection string (instead of ip) or the new multi span. We revert both for now.

### DIFF
--- a/packages/core/src/tracing/instrumentation/database/ioredis.js
+++ b/packages/core/src/tracing/instrumentation/database/ioredis.js
@@ -29,37 +29,16 @@ exports.init = function init() {
   hook.onModuleLoad('ioredis', instrument);
 };
 
-let clusterConnectionString;
-
 function instrument(ioredis) {
-  class InstanaClusterClass extends ioredis.Cluster {
-    constructor() {
-      super(...arguments);
-
-      try {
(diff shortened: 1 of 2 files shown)


packages/core/src/tracing/instrumentation/database/ioredis.js
```diff
@@ -96,8 +75,15 @@ function instrumentSendCommand(original) {
...
+
+      const connection = `${client.options.host}:${client.options.port}`;
+
+      // TODO: https://jsw.ibm.com/browse/INSTA-14540
+      // if (client.isCluster) {
+      //  connection = client.startupNodes.map(node => `${node.host}:${node.port}`).join(',');
+
       span.data.redis = {
-        connection: clusterConnectionString || `${client.options.host}:${client.options.port}`,
+        connection,
```
aryamohanan, line 86: qs: What will be the connection string for the cluster?

kirrg001: IP Address

aryamohanan (approved):

PR #1337, opened 2024-09-19
fix: deprecated kafka-avro

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-1042
refs https://github.com/waldophotos/kafka-avro/issues/120

Tests are already skipped.

### DIFF
--- a/packages/core/src/tracing/instrumentation/messaging/rdkafka.js
+++ b/packages/core/src/tracing/instrumentation/messaging/rdkafka.js
@@ -12,8 +12,12 @@ const constants = require('../../constants');
 const cls = require('../../cls');
 const shimmer = require('../../shimmer');
 const { getFunctionArguments } = require('../../../util/function_arguments');
-let traceCorrelationEnabled = constants.kafkaTraceCorrelationDefault;
+let logger;
+logger = require('../../../logger').getLogger('tracing/rdkafka', newLogger => {
+  logger = newLogger;
+});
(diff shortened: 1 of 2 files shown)


abhilash-sivan (approved):

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -359,3 +364,10 @@ function findInstanaHeaderValues(instanaHeadersAsObject) {
 
   return { level, traceId, longTraceId, parentSpanId };
 }
+
+function logDeprecationKafkaAvroMessage() {
```
aryamohanan, line 368: Suggestion for function name `logKafkaAvroDeprecationWarning`

aryamohanan (approved):
LGTM

aryamohanan:
Qs: Would it be possible to also remove the tests? It seems unnecessary to keep the skipped ones.

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
kirrg001: haha yeah you are right, but I am too lazy to change it now :D

kirrg001:
> Would it be possible to also remove the tests?

I decided to keep them in case we need to test something. We do not have a rule when we remove the tests.

PR #1338, opened 2024-09-20
chore: avoided wrapping libraries twice

kirrg001:
We noticed that redis `wrapExec` fn is called twice:

https://github.com/instana/nodejs/blob/v3.18.2/packages/core/src/tracing/instrumentation/database/redis.js#L69
https://github.com/instana/nodejs/blob/v3.18.2/packages/core/src/tracing/instrumentation/database/redis.js#L109
https://github.com/instana/nodejs/blob/v3.18.2/packages/core/src/tracing/instrumentation/database/redis.js#L142

Client & cluster using the same help function.
But the internal client object in redis is the same. We could probably rewrite the redis instrumentation, but adding a protection against double wrapping is much easier and we need this anyway.

### DIFF
--- a/packages/core/src/tracing/shimmer.js
+++ b/packages/core/src/tracing/shimmer.js
@@ -40,6 +40,16 @@ exports.unwrap = shimmer.unwrap;
  *    instrumented library.
  */
 exports.wrap = (origObject, origMethod, instrumentationWrapperMethod) => {
+  // NOTE: We do not want to wrap a function of a library twice.
+  // CASE: Some instrumentations are really complex such as redis.
+  //       e.g. there is the concept of client and/or cluster connections.
+  //       We might use the same underlying object, but we call `.wrap` twice in the instrumentation
+  //       but we can't control it.
+  if (origObject[origMethod].__wrapped) {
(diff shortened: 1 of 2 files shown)


packages/core/src/tracing/shimmer.js
```diff
@@ -40,6 +40,16 @@ exports.unwrap = shimmer.unwrap;
  *    instrumented library.
  */
 exports.wrap = (origObject, origMethod, instrumentationWrapperMethod) => {
+  // NOTE: We do not want to wrap a function of a library twice.
+  // CASE: Some instrumentations are really complex such as redis.
+  //       e.g. there is the concept of client and/or cluster connections.
+  //       We might use the same underlying object, but we call `.wrap` twice in the instrumentation
+  //       but we can't control it.
+  if (origObject[origMethod].__wrapped) {
```
kirrg001, line 48: See https://www.npmjs.com/package/shimmer

> All monkeypatched functions have an attribute, __wrapped, set to true on them.

abhilash-sivan:
LGTM

abhilash-sivan (approved):

abhilash-sivan (approved):

PR #1340, opened 2024-09-23
chore: refactored redis and added span length asserts

abhilash-sivan:


kirrg001 (approved):

packages/collector/test/tracing/database/redis/test.js
```diff
@@ -235,6 +237,13 @@ const globalAgent = require('../../../globalAgent');
                       ]);
 
                       verifyHttpExit(controls, spans, writeEntrySpan);
+
+                      // TODO: not sure what is the difference in span length with setupType
```
aryamohanan, line 241: Why?

abhilash-sivan: This is from the main. we have to ignore it for now. Might change after the refactoring done in opt-in feature

aryamohanan: See https://github.com/instana/nodejs/blob/main/packages/collector/test/tracing/database/redis/test.js#L31

packages/collector/test/tracing/database/redis/test.js
```diff
@@ -235,6 +237,13 @@ const globalAgent = require('../../../globalAgent');
...
+                      if (setupType === 'cluster') {
+                        expect(spans.length).to.be.eql(2);
```
aryamohanan, line 243: Why the difference in span length for cluster and standalone?

kirrg001: The PR only adds span lengths for the "current state".

kirrg001: We have a similar behavior for `ioredis`.

packages/collector/test/tracing/database/redis/test.js
aryamohanan: Ah I see, the initial comment was confusing.

packages/collector/test/tracing/database/redis/test.js
kirrg001: Extended https://jsw.ibm.com/browse/INSTA-14540

aryamohanan: I see, the initial comment was confusing.

aryamohanan (approved):

PR #1344, opened 2024-09-26
fix(lambda): changed memory warning when using the lambda extension

kirrg001:
Background:
We assume that when installing the lambda layer, but mis-configuring the memory, the extension is still started and running.
There could be significant latency especially during cold starts. That could be caused by:

- waiting for AWS events (invokes, shutdowns), incoming tracer requests or go channels
- slow startup because of low memory

As soon as we increase the memory, the latency problem is gone.

We should discuss if we want to change this behavior e.g. we disable tracing completely. Because **as soon as the extension is installed**, it should be configured correctly.

### DIFF
--- a/packages/aws-lambda/src/wrapper.js
+++ b/packages/aws-lambda/src/wrapper.js
@@ -345,7 +345,7 @@ function shouldUseLambdaExtension() {
       return false;
     }
     if (memorySize < 256) {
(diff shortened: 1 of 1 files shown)


aryamohanan (approved):
LGTM

abhilash-sivan (approved):

PR #1349, opened 2024-09-30
feat: extended typescript support

kirrg001:
## Problems:

> import instana from '@instana/collector';
> console.log('instana', instana);
> instana: undefined

> No declaration file was found for the module "@instana/collector." "node_modules/@instana/collector/src/index.js" implicitly has an "any" type.

## Explanations

Typescript could not load our in-code typedefs. We had to provide types files.
We had no definitions for exported functions. 

The PR is a beginning. We need to do much more. Raised a new ticket.

## Works

```js
import * as instana from '@instana/collector'
```

```js
import instana from '@instana/collector'
```

## Raised

https://jsw.ibm.com/browse/INSTA-15926
https://jsw.ibm.com/browse/INSTA-15926

## Tasks
- [x] fix tests
- [x] add test

### DIFF
--- a/packages/collector/test/tracing/misc/esm-typescript/dist/app.js
+++ b/packages/collector/test/tracing/misc/esm-typescript/dist/app.js
@@ -1,0 +1,11 @@
+"use strict";
+var __importDefault = (this && this.__importDefault) || function (mod) {
+    return (mod && mod.__esModule) ? mod : { "default": mod };
+};
+Object.defineProperty(exports, "__esModule", { value: true });
+const collector_1 = __importDefault(require("@instana/collector"));
+if (!collector_1.default.sdk) {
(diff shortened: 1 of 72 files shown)


packages/collector/test/tracing/misc/esm-typescript/dist/app.js
aryamohanan, line 1: Suggestion: Can we exclude compiled js files from the code?

kirrg001: Yes we could do that. Lets do that after the release 👍

aryamohanan (approved):
LGTM 🤞

abhilash-sivan (approved):

PR #1350, opened 2024-10-01
chore: removed unused file that was no longer needed

aryamohanan:
The file was no longer used. However, there were setups where the versions of @instana/collector and @instana/core did not match, so we kept it around for a while for backwards compatibility. It has now been removed as it is no longer needed.


kirrg001 (approved):
Suggestion:

> chore(core): removed unused buffer file

abhilash-sivan (approved):

PR #1352, opened 2024-10-01
fix: dropped support for Node.js lambda runtime v14 and v16

aryamohanan:
V14 and v16 are officially reached end of support, see  https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-deprecated


kirrg001 (approved):
Please add the breaking change hint in the commit body.

PR #1355, opened 2024-10-03
fix: dropped support for x-instana-service header

aryamohanan:
The capture of the `X-Instana-Service` header value as `span.data.service` has been removed from the tracer. According to the public documentation [here](https://www.ibm.com/docs/en/instana-observability/current?topic=applications-services#specify-the-x-instana-service-http-header):

> By setting the `X-Instana-Service` HTTP header on a call, the destination service (the one receiving the HTTP request) is tagged with the value provided in the header.
> 
> Note: The `X-Instana-Service` HTTP header is not collected automatically. To use it, the Instana agent must be configured, as explained in ["Capturing custom HTTP headers"](https://www.ibm.com/docs/en/instana-observability/current?topic=cha-configuring-host-agents-by-using-agent-configuration-file#capturing-custom-http-headers).

By default, Instana does not capture HTTP headers during tracing. However, you can configure HTTP header capture by modifying the agent's configuration file (<instana-agent-dir>/etc/instana/configuration.yaml) as follows:

```yaml
com.instana.tracing:
  extra-http-headers:
    - 'X-Instana-Service'
```

The backend extracting the service name from the captured header. 
Therefore, implementing this directly within our tracer by inspecting the `X-Instana-Service` header and adding the annotation is redundant. For this reason, the Instana tracer community has chosen not to capture the X-Instana-Service header within the tracer.

### DIFF
--- a/packages/collector/test/tracing/common/test.js
+++ b/packages/collector/test/tracing/common/test.js
@@ -168,23 +168,6 @@ mochaSuiteFn('tracing/common', function () {
       );
     });
 
-    describe('with header when agent is configured to capture the header', function () {
-      const agentControls = new AgentStubControls();
-      const agentConfig = { extraHeaders: ['x-iNsTanA-sErViCe'] };
-
-      before(async () => {
-        await agentControls.startAgent(agentConfig);
-      });
-      after(async () => {
-        await agentControls.stopAgent();
-      });
-
-      registerServiceNameTest.call(this, agentControls, {
-        configMethod: 'X-Instana-Service header',
-        expectServiceNameOnSpans: 'on-entry-span'
-      });
-    });
-
     describe('with header when agent is _not_ configured to capture the header', function () {
       const agentControls = new AgentStubControls();
 
@@ -196,7 +179,6 @@ mochaSuiteFn('tracing/common', function () {
       });
 
       registerServiceNameTest.call(this, agentControls, {
-        configMethod: 'X-Instana-Service header',
         expectServiceNameOnSpans: 'no'
       });
     });
@@ -236,11 +218,6 @@ mochaSuiteFn('tracing/common', function () {
         const req = {
           path: '/with-intermediate-and-exit-spans'
         };
-        if (configMethod === 'X-Instana-Service header') {
-          req.headers = {
-            'x-InsTana-sErvice': 'much-custom-very-wow service'
-          };
-        }
         return controls.sendRequest(req).then(() => verifyServiceName(agentControls, expectServiceNameOnSpans));
       });
     }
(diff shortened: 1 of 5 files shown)


packages/core/src/tracing/instrumentation/protocols/httpServer.js
```diff
@@ -106,12 +106,6 @@ function shimEmit(realEmit) {
         header: getExtraHeadersFromMessage(req, extraHttpHeadersToCapture)
       };
 
-      const incomingServiceName =
```
kirrg001, line 109: qs:

From the docs link you shared:
> By setting the X-Instana-Service HTTP header on a call, the destination (the service, which receives the HTTP request) is tagged with the value that is provided in the header.

How does it work when the tracers do not forward the header?

packages/collector/test/tracing/common/test.js
```diff
@@ -168,23 +168,6 @@ mochaSuiteFn('tracing/common', function () {
       );
     });
 
-    describe('with header when agent is configured to capture the header', function () {
-      const agentControls = new AgentStubControls();
-      const agentConfig = { extraHeaders: ['x-iNsTanA-sErViCe'] };
```
kirrg001, line 173: qs: Should we keep the test but just use a different header?

packages/core/src/tracing/instrumentation/protocols/httpServer.js
aryamohanan: The backend [code](https://github.ibm.com/instana/backend/blob/2a42269ff0d2608a843c0271375d45cd58f8d2a4/label-extractor/src/main/resources/defaultServiceRules.yaml#L37) reads the service name directly from the header, so we don't need to read X-Instana-Service header in our end.

packages/collector/test/tracing/common/test.js
aryamohanan: The test was expecting the service name from the headers, but since we removed the implementation, the test can no longer be reused.

packages/core/src/tracing/instrumentation/protocols/httpServer.js
aryamohanan: The customer have an option to configure the header in agent configuration file. Details explained in the PR description.

kirrg001 (approved):
Thanks!

> Note: The X-Instana-Service HTTP header is not automatically collected. For it to work, the Instana agent must be configured , see ["Capturing custom HTTP headers"](https://www.ibm.com/docs/en/SSE1JP5_current/src/pages/setup_and_manage/host_agent/configuration/index.html#capture-custom-http-headers).

FYI Docs links are broken. 

This is the correct link:
https://www.ibm.com/docs/en/instana-observability/current?topic=cha-configuring-host-agents-by-using-agent-configuration-file#capturing-custom-http-headers

We will fix it soon!

kirrg001:
Suggestion for commit body:
- BREAKING CHANGE: ...
- Migration: Please define the header in the agent config...blabla

PR #1356, opened 2024-10-03
fix: added log to capture instana headers from incoming request

aryamohanan:
This PR adds a log to capture Instana headers from incoming requests. This enhancement improves the application's debugging capabilities by providing visibility into the Instana headers present in the requests.

### DIFF
--- a/packages/core/src/tracing/instrumentation/protocols/httpServer.js
+++ b/packages/core/src/tracing/instrumentation/protocols/httpServer.js
@@ -17,7 +17,10 @@ const {
 } = require('./captureHttpHeadersUtil');
 const shimmer = require('../../shimmer');
 const cls = require('../../cls');
-
+let logger;
+logger = require('../../../logger').getLogger('tracing/httpServer', newLogger => {
+  logger = newLogger;
+});
 let extraHttpHeadersToCapture;
 let isActive = false;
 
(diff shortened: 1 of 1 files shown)


abhilash-sivan (approved):
LGTM 👍

packages/core/src/tracing/instrumentation/protocols/httpServer.js
```diff
@@ -203,3 +206,12 @@ function shimEmit(realEmit) {
     });
   };
 }
+function logInstanaHeaders(req) {
```
aryamohanan, line 209: Added as part INSTA-15032, to check we are getting the trace ids from the header. WE can revert back this later

packages/core/src/tracing/instrumentation/protocols/httpServer.js
```diff
@@ -203,3 +206,12 @@ function shimEmit(realEmit) {
...
+  if (req.headers) {
+    Object.keys(req.headers).forEach(header => {
+      if (header?.toLowerCase()?.startsWith('x-instana')) {
```
kirrg001, line 212: Can we read that from constants? We can add constants.ISNTANA_HEADER_PREFIX='x-instana'

PR #1362, opened 2024-10-04
fix(shared-metrics): removed fs-extra dependency from production

aryamohanan:
The external dependency `fs-extra` has been removed from the production dependencies. Previously, we couldn't use the native `fs.cp` module due to its unavailability in earlier versions of Node.js 16. Now that we're deprecating support for versions below 18, we're switching to the native fs module See https://nodejs.org/api/fs.html#fspromisescpsrc-dest-options
refs https://github.com/instana/nodejs/pull/1265

### DIFF
--- a/packages/collector/test/tracing/database/db2/test.js
+++ b/packages/collector/test/tracing/database/db2/test.js
@@ -877,10 +877,10 @@ mochaSuiteFn('tracing/db2', function () {
         .then(() =>
           testUtils.retry(() =>
             verifySpans(agentControls, controls, {
-              numberOfSpans: 14,
+              numberOfSpans: 15,
               // Spans:
               // 10 queries splitted from the file
-              // 4 fs operations on top (ours + from db2 internally fs-extra)
+              // 5 fs operations on top (ours + from db2 internally fs-extra)
               // https://github.com/ibmdb/node-ibm_db/blob/fb25937524d74d25917e9aa67fb4737971317986/lib/odbc.js#L916
               // If the Otel integration is disabled, we expect 11 spans.
               verifyCustom: (entrySpan, spans) => {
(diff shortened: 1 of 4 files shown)


packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -188,8 +187,8 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
           `Copying the precompiled build for ${opts.nativeModuleName} ${label} from ${sourceDir} to ${targetDir}.`
         );
 
-        fse
-          .copy(sourceDir, targetDir)
+        fs.promises
```
kirrg001, line 190: qs: Where is `fs` required? 🤔

kirrg001 (commented):
> Now that we're deprecating support for versions below 18

Suggestion: Now that we're dropping support for versions below 18 in v4.

---

Suggestion for commit:

> fix(shared-metrics): replaced fs-extra with fs promises

packages/shared-metrics/src/util/nativeModuleRetry.js
aryamohanan: fs already required in this file.

kirrg001 (approved):

packages/shared-metrics/src/util/nativeModuleRetry.js
aryamohanan: https://github.com/instana/nodejs/pull/1362/files#diff-9bccaad2e8829a033823a10f239e5912eb26e892a573aea25a7b111837aa6786R8

> const { logger: Logger, uninstrumentedFs: fs } = require('@instana/core')

packages/collector/test/tracing/database/db2/test.js
```diff
@@ -882,10 +877,10 @@ mochaSuiteFn('tracing/db2', function () {
         .then(() =>
           testUtils.retry(() =>
             verifySpans(agentControls, controls, {
-              numberOfSpans: 14,
+              numberOfSpans: 15,
```
aryamohanan, line 880: I noticed there are now 15 spans, likely because I’m using the native `fs`, which may have caused the span count to increase. If the test continues to fail again, so I will investigate further.

PR #1363, opened 2024-10-04
test: fixed test for typescript

abhilash-sivan:
fixed tests that were failing because of `fetch` lib not supported in node versions < 18


aryamohanan (approved):
Thanks for fixing this ❤️

kirrg001 (approved):
Thanks & Sorry!

PR #1377, opened 2024-10-11
fix: dropped support for q library

aryamohanan:
The [Q](https://www.npmjs.com/package/q) library has been officially deprecated. Last release was 7 years ago.
The maintainers announced this deprecation in the project's [GitHub repository](https://github.com/kriskowal/q?tab=readme-ov-file#note), advising developers to transition to native JavaScript promises. The Q library, an early tool for handling promises in JavaScript, is no longer needed because modern JavaScript now includes built-in support for promises.

We already deprecated the library in v3, see https://github.com/instana/nodejs/pull/1282. So removing the support in V4.

Note: We don't have instrumentation for this and not creating spans for q. 

Commit as a breaking change.


currencies.json
```diff
@@ -779,18 +779,6 @@
     "deprecated": true,
     "core": false
   },
-  {
```
kirrg001, line 782: I see three options:

1. Ask Community and John if this approach is okay.

2. Work with the existing fields
```
"note": "Removed support in Oct 2024",
"ignoreUpdates": true
```

3. Extend the table (preferred):
```
"supportRemovedAt": "date",
"ignoreUpdates": true
```

aryamohanan: Thanks for the options. I will bring this in community.

kirrg001 (approved):

PR #1383, opened 2024-10-14
fix: dropped support for disabling AWS SDK instrumentation in old syntax

aryamohanan:
Added a test to verify that customers can disable AWS SDK v3 instrumentation using `disabledTracers: ['aws-sdk/v3']`.

This highlights a design issue in our earlier system. 
Already documented in the public [documenation](https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-collector-configuration#disabling-individual-tracers)

As part of the updates made in https://github.com/instana/nodejs/pull/866, we've introduced an instrumentationName in our AWS SDK instrumentation. This enhancement allowed us to disable the tracer effectively using the instrumentation name rather than relying on the instrumentation syntax.

BREAKING CHANGE:
- Removed the ability to disable AWS SDK instrumentation using the old syntax disabledTracers: ['aws-sdk/v2/index'].
- Migrate to the new syntax for disabling instrumentation: disabledTracers: ['aws-sdk/v2'].

### DIFF
--- a/packages/core/src/tracing/index.js
+++ b/packages/core/src/tracing/index.js
@@ -41,8 +41,8 @@ let processIdentityProvider = null;
 
 // Note: Also update initializedTooLateHeuristic.js and the accompanying test when adding instrumentations.
 let instrumentations = [
-  './instrumentation/cloud/aws-sdk/v2/index',
-  './instrumentation/cloud/aws-sdk/v3/index',
+  './instrumentation/cloud/aws-sdk/v2',
+  './instrumentation/cloud/aws-sdk/v3',
   './instrumentation/cloud/aws-sdk/v2/sdk',
   './instrumentation/cloud/aws-sdk/v2/sqs',
   './instrumentation/cloud/azure/blob',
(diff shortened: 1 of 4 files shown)


packages/core/src/tracing/index.js
```diff
@@ -41,8 +41,8 @@ let processIdentityProvider = null;
 
 // Note: Also update initializedTooLateHeuristic.js and the accompanying test when adding instrumentations.
 let instrumentations = [
-  './instrumentation/cloud/aws-sdk/v2/index',
-  './instrumentation/cloud/aws-sdk/v3/index',
+  './instrumentation/cloud/aws-sdk/v2',
```
aryamohanan, line 44: As part of the updates made in https://github.com/instana/nodejs/pull/866, we've introduced an instrumentationName in our AWS SDK instrumentation. This enhancement allows us to disable the tracer effectively using the instrumentation name rather than relying on the instrumentation syntax.

kirrg001: I agree that keeping the logic in general might be a good idea, because the path could vary.

Why do we have to keep `exports.instrumentationName = 'aws-sdk/v3';`?
This is already the correct name.

aryamohanan: Yes, so I removed the instrumentationName from aws sdk v2 and v3 and keep the core logic as it is.

kirrg001 (approved):
LGTM

PR #1387, opened 2024-10-14
test: skip test in shared-metrics activeHandles for prelease

abhilash-sivan:


packages/shared-metrics/test/activeHandles_test.js
```diff
@@ -22,7 +22,7 @@ describe('metrics.activeHandles', function () {
   });
 
   it('should update handle count for a setTimeout', () => {
-    if (semver.satisfies(process.versions.node, '>=11')) {
+    if (semver.satisfies(process.versions.node, '>=11', { includePrerelease: true })) {
```
abhilash-sivan, line 25: This test was skipped for versions >=11, the condition was failing for prerelease versions.

Now added the `includePrerelease` tag to include prerelease version in the check

kirrg001: Can we remove this test? 🤔 We only support => v14.

kirrg001: I'd suggest to add `it.skip` on the test.
And add a proper comment.
We can unskip the test if we move away from the deprecated flag `_getActiveHandles`.
The new flag works better https://nodejs.org/api/process.html#processgetactiveresourcesinfo

kirrg001 (approved):
Please follow https://github.com/instana/nodejs/pull/1387/files#r1799469950
You can self merge after resolving 👍

aryamohanan (approved):

PR #1389, opened 2024-10-15
ci: updated save-cache logic to skip for prerelease versions

abhilash-sivan:


.tekton/tasks/save-cache.yaml
```diff
@@ -30,7 +32,10 @@ spec:
         ARTIFACTS_PATH="$(workspaces.output.path)"
         cd $ARTIFACTS_PATH
 
-        if [ ! -e "node-modules.tar.zst" ]; then
+        # we do not want to cache the prerelease node modules as we might change the node version inside container.
+        # because the node modules are stored along with branch and pre defined node version of the container.
```
kirrg001, line 36:
```suggestion
        # The node modules are stored along with branch and pre defined node version of the container (e.g. node-modules-main-20).
```

.tekton/tasks/save-cache.yaml
```diff
@@ -30,7 +32,10 @@ spec:
...
+        # and that might broke other pipeline configurations to use the incompatible node modules
```
kirrg001, line 37:
```suggestion
        # That might brake the target pipeline because its using a wrong node_modules installation.
```

.tekton/tasks/save-cache.yaml
```diff
@@ -30,7 +32,10 @@ spec:
...
+        if [ "$(params.prerelease)" != "true" ] && [ ! -e "node-modules.tar.zst" ]; then
```
kirrg001, line 38: qs: Why do we need "[ ! -e "node-modules.tar.zst" ];"?

.tekton/tasks/save-cache.yaml
```diff
@@ -30,7 +32,10 @@ spec:
         ARTIFACTS_PATH="$(workspaces.output.path)"
         cd $ARTIFACTS_PATH
 
-        if [ ! -e "node-modules.tar.zst" ]; then
+        # we do not want to cache the prerelease node modules as we might change the node version inside container.
```
kirrg001, line 35:
```suggestion
        # We do not want to cache the prerelease node modules as we might change the node version inside container.
```

.tekton/tasks/save-cache.yaml
abhilash-sivan: That condition was already there. I believe it is to prevent caching/overwriting duplicates for the same configuration.

kirrg001: Ah sorry. I did not look close enough.

kirrg001 (approved):

PR #1391, opened 2024-10-15
test: skipped rdkafka on v23

abhilash-sivan:
As node-rdkafka relies heavily on librdkafka and C++ bindings, compatibility with newer Node.js versions typically depends on updates...
Skipping test for now and will investigate as part of https://jsw.ibm.com/browse/INSTA-17132

### DIFF
--- a/packages/collector/test/tracing/messaging/node-rdkafka/test.js
+++ b/packages/collector/test/tracing/messaging/node-rdkafka/test.js
@@ -52,13 +52,11 @@ const SINGLE_TEST_PROPS = {
 const retryTime = 1000;
 const topic = 'rdkafka-topic';
 
-let mochaSuiteFn;
-
-if (!supportedVersion(process.versions.node) || semver.lte(process.versions.node, '16.0.0')) {
-  mochaSuiteFn = describe.skip;
-} else {
-  mochaSuiteFn = describe;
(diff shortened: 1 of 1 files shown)


packages/collector/test/tracing/messaging/node-rdkafka/test.js
```diff
@@ -52,13 +52,11 @@ const SINGLE_TEST_PROPS = {
...
 
-let mochaSuiteFn;
-
-if (!supportedVersion(process.versions.node) || semver.lte(process.versions.node, '16.0.0')) {
-  mochaSuiteFn = describe.skip;
-} else {
-  mochaSuiteFn = describe;
-}
+// TODO: investigate as part of https://jsw.ibm.com/browse/INSTA-17132
+const mochaSuiteFn =
```
aryamohanan, line 56: IMO, we should wait for v23 to be released, and if this case fails, we can go ahead and merge this PR directly.

kirrg001: rdkafka has 45 days support policy. 

IMO no need to fix this as part of the Node v23 support.

But we can investigate straight after the release.

kirrg001 (approved):

PR #1402, opened 2024-10-22
fix: resolved issue with fetching agent key from ssm due to missing aws sdk

aryamohanan:
ref INSTA-17430

> could not fetch instana key from ssm parameter store using "instanakey" because the aws sdk is not available. reason: cannot find module 'aws-sdk'.

Node.js 18.x is using [AWS SDK v3](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/index.html). In our code, we are using AWS SDK v2. The problem is that, NodeJS 18.x runtime [does not come with version 2 of the AWS SDK](https://aws.amazon.com/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/) automatically packaged and ready to use. This is the reason we are getting the error for not being able to find aws-sdk.


packages/aws-lambda/src/ssm.js
```diff
@@ -49,52 +47,46 @@ module.exports.init = ({ logger }) => {
...
-     *
-     * Your code runs in an environment that includes the AWS SDK for JavaScript,
-     * with credentials from an AWS Identity and Access Management (IAM) role that you manage.
+     * As per AWS Lambda Node.js documentation: https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html
+     * The environment includes the AWS SDK for JavaScript, with credentials from an IAM role that you manage.
      */
-    // eslint-disable-next-line instana/no-unsafe-require, import/no-extraneous-dependencies
-    AWS = require('aws-sdk');
+    // eslint-disable-next-line import/no-extraneous-dependencies, instana/no-unsafe-require, prefer-const
+    const { SSMClient, GetParameterCommand } = require('@aws-sdk/client-ssm');
```
aryamohanan, line 54: In SDK v3, instead of requiring the entire `aws-sdk`,  imports only the necessary components: SSMClient and GetParameterCommand from `@aws-sdk/client-ssm`

packages/aws-lambda/src/ssm.js
```diff
@@ -49,52 +47,46 @@ module.exports.init = ({ logger }) => {
 
   try {
     /**
-     * From https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html:
-     *
-     * Your code runs in an environment that includes the AWS SDK for JavaScript,
-     * with credentials from an AWS Identity and Access Management (IAM) role that you manage.
+     * As per AWS Lambda Node.js documentation: https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html
```
aryamohanan, line 50: Rephrased the comments for better readability.

packages/aws-lambda/src/ssm.js
```diff
@@ -103,34 +95,34 @@ module.exports.waitAndGetInstanaKey = callback => {
...
-   * the chance is very high that the interval is not even used.
-   *
-   * In our tests it takes usually ~>150ms (remote call)
+   * The `GetParameterCommand` call in AWS typically takes about 80 to 500ms.
+   * This fetching process starts before the customer's handler is invoked,
+   * and a delay was noticed during a cold start, so this interval may be significant in that case.
+   * However, it's unlikely that this interval will be needed in other scenarios.
    */
-  let stopIntervalAfterMs = 250;
+  let stopIntervalAfterMs = 500;
```
aryamohanan, line 119: Increased the stopIntervalAfterMs, sometimes it seems the response took 80 to 500ms in tests. We can revisit this later.

packages/aws-lambda/test/integration_test/test_definition.js
```diff
@@ -277,48 +277,45 @@ function registerTests(handlerDefinitionPath, reduced) {
 
       let control;
 
-      before(callback => {
-        const AWS = require('aws-sdk');
-        const ssm = new AWS.SSM({ region: awsRegion });
+      before(async () => {
+        const { SSMClient, PutParameterCommand } = require('@aws-sdk/client-ssm');
```
aryamohanan, line 281: In tests, also using AWS SDK v3 format.

abhilash-sivan (approved):
LGTM

PR #1404, opened 2024-10-23
build: removed usage of :latest from v3

aryamohanan:
This change prevents 3.x images as being tagged as latest.


abhilash-sivan (approved):
LGTM

PR #1413, opened 2024-10-25
ci: fetch the prerelease version dynamically for prerelease pipeline

abhilash-sivan:
refs [INSTA-16692](https://jsw.ibm.com/browse/INSTA-16692)

### DIFF
--- a/bin/install-prerelease.sh
+++ b/bin/install-prerelease.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-PRERELEASE_NODE_VERSION=$1
+CONFIGURED_PRERELEASE_NODE_VERSION=$1
 
 fetch_latest_prerelease_node_version() {
   local latest_version
@@ -8,18 +8,11 @@ fetch_latest_prerelease_node_version() {
   latest_version=$(curl -sL https://nodejs.org/dist/index.json | jq -r '[.[]][0].version')
   latest_version=${latest_version#v}
   IFS='.' read -r major minor patch <<< "$latest_version"
-  
   major=$((major + 1))
-  
-  echo "${major}"
+  echo "${major}.0.0-rc"
 }
 
-# Check if PRERELEASE_NODE_VERSION is set; if not, set it to the latest prerelease
-if [ -z "$PRERELEASE_NODE_VERSION" ]; then
-  PRERELEASE_NODE_VERSION=$(fetch_latest_prerelease_node_version)
-fi
-
-echo "Installing Node.js prerelease version $PRERELEASE_NODE_VERSION..."
+PRERELEASE_NODE_VERSION=$(fetch_latest_prerelease_node_version)
 
 if ! command -v nvm &> /dev/null; then
   if ! curl -sSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.3/install.sh | bash &> /dev/null; then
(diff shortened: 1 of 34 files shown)


bin/install-prerelease.sh
```diff
@@ -1,8 +1,18 @@
...
 
-echo "Installing Node.js prerelease version $PRERELEASE_NODE_VERSION..."
+fetch_latest_prerelease_node_version() {
+  local latest_version
+  
+  latest_version=$(curl -sL https://nodejs.org/dist/index.json | jq -r '[.[]][0].version')
+  latest_version=${latest_version#v}
+  IFS='.' read -r major minor patch <<< "$latest_version"
+  major=$((major + 1))
+  echo "${major}.0.0-rc"
```
aryamohanan, line 12:
```suggestion
  echo "${major}"
```
It seems there will be an issue with how the prerelease version is being constructed as `${major}.0.0-rc`, which results in it being set to `24.0.0-rc`. This approach prevents the correct installation of the nightly release, as the prerelease version should simply be `24`.

bin/install-prerelease.sh
```diff
@@ -13,13 +23,27 @@ fi
...
+  export NVM_NODEJS_ORG_MIRROR
+
+  if ! nvm install "$PRERELEASE_NODE_VERSION"; then
+    echo "RC installation failed. Trying Nightly mirror..."
+    NVM_NODEJS_ORG_MIRROR="https://nodejs.org/download/nightly"
+    export NVM_NODEJS_ORG_MIRROR
+
+    if ! nvm install "$PRERELEASE_NODE_VERSION"; then
+      echo "Installation failed with both RC and Nightly mirrors."
+      echo "Trying to install configured version $CONFIGURED_PRERELEASE_NODE_VERSION..."
```
aryamohanan, line 40: What is the purpose of `CONFIGURED_PRERELEASE_NODE_VERSION`? What is its precedence? I believe we may not need it. 

Based on the code, it appears that even if `CONFIGURED_PRERELEASE_NODE_VERSION` is configured, the next major version takes precedence. Additionally, the nightly release for the upcoming version becomes available the day after the major release. Therefore, in my opinion, it seems unnecessary.

bin/install-prerelease.sh
abhilash-sivan: Yes thanks! updated

bin/install-prerelease.sh
abhilash-sivan: If we don't need the configuration, it can be removed without harm!

abhilash-sivan: Yes, I agree. We can remove the config option and relay on the latest prerelease

bin/install-prerelease.sh
```diff
@@ -1,8 +1,18 @@
...
+CONFIGURED_PRERELEASE_NODE_VERSION=$1
 
-echo "Installing Node.js prerelease version $PRERELEASE_NODE_VERSION..."
+fetch_latest_prerelease_node_version() {
+  local latest_version
+  
+  latest_version=$(curl -sL https://nodejs.org/dist/index.json | jq -r '[.[]][0].version')
+  latest_version=${latest_version#v}
+  IFS='.' read -r major minor patch <<< "$latest_version"
+  major=$((major + 1))
```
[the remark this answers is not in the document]
abhilash-sivan: Now we are fetching latest only

bin/install-prerelease.sh
```diff
@@ -13,13 +21,21 @@ fi
 
 source "$HOME/.nvm/nvm.sh"
 
-NVM_NODEJS_ORG_MIRROR="https://nodejs.org/download/rc"
-if ! nvm install "$PRERELEASE_NODE_VERSION" &> /dev/null; then
-  echo "RC installation failed. Trying Nightly mirror..."
-
-  NVM_NODEJS_ORG_MIRROR="https://nodejs.org/download/nightly"
+install_node_version() {
```
aryamohanan, line 24:
```suggestion
install_node_prerelease_version() {
```

aryamohanan (approved):
LGTM

PR #1419, opened 2024-10-28
ci: fix prerelease pipeline for node v24

abhilash-sivan:
refs [INSTA-17150](https://jsw.ibm.com/browse/INSTA-17150)

We are getting `lerna not found` issue in the prerelease pipeline only. All the test suits are scheduled to run after `install npm dependencies` step. However, this issue arises sometimes but it's not reproducible for the moment.

Conclusion:

The `install-npm-dependencies` task was giving a false positive for npm install failures. Updated the logic for npm installation in prerelease pipeline.

Also included changes for incorporating the node (currently v24) prerelease version in the pipeline

### DIFF
--- a/packages/shared-metrics/package.json
+++ b/packages/shared-metrics/package.json
@@ -66,9 +66,7 @@
     "tar": "^6.2.1"
   },
   "devDependencies": {
-    "@types/tar": "^6.1.6",
-    "event-loop-stats": "1.4.1",
-    "gcstats.js": "1.0.0"
+    "@types/tar": "^6.1.6"
   },
   "optionalDependencies": {
     "event-loop-stats": "1.4.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -34562,9 +34562,9 @@
       "version": "1.4.1",
       "resolved": "https://registry.npmjs.org/event-loop-stats/-/event-loop-stats-1.4.1.tgz",
       "integrity": "sha512-Wzohoz1JrQOBjt/shh5Z3/8Ti6SVoGl5nyX952Vcp5N56QVOtjPuJsQa+dEhsDJHu4QAFz45ePXRFq01skb9xA==",
-      "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
+      "optional": true,
       "dependencies": {
         "nan": "^2.14.0"
       },
@@ -36590,9 +36590,9 @@
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/gcstats.js/-/gcstats.js-1.0.0.tgz",
       "integrity": "sha512-gY4x4gWpZtXwIot2js62z4xNNZd+skNzfr4zpK5lVvDqZcqWKP/LhMKKi1Q/VFszJgqPz8ZQpO/OG4SHRgiTng==",
-      "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
+      "optional": true,
       "dependencies": {
         "nan": "^2.0.5"
       }
@@ -57041,9 +57041,7 @@
         "tar": "^6.2.1"
       },
       "devDependencies": {
-        "@types/tar": "^6.1.6",
-        "event-loop-stats": "1.4.1",
-        "gcstats.js": "1.0.0"
+        "@types/tar": "^6.1.6"
       },
       "optionalDependencies": {
         "event-loop-stats": "1.4.1",


.tekton/templates/test-task.yaml.template
```diff
@@ -91,6 +91,11 @@ spec:
           GIT_COMMIT_DESC=$(git log --format=%B -n 1 $(params.revision))
         fi
 
+        if ! command -v lerna &> /dev/null; then
```
abhilash-sivan, line 94: Added a check for lerna and if not installed, then installing it globally

abhilash-sivan:
The issue reappeared in this branch even after the fix code, node modules were not installed completely. The step shows completed (Green tick) in tekton, but in actuality `node-gyp` step failed and the node modules are not properly installed. This could be an issue with the prerelease pipeline and the dependencies. Even though a rerun resolves this issue, the root cause needs to be investigated further.


[pipeline URL](https://cloud.ibm.com/devops/pipelines/tekton/c2cd6a8d-ea5a-47b0-913e-cd172d63833f/runs/06d683e7-9867-486b-92be-24a1450e1cbf/install-npm-dependencies/npm-install?env_id=ibm:yp:eu-de)

packages/shared-metrics/package.json
```diff
@@ -66,9 +66,7 @@
     "tar": "^6.2.1"
   },
   "devDependencies": {
-    "@types/tar": "^6.1.6",
-    "event-loop-stats": "1.4.1",
```
abhilash-sivan, line 70: We should not add these native add-ons to devDependencies.

"These add-ons are "optional npm dependencies". If they can't be installed on your system, the collector still works as expected."
See public doc:
https://www.ibm.com/docs/en/instana-observability/1.0.290?topic=nodejs-collector-installation#native-add-ons

package.json
```diff
@@ -203,16 +203,16 @@
     "nats-v1": "npm:nats@^1.4.12",
     "node-fetch": "^3.3.2",
     "node-fetch-v2": "npm:node-fetch@^2.7.0",
-    "node-gyp": "^9.4.1",
+    "node-gyp": "^11.2.0",
```
abhilash-sivan, line 206: It was kept to v9.4.1 to support the node v14. Now we only support >= v18.

see: https://github.com/instana/nodejs/commit/210e33366b384f5baee3140858f6ae4d21a1f936

.tekton/tasks/install-npm-dependencies.yaml
```diff
@@ -36,21 +36,16 @@ spec:
...
           npm install npm -g
         fi     
 
-        if [ "$(params.prerelease)" == "true" ]; then
-          echo "Installing dependencies for prerelease..."
-
-          source bin/install-prerelease.sh 
-
-          npm install --loglevel verbose          
-          exit 0
```
kirrg001, line 51: > The install-npm-dependencies task was giving a false positive for npm install failures. Updated the logic for npm installation in prerelease pipeline.

I assume you refer to this line?
What I don't understand is, why did this throw a "lerna not found" error? 🤔

abhilash-sivan: Although the logs show Lerna being downloaded from the cache:
`npm http cache lerna@https://registry.npmjs.org/lerna/-/lerna-8.1.8.tgz 0ms (cache hit)`

The `install-npm-dependencies` step actually failed during native module loading. But it reported success (false positive). 
Because of this, Lerna wasn't properly installed in `node_modules`, and downstream tasks that depend on it failed with 'lerna not found' error

kirrg001: I see. So you did not fix "lerna not found", only fixed the false positive. Alrighty 👍

kirrg001: The initial idea was to not run prereleases through the NPM cache. Is this configured on the prerelease pipeline?

abhilash-sivan: Currently, we rely on cached modules to run tests in the prerelease pipeline.

If we disable caching, the installation step may succeed, but the native prebuilt binaries for the prerelease version will not be available. As a result, the pipeline would attempt to build them from source and fail during the native build process.

see: https://github.com/instana/nodejs/blob/221f0356bede4bf3120a5fac9b4def0b2a9bb1fc/packages/autoprofile/package.json#L29

.tekton/tasks/install-npm-dependencies.yaml
```diff
@@ -36,26 +36,24 @@ spec:
...
-        fi
-
         if [ "$(params.skip-cache)" == "true" ]; then
           echo "Skipping npm cache..."
 
           echo "Installing npm dependencies..."
           npm install --loglevel verbose
+
+          # Force non optional dependency installation
+          npm install gcstats.js event-loop-stats -w packages/shared-metrics --no-save
```
abhilash-sivan, line 56: Forcefully installing the native dependencies with `--no-save` rectified the issue in the currency bot mentioned [here](https://github.com/instana/nodejs/commit/605da7bb9d0a8ce1363a0988a9c03c3c474ebdbd).

This doesn't seem to affect the changes identified by the currency bot because of the `--no-save` flag. We can keep monitoring the behavior surely.

Ran a [currency bot pipeline test](https://cloud.ibm.com/devops/pipelines/tekton/c2cd6a8d-ea5a-47b0-913e-cd172d63833f/runs/644f9221-770a-4005-999f-2270b1b155a4/echo-env/execute?env_id=ibm:yp:eu-de) for verifying that the package-lock changes are not affected by this logic.

packages/autoprofile/package.json
```diff
@@ -26,7 +26,8 @@
   "scripts": {
     "audit": "npm audit --omit=dev",
     "node_modules:exists": "mkdir -p node_modules",
-    "install": "node-gyp-build",
+    "install": "node scripts/conditional-install.js",
```
abhilash-sivan, line 29: The autoprofile install has to be limited in prerelease pipeline as prebuilds won't be available in the early stages.

packages/shared-metrics/package.json
abhilash-sivan, line 70: Removed from devDependencies and handled in the `install-npm-dependencies` ci task

kirrg001: Discussed: we keep them

.tekton/tasks/install-npm-dependencies.yaml
```diff
@@ -36,26 +36,24 @@ spec:
...
         else
           npm install npm -g
         fi     
 
-        if [ "$(params.prerelease)" == "true" ]; then
-          echo "Installing dependencies for prerelease..."
-
-          source bin/install-prerelease.sh 
-
-          npm install --loglevel verbose
```
kirrg001, line 50: Discussed: we remove the 2 dev deps here manually for now.

packages/autoprofile/scripts/conditional-install.js
```diff
@@ -0,0 +1,19 @@
...
+ * (c) Copyright IBM Corp. 2025
+ */
+
+'use strict';
+const { execSync } = require('child_process');
+
+const version = process.version;
+
+// NOTE: skip native build step if using a pre-release (e.g., RC or nightly) version of Node.js
+if (/-(rc|nightly)/.test(version)) {
```
kirrg001, line 11: I rem we had a utility to check if a release is a prerelease or?

packages/autoprofile/scripts/conditional-install.js
```diff
@@ -0,0 +1,19 @@
...
+  // eslint-disable-next-line no-console
+  console.log(`Skipping native build for pre-release Node.js version: ${version}`);
+  process.exit(0);
+}
+
+// eslint-disable-next-line no-console
+console.log(`Running node-gyp-build for Node.js version: ${version}`);
+execSync('npm run build:native', { stdio: 'inherit' });
```
kirrg001, line 19: suggestion: npm run node-gyp-build
?

packages/aws-fargate/test/using_api/test.js
```diff
@@ -25,6 +27,8 @@ const instanaAgentKey = 'aws-fargate-dummy-key';
 
 function prelude() {}
 
+const isNodeV24 = semver.gte(semver.coerce(process.versions.node), '24.0.0');
```
kirrg001, line 30: qs: What is coerce?

abhilash-sivan: It trims the version to a comparable one.

> semver.valid(semver.coerce("v2")); // '2.0.0'
> semver.valid(semver.coerce("42.6.7.9.3-alpha")); // '42.6.7'

packages/autoprofile/scripts/conditional-install.js
abhilash-sivan: Sure we can eliminate the build:native script to directly run the node-gyp-build

packages/autoprofile/scripts/conditional-install.js
abhilash-sivan: We don't have a utility but we can use `semver.isPrerelease()`,  but I thought we avoid the import of semver here because this is an npm install step

packages/aws-fargate/test/using_api/test.js
kirrg001: cool

abhilash-sivan:
Closing as the issue is already fixed

PR #1444, opened 2024-11-12
ci: improved handling for failed layer uploads

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-12860

- continue to upload all layers
- if one layer fails, we continue but build will fail

**I did not manually test the changes.**

### DIFF
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -377,7 +377,7 @@ if [[ -z $SKIP_AWS_PUBLISH_LAYER ]]; then
           | jq '.Version' \
       ) || true  # NOTE: If the upload fails, the bash script should not fail.
 
-      if [[ ! $lambda_layer_version =~ ^[0-9]+$ ]]; then
+      if [[ -z $lambda_layer_version ]] || [[ ! $lambda_layer_version =~ ^[0-9]+$ ]]; then
         echo "   + ERROR: Layer upload failed for region $region. Skipping region."
         BUILD_SHOULD_FAIL=1
         continue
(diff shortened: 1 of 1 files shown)


packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -373,7 +375,13 @@ if [[ -z $SKIP_AWS_PUBLISH_LAYER ]]; then
           --output json \
           --compatible-runtimes nodejs18.x nodejs20.x \
           | jq '.Version' \
-      )
+      ) || true  # NOTE: If the upload fails, the bash script should not fail.
+
+      if [[ -z $lambda_layer_version ]] || [[ ! $lambda_layer_version =~ ^[0-9]+$ ]]; then
+        echo "   + ERROR: Layer upload failed for region $region. Skipping region."
```
aryamohanan, line 381: _nitpick_
```suggestion
        echo "   + ERROR: Failed to publish layer in region $region, continuing to the next region."
```

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -373,7 +375,13 @@ if [[ -z $SKIP_AWS_PUBLISH_LAYER ]]; then
...
+        BUILD_SHOULD_FAIL=1
```
aryamohanan, line 382: I'm sharing an idea that could be considered for a future PR or improvement regarding the China region handling. If a publish attempt fails, we could add the region to a `FAILED_REGIONS` list and establish a fresh connection to retry those failed regions at the end.

kirrg001: 👍 Yes. Feel free to raise a card <3

aryamohanan: Created, INSTA-19355

PR #1447, opened 2024-11-13
fix: resolved bunyan npm installation warning 

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-12533

- [x] lock file change
- [x] check and verify concatenated strings using `%s` and correct it
- [x] pass logs to agent stream
- [x] fix tests for pino and logger_test
- [x] fix test for pino/test
- [x] fix instrumentation - We don't want to trace Instana's own loggers ( __in) 
- [x] add additional test cases in pino
- [x] remove unwanted tests and enable only simple app test in bunyan


packages/collector/src/logger.js
```diff
@@ -31,40 +31,57 @@ const registry = {};
...
-      level: 'info'
-    });
+
+  if (isPino(parentLogger)) {
+    const multiStream = {
+      /**
+       * Custom write method to send logs to multiple destinations
+       * @param {string} chunk
+       */
+      write(chunk) {
```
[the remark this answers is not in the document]
abhilash-sivan: I tried implementing this use case using pino.transport, pino.destination and pino.multistream etc. And finally ended up with the current implementation which is working as expected like previous bunyan stream

packages/collector/test/tracing/protocols/graphql/schema.js
```diff
@@ -8,18 +8,18 @@ const graphQL = require('graphql');
 const graphqlSubscriptions = require('graphql-subscriptions');
 const data = require('./data');
 // Pino log spans are used to verify that follow up calls are traced correctly in a GraphQL entry.
-const pinoLogger = require('pino')();
+const bunyanLogger = require('bunyan').createLogger({ name: 'test-logger' });
```
abhilash-sivan, line 11: The reason for not using Winston here is that it is added to the list of packages allowed to load before the core module. In these situations, we might lose some spans that can cause the test to fail. We don't want to re-run the tests suppose this situation occurs

packages/collector/test/tracing/logger/pino/test.js
```diff
@@ -10,8 +10,8 @@ const supportedVersion = require('@instana/core').tracing.supportedVersion;
 const constants = require('@instana/core').tracing.constants;
 const config = require('../../../../../core/test/config');
 const testUtils = require('../../../../../core/test/test_util');
-const ProcessControls = require('../../../test_util/ProcessControls');
 const globalAgent = require('../../../globalAgent');
+const appControls = require('./controls');
```
kirrg001, line 14: issue: Please do not touch this test. Add a new internal logging integration test.

packages/collector/src/logger.js
```diff
@@ -31,40 +31,57 @@ const registry = {};
  * @param {boolean} [isReInit]
  */
 exports.init = function init(config, isReInit) {
+  const consoleStream = pino.destination(1);
```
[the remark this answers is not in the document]
abhilash-sivan: While using multistream, our logger did not write to the console(stdout) by default. 
We explicitly need to specify wherever this logs should go.
So added `stdout` and the `agentStream` as two attributes

packages/collector/src/logger.js
```diff
@@ -31,40 +33,55 @@ const registry = {};
...
   }
-  if (isBunyan(parentLogger)) {
-    // in case we are using a bunyan logger we also forward logs to the agent
-    /** @type {bunyan} */ (parentLogger).addStream({
-      type: 'raw',
-      stream: bunyanToAgentStream,
-      level: 'info'
-    });
+
+  if (isPino(parentLogger)) {
```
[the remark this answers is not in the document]
abhilash-sivan: yes, this is for the internal logger only.

packages/collector/src/agentConnection.js
```diff
@@ -267,9 +267,11 @@ exports.sendSpans = function sendSpans(spans, cb) {
     if (err && !maxContentErrorHasBeenLogged && err instanceof PayloadTooLargeError) {
       logLargeSpans(spans);
     } else if (err) {
-      logger.debug('Failed to send: %s', getSpanLengthInfo(spans));
+      const obj = getSpanLengthInfo(spans);
```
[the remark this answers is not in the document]
kirrg001: > logger.debug('Failed to send: %o', getSpanLengthInfo(spans));

Correction: This won't work for custom bunyan logger or winston.
Lets not use %s and no %o at all. Only normal strings.

If that works for all logger types, we can use this:
> logger.debug('Failed to send spans", getSpanLengthInfo(spans))

If that does not work, you can use:
> blabla = getSpanLengthInfo()
> logger.debug(`Failed to send spans ${blabla}`);

This is what you did already. Its probably better to raise a small PR to just remove this from our codebase. That would be good and makes the review easier.

aryamohanan (commented):
Why can't we use the native console instead of Pino?

packages/core/src/tracing/instrumentation/loggers/bunyan.js
```diff
@@ -29,13 +29,6 @@ function instrument(Logger) {
 function shimLog(markAsError) {
   return originalLog =>
     function () {
-      if (arguments.length === 0 || (this.fields && !!this.fields.__in)) {
```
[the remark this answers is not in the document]
abhilash-sivan: That was removed by mistake. Added it back

packages/collector/src/agentConnection.js
```diff
@@ -267,9 +267,11 @@ exports.sendSpans = function sendSpans(spans, cb) {
...
+      logger.debug({ obj }, 'Failed to transmit.');
```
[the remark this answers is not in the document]
abhilash-sivan: No this does not work with pino

packages/collector/src/logger.js
```diff
@@ -110,11 +127,17 @@ exports.getLogger = function getLogger(loggerName, reInitFn) {
 };
 
 /**
- * @param {import('bunyan') | *} _logger
+ * @param {*} _logger
  * @returns {boolean}
  */
-function isBunyan(_logger) {
-  return _logger instanceof bunyan;
+function isPino(_logger) {
```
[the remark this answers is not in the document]
abhilash-sivan: We can't check instanceof of pino. We need to rely on some internal functions to check this. 
We don't need this check for the new logic. removing the function

packages/collector/src/agentConnection.js
```diff
@@ -267,9 +267,11 @@ exports.sendSpans = function sendSpans(spans, cb) {
     if (err && !maxContentErrorHasBeenLogged && err instanceof PayloadTooLargeError) {
       logLargeSpans(spans);
     } else if (err) {
-      logger.debug('Failed to send: %s', getSpanLengthInfo(spans));
+      const spanInfo = getSpanLengthInfo(spans);
+      logger.debug({ spanInfo }, 'Failed to send.');
```
kirrg001, line 271: As discussed: Please use a standard log message `logger.debug(`Failed to send ${xxx}`).

packages/core/src/uninstrumentedLogger.js
```diff
@@ -0,0 +1,35 @@
...
+//       If pino was required before us, we leave the cache as it is.
+if (!pinoWasRequiredBeforeUs) {
+  Object.keys(require.cache).forEach(key => {
+    if (key.includes('node_modules/pino')) {
+      delete require.cache[key];
+    }
+  });
+}
+
+function createCustomLogger() {
```
kirrg001, line 26: question: Why do we need this? Can't we just exports pino?

packages/collector/test/tracing/logger/bunyan/app.js
```diff
@@ -13,17 +13,7 @@ process.on('SIGTERM', () => {
...
-const instana = require('../../../..');
-instana({
-  agentPort,
-  level: 'warn',
-  tracing: {
-    enabled: process.env.TRACING_ENABLED !== 'false',
-    forceTransmissionStartingAt: 1
-  }
-});
+require('../../../..')();
```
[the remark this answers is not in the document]
abhilash-sivan: This config was used in the integration test for bunyan. Since we changed this, we only need a simple application that uses bunyan from now on. Added these condition bases integration test for pino in separate section.

packages/core/src/tracing/instrumentation/loggers/bunyan.js
```diff
@@ -29,7 +29,7 @@ function instrument(Logger) {
...
+      if (arguments.length === 0) {
         // * arguments.length === 0 -> This is a logger.warn() type of call (without arguments), this will not log
         // anything but simply return whether the log level in question is enabled for this logger.
         // * this.fields.__in -> This is one of Instana's own loggers, we never want to trace those log calls.
```
kirrg001, line 35: Please clean up the comments.

packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/receiveMessage.js
```diff
@@ -32,7 +32,7 @@ const TeeLogger = require('../../../../../test_util/TeeLogger');
 const { sqs } = require('./sqsUtil');
 
 const instanaLogger = require('../../../../../../src/logger').getLogger('SQS receiver');
-instanaLogger.level('warn');
+instanaLogger.level = 'warn';
```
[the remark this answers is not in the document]
kirrg001: @abhilash-sivan ^

packages/core/src/uninstrumentedLogger.js
abhilash-sivan: We can do that, but if more customizations are needed, we can simply add to the custom object. Also, we are following this same format for  already existing uninstrumented packages

packages/collector/src/logger.js
```diff
@@ -35,36 +35,45 @@ exports.init = function init(config, isReInit) {
...
+
+    // No custom logger has been provided via config, we create a new pino logger as the parent logger for all loggers
     // we create later on.
-    parentLogger = bunyan.createLogger({
-      name: '@instana/collector',
-      thread: threadId,
-      __in: 1
-    });
-  }
-  if (isBunyan(parentLogger)) {
```
kirrg001, line 53: I reconsidered. Can we please keep the logic?

The agent log stream is pretty important and we do not want to break customer applications if they use a custom bunyan logger.

```js
if (config.logger && typeof config.logger.child === 'function') {
  ...  
  
  addAgentLogStream(...)
} else if (config.logger && hasLoggingFunctions(config.logger)) {
  ....
  
  addAgentLogStream(...)
}  else {
  ....
}

if (process.env['INSTANA_DEBUG']) {
...


function addAgentLogStream() {
  if (isBunyan(parentLogger)) {
  }
  
  # TODO: Add support for custom winston and custom pino logger etc.
}

Comment (kirrg001) [reply packages/collector/src/logger.js:53]: Streams should be covered in packages/collector/test/logger_test.js.
A simple test which checks the number of streams.

I'd suggest, extend this test on main and raise a PR. We merge it. Rebase. Thanks!

File: packages/collector/test/logger_test.js [hunk 16]
```diff
@@ -23,68 +31,75 @@ describe('logger', () => {
...
-    expect(logger.level()).to.equal(40);
+    expect(logger.level).to.equal(logLevel[40]);
   });
 
   it('should use debug log level when INSTANA_DEBUG is set', () => {
     process.env.INSTANA_DEBUG = 'true';
     log.init({});
     const logger = log.getLogger('childName');
-    expect(logger.level()).to.equal(20);
+    expect(logger.level).to.equal(logLevel[20]);
```
kirrg001, line 95: suggestion: Its a nice idea to add the log levels on top. To increase the readability I'd suggest to do

```js

const logLevels = {
  warn: 20
}

logLevel.warn
```

Just the other way around.

packages/collector/test/tracing/logger/instana-logger/app-instana-receives-non-pino-logger.js
```diff
@@ -45,7 +44,7 @@ const express = require('express');
 const morgan = require('morgan');
 const port = require('../../../test_util/app-port')();
 const app = express();
-const logPrefix = `Bunyan App [Instana receives non-Bunyan logger] (${process.pid}):\t`;
+const logPrefix = `Pino App [Instana receives non-Pino logger] (${process.pid}):\t`;
```
kirrg001, line 47: suggestion: Instana receives custom dummy logger

packages/collector/test/tracing/logger/instana-logger/app-instana-receives-pino-logger.js
```diff
@@ -14,7 +13,7 @@ process.on('SIGTERM', () => {
 });
 
 const agentPort = process.env.AGENT_PORT;
-const buynan = require('bunyan');
+const pino = require('pino');
```
kirrg001, line 16: Please add: app-instana-receives-bunyan-logger

packages/collector/test/tracing/logger/instana-logger/controls.js
```diff
@@ -19,20 +17,18 @@ const agentControls = require('../../../globalAgent').instance;
...
 
-// TODO: transform into class
-
 exports.registerTestHooks = (opts = {}) => {
   let appName = 'app.js';
   if (opts.instanaLoggingMode) {
     switch (opts.instanaLoggingMode) {
-      case 'instana-creates-bunyan-logger':
-        appName = 'app-instana-creates-bunyan-logger.js';
+      case 'instana-creates-pino-logger':
```
kirrg001, line 24: suggestion: you can also use generic names such as "instana-uses-default-logger" and "instana-uses-custom-pino-logger" etc. With this, we never have to change the filenames or strings when going away from pino.

packages/collector/test/tracing/logger/instana-logger/controls.js
```diff
@@ -84,3 +81,7 @@ function waitUntilServerIsUp() {
 exports.getPid = () => appProcess.pid;
 
 exports.trigger = (level, headers = {}) => fetch(`http://127.0.0.1:${appPort}/${level}`, { headers });
+
+exports.stop = async () => {
```
kirrg001, line 85: ?

packages/collector/test/tracing/logger/instana-logger/test.js
```diff
@@ -0,0 +1,67 @@
...
+        agentControls.getSpans().then(spans => {
+          testUtils.expectAtLeastOneMatching(spans, [
+            span => expect(span.n).to.equal('node.http.server'),
+            span => expect(span.f.e).to.equal(String(appControls.getPid())),
+            span => expect(span.f.h).to.equal('agent-stub-uuid')
+          ]);
+
+          // verify that nothing logged by Instana has been traced
+          const allpinoSpans = testUtils.getSpansByName(spans, 'log.pino');
+          expect(allpinoSpans).to.be.empty;
```
kirrg001, line 62: We always add a tiny delay before the initial `getSpans` to avoid false positives.

> delay(250)
> testUtils.retry(() =>

packages/collector/src/logger.js
```diff
@@ -99,6 +109,11 @@ exports.getLogger = function getLogger(loggerName, reInitFn) {
     _logger = parentLogger;
   }
 
+  // Extended method to set log level
+  _logger.setLoggerLevel = function (/** @type {string | number} */ level) {
```
kirrg001, line 113: IMO this is a wrong semantical place. Define the fn once in the init fn. Not here please

packages/core/src/uninstrumentedLogger.js
kirrg001: I'd suggest to delete it because it does nothing at the moment. If I am wrong, let me know 👍 

> Also, we are following this same format for already existing uninstrumented packages

Not really, because the existing uninstrumented libraries are core libraries. And yes, for core libraries we need to clone. For pino, I don't think we have to clone/copy 🤔

packages/collector/src/logger.js
```diff
@@ -31,40 +33,55 @@ const registry = {};
  * @param {boolean} [isReInit]
  */
 exports.init = function init(config, isReInit) {
+  const consoleStream = pino.destination(1);
+
   if (config.logger && typeof config.logger.child === 'function') {
-    // A bunyan or pino logger has been provided via config. In either case we create a child logger directly under the
-    // given logger which serves as the parent for all loggers we create later on.
+    // A pino logger has been provided via config; create a child logger directly under it.
```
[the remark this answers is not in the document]
abhilash-sivan: Its true for both bunyan and pino.. updated the comments

packages/core/src/uninstrumentedLogger.js
abhilash-sivan: In that case, should we remove the unintstrumented file from core and place it in collector ? That way pino won't be our core dependency

kirrg001:
> Why can't we use the native console instead of Pino?

@aryamohanan Because native console log has no features others than logging to stdout or stderr. How do you want to handle streams, log levels, child loggers etc?

packages/core/src/uninstrumentedLogger.js
kirrg001: Yeah makes sense 👍 Good idea

aryamohanan:
> @aryamohanan Because native console log has no features others than logging to stdout or stderr. How do you want to handle streams, log levels, child loggers etc?

I haven't explored this in depth, but I was wondering if it's possible to extend the native console to build a custom logger. By overriding console, could we incorporate additional features like log levels, timestamps, or custom streams? So, we could avoid using an external library. Just asking about the possibilities.

packages/collector/src/logger.js
abhilash-sivan: Now we can add custom pino log stream here, but if we are to consider adding that for bunyan too, then we will have to require both pino and bunyan in this file. Do we want to use `addAgentLogStream` to send logs from all the loggers ?

abhilash-sivan: Also, I have added assertion for bunyan log streams, but it is found that  

> "When using pino with a multi-stream setup, the logger's streams aren't directly exposed",

 so we can only check the properties and custom functions we added in the multistream.

packages/collector/test/tracing/logger/instana-logger/app-instana-receives-bunyan-logger.js
```diff
@@ -1,6 +1,5 @@
 /*
- * (c) Copyright IBM Corp. 2021
- * (c) Copyright Instana Inc. and contributors 2018
+ * (c) Copyright IBM Corp. 2024
```
kirrg001, line 2: FYI We do not change the copyright year when renaming/moving a file.

abhilash-sivan: okay. reverting back

packages/collector/src/logger.js
```diff
@@ -39,33 +39,68 @@ exports.init = function init(config, isReInit) {
...
   } else if (config.logger && hasLoggingFunctions(config.logger)) {
-    // A custom non-bunyan logger has been provided via config. We use it as is.
+    // A custom non-bunyan/non-pino logger has been provided via config. We use it as is.
     parentLogger = config.logger;
   } else {
-    // No custom logger has been provided via config, we create a new bunyan logger as the parent logger for all loggers
+    // No custom logger has been provided via config, we create a new pino logger as the parent logger for all loggers
     // we create later on.
-    parentLogger = bunyan.createLogger({
+    // @ts-ignore
```
kirrg001, line 47: Why do we need ts-ignore suddenly?

packages/collector/src/logger.js
```diff
@@ -39,33 +39,68 @@ exports.init = function init(config, isReInit) {
...
     });
   }
-  if (isBunyan(parentLogger)) {
+
+  // Passing the log stream to agentStream for Debugging purposes
+  // TODO: consider adding winston and other major loggers also for this
+  if (isPinoLogger(parentLogger)) {
+    // This consoleStream creates a destination stream for the logger that writes log data to the standard output.
+    // Since we are using multistream here, this needs to be specified explicitly
+    const consoleStream = pino.destination(process.stdout);
```
kirrg001, line 61: Why do we attach a console stream logger for a custom pino logger? I expected this logic only for our internal pino instance.

I think it has to work like this:

Create internal default pino instance with stdout only.
From the docs
>  // By default, Pino logs to stdout unless a 'destination' is specified

if (logger is pino )
const streams = [
  { stream: parentLogger.destination }, // EXISTING stream
  { stream: customStream }              // Custom stream
];

// Create a new multistream logger
parentLogger = pino.multistream(streams, { level: 'info' });

packages/collector/src/logger.js
```diff
@@ -39,33 +39,68 @@ exports.init = function init(config, isReInit) {
...
-      /** @type {bunyan} */ (parentLogger).level(/** @type {import('bunyan').LogLevel} */ (config.level));
-    } else if (process.env['INSTANA_LOG_LEVEL']) {
-      /** @type {bunyan} */ (parentLogger).level(
-        /** @type {import('bunyan').LogLevel} */ (process.env['INSTANA_LOG_LEVEL'].toLowerCase())
-      );
-    }
+  }
+
+  // attaching custom method after reinitializing logger
+  parentLogger.setLoggerLevel = function (/** @type {string | number} */ level) {
```
kirrg001, line 94: Please use a clear fn name such as _setInstanaLogLevel . We extend external logger instances.

kirrg001: It should be clear where its coming from.

packages/collector/src/logger.js
```diff
@@ -39,33 +39,68 @@ exports.init = function init(config, isReInit) {
       __in: 1
```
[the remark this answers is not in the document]
abhilash-sivan: Okay added it back. I thought this might be used in agent side

packages/collector/src/logger.js
abhilash-sivan: Since we removed the custom object export from unInstrumented logger and replaced it with pino library itself, the type warning is showing

packages/collector/src/logger.js
abhilash-sivan: Yeah it makes sense. I have changed the naming 👍

packages/core/src/tracing/instrumentation/loggers/pino.js
```diff
@@ -18,6 +18,9 @@ const cls = require('../../cls');
 let isActive = false;
 
 exports.init = function init() {
+  // NOTE: There is a bug in the pino instrumentation. As soon as you require pino twice,
```
aryamohanan, line 21:
```suggestion
 // TODO: Fix the issue with Pino instrumentation. If Pino is required multiple times,
 //       only the first instance gets instrumented. This behavior is caused by `onFileLoad`.
 //       Fix is being tracked in https://jsw.ibm.com/browse/INSTA-23066.
```

packages/core/src/uninstrumentedLogger.js
```diff
@@ -0,0 +1,38 @@
...
+  const customLogger = Object.assign(function (/** @type {any} */ ...args) {
+    // @ts-ignore
```
aryamohanan, line 31:
```suggestion
```

packages/core/src/uninstrumentedLogger.js
```diff
@@ -0,0 +1,38 @@
+/*
```
aryamohanan, line 1: I couldn’t find any references to this file. Where is it being used? I only found the usage of `uninstrumentedLogger` in the collector.

packages/core/src/uninstrumentedLogger.js
```diff
@@ -0,0 +1,38 @@
...
+ * (c) Copyright IBM Corp. 2024
+ */
+
+'use strict';
+
+const pinoWasRequiredBeforeUs = Object.keys(require.cache).some(key => key.includes('node_modules/pino'));
+
+// eslint-disable-next-line import/no-extraneous-dependencies, instana/no-unsafe-require
+const logger = require('pino');
```
aryamohanan, line 10: Is this a prod dependency of core?

packages/collector/test/tracing/logger/instana-logger/controls.js
```diff
@@ -1,12 +1,10 @@
 /*
- * (c) Copyright IBM Corp. 2021
- * (c) Copyright Instana Inc. and contributors 2016
+ * (c) Copyright IBM Corp. 2024
```
aryamohanan, line 2: We should not edit this.

packages/collector/test/tracing/logger/instana-logger/app-instana-uses-default-logger.js
```diff
@@ -1,6 +1,5 @@
 /*
- * (c) Copyright IBM Corp. 2021
- * (c) Copyright Instana Inc. and contributors 2018
+ * (c) Copyright IBM Corp. 2024
```
aryamohanan, line 2: Here also

packages/collector/src/logger.js
```diff
@@ -129,3 +157,25 @@ function hasLoggingFunctions(_logger) {
...
+  } else {
+    _logger.level = level;
+  }
+}
+
+/**
+ * @param {*} _logger
+ * @returns {boolean}
+ */
+function isPinoLogger(_logger) {
```
aryamohanan, line 175: Consider enhancing the `isPinoLogger` function to specifically validate Pino loggers by checking for Pino-specific properties. The current implementation might return a positive result for other loggers that have similar method names (like child and level). Adding more precise checks would improve accuracy and avoid false positives.

packages/collector/src/uninstrumentedLogger.js
```diff
@@ -0,0 +1,29 @@
+/*
+ * (c) Copyright IBM Corp. 2024
+ */
+
+'use strict';
+
+const pinoWasRequiredBeforeUs = Object.keys(require.cache).some(key => key.includes('node_modules/pino'));
```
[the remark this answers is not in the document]
abhilash-sivan: Yes, we are only using from collector now. Removed the core dependency . Updated the code

packages/core/src/uninstrumentedLogger.js
abhilash-sivan: Removed unused file

packages/collector/src/logger.js
abhilash-sivan: I think we are currently following this logic.
We are attaching multistream only if the `if (logger is pino )` .

Please let me know if there is something I am missing out.

packages/collector/src/logger.js
abhilash-sivan: Yes thanks for suggesting, I also had this doubt. I will look into it and update with more checks if possible

packages/collector/src/agentConnection.js
```diff
@@ -267,9 +267,11 @@ exports.sendSpans = function sendSpans(spans, cb) {
     if (err && !maxContentErrorHasBeenLogged && err instanceof PayloadTooLargeError) {
       logLargeSpans(spans);
     } else if (err) {
-      logger.debug('Failed to send: %s', getSpanLengthInfo(spans));
+      const spanInfo = getSpanLengthInfo(spans);
+      logger.debug(`Failed to sent: ${JSON.stringify(spanInfo)}`);
```
aryamohanan, line 271: Just a small correction: 'Failed to sent' should be 'Failed to send'

packages/collector/src/logger.js
```diff
@@ -35,37 +35,70 @@ exports.init = function init(config, isReInit) {
...
+    // @ts-ignore
+    parentLogger = pino(
+      {
+        ...parentLogger.levels,
+        level: parentLogger.level || 'info',
+        base: parentLogger.bindings()
+      },
+      multiStream
+    );
+  } else if (parentLogger && parentLogger.addStream) {
```
aryamohanan, line 82: @qs: What is this else case? how do we verify this as bunyan logger?

packages/collector/src/logger.js
aryamohanan: Were you able to check this https://github.com/pinojs/pino/blob/HEAD/docs/api.md#pino-multistream?

packages/collector/src/logger.js
abhilash-sivan: currently, the addstream method is directly available for bunyan only

abhilash-sivan:
> Were you able to check this https://github.com/pinojs/pino/blob/HEAD/docs/api.md#pino-multistream?

Yes I checked this and fine tuned it to support the existing agentStream functionality without major code changes

packages/collector/src/logger.js
aryamohanan: In this case, any logger that supports the addStream method will be accepted. Our future goal is to implement this logic for other loggers as well, correct?

abhilash-sivan: Yes that is correct!

packages/collector/test/tracing/logger/instana-logger/app-instana-receives-custom-dummy-logger.js
```diff
@@ -1,6 +1,5 @@
 /*
- * (c) Copyright IBM Corp. 2021
- * (c) Copyright Instana Inc. and contributors 2018
+ * (c) Copyright IBM Corp. 2024
```
aryamohanan, line 2: here

packages/collector/test/tracing/logger/instana-logger/controls.js
```diff
@@ -1,12 +1,11 @@
 /*
  * (c) Copyright IBM Corp. 2021
- * (c) Copyright Instana Inc. and contributors 2016
+ * (c) Copyright Instana Inc. and contributors 2018
```
aryamohanan, line 3: here

aryamohanan (approved):
The PR looks in good shape now. A few minor comments need to be addressed, I'm pre-approving it for now.

packages/collector/src/logger.js
abhilash-sivan: Yes I checked this and it was not compatible with our current logic in sending streams to agent. So I used multistream in a custom way

kirrg001:
Please always try to think in the perspective of the customer.

Commit suggestion
> fix: resolved bunyan npm installation warning #1447

packages/collector/src/logger.js
kirrg001: Yeah but why?

I assume its because our custom typescript GenericLogger is no longer compatible with pino

export interface GenericLogger {
  log?: (...args: any[]) => void;
  error: (...args: any[]) => void;
  warn: (...args: any[]) => void;
  info: (...args: any[]) => void;
  [key: string]: any;
  child?: (fields: Record<string, any>) => GenericLogger;
  /**
   * @param level - desired logging level (e.g., 'debug', 'warn').
   */
  setLoggerLevel?: (level: string | number) => void;
}

packages/collector/src/logger.js
```diff
@@ -35,37 +35,70 @@ exports.init = function init(config, isReInit) {
     // A bunyan or pino logger has been provided via config. In either case we create a child logger directly under the
     // given logger which serves as the parent for all loggers we create later on.
     parentLogger = config.logger.child({
-      module: 'instana-nodejs-logger-parent',
-      __in: 1
```
kirrg001, line 39: Did you test that a custom logger won't be instrumented?

```
const instana = require('@instana/collector')();

// can be pino or winston as well
const bunyan = require('bunyan');
const logger = bunyan.createLogger(...);
instana.setLogger(logger);

-> any internal log should not be traced
```

packages/collector/src/logger.js
kirrg001: Okay thanks for checking.

Again, it feels like this is wrong:
> const consoleStream = pino.destination(process.stdout);

Can't we do?
> const consoleStream = pino.destination(parentLogger.destination);

packages/collector/src/logger.js
kirrg001: Also: the fn name is wrong. It must be _setInstanaLogLevel

packages/collector/test/logger_test.js
```diff
@@ -128,46 +135,87 @@ describe('logger', () => {
...
-
     expectAtLeastOneMatching(
       logger.streams,
       stream => expect(stream.type).to.equal('raw'),
       stream => expect(stream.level).to.equal('info')
     );
   });
+
+  // TODO: This test case is not consistent.
+  // Need to investigate further why its failing in the initial run and succeeding after that
```
kirrg001, line 189: This is because you are testing against the wrong pino instance.

First run:
- we require pino internally, we remove the instance from the cache
- second run, a pino logger is already in the require cache, we leave it there and we use the existing cached one

Please keep both tests:
1. This test and multiStream should be null
2. Cop the test and use the symbol from the right internal logger

packages/collector/src/logger.js
kirrg001: Cant we set this fn in the uninstrumtnedLogger file? I guess it belongs there anyway?

aryamohanan:
Putting this here —please check the public documentation and update it as accordingly: [Setting the Logger After Initialization](https://www.ibm.com/docs/en/SSE1JP5_current/src/pages/ecosystem/node-js/api.html#setting-the-logger-after-initialization)

packages/collector/test/logger_test.js
abhilash-sivan: We are not encountering the issue right now.
previously the uninstrumentedLogger was exporting a custom function

```
function createCustomLogger() {
  const customPino = Object.assign(function (...args) {
    return logger(...args);
  }, logger);

  return customPino;
}
```

Now we are directly exporting the package itself. That could be the reason the tests are passing now. 
Please share your thoughts

packages/collector/src/logger.js
abhilash-sivan: This test case is already covered in this [file](https://github.com/instana/nodejs/blob/77387114ddc7918e7abb1fdc50a514e176a76d6b/packages/collector/test/tracing/logger/instana-logger/app-instana-receives-custom-dummy-logger.js).

please let me know whether this is enough

packages/collector/src/logger.js
abhilash-sivan: I tried this and not seeing any issues with logging. I will push this to see if all the tests are passing or not and we can decide. Thanks

packages/collector/src/logger.js
abhilash-sivan: That is doable. But IMO there is already a utility function isPinoLogger present in the logger.js file.

also we intend to write this `function setLoggerLevel` for any loggers thats coming to the logger init function.
So I prefer this location. If it is semantically wrong, I can make adjustments

kirrg001: But _setInstanaLogLevel is only used in our tests at the moment. It does not make sense to me somehow.

I'd suggest, remove the fn from the typescript definition. And move the fn to uninstrumentedLogger. This should only be an internal fn for now.

packages/collector/src/logger.js
abhilash-sivan: I have added rerun and delay in the latest [changes](https://github.com/instana/nodejs/blob/2ad2ab70d346381cf6f1993355b6ea3528db730f/packages/collector/test/tracing/logger/instana-logger/test.js#L64). Could you please reverify ?

packages/collector/src/logger.js
```diff
@@ -35,37 +35,66 @@ exports.init = function init(config, isReInit) {
...
     parentLogger = config.logger;
   } else {
-    // No custom logger has been provided via config, we create a new bunyan logger as the parent logger for all loggers
+    // No custom logger has been provided via config, we create a new pino logger as the parent logger for all loggers
     // we create later on.
-    parentLogger = bunyan.createLogger({
+    // @ts-ignore
+    parentLogger = pino({
       name: '@instana/collector',
       thread: threadId,
```
kirrg001, line 49: issue: There is no field thread

https://github.com/pinojs/pino/blob/main/pino.d.ts

kirrg001: issue: @ts-ignore is not right

kirrg001: https://github.com/instana/nodejs/pull/1447/commits/ffdc7750aa76a391c3e872b2b9936b01c7975ad3

kirrg001 (approved):
I think this is good to go 😅

kirrg001:
> Putting this here —please check the public documentation and update it as accordingly: [Setting the Logger After Initialization](https://www.ibm.com/docs/en/SSE1JP5_current/src/pages/ecosystem/node-js/api.html#setting-the-logger-after-initialization)

@abhilash-sivan Please add a TODO to your JIRA. Thanks!

PR #1453, opened 2024-11-19
ci: added env variable to handle publish failure

aryamohanan:


packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -448,7 +459,7 @@ echo "step 9/9: cleaning up"
 rm -rf $TMP_ZIP_DIR
 rm -rf $ZIP_NAME
 
-if [[ $BUILD_SHOULD_FAIL -eq 1 ]]; then
+if [[ $BUILD_X86_64_FAIL -eq 1 && $BUILD_arm64_FAIL -eq 1 ]]; then
```
abhilash-sivan, line 462: This condition will kick in only if both `BUILD_X86_64_FAIL` and `BUILD_arm64_FAIL` fails.

We should exit if either one fails right?

aryamohanan: That is happening now and it blocks the execution of the other task from running.

kirrg001 (approved):
Suggestion

> ci: handled partial failure in publish-layer script for arm64 & x64 #1453

PR #1457, opened 2024-11-21
build: bumped node-rdkafka from 3.1.1 to 3.2.0

aryamohanan:
The tests were failing after upgrading node-rdkafka from v3.1.1 to v3.2.0 due to an underlying [issue](https://github.com/confluentinc/librdkafka/issues/4870) with librdkafka v2.6.0, which causes an error for Kafka versions earlier than 2.7.
 I have raised an issue regarding this [here](https://github.com/Blizzard/node-rdkafka/issues/1105). In the meantime, as we are using an older Kafka image version, I have updated the image to a version greater than 2.7 to work around the issue.

### DIFF
--- a/docker-compose-base.yaml
+++ b/docker-compose-base.yaml
@@ -80,7 +80,7 @@ services:
       - 2181:2181
 
   kafka:
-    image: wurstmeister/kafka:2.12-2.2.1
+    image: wurstmeister/kafka:2.13-2.8.1
     platform: linux/amd64
     ports:
       - 9092:9092
--- a/.tekton/assets/sidecars.json
+++ b/.tekton/assets/sidecars.json
@@ -46,7 +46,7 @@
     },
     {
       "name": "kafka",
-      "image": "wurstmeister/kafka:2.12-2.2.1",
+      "image": "wurstmeister/kafka:2.13-2.8.1",
       "env": [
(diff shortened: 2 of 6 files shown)


docker-compose-base.yaml
```diff
@@ -80,7 +80,7 @@ services:
       - 2181:2181
 
   kafka:
-    image: wurstmeister/kafka:2.12-2.2.1
+    image: wurstmeister/kafka:2.13-2.8.1
```
aryamohanan, line 83: We were using an older version of the image, but I have now updated it to the latest available version https://hub.docker.com/r/wurstmeister/kafka/tags.

kirrg001 (approved):
Further context:

[wurstmeister/kafka:2.12-2.2.1](https://hub.docker.com/layers/wurstmeister/kafka/2.12-2.2.1/images/sha256-1ab3df91e1e8070e13a101d98c1df41d19a4e5bade3ab98266c7902866fbfdf5?context=explore) is 5y old and is using kafka version 2.4.0.

> ARG kafka_version=2.4.0

[wurstmeister/kafka:2.13-2.8.1](https://hub.docker.com/layers/wurstmeister/kafka/2.13-2.8.1/images/sha256-1e05be28055c8c3527ba4f892b5574411ee713899d1181d43dfdbc4dffcf78df?context=explore) is using

> ARG kafka_version=2.8.1

The update resolves https://github.com/confluentinc/librdkafka/pull/4871/files, because the error occurs in Kafka < 2.7

👍

PR #1461, opened 2024-11-26
ci: automated currency-bot branch deletion if no PRs existed   

aryamohanan:
At the moment, we are manually deleting the `currency-bot` branch after merging the currency PRs. However, there are times when we forget to delete the branch. In such cases, the new PR cannot be created because the branch already exists. To eliminate the need for manual intervention, introduced the  automated the deletion of the currency-bot branch if no open PRs exist.

### DIFF
--- a/.tekton/tasks/update-dependencies.yaml
+++ b/.tekton/tasks/update-dependencies.yaml
@@ -36,17 +36,25 @@ spec:
         && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
         && apt update \
         && apt install gh -y
-
+        apt install jq -y
+        
         gh auth login --with-token
-
+        
+        OPEN_PRS=$(gh pr list --head currency-bot --base main --state open --json number | jq 'length')
+  
         git ls-remote --exit-code --heads origin currency-bot >/dev/null 2>&1
         EXIT_CODE=$?
 
-        if [[ $EXIT_CODE != '0' ]]; then          
+        if [[ $EXIT_CODE -eq 0 && $OPEN_PRS -eq 0]]; then
+          echo "No open PRs found. Deleting branch currency-bot."
+          git push origin --delete currency-bot
+        fi
+
+        if [[ $EXIT_CODE != 0 ]]; then
           echo "Running patch and minor updates..."
           MAJOR_UPDATES_MODE=false BRANCH=currency-bot node bin/currency/update-currencies.js    
         else
-          echo "Skipping patch/minor updates. Branch exists."
+          echo "Open PRs exist for branch currency-bot. Skipping patch/minor updates."
         fi        
 
         echo "Running major updates..."


.tekton/tasks/update-dependencies.yaml
```diff
@@ -36,17 +36,27 @@ spec:
...
-        EXIT_CODE=$?
+        BRANCH_EXISTS=$?
 
-        if [[ $EXIT_CODE != '0' ]]; then          
+        if [[ $BRANCH_EXISTS != 0 ]]; then
           echo "Running patch and minor updates..."
           MAJOR_UPDATES_MODE=false BRANCH=currency-bot node bin/currency/update-currencies.js    
         else
-          echo "Skipping patch/minor updates. Branch exists."
+          echo "currency-bot branch exists. Checking for open pull requests.."
```
kirrg001, line 50: issue: The logic will only work for the major updates.

See line 63
> MAJOR_UPDATES_MODE=true BRANCH=currency-bot-major node bin/currency/update-currencies.js

The logic needs to be added before line 46.

kirrg001 (changes requested):
Cool idea 👍

.tekton/tasks/update-dependencies.yaml
aryamohanan: Good catch ❤️

.tekton/tasks/update-dependencies.yaml
```diff
@@ -36,17 +36,29 @@ spec:
...
+
+          OPEN_PRS=$(gh pr list --head currency-bot --base main --state open --json number | jq 'length')
+
+          if [[ $OPEN_PRS -eq 0 ]]; then
+            echo "No open PRs found. Deleting branch currency-bot..."
+            git push origin --delete currency-bot
+            echo "Running patch and minor updates..."
+            MAJOR_UPDATES_MODE=false BRANCH=currency-bot node bin/currency/update-currencies.js
```
kirrg001, line 58: suggestion: To avoid having to use the same command twice:

in line 45:
```
OPEN_PRS=$(gh pr list --head currency-bot --base main --state open --json number | jq 'length')
git ls-remote --exit-code --heads origin currency-bot >/dev/null 2>&1
BRANCH_EXISTS=$?

# auto cleanup in case we forgot to manually delete the branch
if BRANCH EXISTS AND NO OPEN PULL REQUESTS
  git push origin --delete currency-bot

...continue with normal logic...
```

abhilash-sivan (approved):
LGTM

.tekton/tasks/update-dependencies.yaml
```diff
@@ -36,17 +36,25 @@ spec:
...
         gh auth login --with-token
-
+        
+        OPEN_PRS=$(gh pr list --head currency-bot --base main --state open --json number | jq 'length')
+  
         git ls-remote --exit-code --heads origin currency-bot >/dev/null 2>&1
-        EXIT_CODE=$?
+        BRANCH_EXISTS=$?
+
+        if [[ $BRANCH_EXISTS -eq 0 && $OPEN_PRS -eq 0]]; then
```
kirrg001, line 48: suggestion: The renaming from `EXIT_CODE` to `BRANCH_EXISTS` is semantically wrong.
`BRANCH_EXISTS` of the value 0 means it exists. But it should be 1 to be readable. The usage of `EXIT_CODE` was more readable because its a known concept that 0 is positive.

```sh
if git ls-remote --exit-code --heads origin currency-bot >/dev/null 2>&1; then
    BRANCH_EXISTS=true
else
    BRANCH_EXISTS=false
fi

if [[ "$BRANCH_EXISTS" = true && $OPEN_PRS -eq 0 ]]; then
...
```

kirrg001 (approved):

PR #1470, opened 2024-12-03
test: added span length assertions for mysql tests

aryamohanan:
Noticed that mysql tests lack strict assertions regarding the span lengths.


kirrg001 (approved):
LGTM

PR #1473, opened 2024-12-04
ci: added separate task for china region lambda layer publish

aryamohanan:
ref INSTA-12860

### DIFF
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -44,7 +44,6 @@ if [[ -z $PUBLISH_TO_CHINA_REGIONS ]]; then
   PUBLISH_TO_CHINA_REGIONS=false
 fi
 
-
 PACKAGE_NAMES="@instana/aws-lambda@$PACKAGE_VERSION instana-aws-lambda-auto-wrap@$PACKAGE_VERSION"
 
 # The default layer name is instana-nodejs. If you want to push experimental changes under a different layer name, you
@@ -288,7 +287,7 @@ if [[ -z $SKIP_AWS_PUBLISH_LAYER ]]; then
   echo "step 6/9: publishing $ZIP_NAME as AWS Lambda layer $LAYER_NAME to specifed regions"
   
   if [[ "$PUBLISH_TO_CHINA_REGIONS" == "true" ]]; then
-    echo "Publishing only to Chinese regions: $CHINESE_REGIONS"
+    echo "Publishing to Chinese regions: $CHINESE_REGIONS"
     REGIONS="$CHINESE_REGIONS"
   else
     # Remove Chinese regions from REGIONS
@@ -303,7 +302,7 @@ if [[ -z $SKIP_AWS_PUBLISH_LAYER ]]; then
       # Update REGIONS with the filtered list
       REGIONS="$FILTERED_REGIONS"
     done
-    echo "Publishing to  regions except Chinese regions.."
+    echo "Publishing to regions except Chinese regions.."
   fi
 
 
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -16,7 +16,7 @@ EOF
   }
 }
 
-
+# Ensure required tools are installed.
 check_installation npm
 check_installation aws
 check_installation docker
@@ -24,7 +24,6 @@ check_installation jq
 check_installation zip
 
 echo "Using AWS CLI version: $(aws --version)"
-
 
 cd "$(dirname "$BASH_SOURCE")/.."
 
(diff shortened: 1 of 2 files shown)


packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -7,54 +7,26 @@
...
 
-Aborting.
-EOF
-  exit 1
-}
+check_installation npm
+check_installation aws
+check_installation docker
+check_installation jq
+check_installation zip
```
kirrg001, line 24: <3

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -315,54 +283,53 @@ popd > /dev/null
...
-      echo " - publishing to region $region:"
-
-      is_chinese_region=0
-      while read -r chinese_region; do
-        if [[ "$region" == "$chinese_region" ]]; then
-          is_chinese_region=1
-          break
-        fi
-      done <<< "$CHINESE_REGIONS"
+    # Publish to AWS, handle Chinese regions
```
kirrg001, line 315:
```suggestion
```

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -315,54 +283,53 @@ popd > /dev/null
...
-          printf "Error: Trying to publish to Chinese region $region, but at least one of the environment variables\n"
-          printf "AWS_ACCESS_KEY_ID_CHINA or AWS_SECRET_ACCESS_KEY_CHINA is not set.\n"
-          exit 1
-        fi
-        # Publishing to a Chinese regions requires different credentials, because it is a different AWS account. The
-        # AWS credential environment variables will be reverted after the publish for this region is done.
-        AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID_CHINA
-        AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY_CHINA
-        aws_cli_timeout_options="--cli-read-timeout $AWS_CLI_TIMEOUT_FOR_CHINA --cli-connect-timeout $AWS_CLI_TIMEOUT_FOR_CHINA"
+    if [[ "$PUBLISH_TO_CHINA_REGIONS" == "true" && "$CHINESE_REGIONS" == *"$region"* ]]; then
```
kirrg001, line 320: qs: What does this check ""$CHINESE_REGIONS" == *"$region"*"? I assume its checking if chinese_regions is empty? If its empty and PUBLISH_TO_CHINA_REGIONS is true, we should quit the script or?

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -315,54 +283,53 @@ popd > /dev/null
...
-fi
-
-if [[ -z "$BUILD_arm64_FAIL" ]]; then
-    export BUILD_arm64_FAIL=0
-fi
-
 if [[ -z $SKIP_AWS_PUBLISH_LAYER ]]; then
   echo "step 6/9: publishing $ZIP_NAME as AWS Lambda layer $LAYER_NAME to specifed regions"
+  
+  if [[ "$PUBLISH_TO_CHINA_REGIONS" == "true" ]]; then
```
kirrg001, line 289: suggestion: You could also remove the list of CHINESE_REGIONS on top. And make it dynamic. This will work out of the box if new cn regions are added.

```bash
REGIONS=$(aws ssm get-parameters-by-path --path /aws/service/global-infrastructure/services/lambda/regions --output text --query "Parameters[].Value" | tr '\t' '\n' | sort)

REGIONS_CN=()
REGIONS_OTHERS=()

while IFS= read -r region; do
    if [[ $region == cn-* ]]; then
        REGIONS_CN+=("$region")
    else
        REGIONS_OTHERS+=("$region")
    fi
done <<< "$REGIONS"
```

packages/aws-lambda/layer/bin/publish-layer.sh
aryamohanan: It does not check if `CHINESE_REGIONS` is empty. The condition `"$CHINESE_REGIONS" == *"$region"*" `checks if the value of region matches one of the values listed in CHINESE_REGIONS. 
I have refactored the code.

packages/aws-lambda/layer/bin/publish-layer.sh
aryamohanan: Thank you for the suggestion ❤️, I’ve implemented it!

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -315,109 +294,75 @@ popd > /dev/null
...
-      else
-        echo "   + WARNING: Lambda layer version $lambda_layer_version does not seem to be numeric, will not set permissions in region $region"
-      fi
+    if [[ "$PUBLISH_TO_CHINA_REGIONS" == "true" ]]; then
 
-      if [[ $is_chinese_region -eq 1 ]]; then
-        # Revert access key swap for publishing to a Chinese AWS region:
-        AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID_BACKUP
-        AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY_BACKUP
+      if [[ -z $AWS_ACCESS_KEY_ID_CHINA ]] || [[ -z $AWS_SECRET_ACCESS_KEY_CHINA ]]; then
```
aryamohanan, line 313: There is no changes from line 314 to 365, I have corrected the intention.

kirrg001 (approved):
LGTM. Thanks for this initiative

PR #1478, opened 2024-12-06
build: bumped ibm_db from 3.2.5 to 3.3.0

kirrg001:


packages/collector/test/tracing/database/db2/app.js
```diff
@@ -34,20 +36,32 @@ const connStr2 =
  *
  * We are currently using the IBM DB2 cloud service, because we had trouble on Circleci.
```
aryamohanan, line 37:
```suggestion
```
we can remove this statement as this is not true.

packages/collector/test/tracing/database/db2/app.js
```diff
@@ -34,20 +36,32 @@ const connStr2 =
...
  *
- * Example connection string for ENV:
- * DATABASE=bludb;HOSTNAME=*.databases.appdomain.cloud;UID=msv01866;PWD=xxx;PORT=31198;PROTOCOL=TCPIP;SECURITY=SSL
+ * The db2 instance is hosted on IBM Cloud. To test the cloud instance locally, you need to enable
+ * public endpoint temporarly:
+ *
+ * - Go to https://cloud.ibm.com/resources. Click on your instance.
```
aryamohanan, line 42: It seems I don't have sufficient permissions to check the "Connections" section on the UI. Is there an alternative way we can test this? Also, can we store this in 1Password? I came across `Team Node.js: DB2 IBM Cloud Instance`, but it doesn't appear to be working in local.

aryamohanan:
I am now able to install the `ibm_db` on my machine (arm64), which was not possible before. The release [v3.3.0](https://github.com/ibmdb/node-ibm_db/releases/tag/v3.3.0) added support for installing `ibm_db` on arm64, as noted in the [installation commit](https://github.com/ibmdb/node-ibm_db/commit/dc14be9861d50cd48f0bab066fa877105caf806d). I will update the card [INSTA-796](https://jsw.ibm.com/browse/INSTA-796).

packages/collector/test/tracing/database/db2/app.js
kirrg001: We cannot put the public creds into 1pwd, because the public endpoint is removed. These instructions will help the team in case we have to test it locally against the remote db2 instance.
Locally, we test against docker by default.

> It seems I don't have sufficient permissions

Please ask Chris 👍

aryamohanan: Verified the tests 👍

aryamohanan (approved):
LGTM

PR #1481, opened 2024-12-09
chore: updated mocharc for mocha watch

kirrg001:
When you saved a src file in another pkg, mocha watch did not re-run. Improved the mocha watch files option.
I guess we could add one mocharc.js file for all packages, but can be changed later.

### DIFF
--- a/packages/autoprofile/test/.mocharc.js
+++ b/packages/autoprofile/test/.mocharc.js
@@ -1,7 +1,8 @@
 'use strict';
 
 const mochaOptions = {
-  ignore: ['node_modules/**/*', 'test/**/node_modules/**/*']
+  ignore: ['node_modules/**/*', 'test/**/node_modules/**/*'],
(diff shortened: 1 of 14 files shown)


packages/autoprofile/test/.mocharc.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
 
 const mochaOptions = {
-  ignore: ['node_modules/**/*', 'test/**/node_modules/**/*']
+  ignore: ['node_modules/**/*', 'test/**/node_modules/**/*'],
+  watchFiles: ['test/**/*', '../*/src/**/*']
```
kirrg001, line 5: Watch all packages src folders.

kirrg001:
> qs: Are you intentionally not making the changes in the aws-lambda-auto-wrap/test/.mocharc.js file?

I missed this one. Thanks <3

PR #1448, opened 2024-11-13
feat(redis): added endpoint filtering

aryamohanan:
Introduced a configuration to exclude specific Redis commands from tracing. 

This can be configured via the agent arguments as follows:
```
com.instana.tracing:
    ignore-endpoints: 
       redis: type|get
```
Tracer configuration:

```
tracing: {
  ignoreEndpoints: {
    redis: ['type', 'get']
  }
}
```

With this setup, the tracer will ignore Redis commands such as Type or Get, and any other Redis command can be added to the list for exclusion. This enhancement improves performance by reducing unnecessary tracing for specific Redis operations.

TODO
- [x] Generic implementation Design
- [x]  Implementation
- [ ] Public documentation
- [ ] Check the technical documentation update

ref INSTA-16323

PR #1484, opened 2024-12-12
feat(dynamodb): added endpoint filtering

aryamohanan:
Enabled configuration to exclude specific `dynamodb` commands from tracing. 

This can be configured via following ways:

Add the configuration to the agent's `configuration.yaml` file:
```
com.instana.tracing:
    ignore-endpoints: 
       dynamodb: 
         - query
         - listTables
```
Tracer configuration:
```
tracing: {
  ignoreEndpoints: {
    dynamodb: ['query']
  }
}
```
Using an Environment Variable: 

  `INSTANA_IGNORE_ENDPOINTS=dynamodb:query,listTables;`
  
With this setup, the tracer will ignore dynamodb commands such as `QUERY` or `LISTTABLES`, and any other dynamodb command can be added to the list for exclusion. This enhancement improves performance by reducing unnecessary tracing for specific dynamodb operations.

TODO
- [x] Code and Test Implementation
- [x] Public documentation https://github.ibm.com/instana/docs/pull/12356

ref INSTA-16323
ref https://github.com/instana/nodejs/pull/1448


packages/core/src/util/spanFilter.js
```diff
@@ -17,15 +17,18 @@ function shouldIgnore(span, endpoints) {
...
-    const endpoint = endpoints[span.n];
-    if (Array.isArray(endpoint)) {
-      return endpoint.some(op => op === operation);
-    } else if (typeof endpoint === 'string') {
-      return endpoint === operation;
-    }
+  const endpoint = endpoints[span.n];
+  if (!endpoint) return false;
+
+  const operation = span.data?.[span.n]?.operation?.toLowerCase();
```
kirrg001, line 23: suggestion: How about we add some comments?

This case looks like an internal usage bug, right?

packages/core/src/util/spanFilter.js
```diff
@@ -17,15 +17,18 @@ function shouldIgnore(span, endpoints) {
...
+  if (!operation) return false;
+
+  if (Array.isArray(endpoint)) {
```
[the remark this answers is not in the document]
kirrg001: Ah ignore. We have both supports. As suggested, could u please make the cases clear? Thanks :)

packages/core/src/tracing/backend_mappers/mapper.js
```diff
@@ -0,0 +1,59 @@
...
+ * Transforms span data fields to match the backend format.
+ *
+ * @param {import('../../core').InstanaBaseSpan} span
+ * @returns {import('../../core').InstanaBaseSpan} The transformed span.
+ */
+module.exports.transform = span => {
+  const spanName = span.n;
+  if (!fieldMappings[spanName]) return span;
+
+  const mappings = fieldMappings[spanName];
```
kirrg001, line 43: Suggestion:

Access obj once.

```js
const mappings = fieldMappings[spanName];
if (!mappings) return span;
```

packages/core/src/tracing/backend_mappers/mapper.js
```diff
@@ -0,0 +1,59 @@
...
+  if (span.data[spanName]) {
+    Object.entries(mappings).forEach(([internalField, backendField]) => {
+      if (span.data[spanName][internalField]) {
```
kirrg001, line 46: issue: IMO its more efficient to iterate over the span.data keys. By that we iterate once in best case. Most of our spans only have one key.

aryamohanan: a good point! 💵

packages/core/src/tracing/backend_mappers/mapper.js
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param {import('../../core').InstanaBaseSpan} span
+ * @returns {import('../../core').InstanaBaseSpan} The transformed span.
+ */
+module.exports.transform = span => {
+  const spanName = span.n;
+  const mappings = fieldMappings[spanName];
+  if (!mappings || !span.data[spanName]) return span;
+
+  Object.keys(span.data[spanName]).forEach(internalField => {
```
kirrg001, line 44: In general: It could be that a span contains multiple data keys. But I have never seen a case where two instrumentation keys were present. I think only with the SDK 🤔 
But I would ignore that for now I guess.

kirrg001 (approved):
LGTM

PR #1485, opened 2024-12-12
test: fix couchbase test for esm

abhilash-sivan:
running ESM app test for latest version only due to limitations of requiremock


aryamohanan (approved):
LGTM

PR #1509, opened 2025-01-13
fix: resolved logging objects being undefined or missing

abhilash-sivan:
- refs [INSTA-12533](https://jsw.ibm.com/browse/INSTA-12533)
- some log messages did not log the corresponding objects

note:
In some places, I have used `%s`, and `%o` notations.
In some places where JSON.stringify is already used, I followed the same to keep the logs clean and more readable.

### DIFF
--- a/packages/aws-fargate/src/activate.js
+++ b/packages/aws-fargate/src/activate.js
@@ -70,8 +70,9 @@ function init() {
       // eslint-disable-next-line no-unused-expressions
       process.send && process.send('instana.aws-fargate.initialized');
     } catch (e) {
-      logger.error(`Initializing @instana/aws-fargate failed. This fargate task will not be monitored. 
-        ${JSON.stingify(e)}`);
+      logger.error(
+        `Initializing @instana/aws-fargate failed. This fargate task will not be monitored. ${e.message} ${e.stack}`
+      );
     }
   });
 }
--- a/packages/aws-fargate/src/activate.js
+++ b/packages/aws-fargate/src/activate.js
@@ -30,7 +30,9 @@ function init() {
 
   metrics.init(config, function onReady(err, ecsContainerPayload) {
     if (err) {
-      logger.error('Initializing @instana/aws-fargate failed. This fargate task will not be monitored.', err);
+      logger.error(
+        `Initializing @instana/aws-fargate failed. This fargate task will not be monitored. ${err.message} ${err.stack}`
+      );
       metrics.deactivate();
       return;
     }
@@ -68,7 +70,8 @@ function init() {
       // eslint-disable-next-line no-unused-expressions
       process.send && process.send('instana.aws-fargate.initialized');
     } catch (e) {
-      logger.error('Initializing @instana/aws-fargate failed. This fargate task will not be monitored.', e);
+      logger.error(`Initializing @instana/aws-fargate failed. This fargate task will not be monitored. 
+        ${JSON.stingify(e)}`);
     }
   });
 }
--- a/packages/aws-lambda/src/metrics/npmPackageDescription.js
+++ b/packages/aws-lambda/src/metrics/npmPackageDescription.js
@@ -16,7 +16,7 @@ exports.currentPayload = undefined;
 exports.activate = function activate() {
   coreUtil.applicationUnderMonitoring.getMainPackageJsonStartingAtDirectory(rootDir.root, (err, pckg) => {
     if (err) {
-      logger.warn('Failed to determine main package json. Reason: ', err.message, err.stack);
+      logger.warn(`Failed to determine main package json. Reason: ${err.message} ${err.stack}`);
     }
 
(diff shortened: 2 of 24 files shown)


kirrg001:
> fix: log object structure made compatible with major loggers #1509

Suggestion

> fix: resolved logging objects being undefined or missing
>
> refs xxx
> - some log messages did not log the corresponding objects

packages/aws-fargate/src/activate.js
```diff
@@ -30,7 +30,8 @@ function init() {
 
   metrics.init(config, function onReady(err, ecsContainerPayload) {
     if (err) {
-      logger.error('Initializing @instana/aws-fargate failed. This fargate task will not be monitored.', err);
+      logger.error(`Initializing @instana/aws-fargate failed. This fargate task will not be monitored.
```
kirrg001, line 33: issue: Does not work. Please use

> logger.error(`Will do fail ${new Error('bhbu')}`)

No need to stringify an Error object.

packages/aws-lambda/src/metrics/npmPackageDescription.js
```diff
@@ -16,7 +16,7 @@ exports.currentPayload = undefined;
 exports.activate = function activate() {
   coreUtil.applicationUnderMonitoring.getMainPackageJsonStartingAtDirectory(rootDir.root, (err, pckg) => {
     if (err) {
-      logger.warn('Failed to determine main package json. Reason: ', err.message, err.stack);
+      logger.warn('Failed to determine main package json. Reason: %s, %s', err.message, err.stack);
```
kirrg001, line 19: issue: Why are we using multiple formats?

> logger.warn('Failed to determine main package json. Reason: ${err.stack});

We should probably use `err.stack || err.message` or ${err.message} ${err.stack} everywhere.

packages/collector/src/announceCycle/unannounced.js
```diff
@@ -93,11 +93,9 @@ function tryToAnnounce(ctx, retryDelay = initialRetryDelay) {
...
     } catch (e) {
       logger.error(
-        "Failed to parse the JSON payload from the Instana host agent's response. Establishing the " +
-          'connection to the Instana host agent will be retried in %s ms. The response payload was %s.',
-        retryDelay,
-        rawResponse,
-        e
+        "Failed to parse the JSON payload from the Instana host agent's response. Establishing the connection" +
+          `to the Instana host agent will be retried in ${retryDelay} ms. The response payload was ${rawResponse}.` +
+          `${e?.message} ${e?.stack}`
```
kirrg001, line 98: We use `e?` here but nowhere else. Any reason?

packages/core/src/tracing/cls.js
```diff
@@ -310,11 +310,11 @@ function putPseudoSpan(spanName, kind, traceId, spanId) {
   span.k = kind;
 
   if (!traceId) {
-    logger.warn('Cannot start a pseudo span without a trace ID', spanName, kind);
+    logger.warn(`Cannot start a pseudo span without a trace ID": ${spanName}, ${kind}`);
```
kirrg001, line 313:
```suggestion
    logger.warn(`Cannot start a pseudo span without a trace ID: ${spanName}, ${kind}`);
```

packages/core/src/tracing/cls.js
```diff
@@ -310,11 +310,11 @@ function putPseudoSpan(spanName, kind, traceId, spanId) {
...
     return;
   }
   if (!spanId) {
-    logger.warn('Cannot start a pseudo span without a span ID', spanName, kind);
+    logger.warn(`Cannot start a pseudo span without a span ID": ${spanName}, ${kind}`);
```
kirrg001, line 317:
```suggestion
    logger.warn(`Cannot start a pseudo span without a span ID: ${spanName}, ${kind}`);
```

packages/google-cloud-run/src/activate.js
```diff
@@ -37,10 +37,8 @@ function init() {
 
   metrics.init(config, function onReady(err, serviceRevisionPayload) {
     if (err) {
-      logger.error(
-        'Initializing @instana/google-cloud-run failed. This container instance will not be monitored.',
-        err
-      );
+      logger.error('Initializing @instana/google-cloud-run failed. This container instance will not be monitored.');
+      logger.error(`Error: ${err.message} ${err.stack}`);
```
kirrg001, line 41: Can we keep please one log instead of two? There was another place.

packages/collector/src/announceCycle/unannounced.js
abhilash-sivan: We are already using object name `err` in the same code block, that's why here only `e` is used

kirrg001 (approved):
LGTM

We should probably use `error?.stack` & `error?.message` everywhere as an extra protection. If we name all errors `error` or `err`, we can more easily refactor things.

PR #1514, opened 2025-01-15
docs: updated branch name policy

kirrg001:
I will manually cleanup the branches for now. @aryamohanan pointed out that we already have over 80 branches.

When we start using prefixes, we can easily cleanup the branches.


aryamohanan (approved):
Thank you ❤️

abhilash-sivan (approved):
LGTM

PR #1520, opened 2025-01-17
fix: restrict root exit span in serverless

abhilash-sivan:
Restrict usage of root exit spans in serverless services like azure container, was lambda etc

refs [INSTA-22702](https://jsw.ibm.com/browse/INSTA-22702)


abhilash-sivan:
Closing as the feature is working in the serverless environments with only some minor changes are required which we will do as part of [INSTA-25790](https://jsw.ibm.com/browse/INSTA-25790)

PR #1525, opened 2025-01-20
chore: added ci- and build- prefixed branches for auto deletion after 60 days

aryamohanan:
We missed adding ci- and build- prefixed branches to the auto deletion process. 
As we may have branches starting with these prefixes, this update ensures that 
they will be deleted if they are older than 60 days.


abhilash-sivan (approved):
LGTM

kirrg001 (approved):

PR #1346, opened 2024-09-27
fix: connected deferred exit parent span with original entry span

kirrg001:
Found while working on #1297 

Problem: the deferred http request was not traced correctly. It was not connected to the entry http server span.

- for httpClient: was already traced, but not correctly connected to the entry parent id
- for nativeFetch: was not working at all

PR #1531, opened 2025-01-27
fix: resolved tracing for deferred exit spans in http2

aryamohanan:
The HTTP2 client request was traced but didn't connect to the entry parent ID. 
A test has been added to verify this behavior.

related https://github.com/instana/nodejs/pull/1346
ref [INSTA-15647](https://jsw.ibm.com/browse/INSTA-15647)

### DIFF
--- a/packages/collector/test/tracing/protocols/http2/test.js
+++ b/packages/collector/test/tracing/protocols/http2/test.js
@@ -17,7 +17,7 @@ const { AgentStubControls } = require('../../../apps/agentStubControls');
 
 const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : describe.skip;
 
-mochaSuiteFn.only('tracing/http2', function () {
+mochaSuiteFn('tracing/http2', function () {
   this.timeout(config.getTestTimeout() * 2);
 
   const agentControls = new AgentStubControls();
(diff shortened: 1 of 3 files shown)


packages/collector/test/tracing/protocols/http2/test.js
```diff
@@ -315,45 +371,56 @@ function constructPath(basePath, withQuery) {
...
-  );
+  });
   checkQuery(entryInServer, withQuery, erroneous);
   expect(spans).to.have.lengthOf.greaterThanOrEqual(3);
 }
 
-function verifyRootHttpEntry(spans, host, url, method = 'GET', status = 200, erroneous = false, synthetic = false) {
-  return verifyHttpEntry(spans, null, host, url, method, false, status, erroneous, synthetic);
+function verifyRootHttpEntry({ spans, host, url, method = 'GET', status = 200, erroneous = false, synthetic = false }) {
+  return verifyHttpEntry({
```
aryamohanan, line 410: I refactored the function call to use object destructuring for the parameters instead of positional arguments. Passing such a large number of parameters in this way is not ideal, as it can become difficult to manage.

kirrg001 (approved):
LGTM

We need to remove the `.only`

Commit suggestion:

fix: resolved tracing for deferred exit spans in http2

abhilash-sivan (approved):

Issue #1451, opened 2024-11-18
[Bug]: Unable to add custom spans with layer arm64 > 96 and x86 > 213


janritter:
### Problem Description

When using the newest versions of the Instana Layer in Nodejs with AutoTrace (wrapping the handler), the auto instrumentation part works fine, but trying to access the current span or adding intermediate spans fails (they are null and intermediate spans are not send to Instana).

After creating a Lambda Function with autoTrace and a function URL using the code blow, when executing the Lambda, we can see the following in the logs:

```text
  | 2024-11-18T15:35:33.933Z | INIT_START Runtime Version: nodejs:20.v43 Runtime Version ARN: arn:aws:lambda:eu-central-1::runtime:4a87b8fceb008240eb9bfe7dc3126c2dec6fe75f06c6a4d8977d0267eb5da2bd
  | 2024-11-18T15:35:34.169Z | 2024-11-18T15:35:34.169Z undefined INFO @instana/aws-lambda module version: 3.21.0
  | 2024-11-18T15:35:34.693Z | 2024-11-18T15:35:34.693Z undefined INFO @instana/aws-lambda module version: 4.0.1
  | 2024-11-18T15:35:34.843Z | 2024-11-18T15:35:34.843Z undefined INFO Method warn of [object console] was already wrapped. Unwrapping.
  | 2024-11-18T15:35:34.844Z | 2024-11-18T15:35:34.844Z undefined INFO Method error of [object console] was already wrapped. Unwrapping.
  | 2024-11-18T15:35:34.869Z | EXTENSION Name: instana-lambda-extension State: Ready Events: [INVOKE, SHUTDOWN]
```
  
It looks like the original handler is wrapped in version 3.21.0, while 4.0.1 is used when requiring the package in the code.

### Short, Self Contained Example

Lambda code: 
```javascript
const instana = require('@instana/aws-lambda');

exports.handler = async (event, context) => {
    const span = instana.currentSpan();
    const response = {
        statusCode: 200,
        body: JSON.stringify({
            currentSpanId: span ? span.getSpanId() : null,
        }),
    };
    return response;
};
```

### Node.js Version

Lambda nodejs20.x

### package.json

```shell
-
```

### package-lock.json

```shell
-
```

kirrg001:
Hi @janritter Thanks for your report! We will take a look asap!

aryamohanan:
Hi @janritter 

[...]

PR #1532, opened 2025-01-27
docs: removed outdated contents from readme

aryamohanan:


packages/aws-lambda-auto-wrap/esm/README.md
```diff
@@ -10,6 +10,4 @@ AWS Lambda began supporting ES modules with Node v14. Learn more on the [AWS Lam
   - Transitioning the handler to pure ES modules does not provide immediate benefits, but we are monitoring AWS updates closely.
 
 - **Roadmap for Improvements**:
-  - We've initiated an AWS support ticket (172069954500883) and engaged with the AWS SDK community to advocate for ES module support in Lambda layers. For more details, see the [GitHub discussion](https://github.com/aws/aws-sdk-js/discussions/4651).
-  - In 4.x we will remove the `aws-lambda-auto-wrap` npm package, because this is an internal package. It got published by mistake. We copy the necessary files during the layer publishing script into the layer.
```
kirrg001, line 14: We remember we re-added the pkg and we we have a JIRA for it or?

aryamohanan: In the AWS Lambda layer publish script, we are installing the `aws-lambda-auto-wrap` npm package when building with npm. This was previously overlooked, which is why we made this statement.  After removing the `aws-lambda-auto-wrap` npm package, which caused bug #1451, so we reverted the changes. 
We may consider removing this in the next major v5 update. I will add a card for this.

kirrg001 (approved):

abhilash-sivan (approved):

packages/aws-lambda-auto-wrap/esm/README.md
aryamohanan: created card INSTA-24914

PR #1543, opened 2025-02-04
test: updated deprecated url.parse with new WHATWG URL API

abhilash-sivan:
refs INSTA-18579

This PR replaces the deprecated `url.parse` with `new URL` API

### DIFF
--- a/packages/aws-lambda/test/lambdas/callback.js
+++ b/packages/aws-lambda/test/lambdas/callback.js
@@ -8,7 +8,6 @@
 'use strict';
 
 const instana = require('../..');
-const url = require('url');
 
 // In production, the package @instana/aws-lambda is located in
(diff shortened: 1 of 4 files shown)


kirrg001:
There are more occurrences in the codebase. Is the plan to do another PR for non-test replacements?

abhilash-sivan:
> There are more occurrences in the codebase. Is the plan to do another PR for non-test replacements?

Did you mean all the usage of `url` library?

eg: https://github.com/instana/nodejs/blob/7e7620c5f340b144fb208b2d7467af2b55342829/packages/core/src/tracing/instrumentation/protocols/httpClient.js#L10

kirrg001:
Yeah. I guess these are not affected by this warning. Ok 👍

kirrg001 (approved):
LGTM its only a test env change. I thought this will have an affect on production code.

abhilash-sivan:
Yeah, only url.parse() method is deprecated!

PR #1545, opened 2025-02-05
ci: fixed slack alert issue related to AWS Lambda layer publishing

aryamohanan:
This PR addresses two issues:

- The current implementation causes all remaining tasks to be skipped if one task fails.
- Slack notifications were not working for each tasks. 

ref [INSTA-22682](https://jsw.ibm.com/browse/INSTA-22682)

### DIFF
--- a/packages/serverless/ci/pipeline.yaml
+++ b/packages/serverless/ci/pipeline.yaml
@@ -401,23 +401,23 @@ jobs:
                 AWS_ACCESS_KEY_ID_CHINA=((aws-ci-manage-lambda-layers-china.aws_access_key_id)) \
                 AWS_SECRET_ACCESS_KEY_CHINA=((aws-ci-manage-lambda-layers-china.aws_secret_access_key)) \
                 nodejs-repository/packages/aws-lambda/layer/bin/publish-layer.sh
-          on_success:
-            put: slack-alert
-            params:
-              channel: '#team-node'
-              icon_emoji: ':white_check_mark:'
-              text: |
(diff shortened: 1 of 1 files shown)


packages/serverless/ci/pipeline.yaml
```diff
@@ -378,171 +378,171 @@ jobs:
...
-                NO_PROMPT=true \
-                CONTAINER_REGISTRY_USER=iamapikey \
-                CONTAINER_REGISTRY_PASSWORD=((concourse-icr-containers-public.password)) \
-                AWS_ACCESS_KEY_ID=((aws-ci-manage-lambda-layers.aws_access_key_id)) \
-                AWS_SECRET_ACCESS_KEY=((aws-ci-manage-lambda-layers.aws_secret_access_key)) \
-                AWS_ACCESS_KEY_ID_CHINA=((aws-ci-manage-lambda-layers-china.aws_access_key_id)) \
-                AWS_SECRET_ACCESS_KEY_CHINA=((aws-ci-manage-lambda-layers-china.aws_secret_access_key)) \
-                nodejs-repository/packages/aws-lambda/layer/bin/publish-layer.sh
+      
+      - in_parallel:
```
aryamohanan, line 382: Here, tasks are being run in parallel to ensure each operates independently, allowing the others to continue even if one fails, without causing any blockage. Previously, tasks were running sequentially, meaning if one task failed, the remaining tasks would be skipped.

kirrg001 (approved):

PR #1550, opened 2025-02-10
refactor: added new span class for ignored span

aryamohanan:
The addition of `InstanaIgnoredSpan` serves to address case where a span is intentionally ignored via configuration and should not be transmitted or placed into the span buffer.

### DIFF
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -219,7 +219,9 @@ class InstanaPseudoSpan extends InstanaSpan {
   }
 }
 /**
- * This class represents an ignored span when accessing the `currentSpan`.
+* This class represents an ignored span. We need this representation for the endpoint filtering feature because
+ * when a customer accesses the current span via `instana.currentSpan()`, we do not want to return a "NoopSpan".
+ * Instead, we return this ignored span instance so the trace ID remains accessible.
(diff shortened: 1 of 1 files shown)


abhilash-sivan (approved):

packages/core/src/tracing/cls.js
```diff
@@ -219,6 +219,32 @@ class InstanaPseudoSpan extends InstanaSpan {
   }
 }
 
+// This class represents an ignored span and overrides the `transmit` and `cancel` methods
```
kirrg001, line 222: suggestion: Please describe the purpose of this class.

kirrg001 (approved):
Pre-approving. Added one suggestion.

packages/core/src/tracing/cls.js
```diff
@@ -218,6 +218,33 @@ class InstanaPseudoSpan extends InstanaSpan {
     }
   }
 }
+/**
+ * This class represents an ignored span when accessing the `currentSpan`.
```
kirrg001, line 222:
```suggestion
* This class represents an ignored span. We need this representation for the endpoint filtering feature because
 * when a customer accesses the current span via `instana.currentSpan()`, we do not want to return a "NoopSpan".
 * Instead, we return this ignored span instance so the trace ID remains accessible.
```

PR #1552, opened 2025-02-11
refactor: replaced startSpan positional arguments with optional object pattern

aryamohanan:
Refactored the `startSpan` function to use an optional object pattern instead of multiple positional arguments. This improves readability, allows passing only the necessary values, and eliminates the need to remember the argument order.

ref INSTA-26148


kirrg001:
https://medium.com/@bchadwickfrance/the-options-object-pattern-with-typescript-a14f33306ec8

IMO the official name is "options object pattern" - short "options pattern".

kirrg001 (approved):
<3

abhilash-sivan (approved):

PR #1553, opened 2025-02-11
chore: udpated esm app test execution logic

abhilash-sivan:
refs [INSTA-26666](https://jsw.ibm.com/browse/INSTA-26666)

Currently, we are processing only `app.mjs` file name for ESM test apps as this is hardcoded to default.
Now we can pass the file name dynamically to processControls to execute.

- [x] update the naming strategy
- [x] add example test (at least one instr)

We need this in `bull` as there are no app.js but the logic is separated into receiver.js and sender.js etc


abhilash-sivan:
I have extended the ESM test to Bunyan logger using the newly updated naming strategy.

see the filename: app_esm.mjs being loaded:

Screenshot 2025-02-12 at 4 44 40 PM

packages/collector/test/tracing/logger/bunyan/app_esm.mjs
```diff
@@ -0,0 +1,118 @@
+/*
+ * (c) Copyright IBM Corp. 2021
+ * (c) Copyright Instana Inc. and contributors 2018
```
aryamohanan, line 3: Please take care of the headers ☝️

packages/collector/test/test_util/ProcessControls.js
```diff
@@ -70,6 +70,17 @@ class ProcessControls {
           // eslint-disable-next-line no-console
           console.log('Unable to load the target app.mjs', err);
         }
+      } else if (opts.appPath) {
```
kirrg001, line 73: Suggestion: We do too many checks for `RUN_ESM` in the test with this solution.

ProcessControls.js
```
if (opts.customEsmPath) {
} else if (opts.dirname) {
}
```

test.js

```
controls = new ProcessControls({dirname: __dirname, customEsmPath: 'PATH'});
```

packages/collector/test/tracing/logger/bunyan/test.js
```diff
@@ -78,16 +89,19 @@ mochaSuiteFn('tracing/logger/bunyan', function () {
...
-      runTest(
-        'error-large-object-only',
-        true,
-        // eslint-disable-next-line max-len
-        '{"_id":"638dea148cff492d47e792ea","index":0,"guid":"01b61bfa-fe4c-4d75-9224-389c4c04de10","isActive":false,"balance":"$1,919.18","picture":"http://placehold.it/32x32","age":37,"eyeColor":"blue","name":"Manning Brady","gender":"male","company":"ZYTRAC","email":"manningbrady@zytrac.com","phone":"+1 (957) 538-2183","address":"146 Bushwick Court, Gilgo, New York, 2992","about":"Ullamco cillum reprehenderit eu proident veniam laboris tempor voluptate. Officia deserunt velit incididunt consequat la...',
-        controls,
-        500,
-        4
-      ));
+    // TODO: spans are missing in ESM for this API
```
kirrg001, line 92: Could you please try to explain why?

abhilash-sivan: The following is the log code which is not being traced here
 
 > logger.error(JSON.parse(fs.readFileSync(path.join(__dirname, 'assets', 'large_obj.json'), 'utf8')));

I debugged this and found that the instrumentation code did not run for this case.

Since `bunyan` is deprecated and only 1 complex test case has failed I chose to ignore it for now.

packages/collector/test/test_util/ProcessControls.js
abhilash-sivan: Yeah we have a lot of logics using `dirname` and `appPath`. I will try to optimize it

kirrg001: The goal should be to not use. This is hard to maintain.
The pipeline forwards `RUN_ESM=true`. The ProcessControls automatically goes into ESM mode. There is IMO no need to do an extra check in the test.

```
      let configObj = {};
      if (process.env.RUN_ESM) {
        configObj = {
          appPath: path.join(__dirname, 'app_esm'),
          useGlobalAgent: true
        };
      } else {
        configObj = {
          dirname: __dirname,
          useGlobalAgent: true
        };
      }
```

abhilash-sivan: PS: I only used different names here to prove the dynamic naming changes

abhilash-sivan: > bunyanApp.js & bunyanApp.mjs?

done

abhilash-sivan:
bunyanApp.mjs:

Screenshot 2025-02-12 at 8 41 58 PM

bunyanApp.js:

Screenshot 2025-02-12 at 8 45 21 PM

kirrg001 (approved):
LGTM

aryamohanan (approved):

PR #1558, opened 2025-02-13
refactor: improved logic for ignoring endpoints

aryamohanan:
Refactored the implementation of ignored endpoints by moving the filtering logic to startSpan. This change makes it easier to extend in the future, simplifies modification, and resolves the issue with accessing the current span for ignored span.

Tasks

- [x] - Refactor the existing span filter logic within the `startSpan` function.  
- [x] - Get the ignore endpoints config in cls
- [x] - Perform the span mapping within the span buffer while adding a span.  

Pending 

- [x] - Apply filtering based on  operation as well as endpoints https://github.com/instana/nodejs/pull/1580.  
- [ ] - An integration test for the ignore endpoint (New PR)

ref INSTA-26148

### DIFF
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -272,19 +272,18 @@ class InstanaIgnoredSpan extends InstanaSpan {
  * @param {string} [spanAttributes.traceId]
  * @param {string} [spanAttributes.parentSpanId]
  * @param {import('./w3c_trace_context/W3cTraceContext')} [spanAttributes.w3cTraceContext]
- * @param {Object} [spanAttributes.spanContextData]
+ * @param {Object} [spanAttributes.spanData]
  * @returns {InstanaSpan}
  */
 
 function startSpan(spanAttributes = {}) {
-  let { spanName, kind, traceId, parentSpanId, w3cTraceContext, spanContextData } = spanAttributes;
+  let { spanName, kind, traceId, parentSpanId, w3cTraceContext, spanData } = spanAttributes;
 
   tracingMetrics.incrementOpened();
   if (!kind || (kind !== ENTRY && kind !== EXIT && kind !== INTERMEDIATE)) {
     logger.warn(`Invalid span (${spanName}) without kind/with invalid kind: ${kind}, assuming EXIT.`);
     kind = EXIT;
   }
-  const spanData = spanContextData || {};
 
   const span = new InstanaSpan(spanName, spanData);
   span.k = kind;
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -402,7 +402,7 @@ function putPseudoSpan(spanName, kind, traceId, spanId) {
 /**
  * Adds an ignored span to the CLS context, serving as a holder for a trace ID and span ID.
  * Customers can access the current span via `instana.currentSpan()`, and we avoid returning a "NoopSpan".
- * Instead, we need this ignored span instance to ensure the trace ID remains accessible.
+ * We need this ignored span instance to ensure the trace ID remains accessible for future cases such as propagating the trace ID although suppression is on to reactivate a trace.
  * These spans will not be sent to the backend.
  * @param {Object} options - The options for the span.
  * @param {string} options.spanName
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -402,7 +402,8 @@ function putPseudoSpan(spanName, kind, traceId, spanId) {
 /**
  * Adds an ignored span to the CLS context, serving as a holder for a trace ID and span ID.
  * Customers can access the current span via `instana.currentSpan()`, and we avoid returning a "NoopSpan".
- * We need this ignored span instance to ensure the trace ID remains accessible for future cases such as propagating the trace ID although suppression is on to reactivate a trace.
+ * We need this ignored span instance to ensure the trace ID remains accessible for future cases such as propagating
+ * the trace ID although suppression is on to reactivate a trace.
  * These spans will not be sent to the backend.
  * @param {Object} options - The options for the span.
  * @param {string} options.spanName
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -63,7 +63,7 @@ function init(config, _processIdentityProvider) {
  * @param {import('../util/normalizeConfig').AgentConfig} extraConfig
  */
 function activate(extraConfig) {
-  if (extraConfig.tracing.ignoreEndpoints) {
+  if (extraConfig?.tracing?.ignoreEndpoints) {
     ignoreEndpoints = extraConfig.tracing.ignoreEndpoints;
   }
 }
@@ -705,6 +705,7 @@ function sanitizeSpan(span) {
   if (ignoreEndpoints) {
     return applyFilter({ span, ignoreEndpoints });
   }
+  return span;
 }
 
 module.exports = {
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -339,7 +339,8 @@ function startSpan(spanAttributes = {}) {
       spanName: span.n,
       kind: span.k,
       traceId: span.t,
-      spanId: span.p
+      parentId: span.p,
+      data: span.data
     });
   }
 
@@ -403,18 +404,19 @@ function putPseudoSpan(spanName, kind, traceId, spanId) {
  * Instead, we need this ignored span instance to ensure the trace ID remains accessible.
  * These spans will not be sent to the backend.
  * @param {Object} options - The options for the span.
- * @param {string} options.spanName - The name of the span.
- * @param {number} options.kind - The kind of span (ENTRY, EXIT, or INTERMEDIATE).
- * @param {string} options.traceId - The trace ID associated with the span.
- * @param {string} options.spanId - The span ID.
+ * @param {string} options.spanName
+ * @param {number} options.kind
+ * @param {string} options.traceId
+ * @param {string} options.parentId
+ * @param {Object} options.data
  */
-function setIgnoredSpan({ spanName, kind, traceId, spanId }) {
+function setIgnoredSpan({ spanName, kind, traceId, parentId, data = {} }) {
   if (!kind || (kind !== ENTRY && kind !== EXIT && kind !== INTERMEDIATE)) {
     logger.warn(`Invalid ignored span (${spanName}) without kind/with invalid kind: ${kind}, assuming EXIT.`);
     kind = EXIT;
   }
 
-  const span = new InstanaIgnoredSpan(spanName);
+  const span = new InstanaIgnoredSpan(spanName, data);
   span.k = kind;
 
   if (!traceId) {
@@ -422,13 +424,14 @@ function setIgnoredSpan({ spanName, kind, traceId, spanId }) {
     return;
   }
 
-  if (!spanId) {
-    logger.warn(`Cannot start an ignored span without a span ID: ${spanName}, ${kind}`);
-    return;
-  }
-
   span.t = traceId;
-  span.s = spanId;
+  span.p = parentId;
+
+  // Setting the 'parentId' of the span to 'span.s' to ensure trace continuity.
+  // Although this span doesn't physically exist, we are ignoring it, but retaining its parentId.
+  // This parentId is propagated downstream.
+  // The spanId does not need to be retained.
+  span.s = parentId;
 
   if (span.k === ENTRY) {
     // Make the entry span available independently (even if getCurrentSpan would return an intermediate or an exit at
@@ -440,7 +443,6 @@ function setIgnoredSpan({ spanName, kind, traceId, spanId }) {
   // Set the span object as the currently active span in the active CLS context and also add a cleanup hook for when
   // this span is transmitted.
   span.addCleanup(ns.set(currentSpanKey, span));
-
   return span;
 }
 
--- a/packages/core/src/tracing/cls.js
+++ b/packages/core/src/tracing/cls.js
@@ -332,8 +332,9 @@ function startSpan(spanAttributes = {}) {
     span.addCleanup(ns.set(w3cTraceContextKey, w3cTraceContext));
   }
 
-  const sanitizedspan = sanitizeSpan(span);
-  if (!sanitizedspan) {
+  const filteredSpan = applySpanFilter(span);
+  // If the span is filtered out, we return 'InstanaIgnoredSpan' class to indicate the span is ignored.
+  if (!filteredSpan) {
     return setIgnoredSpan({
       spanName: span.n,
       kind: span.k,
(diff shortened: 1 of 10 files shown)


packages/core/src/tracing/index.js
```diff
@@ -265,6 +264,7 @@ exports.activate = function activate(extraConfig = {}) {
     spanBuffer.activate(extraConfig);
     opentracing.activate();
     sdk.activate();
+    cls.activate(extraConfig);
```
aryamohanan, line 267: We need the agent config available at the cls.

packages/core/src/tracing/cls.js
```diff
@@ -256,18 +272,21 @@ class InstanaIgnoredSpan extends InstanaSpan {
...
 
   tracingMetrics.incrementOpened();
   if (!kind || (kind !== ENTRY && kind !== EXIT && kind !== INTERMEDIATE)) {
     logger.warn(`Invalid span (${spanName}) without kind/with invalid kind: ${kind}, assuming EXIT.`);
     kind = EXIT;
   }
-  const span = new InstanaSpan(spanName);
+  const spanData = spanContextData || {};
+
+  const span = new InstanaSpan(spanName, spanData);
```
kirrg001, line 289: suggestion: Forward spanContextData straight into the class. Rename spanContextData to spanData in function argument

packages/core/src/tracing/cls.js
```diff
@@ -367,6 +398,54 @@ function putPseudoSpan(spanName, kind, traceId, spanId) {
...
+ * @param {string} options.parentId
+ * @param {Object} options.data
+ */
+function setIgnoredSpan({ spanName, kind, traceId, parentId, data = {} }) {
+  if (!kind || (kind !== ENTRY && kind !== EXIT && kind !== INTERMEDIATE)) {
+    logger.warn(`Invalid ignored span (${spanName}) without kind/with invalid kind: ${kind}, assuming EXIT.`);
+    kind = EXIT;
+  }
+
+  const span = new InstanaIgnoredSpan(spanName, data);
```
kirrg001, line 419: qs: Did u think about InstanaFilteredSpan?

packages/core/src/tracing/cls.js
```diff
@@ -367,6 +398,54 @@ function putPseudoSpan(spanName, kind, traceId, spanId) {
...
+  span.k = kind;
+
+  if (!traceId) {
+    logger.warn(`Cannot start an ignored span without a trace ID: ${spanName}, ${kind}`);
+    return;
```
kirrg001, line 424: qs: I dont think we can return null for any case?

packages/core/src/tracing/index.js
kirrg001: Inwas wondering if we should activate cls as early as possble because its our core module

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/dynamodb.js
```diff
@@ -19,16 +19,18 @@ class InstanaAWSDynamoDB extends InstanaAWSProduct {
     }
 
     const command = smithySendArgs[0];
-
+    const spanData = {
```
kirrg001, line 22: qs: Are u sure we can build the data before the cls call?

packages/core/src/tracing/spanBuffer.js
```diff
@@ -191,8 +182,8 @@ exports.addSpan = function (span) {
     return;
   }
 
-  // Process the span, apply any transformations, and implement filtering if necessary.
-  const processedSpan = processSpan(span);
+  // Apply necessary transformations to the span (e.g., renaming properties, modifying data).
+  const processedSpan = applySpanTransformation(span);
   if (!processedSpan) {
```
kirrg001, line 187: qs: processSpan should never be null here or

kirrg001 (commented):
Some first questions 👏🏻👏🏻👏🏻

packages/core/src/tracing/cls.js
aryamohanan: Yes, I considered InstanaFilteredSpan, but InstanaIgnoredSpan seems clearer since the span is fully excluded from tracing. Filtered might suggest some level of processing and doesn’t necessarily imply exclusion, whereas Ignored clearly conveys that it’s completely skipped. We can refine it further if needed.

packages/core/src/tracing/cls.js
aryamohanan: The trace ID is always present since we generate it if it's missing in the upper layer [reference](https://github.com/instana/nodejs/blob/main/packages/core/src/tracing/cls.js#L298). So, I believe the condition may be unnecessary. I will remove this.

packages/core/src/tracing/spanBuffer.js
aryamohanan: We only transform the span data fields, so this check is unnecessary. I'll remove it—it was originally left over from when we had filtering logic. Thanks for pointing it out!

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/dynamodb.js
aryamohanan: In startSpan, we typically create the span structure without data, so IMO preparing the data beforehand is fine. The data is solely dependent on the package specific information. I believe this approach is okay, but please share any use cases if you have any.

packages/core/src/tracing/cls.js
kirrg001: Ok 👍

packages/core/src/tracing/cls.js
```diff
@@ -367,6 +399,48 @@ function putPseudoSpan(spanName, kind, traceId, spanId) {
   return span;
 }
 
+/**
+ * Adds an ignored span to the CLS context, serving as a holder for a trace ID and span ID.
+ * Customers can access the current span via `instana.currentSpan()`, and we avoid returning a "NoopSpan".
+ * Instead, we need this ignored span instance to ensure the trace ID remains accessible.
```
kirrg001, line 405:
```suggestion
 * We need this ignored span instance to ensure the trace ID remains accessible for future cases such as propagating the trace ID although suppression is on to reactivate a trace.
```

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v2/dynamodb.js
```diff
@@ -31,17 +31,20 @@ class InstanaAWSDynamoDB extends InstanaAWSProduct {
     if (cls.skipExitTracing({ isActive })) {
       return originalMakeRequest.apply(ctx, originalArgs);
     }
+    // Data attribs: op and table
```
kirrg001, line 34:
```suggestion
    // Data attributes: operation, table
```

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v2/dynamodb.js
```diff
@@ -31,17 +31,20 @@ class InstanaAWSDynamoDB extends InstanaAWSProduct {
...
+      [this.spanName]: this.buildSpanData(ctx, originalArgs[0], originalArgs[1])
+    };
 
     return cls.ns.runAndReturn(() => {
       const self = this;
       const span = cls.startSpan({
         spanName: this.spanName,
-        kind: EXIT
+        kind: EXIT,
+        spanData: spanData
```
kirrg001, line 44:
```suggestion
        spanData
```

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/dynamodb.js
kirrg001: IMO it breaks the pattern for all instrumentations. We usually collect the data after 

> return cls.ns.runAndReturn(() => {

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/dynamodb.js
```diff
@@ -19,16 +19,18 @@ class InstanaAWSDynamoDB extends InstanaAWSProduct {
...
+      [this.spanName]: this.buildSpanData(command.constructor.name, command.input)
+    };
     return cls.ns.runAndReturn(() => {
       const self = this;
       const span = cls.startSpan({
         spanName: this.spanName,
-        kind: EXIT
+        kind: EXIT,
+        spanData: spanData
```
kirrg001, line 30:
```suggestion
        spanData
```

packages/core/src/tracing/instrumentation/database/ioredis.js
```diff
@@ -156,15 +158,18 @@ function instrumentMultiOrPipelineCommand(commandName, original) {
...
+      redis: {
+        connection,
+        operation: commandName
+      }
+    };
     const span = cls.startSpan({
       spanName: exports.spanName,
-      kind: constants.EXIT
+      kind: constants.EXIT,
+      spanData: spanData
```
kirrg001, line 170:
```suggestion
      spanData
```

packages/core/src/tracing/instrumentation/database/redis.js
```diff
@@ -234,19 +234,20 @@ function instrumentCommand(original, command, address, cbStyle) {
...
+        connection: address || origCtx.address,
+        operation: command
+      }
+    };
     return cls.ns.runAndReturn(() => {
       const span = cls.startSpan({
         spanName: exports.spanName,
-        kind: constants.EXIT
+        kind: constants.EXIT,
+        spanData: spanData
```
kirrg001, line 247:
```suggestion
        spanData
```

packages/core/src/tracing/instrumentation/database/redis.js
```diff
@@ -315,27 +316,30 @@ function instrumentMultiExec(origCtx, origArgs, original, address, isAtomic, cbS
...
+        // pipeline = batch
+        operation: isAtomic ? 'multi' : 'pipeline'
+      }
+    };
     if (skipExitResult.allowRootExitSpan) {
       span = cls.startSpan({
         spanName: exports.spanName,
-        kind: constants.EXIT
+        kind: constants.EXIT,
+        spanData: spanData
```
kirrg001, line 330:
```suggestion
        spanData
```

packages/core/src/tracing/instrumentation/database/redis.js
```diff
@@ -315,27 +316,30 @@ function instrumentMultiExec(origCtx, origArgs, original, address, isAtomic, cbS
...
       });
     } else {
       span = cls.startSpan({
         spanName: exports.spanName,
         kind: constants.EXIT,
         traceId: parentSpan.t,
-        parentSpanId: parentSpan.s
+        parentSpanId: parentSpan.s,
+        spanData: spanData
```
kirrg001, line 338:
```suggestion
        spanData
```

packages/core/src/tracing/spanBuffer.js
```diff
@@ -191,12 +182,8 @@ exports.addSpan = function (span) {
     return;
   }
 
-  // Process the span, apply any transformations, and implement filtering if necessary.
-  const processedSpan = processSpan(span);
-  if (!processedSpan) {
-    return;
-  }
-  span = processedSpan;
+  // Apply necessary transformations to the span (e.g., renaming properties, modifying data).
```
kirrg001, line 185:
```suggestion
  // Transform internal span data format into external (backend) readable format.
```

packages/core/test/tracing/cls_test.js
```diff
@@ -369,4 +369,58 @@ describe('tracing/cls', () => {
...
+          }
+        }
+      });
+      expect(span.constructor.name).to.equal('InstanaIgnoredSpan');
+      expect(span).to.be.an('object');
+      expect(span.n).to.equal('redis');
+      expect(span.t).to.be.a('string');
+      expect(span.k).to.equal(constants.EXIT);
+      expect(span.data).to.be.an('object');
+      expect(Object.keys(span.data)).to.have.lengthOf(1);
```
kirrg001, line 423: Same here.

packages/core/test/tracing/spanBuffer_test.js
```diff
@@ -566,12 +566,12 @@ describe('tracing/spanBuffer', () => {
...
       });
     });
-    describe('when ignoreEndpoints is configured', () => {
+    describe('when applying span transformations', () => {
       before(() => {
         spanBuffer.init(
           {
             tracing: {
-              ignoreEndpoints: { redis: ['get'] }
+              maxBufferedSpans: 1000
```
kirrg001, line 574: ?

packages/core/src/tracing/cls.js
```diff
@@ -313,6 +331,20 @@ function startSpan(spanAttributes = {}) {
     span.addCleanup(ns.set(w3cTraceContextKey, w3cTraceContext));
   }
 
+  const filteredSpan = applySpanFilter(span);
+
+  // If the span was filtered out, we do not process it further.
+  // Instead, we return an 'InstanaIgnoredSpan' instance to explicitly indicate that it was excluded from tracing.
+  if (!filteredSpan) {
+    return setIgnoredSpan({
```
kirrg001, line 339: Putting it here: I am missing a test for the SDK. 

1. Ignore endpoints configured
2. normal app.js
3. SDK requests currentSpan

aryamohanan: I have planned this in a separate PR, as stated in the description.

kirrg001: Hmm but we cannot merge this into main without this test I think. This PR introduces a new class instance, which you will receive a customer when using the SDK + ignore endpoints.

packages/core/test/tracing/cls_test.js
```diff
@@ -369,4 +369,58 @@ describe('tracing/cls', () => {
...
+      expect(span.t).to.be.a('string');
+      expect(span.s).to.be.a('string');
+      expect(span.k).to.equal(constants.EXIT);
+      expect(span.error).to.not.exist;
+      expect(span.ec).to.equal(0);
+      expect(span.ts).to.be.a('number');
+      expect(span.d).to.equal(0);
+      expect(span.stack).to.deep.equal([]);
+      expect(span.data).to.be.an('object');
+      expect(Object.keys(span.data)).to.have.lengthOf(1);
```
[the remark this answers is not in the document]
aryamohanan: I don't get this.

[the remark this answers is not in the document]
kirrg001: I thought "serviceName" has a default value. 

Then

>       expect(Object.keys(span.data)).to.have.lengthOf(1);

Is correct. The one key is "redis".
Maybe it would be good to check this. It can lead to false positives.

packages/core/src/tracing/instrumentation/cloud/aws-sdk/v3/dynamodb.js
aryamohanan: I have incorporated the `spandata` logic within `return cls.ns.runAndReturn(() => {`.

packages/core/src/tracing/cls.js
kirrg001: Discussed: We add tests in the other kafka PR

kirrg001 (approved):
LGTM

PR #1561, opened 2025-02-17
test: enabled ESM test for bull

abhilash-sivan:
refs [INSTA-13393](https://jsw.ibm.com/browse/INSTA-13393)

This PR enables tests for bull in ESM.

There is an existing issue with bull when using separate processes in the ES module syntax
see: https://github.com/OptimalBits/bull/issues/924


packages/collector/test/tracing/messaging/bull/test.js
```diff
@@ -108,211 +108,215 @@ mochaSuiteFn('tracing/messaging/bull', function () {
       await senderControls.clearIpcMessages();
     });
 
-    describe('receiving via "Process" API', () => {
-      let receiverControls;
-      const receiveMethod = 'Process';
+    // NOTE: Bull doesn't officially support Process API when using processor with ES module syntax
+    //       See: https://github.com/OptimalBits/bull/blob/489c6ab8466c1db122f92af3ddef12eacc54179e/lib/queue.js#L712
+    //       Related issue: https://github.com/OptimalBits/bull/issues/924
+    if (!process.env.RUN_ESM) {
```
abhilash-sivan, line 114: Bull doesn't support ESM with process API.
So ignoring the test cases for the same

kirrg001:
Could you please run the tests with ESM?

abhilash-sivan:
Ran against ESM true:
https://cloud.ibm.com/devops/pipelines/tekton/c2cd6a8d-ea5a-47b0-913e-cd172d63833f/runs/ae525924-6058-4462-897d-b42f54151a65/extract-repository-url/extract-value-jq?env_id=ibm:yp:eu-de

packages/collector/test/tracing/messaging/amqp/test.js
```diff
@@ -23,7 +23,8 @@ const agentControls = globalAgent.instance;
 let publisherControls;
 let consumerControls;
 
-const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : describe.skip;
+// Note: Skipping ESM tests since we don't have any, but this interferes with the Bull ESM test
+const mochaSuiteFn = supportedVersion(process.versions.node) && !process.env.RUN_ESM ? describe : describe.skip;
```
abhilash-sivan, line 27: AMQP doesn't have ESM files yet, so it was skipped in ESM because an extra 'rabbitmq' span appeared during the evaluation of Bull ESM tests.
We can revisit this when adding esm support for amqp

packages/collector/test/tracing/messaging/bull/receiver.mjs
```diff
@@ -0,0 +1,81 @@
...
+const logPrefix = `Bull Receiver ESM (${process.pid}):\t`;
+
+import express from 'express';
+import Queue from 'bull';
+
+const app = express();
+import portFactory from '../../../test_util/app-port.js';
+const port = portFactory();
+
+import { ProcessTypes, buildReceiver } from './util.mjs';
```
kirrg001, line 22: issue: What was the reason to add a copy of "utils.js"? Did you do any customizations in utils.mjs?

If not, please use:
> import { ProcessTypes, buildReceiver } from './util.js';

We can import CJS from ESM.

packages/collector/test/tracing/messaging/amqp/test.js
kirrg001: qs: I have two questions:

1. Why is this part of this PR?
2. If a test does not have ".mjs" files, the tests are automatically skipped. Why do we need this new condition?

packages/collector/test/tracing/messaging/bull/receiver.mjs
abhilash-sivan: No customizations were added. I will revert back to js

kirrg001 (changes requested):
From https://jsw.ibm.com/browse/INSTA-13393

> We append --require. We do not check if CJS or ESM

I am missing a statement for ^.

packages/collector/test/tracing/messaging/bull/test.js
```diff
@@ -39,7 +39,7 @@ const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : descri
 
 const retryTime = 1000;
 
-mochaSuiteFn('tracing/messaging/bull', function () {
+mochaSuiteFn('tracing/messaging/bull', async function () {
```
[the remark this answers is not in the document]
abhilash-sivan: added an if statement for restricting the bull tests in ESM for process API. 

It's just a wrap code.


if (process.env) {
  .....
}

These are the changes you are seeing

packages/collector/test/tracing/messaging/amqp/test.js
abhilash-sivan: 1. Removed it as it's not needed
2. Yeah correct. However the `allowRootExitSpanApp.js` succeeded in the ESM check because of the `App.js` keyword presence. That created some issues while running this in the test suit.

Anyway I have the solution now without skipping the test

Now bull ESM tests are running properly locally. 
Let's see how it behaves in the pipeline

packages/collector/test/tracing/messaging/amqp/test.js
```diff
@@ -53,7 +53,7 @@ const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : descri
           await controls.start(null, null, true);
         });
 
-        beforeEach(async () => {
+        after(async () => {
```
aryamohanan, line 56: why this change?

abhilash-sivan:
> From https://jsw.ibm.com/browse/INSTA-13393
> 
> > We append --require. We do not check if CJS or ESM
> 
> I am missing a statement for ^.

_packages/collector/src/tracing/instrumentation/process/childProcess.js_
_We append --require. We do not check if CJS or ESM_

This is because here we are not injecting the esm-loader or esm-register rather we are injecting the following file:
https://github.com/instana/nodejs/blob/main/packages/collector/src/immediate.js

This injection occurs in child processes to ensure that if a process is spawned by an instrumented parent, it can inherit the agent UUID. This enables the child process to generate and collect spans immediately without waiting to establish a connection to the agent.

packages/collector/test/tracing/messaging/amqp/test.js
abhilash-sivan: It has also passed successfully in the pipeline!

packages/collector/test/tracing/messaging/amqp/test.js
abhilash-sivan: This was adding clutter to the trace data because it was running for ESM as well but wasn't being cleared properly.
Sio I changed the logic to clear trace data after the current test suit.

kirrg001:
> This is because here we are not injecting the esm-loader or esm-register rather we are injecting the following file:
https://github.com/instana/nodejs/blob/main/packages/collector/src/immediate.js
> This injection occurs in child processes to ensure that if a process is spawned by an instrumented parent, it can inherit the agent UUID. This enables the child process to generate and collect spans immediately without waiting to establish a connection to the agent.

Thats a good general description, but lets be a bit more specific here.

Lets add a comment here https://github.com/instana/nodejs/blob/v4.6.2/packages/collector/src/tracing/instrumentation/process/childProcess.js#L67-L68

> Bull is not using ESM internally yet. They fork a childProcess for "master.js" (CJS) internally here 
> https://github.com/OptimalBits/bull/blob/489c6ab8466c1db122f92af3ddef12eacc54179e/lib/process/child-pool.js#L52
> Currently, we only support `childProcess.fork` for Bull. If Bull migrates from CJS to ESM, we have to change the fork implementation. 

Furthermore:

Please move the comment from https://github.com/instana/nodejs/blob/v4.6.2/packages/collector/src/tracing/instrumentation/process/childProcess.js#L24-L29 to https://github.com/instana/nodejs/blob/v4.6.2/packages/collector/src/tracing/instrumentation/process/childProcess.js#L34-L35

And extend with

TODO: We have not investigated if we have to add ESM support for edgemicro. TICKET-REF

packages/collector/test/tracing/messaging/amqp/test.js
abhilash-sivan: I did a refactoring to remove this change

kirrg001 (approved):

PR #1556, opened 2025-02-11
refactor: removed "getLogger" usage

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-13498

Reasons why we have to move away from `getLogger('tracing/xxx')`:

- Overhead / Performance. We create and store (!) more then 70 logger child instances. See registry logic in logger.js.
- We do not follow dependency injection, which makes it very hard to maintain the codebase. We always request a new instance on top of each file using `getLogger('tracing/blubb')` - without knowing what the actual logging instance is. We should only access `config.logger` in the init function of each module. This gives us control over the codeflow.
- Logs disappear in the beginning of the application if you call `getLogger` too early e.g. logs from `normalizeConfig` did not appear from serverless correctly.
- I have never used this feature. If we receive logs, I just copy the error message and search for it in the codebase.

I have not found a good reason to keep the child logger logic. We now create only one child logger: the instana internal logger if you provide a custom logger. Otherwise we only have one logger instance. 

If we want to re-add log prefixes, we can easily do that by adding e.g. `logger = config.logger.prefix('xxx')`. This helper fn will wrap the log calls and automatically adds a prefix such as `tracing/kafka: error msg`. 

- [x] ~~measure bootstrap time (AWS + Locally)~~ too hard to measure
- [x] fallback for `let logger` to prevent bugs
- [x] extract some changes in smaller PR's if possible

PR #1565, opened 2025-02-17
fix: ensure the logger always applies the correct log level

kirrg001:
no issue

Extracted from #1556 

See comment.

### DIFF
--- a/packages/collector/src/logger.js
+++ b/packages/collector/src/logger.js
@@ -163,10 +163,23 @@ function hasLoggingFunctions(_logger) {
 /**
  * @param {import("@instana/core/src/core").GenericLogger} _logger
  * @param {string|number} level
+ *
+ * Bunyan uses `level()` function to set the log level.
+ * https://www.npmjs.com/package/bunyan#levels
+ *
+ * Pino uses `level = LEVEL`.
+ * https://github.com/pinojs/pino/blob/main/docs/api.md#logger-level
+ *
+ * As far as I could figure out, Winston uses `level = LEVEL` or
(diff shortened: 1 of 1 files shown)


packages/collector/src/logger.js
```diff
@@ -163,10 +163,23 @@ function hasLoggingFunctions(_logger) {
...
+ * Bunyan uses `level()` function to set the log level.
+ * https://www.npmjs.com/package/bunyan#levels
+ *
+ * Pino uses `level = LEVEL`.
+ * https://github.com/pinojs/pino/blob/main/docs/api.md#logger-level
+ *
+ * As far as I could figure out, Winston uses `level = LEVEL` or
+ * `transports.console.level = 'info';`
+ *
+ * I could not figure out who uses `setLevel`.
```
aryamohanan, line 176:
```suggestion
```

aryamohanan, line 176: 🤔

packages/collector/src/logger.js
```diff
@@ -163,10 +163,23 @@ function hasLoggingFunctions(_logger) {
...
  */
 function setLoggerLevel(_logger, level) {
   if (typeof _logger.setLevel === 'function') {
     _logger.setLevel(level);
+  } else if (typeof _logger.level === 'function') {
```
aryamohanan, line 181: Would it make more sense to move this condition to the beginning then?

kirrg001: Feel free to create a PR to reorder 👌

aryamohanan (approved):

PR #1573, opened 2025-02-17
refactor: core utils usage

kirrg001:
Extracted from #1556 

Cleanup readability for core utils.

### DIFF
--- a/packages/core/src/tracing/index.js
+++ b/packages/core/src/tracing/index.js
@@ -15,9 +15,7 @@ const tracingUtil = require('./tracingUtil');
 const spanBuffer = require('./spanBuffer');
 const supportedVersion = require('./supportedVersion');
 const { otelInstrumentations } = require('./opentelemetry-instrumentations');
-const { isLatestEsmSupportedVersion, hasExperimentalLoaderFlag, isESMApp } = require('../util/esm');
-const iitmHook = require('../util/iitmHook');
-const { getPreloadFlags } = require('../util/getPreloadFlags');
+const coreUtil = require('../util');
 
 let tracingEnabled = false;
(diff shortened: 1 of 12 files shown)


packages/core/src/tracing/index.js
```diff
@@ -135,7 +133,6 @@ exports.sdk = sdk;
 exports.spanBuffer = spanBuffer;
 exports.supportedVersion = supportedVersion;
 exports.util = tracingUtil;
-exports.isLatestEsmSupportedVersion = isLatestEsmSupportedVersion;
```
kirrg001, line 138: This was absolutely unnecessary. The ESM util can be accessed via `core.util.esm` now.

packages/core/src/tracing/index.js
```diff
@@ -15,9 +15,7 @@ const tracingUtil = require('./tracingUtil');
 const spanBuffer = require('./spanBuffer');
 const supportedVersion = require('./supportedVersion');
 const { otelInstrumentations } = require('./opentelemetry-instrumentations');
-const { isLatestEsmSupportedVersion, hasExperimentalLoaderFlag, isESMApp } = require('../util/esm');
-const iitmHook = require('../util/iitmHook');
-const { getPreloadFlags } = require('../util/getPreloadFlags');
+const coreUtil = require('../util');
```
kirrg001, line 18: We have to make clear that these are the core utils.
They will get initialized in the core/src/index module in #1556 
But they can be used & activated anywhere.
There is always more room for further optimizations later.

e.g. We should rename `getPreloadFlags` to simply `preloadFlags`. But it will create a large conflict between the pull requests.

aryamohanan (approved):

PR #1578, opened 2025-02-19
test: extended bull test to support esm app

abhilash-sivan:
This is a basic PR that ensures that Bull is working with ESM.

I will expand the test base in another PR

### DIFF
--- a/packages/collector/test/tracing/messaging/bull/app.js
+++ b/packages/collector/test/tracing/messaging/bull/app.js
@@ -1,0 +1,54 @@
+/*
+ * (c) Copyright IBM Corp. 2025
+ */
+
+'use strict';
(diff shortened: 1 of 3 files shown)


kirrg001 (commented):
I thought with https://github.com/instana/nodejs/pull/1553, we can add receiver.mjs and sender.mjs.

What is your plan? I would like to avoid having to maintain two apps for the same thing.

Could you please add the JIRA reference to the PR body?

abhilash-sivan:
This was an initial test case for verifying bull working with ESM.
I planned to remove this and update it with proper files in https://github.com/instana/nodejs/pull/1553 .
It's completed now. We can ignore this PR. Closing

PR #1580, opened 2025-02-19
refactor: updated the filtering logic to support advanced configurations

aryamohanan:
Refactored span filtering logic to support advanced configurations for future Kafka support

**Changes:**

- Replaced the original simple endpoint filtering logic with a more flexible, advanced configuration
  that supports both simple string/array and object-based configs.
- Introduced new helper functions:
  • `matchEndpoints`
  • `matchMethods`
- Extended the tests to cover more cases.

Example config we expected:

```
ignoreEndpoints = {
    <service1>: [<methods_values>],
    <service2>: [{ methods: [<methods_values>], endpoints: [<endpoint_values>] }]
 };
```

**Next**

- [x] Parsing logic for config  #1585 (RFR)

This PR merge against branch `feat-ignore-endpoints-kafka`

ref INSTA-26148

### DIFF
--- a/packages/core/src/tracing/index.js
+++ b/packages/core/src/tracing/index.js
@@ -123,8 +123,8 @@ if (customInstrumentations.length > 0) {
 
 /**
  * @typedef {Object} IgnoreEndpointConfig
- * @property {string | string[]} methods
- * @property {string[]=} [endpoints]
+ * @property {string | string[]} [methods]
+ * @property {string | string[]} [endpoints]
  */
 /** @type {Array.<InstanaInstrumentedModule>} */
 let additionalInstrumentationModules = [];
--- a/packages/core/src/tracing/index.js
+++ b/packages/core/src/tracing/index.js
@@ -118,9 +118,14 @@ if (customInstrumentations.length > 0) {
  */
 
 /**
- * @typedef {Object.<string, string[]|string>}  IgnoreEndpoints
+ * @typedef {Object.<string, string[]|string|IgnoreEndpointConfig[]>}  IgnoreEndpoints
  */
 
+/**
+ * @typedef {Object} IgnoreEndpointConfig
+ * @property {string | string[]} method - A method or list of methods to ignore.
+ * @property {string[]=} [endpoints] - A list of endpoints to ignore (optional).
+ */
 /** @type {Array.<InstanaInstrumentedModule>} */
 let additionalInstrumentationModules = [];
 /** @type {Object.<string, InstanaInstrumentedModule>} */
--- a/packages/core/src/tracing/index.js
+++ b/packages/core/src/tracing/index.js
@@ -123,8 +123,8 @@ if (customInstrumentations.length > 0) {
 
 /**
  * @typedef {Object} IgnoreEndpointConfig
- * @property {string | string[]} [methods]
- * @property {string | string[]} [endpoints]
+ * @property {string[]} [methods]
+ * @property {string[]} [endpoints]
  */
 /** @type {Array.<InstanaInstrumentedModule>} */
 let additionalInstrumentationModules = [];
--- a/packages/core/src/tracing/index.js
+++ b/packages/core/src/tracing/index.js
@@ -123,8 +123,8 @@ if (customInstrumentations.length > 0) {
 
 /**
  * @typedef {Object} IgnoreEndpointConfig
- * @property {string | string[]} method - A method or list of methods to ignore.
- * @property {string[]=} [endpoints] - A list of endpoints to ignore (optional).
+ * @property {string | string[]} methods
+ * @property {string[]=} [endpoints]
  */
 /** @type {Array.<InstanaInstrumentedModule>} */
 let additionalInstrumentationModules = [];
--- a/packages/core/src/tracing/index.js
+++ b/packages/core/src/tracing/index.js
@@ -118,11 +118,10 @@ if (customInstrumentations.length > 0) {
  */
 
 /**
- * @typedef {Object.<string, string[]|string|IgnoreEndpointConfig[]>}  IgnoreEndpoints
+ * @typedef {Object.<string,IgnoreEndpointsFields[]>}  IgnoreEndpoints
  */
-
 /**
- * @typedef {Object} IgnoreEndpointConfig
+ * @typedef {Object} IgnoreEndpointsFields
  * @property {string[]} [methods]
  * @property {string[]} [endpoints]
  */
--- a/packages/core/src/util/spanFilter.js
+++ b/packages/core/src/util/spanFilter.js
@@ -8,7 +8,6 @@
 const IGNORABLE_SPAN_TYPES = ['redis', 'dynamodb', 'kafka'];
 
 /**
- *
  * @param {import('../core').InstanaBaseSpan} span
  * @param {import('../tracing').IgnoreEndpoints} ignoreEndpointsConfig
  * @returns {boolean}
@@ -22,36 +21,34 @@ function shouldIgnore(span, ignoreEndpointsConfig) {
   const ignoreConfigs = ignoreEndpointsConfig[span.n];
   if (!ignoreConfigs) return false;
 
-  const operation = span.data[span.n]?.operation;
-
-  const endpoints = Array.isArray(span.data[span.n]?.endpoints)
+  const spanOperation = span.data[span.n]?.operation?.toLowerCase();
+  const spanEndpoints = Array.isArray(span.data[span.n]?.endpoints)
     ? span.data[span.n].endpoints
     : [span.data[span.n]?.endpoints].filter(Boolean);
 
-  if (!operation && !endpoints.legth) {
+  if (!spanOperation && !spanEndpoints.legth) {
     return false;
   }
 
   // We support string, array, object formats for the config, but the string format is not shared publicly.
   // If the ignoreConfigs is a simple string, compare it with the operation.
   if (typeof ignoreConfigs === 'string') {
-    return ignoreConfigs?.toLowerCase() === operation?.toLowerCase();
+    return ignoreConfigs?.toLowerCase() === spanOperation;
   }
 
   return ignoreConfigs.some(config => {
     if (typeof config === 'string') {
-      return config?.toLowerCase() === operation?.toLowerCase();
+      return config?.toLowerCase() === spanOperation;
     }
 
     // If the config is an object, proceed with methods and endpoints checks.
-    if (typeof config === 'object' && config) {
+    if (typeof config === 'object') {
       let methodMatches = false;
       if (config.methods) {
         if (Array.isArray(config.methods)) {
-          methodMatches =
-            config.methods.includes('*') || config.methods.some(m => m?.toLowerCase() === operation?.toLowerCase());
+          methodMatches = config.methods.includes('*') || config.methods.some(m => m?.toLowerCase() === spanOperation);
         } else if (typeof config.methods === 'string') {
-          methodMatches = config.methods === '*' || config.methods?.toLowerCase() === operation?.toLowerCase();
+          methodMatches = config.methods === '*' || config.methods?.toLowerCase() === spanOperation;
         }
       }
 
@@ -64,7 +61,9 @@ function shouldIgnore(span, ignoreEndpointsConfig) {
       if (configuredEndpoints.includes('*')) {
         return true;
       }
-      return endpoints.some((/** @type {string} */ endpoint) => configuredEndpoints.includes(endpoint));
+      return spanEndpoints.some((/** @type {string} */ endpoint) =>
+        configuredEndpoints.some(e => e?.toLowerCase() === endpoint?.toLowerCase())
+      );
     }
 
     return false;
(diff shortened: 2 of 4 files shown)


packages/core/src/util/spanFilter.js
```diff
@@ -5,36 +5,103 @@
 'use strict';
 
 // List of span types to allowed to ignore
-const IGNORABLE_SPAN_TYPES = ['redis', 'dynamodb'];
+// 'kafka_placeholder' used as a temporary span type for testing, until the Kafka support is implemented.
+const IGNORABLE_SPAN_TYPES = ['redis', 'dynamodb', 'kafka_placeholder'];
```
aryamohanan, line 9: I don't get any other better name instead of `kafka_placeholder`, I'm open to alternative names if anyone has a better suggestion

packages/core/src/util/spanFilter.js
```diff
@@ -5,36 +5,103 @@
...
     return false;
   }
-  const endpoint = endpoints[span.n];
-  if (!endpoint) return false;
 
-  // The endpoint already been converted to lowercase during parsing, so here we are normalizing
-  // the operation for case-insensitive comparison
-  const operation = span.data?.[span.n]?.operation?.toLowerCase();
-  if (!operation) return false;
+  const configuredEndpoints = Array.isArray(config.endpoints) ? config.endpoints : [config.endpoints];
```
abhilash-sivan, line 25: suggestion: we can simplify this.

`const configuredEndpoints = [].concat(config.endpoints);`

OR 

`const configuredEndpoints = [config.endpoints].flat();`

aryamohanan: Thanks for the suggestion, Abhilash! While both .concat() and .flat() work, I think using:
```
const configuredEndpoints = Array.isArray(config.endpoints) ? config.endpoints : [config.endpoints];
```
is a bit more explicit, readable and avoids unnecessary operations. It explicitly checks if config.endpoints is already an array, avoiding unnecessary operations. Let me know your thoughts!

packages/core/test/util/spanFilter_test.js
```diff
@@ -19,7 +19,8 @@ const span = {
     }
   }
 };
-const ignoreEndpoints = {
+
+let ignoreEndpoints = {
```
kirrg001, line 23: suggestion: Move this to line 29 with keeping const.

packages/core/test/util/spanFilter_test.js
```diff
@@ -57,4 +58,202 @@ describe('util.spanFilter', () => {
     span.data.dynamodb.operation = 'PutItem';
     expect(applyFilter({ span, ignoreEndpoints })).equal(span);
   });
+
+  describe('advanced filtering with mixed config', () => {
```
kirrg001, line 62:
```suggestion
  describe('util.spanFilter: advanced filtering', () => {
```

packages/core/src/util/spanFilter.js
```diff
@@ -5,36 +5,103 @@
...
+ * @returns {boolean}
+ */
+function matchMethods(span, config) {
+  const spanOperation = span.data[span.n]?.operation?.toLowerCase();
+  let methodMatches = false;
+
+  if (config.methods) {
+    if (Array.isArray(config.methods)) {
+      methodMatches = config.methods.includes('*') || config.methods.some(m => m?.toLowerCase() === spanOperation);
+    } else if (typeof config.methods === 'string') {
```
kirrg001, line 49: This is not covered in the unit test. Could we add a test case? Thanks!

aryamohanan: Added the test cases, we have now 99.15% coverage for the file

packages/core/src/util/spanFilter.js
abhilash-sivan: Yes, I agree with this. Let's go with the initial approach, which is more readable!

abhilash-sivan (approved):

packages/core/src/util/spanFilter.js
kirrg001: IMO its fine because you will change it soon.
Will you merge this PR into a feature branch? I am not sure we should merge this into main.
But I am also not against it.

packages/core/src/util/spanFilter.js
```diff
@@ -5,36 +5,103 @@
...
  * @param {import('../core').InstanaBaseSpan} span
- * @param {import('../tracing').IgnoreEndpoints} endpoints
+ * @param {import('../tracing').IgnoreEndpointConfig} config
  * @returns {boolean}
  */
-function shouldIgnore(span, endpoints) {
-  // Skip if the span type is not in the ignored list
-  if (!IGNORABLE_SPAN_TYPES.includes(span.n)) {
+function matchEndpoints(span, config) {
+  const spanEndpoints = Array.isArray(span.data[span.n]?.endpoints)
```
kirrg001, line 17: qs: Where is "span.data.kafka.endpoints" coming from? I think this must be **service**!

packages/core/test/util/spanFilter_test.js
```diff
@@ -57,4 +57,244 @@ describe('util.spanFilter', () => {
...
+
+    it('returns null when both method and endpoint criteria match', () => {
+      ignoreEndpoints = {
+        kafka_placeholder: [{ methods: ['send', 'consume'], endpoints: ['topic3'] }]
+      };
+      span.n = 'kafka_placeholder';
+      span.data = {
+        kafka_placeholder: {
+          operation: 'send',
+          endpoints: ['topic3']
```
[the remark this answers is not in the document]
aryamohanan: In case of kafka we can expect multiple topics in span. When adding the kafka support I can double check this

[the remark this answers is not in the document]
kirrg001: See https://github.com/instana/nodejs/pull/1580/files#r1975080397

packages/core/src/tracing/index.js
```diff
@@ -118,9 +118,14 @@ if (customInstrumentations.length > 0) {
...
 
 /**
- * @typedef {Object.<string, string[]|string>}  IgnoreEndpoints
+ * @typedef {Object.<string, string[]|string|IgnoreEndpointConfig[]>}  IgnoreEndpoints
  */
 
+/**
+ * @typedef {Object} IgnoreEndpointConfig
+ * @property {string | string[]} [methods]
+ * @property {string | string[]} [endpoints]
```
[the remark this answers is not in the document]
aryamohanan: updated

packages/core/src/util/spanFilter.js
```diff
@@ -5,36 +5,99 @@
...
 /**
  * @param {import('../core').InstanaBaseSpan} span
- * @param {import('../tracing').IgnoreEndpoints} endpoints
+ * @param {import('../tracing').IgnoreEndpointConfig} config
  * @returns {boolean}
  */
-function shouldIgnore(span, endpoints) {
-  // Skip if the span type is not in the ignored list
-  if (!IGNORABLE_SPAN_TYPES.includes(span.n)) {
+function matchEndpoints(span, config) {
```
kirrg001, line 16:
```suggestion
function matchEndpoints(span, ignoreConfig) {
```

Its confusing to read in line 26, because I thought this is the whole instana config object

> config.endpoints.includes

packages/core/src/tracing/index.js
```diff
@@ -118,9 +118,14 @@ if (customInstrumentations.length > 0) {
  */
 
 /**
- * @typedef {Object.<string, string[]|string>}  IgnoreEndpoints
+ * @typedef {Object.<string, string[]|string|IgnoreEndpointConfig[]>}  IgnoreEndpoints
```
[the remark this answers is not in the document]
kirrg001: I understand now that we need to support "string" and "string[]" because the normalization is not done yet. 

But why three? 🤔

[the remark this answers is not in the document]
aryamohanan: We have 3 options
- string // basic filtering with a single operation
- string[] // basic filtering with an array of operation
- [{methods: []: endpoints: []}}

We can normalize this to
- string[] // basic filtering with an array of operation
- [{methods: []: endpoints: []}}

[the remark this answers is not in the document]
kirrg001: Its hard to follow

I expected to only have something like this:

> Object.<IgnoreEndpointConfigOld[] | IgnoreEndpointConfig[]>

> /**
> * @typedef {Object} IgnoreEndpointConfigOld
> * @property {string} [method]
> */

till we have merged the next PR for having a normalize helper which will transform any input into the same format for the internal processing.

packages/core/src/util/spanFilter.js
aryamohanan: If `span.data[span.n]?.endpoints` exists, filtering will be based solely on the `endpoints` property. Currently, in the case of Kafka, the span data does not have an `endpoints` field; instead, it have `service`. When we add support for ignoring Kafka endpoints, similar to `operation`, we will need to map the `service` to `endpoints.` In `spanBuffer`, we can then map it back to `service`, just as we do for `operation`.
Sorry I missed to add this.

kirrg001: Oki 👍 How about we add a comment to explain it shortly? Especially regarding kafka.service and any upcoming filtering such as http.

packages/core/src/util/spanFilter.js
```diff
@@ -5,36 +5,100 @@
...
   }
-  const endpoint = endpoints[span.n];
-  if (!endpoint) return false;
 
-  // The endpoint already been converted to lowercase during parsing, so here we are normalizing
-  // the operation for case-insensitive comparison
-  const operation = span.data?.[span.n]?.operation?.toLowerCase();
-  if (!operation) return false;
+  // If * is present, any endpoint will match.
+  if (ignoreconfig.endpoints.includes('*')) {
```
kirrg001, line 26: qs: I thought endpoints is an array or do we support both already?
I cannot find a test for this.

> endpoints: [*]
> endpoints: *

packages/core/src/util/spanFilter.js
kirrg001: We can merge it into main if the code is clean. Closing convo

packages/core/src/util/spanFilter.js
aryamohanan: Its already covered here https://github.com/instana/nodejs/pull/1580/files#diff-30b8cf03d0e00401cbc7d729a485dff0baef47ac7af17023f174e54d622b1540R104

aryamohanan: internally endpoints always an array

packages/core/src/util/spanFilter.js
```diff
@@ -5,36 +5,100 @@
...
+    return true;
+  }
 
-  // We support both array and string formats for endpoints, but the string format is not shared publicly.
-  if (Array.isArray(endpoint)) {
-    return endpoint.some(op => op.toLowerCase() === operation);
+  return spanEndpoints.some((/** @type {string} */ endpoint) =>
```
kirrg001, line 30: Please remove these types here and use:

  /**
   * @type {string[]}
   */
     const spanEndpoints = Array.isArray(span.data[span.n]?.endpoints)
    ? span.data[span.n].endpoints
    : [span.data[span.n]?.endpoints].filter(Boolean);

packages/core/src/util/spanFilter.js
kirrg001: Thanks!

packages/core/src/tracing/index.js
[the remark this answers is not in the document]
aryamohanan: IMO string[] is not old its basic filtering, adding methods and endpoints are advanced filtering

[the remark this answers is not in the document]
kirrg001: Then IgnoreEndpointConfigSimple vs IgnoreEndpointConfigExtended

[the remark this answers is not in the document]
kirrg001: But wait, I thought we normalize this and then we only need one.
I am so confused

packages/core/src/util/spanFilter.js
```diff
@@ -5,36 +5,100 @@
...
+  }
+
+  const ignoreConfigs = ignoreEndpointsConfig[span.n];
+  if (!ignoreConfigs) {
+    return false;
+  }
+  const spanOperation = span.data[span.n]?.operation?.toLowerCase();
+  // For basic filtering: when the configuration is provided as a simple string
+  // (e.g., "redis.get" or "dynamodb.query"), directly compare it to the span's operation.
+  if (typeof ignoreConfigs === 'string') {
```
kirrg001, line 72: So this will get **removed** when we add the normalize helper RIGHT?
Because we always receive **internally** (after normalazing the input from various places):

technology
  - methods: []
     endpoints: []

packages/core/src/tracing/index.js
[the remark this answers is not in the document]
aryamohanan: Will be resolved in https://github.com/instana/nodejs/pull/1580

packages/core/src/util/spanFilter.js
aryamohanan: Yes

kirrg001 (approved):

PR #1586, opened 2025-02-24
ci: updated the slack notification message for aws lambda layer publish

aryamohanan:
Added the clear message for slack notifications to scope the regions, whether its cn or non-cn.


kirrg001 (approved):
We could highlight "to the cn" and "to non cn" a bit more.
Otherwise LGTM

PR #1588, opened 2025-02-25
chore: added another docker example app

kirrg001:
Just another example app.
Added an improvment how to start the agent with

> npm run agent-run

We can also merge collector-latest & collector-v2 together.
v2 runs with v14.

Not important right now.


aryamohanan (approved):
LGTM

PR #1589, opened 2025-02-25
fix: reduced noise for dependency calculation debug logs

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-27904

We often have seen these debug logs in customer logs

> No main module path for fsevents

e.g. `fsevents` for example is an optional dependency (see package-lock.json). If it cannot be installed, the customer sees this log. Its super useless.

TBH we don't care if the module in your package.json was installed or not.

The easiest fix for now is to reduce the log to trace. It won't show up when using `INSTANA_DEBUG=true`.
Furthermore I pimped the log.

`INSTANA_LOG_LEVEL=trace` can be used instead. We should make more use of trace vs debug.

I want to merge #1556 first.

### DIFF
--- a/packages/core/test/test_util/index.js
+++ b/packages/core/test/test_util/index.js
@@ -22,7 +22,8 @@ module.exports = {
       debug: () => {},
       info: () => {},
       warn: () => {},
-      error: () => {}
+      error: () => {},
+      trace: () => {}
     };
   },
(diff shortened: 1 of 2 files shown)


aryamohanan (approved):
LGTM, please remove the blocked label.

PR #1592, opened 2025-02-26
refactor: added options pattern for backend connector init

kirrg001:
refs #1315

### DIFF
--- a/packages/serverless/src/backend_connector.js
+++ b/packages/serverless/src/backend_connector.js
@@ -16,37 +16,38 @@ const layerExtensionHostname = 'localhost';
 const layerExtensionPort = process.env.INSTANA_LAYER_EXTENSION_PORT
   ? Number(process.env.INSTANA_LAYER_EXTENSION_PORT)
   : 7365;
-let useLambdaExtension = false;
 
 const timeoutEnvVar = 'INSTANA_TIMEOUT';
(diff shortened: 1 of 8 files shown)


packages/serverless/src/backend_connector.js
```diff
@@ -65,39 +66,33 @@ exports.init = function init(
...
       );
-      backendTimeout = defaultTimeout;
+
+      options.backendTimeout = 500;
     }
   }
 
-  if (identityProvider) {
-    hostHeader = identityProvider.getHostHeader();
+  if (options.identityProvider) {
```
abhilash-sivan, line 82: Suggestion:
Can we optimize this to:

```
if (options.identityProvider) {
    hostHeader = options.identityProvider.getHostHeader() || 'nodejs-serverless';
} else {
    hostHeader = 'nodejs-serverless';
}
```

Or should we maintain the existing logic to allow **hostHeader = ''** if getHostHeader() returns an empty string?

aryamohanan (approved):
🫶

packages/serverless/src/backend_connector.js
kirrg001: Agree. But I have not touched this code. Feel free to raise a new PR to optimize the condition 👍

PR #1594, opened 2025-02-27
refactor: updated the normalize config to load agent config

aryamohanan:


aryamohanan:
This is not required for now. Closing

PR #1595, opened 2025-02-27
fix: fixed repeated debug logs for announceNodeCollector

abhilash-sivan:
refs [INSTA-27881](https://jsw.ibm.com/browse/INSTA-27881)

- [ ] verify if the retry agent connection works fine as before

### DIFF
--- a/packages/collector/src/agentConnection.js
+++ b/packages/collector/src/agentConnection.js
@@ -28,6 +28,8 @@ let maxContentErrorHasBeenLogged = false;
 const http = uninstrumentedHttp.http;
 let isConnected = false;
 
+let callbackWasCalled = false;
+
 /** @type {string | null} */
 let cpuSetFileContent = null;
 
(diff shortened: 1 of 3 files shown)


packages/collector/src/agentConnection.js
```diff
@@ -151,12 +151,14 @@ exports.announceNodeCollector = function announceNodeCollector(cb) {
   );
 
   req.setTimeout(agentOpts.requestTimeout, function onTimeout() {
-    cb(new Error('Announce request to agent failed due to timeout'));
+    // even though the cb here should point to the callee of the current fn, ie, `function tryToAnnounce`
+    // it gets pointed to the atmostOnce fucntion (wrapped call), but why the wrapped call executes the log inside ?
+    logger.error('Announce request to agent failed due to timeout');
```
abhilash-sivan, line 156: The issue we got because of this line got executed and also the 
https://github.com/instana/nodejs/pull/1595/files#diff-8ff40b8dde5d139718e5cd429da1d02c4aac512218fa125e3947c0b0e049df8cL138
got executed at the same execution cycle.

These two events caused the callback to execute twice and we got the debug log.

We only get this issue if the agent host specified is not accessible and the agent is not ready.
It keeps on printing out the logs because of this issue.

If the agent is not ready, but we try to connect to the agent with the proper hostname and port, the debug logs are different and we only get what we expect!

We can ignore this use case as this is just a debug log and the config was wrong in the first place.

abhilash-sivan:
this will be resolved in the following PRs:

https://github.com/instana/nodejs/pull/1600

https://github.com/instana/nodejs/pull/1601

PR #1597, opened 2025-02-28
test: fixed esm app run logic

abhilash-sivan:
This PR ensures the app.mjs exists in the path for proper ESM execution
see: `opts.appPath = path.join(opts.dirname, 'app.mjs');`

### DIFF
--- a/packages/collector/test/test_util/ProcessControls.js
+++ b/packages/collector/test/test_util/ProcessControls.js
@@ -73,7 +73,7 @@ class ProcessControls {
             opts.appPath = updatedPath;
           }
         } else if (opts?.dirname) {
-          const esmApp = fs.readdirSync(opts.dirname).find(f => f.endsWith('.mjs'));
+          const esmApp = fs.readdirSync(opts.dirname).includes('app.mjs');
           if (esmApp) {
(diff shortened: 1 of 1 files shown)


kirrg001 (approved):
Thanks!

packages/collector/test/test_util/ProcessControls.js
```diff
@@ -73,7 +73,7 @@ class ProcessControls {
             opts.appPath = updatedPath;
           }
         } else if (opts?.dirname) {
-          const esmApp = fs.readdirSync(opts.dirname).find(f => f.endsWith('.mjs'));
+          const esmApp = fs.readdirSync(opts.dirname).includes('app.mjs');
```
aryamohanan, line 76: Qs: What problem this update resolves?

abhilash-sivan: The previous logic only checks for .esm files and assigning the esmApp path as app.mjs. That seems to be wrong because if I have `sender.mjs` present in the folder and not an `app.mjs`, this logic fails

abhilash-sivan:
Closing this PR as we already covered this in https://github.com/instana/nodejs/pull/1610

PR #1603, opened 2025-03-03
chore(serverless): called log.init in preactivate

kirrg001:
- we cannot use the "console" logger in environment.js because then the customer will see debug logs on boostrap
- call log.init twice for now
- refactor later


packages/serverless/src/environment.js
```diff
@@ -9,16 +9,7 @@ const instanaEndpointUrlEnvVar = 'INSTANA_ENDPOINT_URL';
...
-/**
- * TODO:
- *
- * This module is called BEFORE the serverless packages is required - see preactivate.js.
- * At this stage, we don't have any logger initialized yet.
- * For now, we just assign the Node.js inbuild `console` instance.
- *
- * We have to move the validation into the initialization of the serverless packages.
- */
-const logger = console;
```
kirrg001, line 21: This does not work, because we see debug logs when requiring e.g. the aws lambda pkg.

abhilash-sivan (approved):

PR #1605, opened 2025-03-03
feat: added environment variable INSTANA_IGNORE_ENDPOINTS_PATH to configure ignoring endpoints.

aryamohanan:


packages/core/src/util/readFromYaml.js
```diff
@@ -0,0 +1,25 @@
+/*
+ * (c) Copyright IBM Corp. 2025
+ */
+
+'use strict';
+
+const yaml = require('yaml');
```
aryamohanan, line 7: The two most commonly used packages for YAML parsing are [js-yaml](https://www.npmjs.com/package/js-yaml) and [yaml](https://www.npmjs.com/package/yaml?activeTab=readme). Both of these packages have a huge downloads. 
I preferred `yaml` because it has more recent support; the last release of `js-yaml` was four years ago, whereas `yaml` was updated just two months ago.  

[Comparison on npm trends](https://npmtrends.com/js-yaml-vs-yaml)

packages/collector/test/tracing/database/redis/test.js
```diff
@@ -1005,6 +1005,88 @@ const globalAgent = require('../../../globalAgent');
                   });
               });
             });
+
+            describe('Ignore Endpoints loaded from env variable INSTANA_IGNORE_ENDPOINTS_PATH', async () => {
```
aryamohanan, line 1009: Integration test added for redis in case of INSTANA_IGNORE_ENDPOINTS_PATH is set

packages/core/src/util/configNormalizers/ignoreEndpoints.js
```diff
@@ -110,6 +111,34 @@ exports.fromEnv = function fromEnv(ignoreEndpointsEnv) {
   }
 };
 
+/**
+ * Reads and normalizes the `INSTANA_IGNORE_ENDPOINTS_PATH` configuration from a YAML file.
+ *
+ * - The YAML file must have a root key of either `tracing` or `com.instana.tracing`.
+ * - If `com.instana.tracing` is detected, it will be logged and automatically mapped to `tracing`.
```
aryamohanan, line 118: We support both "tracing" and "com.instana.tracing" internally. However, if a customer configures "com.instana.tracing", we will internally treat it as "tracing" and log a message.

see discussion https://github.ibm.com/instana/technical-documentation/pull/333#discussion_r10251165

packages/core/src/util/yamlReader.js
```diff
@@ -0,0 +1,34 @@
+/*
+ * (c) Copyright IBM Corp. 2025
+ */
+
+'use strict';
+
+// eslint-disable-next-line instana/no-unsafe-require, import/no-extraneous-dependencies
+const yaml = require('yaml');
```
aryamohanan, line 8: The two most commonly used packages for YAML parsing are [js-yaml](https://www.npmjs.com/package/js-yaml) and [yaml](https://www.npmjs.com/package/yaml?activeTab=readme). Both of these packages have a huge downloads.
I preferred yaml because it has more recent support; the last release of js-yaml was four years ago, whereas yaml was updated just two months ago.

[Comparison on npm trends](https://npmtrends.com/js-yaml-vs-yaml)
https://npm-compare.com/js-yaml,yaml,yamljs

packages/core/src/util/index.js
```diff
@@ -35,6 +36,7 @@ exports.init = function init(config) {
   requireHook.init(config);
   preloadFlags.init(config);
   iitmHook.init(config);
+  yamlReader.init(config);
```
aryamohanan, line 39: This file can be used in future for reading config from Yaml file.

packages/core/package.json
```diff
@@ -70,6 +70,7 @@
     "methods": "^1.1.2",
     "opentracing": "^0.14.5",
     "semver": "^7.5.4",
-    "shimmer": "^1.2.1"
+    "shimmer": "^1.2.1",
+    "yaml": "^2.7.0"
```
kirrg001, line 74: qs: Did you check if there is a lightweight yaml pkg? This pkg is quite big (681.1kb).
I found https://github.com/asab-se/yaml-reader. It has only 26kb. But not sure we can use it.

packages/core/src/util/normalizeConfig.js
```diff
@@ -718,18 +718,33 @@ function normalizeIgnoreEndpoints(config) {
...
     config.tracing.ignoreEndpoints = configNormalizers.ignoreEndpoints.fromEnv(process.env.INSTANA_IGNORE_ENDPOINTS);
-  } else {
     return;
   }
 
+  // Default Case: No ignore endpoints configured
+  // If none of the above conditions are met, tracing will capture all operations by default.
+  config.tracing.ignoreEndpoints = {};
+
   logger.debug(`Ignore endpoints have been configured: ${JSON.stringify(config.tracing.ignoreEndpoints)}`);
```
kirrg001, line 749: qs: I think this log msg should only appear if you configured ignoreEndpoints or?

packages/core/src/util/normalizeConfig.js
```diff
@@ -718,18 +718,33 @@ function normalizeIgnoreEndpoints(config) {
...
+  // Provides a simple way to configure ignored operations via environment variables.
+  if (process.env.INSTANA_IGNORE_ENDPOINTS) {
     config.tracing.ignoreEndpoints = configNormalizers.ignoreEndpoints.fromEnv(process.env.INSTANA_IGNORE_ENDPOINTS);
-  } else {
     return;
   }
 
+  // Default Case: No ignore endpoints configured
+  // If none of the above conditions are met, tracing will capture all operations by default.
+  config.tracing.ignoreEndpoints = {};
```
kirrg001, line 747: qs: Why do we have to redefine the default value? 🤔

packages/core/package.json
aryamohanan: The most commonly used packages for YAML parsing are [js-yaml](https://www.npmjs.com/package/js-yaml) and [yaml](https://www.npmjs.com/package/yaml?activeTab=readme). In comparison, [yaml-reader](https://github.com/asab-se/yaml-reader) is significantly less popular, with very low download numbers.  So not a good option for us.

[See usage trends](https://npmtrends.com/js-yaml-vs-yaml-vs-yaml-reader).

packages/core/src/util/yamlReader.js
```diff
@@ -0,0 +1,34 @@
...
+  logger = config.logger;
+};
+
+/**
+ * Loads and parses a YAML file from the given path.
+ * @param {string} filePath - The relative or absolute path to the YAML file.
+ */
+exports.read = function read(filePath) {
+  try {
+    const absolutePath = path.resolve(__dirname, filePath);
```
kirrg001, line 27: qs: Is this right? The env path input is always absolute, right? I am missing a clear handling for absolute vs relative. Maybe I have overseen it.

packages/core/src/util/yamlReader.js
```diff
@@ -0,0 +1,34 @@
...
+    const fileContent = fs.readFileSync(absolutePath, 'utf8');
+    return yaml.parse(fileContent);
+  } catch (error) {
+    logger?.warn('Error loading YAML file', error?.message);
```
kirrg001, line 31: suggestion: log the input path as well.

packages/core/src/util/yamlReader.js
```diff
@@ -0,0 +1,34 @@
...
+};
+
+/**
+ * Loads and parses a YAML file from the given path.
+ * @param {string} filePath - The relative or absolute path to the YAML file.
+ */
+exports.read = function read(filePath) {
+  try {
+    const absolutePath = path.resolve(__dirname, filePath);
+    const fileContent = fs.readFileSync(absolutePath, 'utf8');
```
kirrg001, line 28: qs: Did you research if its possible to read the file async? Sync is blocking.

packages/core/src/util/yamlReader.js
aryamohanan: Good point! I’ve updated the implementation to properly handle both absolute and relative paths.

packages/core/src/util/normalizeConfig.js
aryamohanan: removed

packages/core/src/util/yamlReader.js
aryamohanan: I used the synchronous method (`readFileSync`) because the configuration needs to be loaded before proceeding, ensuring availability at runtime. However, it’s definitely possible to use an asynchronous approach ([fs.promises.readFile](https://nodejs.org/api/fs.html#fspromisesreadfilepath-options)) if we refactor the logic to handle async operations properly. If non-blocking behavior is a priority, I can explore converting this into an async function. Let me know your thoughts!

packages/core/src/util/yamlReader.js
aryamohanan: done

packages/core/package.json
kirrg001: I saw the trends, thanks for sharing. Why is popularity important? Is yaml-reader maintained?

packages/collector/test/tracing/database/redis/test.js
```diff
@@ -1005,6 +1005,88 @@ const globalAgent = require('../../../globalAgent');
...
+
+              before(async () => {
+                controls = new ProcessControls({
+                  useGlobalAgent: true,
+                  appPath:
+                    redisVersion === 'latest' ? path.join(__dirname, 'app.js') : path.join(__dirname, 'legacyApp.js'),
+                  env: {
+                    REDIS_VERSION: redisVersion,
+                    REDIS_PKG: redisPkg,
+                    INSTANA_IGNORE_ENDPOINTS_PATH: path.resolve(__dirname, './files/tracing.yaml')
```
kirrg001, line 1021:
```suggestion
                    INSTANA_IGNORE_ENDPOINTS_PATH: path.join(__dirname, 'files', 'tracing.yaml')
```

packages/core/package.json
aryamohanan: It is not actively maintained, no proper release notes. The last release was over two years ago, and the major version was published six years ago. Furthermore, there are no open issues, and the project lacks a community presence. Only one contributor ever. This indicates poor maintenance and the absence of an active user base, potentially leading to security risks and a lack of future support.
IMO this is not a good option.

kirrg001: Its not completely unmaintained, but I agree its not ideal. Especially because of yaml specification support 1.0, 2.0 etc.

The problem is still: 681kb is a lot.

kirrg001: What about https://www.npmjs.com/package/read-yaml-file

packages/core/src/util/yamlReader.js
```diff
@@ -0,0 +1,34 @@
...
+  logger = config.logger;
+};
+
+/**
+ * Loads and parses a YAML file from the given path.
+ * @param {string} filePath - The relative or absolute path to the YAML file.
+ */
+exports.read = function read(filePath) {
+  try {
+    const absolutePath = path.isAbsolute(filePath) ? filePath : path.resolve(__dirname, filePath);
```
kirrg001, line 27: issue: That will not work.

Okay lets say you configure a relative path like

> INSTANA_IGNORE_ENDPOINTS_PATH=./my/folder/bla.yaml

How do you know where this file is? You cannot just resolve with __dirname.

We have to tell the customer to use absolute.

packages/core/package.json
aryamohanan: We have done a performance test in local and choose 'read-yaml-file' which is light weight and faster compared to other libraries.

packages/core/src/util/yamlReader.js
aryamohanan: done

packages/core/src/util/yamlReader.js
```diff
@@ -0,0 +1,41 @@
...
+
+'use strict';
+
+// eslint-disable-next-line instana/no-unsafe-require, import/no-extraneous-dependencies
+const readYamlFile = require('read-yaml-file');
+/**
+ * We chose 'read-yaml-file' because it is lightweight (3.34 kB) and faster
+ * compared to other YAML parsers. At the time of adding this, we considered
+ * alternatives like 'yaml', 'js-yaml', etc., but found them to be larger in size
+ * and slower in performance.
```
kirrg001, line 13:
```suggestion
 * and slower in performance. 'read-yaml-file' and "js-yaml" had no releases for 4y. If we have to replace the library in the future, we can easily do so.
```

packages/core/src/util/configNormalizers/ignoreEndpoints.js
```diff
@@ -110,6 +111,35 @@ exports.fromEnv = function fromEnv(ignoreEndpointsEnv) {
...
+exports.fromYaml = function fromYaml(yamlFilePath) {
+  /** @type {Record<string, any>} */
+  const endpointsConfig = read(yamlFilePath);
+  if (!endpointsConfig || typeof endpointsConfig !== 'object') {
+    logger?.debug(`Invalid YAML content. Expected an object, but got: ${typeof endpointsConfig}`);
+    return {};
+  }
+
+  if (!endpointsConfig.tracing && !endpointsConfig['com.instana.tracing']) {
+    logger?.debug('Invalid YAML structure. The root key must be either "tracing" or "com.instana.tracing".');
```
kirrg001, line 132:
```suggestion
    logger?.debug('Invalid YAML structure. The root key must be "tracing".');
```

packages/core/src/util/configNormalizers/ignoreEndpoints.js
```diff
@@ -110,6 +111,35 @@ exports.fromEnv = function fromEnv(ignoreEndpointsEnv) {
...
+    return {};
+  }
+
+  if (endpointsConfig['com.instana.tracing']) {
+    logger?.info('Detected "com.instana.tracing" in the YAML file. Using "tracing" as the accepted format.');
+    return exports.normalizeConfig(endpointsConfig['com.instana.tracing']['ignore-endpoints'] || {});
```
kirrg001, line 138:
```suggestion
   endpointsConfig['tracing'] = endpointsConfig['com.instana.tracing'];
   delete endpointsConfig['com.instana.tracing'];
```

packages/core/src/util/normalizeConfig.js
```diff
@@ -718,18 +718,30 @@ function normalizeIgnoreEndpoints(config) {
...
-  // This was introduced in Phase 1 for basic filtering based on operations only (without advanced filtering).
-  // Use this to filter spans by operation or method, e.g., `redis.get`, `kafka.consume`, etc.
-  else if (process.env.INSTANA_IGNORE_ENDPOINTS) {
-    config.tracing.ignoreEndpoints = configNormalizers.ignoreEndpoints.fromEnv(process.env.INSTANA_IGNORE_ENDPOINTS);
-  } else {
+
+  // Case 2: Load from a YAML file if `INSTANA_IGNORE_ENDPOINTS_PATH` is set
+  // Introduced in Phase 2 for advanced filtering based on both methods and endpoints.
+  // Also supports basic filtering for endpoints.
+  if (process.env.INSTANA_IGNORE_ENDPOINTS_PATH) {
```
kirrg001, line 731: issue: I am missing the absolute path check validation.

aryamohanan: re added

packages/core/src/util/configNormalizers/ignoreEndpoints.js
```diff
@@ -110,6 +111,35 @@ exports.fromEnv = function fromEnv(ignoreEndpointsEnv) {
...
+    return {};
+  }
+
+  if (!endpointsConfig.tracing && !endpointsConfig['com.instana.tracing']) {
+    logger?.debug('Invalid YAML structure. The root key must be either "tracing" or "com.instana.tracing".');
+    return {};
+  }
+
+  if (endpointsConfig['com.instana.tracing']) {
+    logger?.info('Detected "com.instana.tracing" in the YAML file. Using "tracing" as the accepted format.');
```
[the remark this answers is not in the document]
aryamohanan: done

packages/core/src/util/yamlReader.js
```diff
@@ -0,0 +1,48 @@
...
+};
+
+/**
+ * Loads and parses a YAML file from the given path.
+ * @param {string} filePath - The absolute path to the YAML file.
+ */
+exports.read = function read(filePath) {
+  try {
+    if (!path.isAbsolute(filePath)) {
+      logger?.warn(`The file path is not absolute. Expected an absolute path, but received: ${filePath}`);
```
kirrg001, line 40: This is not good enough. The customer will not know that it belongs to input validation for ignore endpoints.

In general: We validate in normalizeConfig, not here.

packages/core/src/util/configNormalizers/ignoreEndpoints.js
aryamohanan: I have manually updated the suggestion, sorry, if I do this there is a type error.

packages/core/src/util/configNormalizers/ignoreEndpoints.js
```diff
@@ -110,6 +111,39 @@ exports.fromEnv = function fromEnv(ignoreEndpointsEnv) {
...
+
+  if (endpointsConfig['com.instana.tracing']) {
+    logger?.info(
+      // eslint-disable-next-line max-len
+      `Detected the root key "com.instana.tracing" in the YAML file at "${yamlFilePath}". This format is accepted, but please migrate to using "tracing" as the root key.`
+    );
+    endpointsConfig.tracing = endpointsConfig['com.instana.tracing'];
+    delete endpointsConfig['com.instana.tracing'];
+  }
+  return exports.normalizeConfig(endpointsConfig.tracing['ignore-endpoints']) || {};
```
kirrg001, line 144: qs: Why do we need "|| {};"? normalizeConfig should return {} if anything is broken.

kirrg001 (approved):
Pre-approving 🥇

PR #1608, opened 2025-03-04
chore(aws-lambda): fixed and improved zip utility

kirrg001:
Extracted from #1315 

- cleanup, improve and fix

### DIFF
--- a/packages/aws-lambda/lambdas/bin/create-zip-util
+++ b/packages/aws-lambda/lambdas/bin/create-zip-util
@@ -1,6 +1,6 @@
 function createZip {
   name=${PWD##*/}
-  echo "create-zip-util: $name.zip"
+  echo "create-zip-util: $name.zip with build option '$BUILD_LAMBDAS_WITH'"
   rm -f $name.zip
   mkdir -p ../zip
   rm -f ../zip/$name.zip
@@ -16,7 +16,12 @@ function createZip {
       cp package-lock.json package-lock.json.backup
     fi
 
-    if [[ "${BUILD_LAMBDAS_WITH-}" == "layer" ]]; then
+    # if package-lock.json.backup exists, the script crashed. Bring it back
+    if [[ -f package-lock.json.backup ]]; then
+      cp package-lock.json.backup package-lock.json
+    fi
+
+    if [[ "${BUILD_LAMBDAS_WITH}" == "layer" ]]; then
       # We want to use the lambda layer and not the locally build npm package tar file.
       # If the lambda's package.json currently contains a dependency to the local tar files, like:
       # "dependencies": {
@@ -30,7 +35,7 @@ function createZip {
       npm uninstall --production -S @instana/serverless
       npm uninstall --production -S instana-aws-lambda-auto-wrap
 
-    elif [[ "${BUILD_LAMBDAS_WITH-}" == "local" ]]; then
+    elif [[ "${BUILD_LAMBDAS_WITH}" == "local" ]]; then
       # The lambda's package.json might or might not already have a dependency to the local tar files:
       # "dependencies": {
       #   "@instana/aws-lambda": "file:../../instana-aws-lambda.tgz",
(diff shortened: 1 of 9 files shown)


packages/aws-lambda/lambdas/bin/create-zip-util
```diff
@@ -30,7 +35,7 @@ function createZip {
       npm uninstall --production -S @instana/serverless
       npm uninstall --production -S instana-aws-lambda-auto-wrap
 
-    elif [[ "${BUILD_LAMBDAS_WITH-}" == "local" ]]; then
```
kirrg001, line 33: Got rid of "BUILD_LAMBDAS_WITH-". Not sure how this works. Its hard to understand. And there was also bugs e.g. when you did not provide the env at all.

packages/aws-lambda/lambdas/bin/create-zip.sh
```diff
@@ -37,37 +37,41 @@ echo
...
 # packages we will call "npm uninstall -S $package-name"  and if the package.json points to the tar file it needs
 # to exist so npm can uninstall it and all its transitive dependencies.
+
+version=$(jq -r '.version' package.json)
+
 echo "Building local tar.gz for @instana/serverless."
 cd ../serverless
 npm --loglevel=warn pack
-mv instana-serverless-*.tgz ../aws-lambda/instana-serverless.tgz
+mv instana-serverless-${version}.tgz ../aws-lambda/instana-serverless.tgz
```
kirrg001, line 54: Its better to read the exact version. Because sometimes you have multiple tgz files and then the "mv" cmd does not work. Have seen this a couple of times.

packages/aws-lambda/lambdas/bin/deploy-zip.sh
```diff
@@ -5,16 +5,7 @@
...
-
-if [[ -z "${1-}" ]]; then
-  echo "Usage $0 <lambda-folder-name>"
-  echo
-  echo "The mandatory argument <lambda-folder-name> is missing."
-  exit 1
-fi
-
-cd `dirname $BASH_SOURCE`/../zip
+set -eEo pipefail
```
kirrg001, line 9: I gave the utility love and fixed some bugs. You don't have to review the diff. Its just a test utility.

packages/aws-lambda/lambdas/bin/create-zip-util
abhilash-sivan: Maybe we should keep this:

> "${BUILD_LAMBDAS_WITH-}"
> 
> The - after the variable name (${BUILD_LAMBDAS_WITH-}) ensures that:
> If BUILD_LAMBDAS_WITH is unset, it expands to an empty string ("").
> This prevents errors from using an unset variable.

abhilash-sivan: If its buggy for us, then no need to stick on with the syntax

packages/aws-lambda/lambdas/bin/create-zip.sh
```diff
@@ -37,37 +37,41 @@ echo
 echo Building all local tar.gz files.
 
 if [[ -z "${BUILD_LAMBDAS_WITH-}" ]]; then
```
abhilash-sivan, line 39: Here also we are relying on - syntax, Let's keep it generic. 
Pls consider removing this if we are planning to do the same in https://github.com/instana/nodejs/pull/1608/files#diff-7206912f97379b43fc548475cc72f54614411a8e9eece3431708b81ecb075102L33

"${BUILD_LAMBDAS_WITH-}"

packages/aws-lambda/lambdas/bin/create-zip-util
kirrg001: Thanks for reminding me. I will remove the docs section. 

The logic with BUILD_LAMBDAS_WITH- did not work for me and was really hard to read.

Any reason why do you want to keep it?
If BUILD_LAMBDAS_WITH is not defined, we fallback to 

> BUILD_LAMBDAS_WITH = local

See https://github.com/instana/nodejs/pull/1608/files#diff-66728008d66627e1fb55c0c1437999d221cfb686059e9d6c34aba069a2dd5c80R41

packages/aws-lambda/lambdas/bin/create-zip.sh
kirrg001: Good catch. I need to cleanup more 👍

packages/aws-lambda/lambdas/bin/create-zip-util
abhilash-sivan: No worries, we can go with the more readable approach

kirrg001: Sweet. Thx 👍

abhilash-sivan (approved):
LGTM

aryamohanan (approved):
LGTM ❤️

PR #1610, opened 2025-03-04
chore: add utility function for checking ESM apps

abhilash-sivan:
This PR ensures that only .mjs files are run for tests when ESM is enabled.

This potentially blocks the wrong app executions and reduces noise in ESM.


packages/collector/test/test_util/ProcessControls.js
```diff
@@ -63,17 +64,18 @@ class ProcessControls {
...
             ? opts.appPath.replace(/\.js$/, '.mjs')
             : `${opts.appPath}.mjs`;
+
           if (fs.existsSync(updatedPath)) {
             opts.execArgv = resolveEsmLoader();
             opts.appPath = updatedPath;
           }
         } else if (opts?.dirname) {
-          const esmApp = fs.readdirSync(opts.dirname).find(f => f.endsWith('.mjs'));
+          const esmApp = checkESMApp(opts.dirname);
```
kirrg001, line 78: qs: Any reason why we cannot put the logic from line 69-76 into the utility?
Its a bit weird that we leave some of the logic here and some of the logic got outsourced. I understand why you did, but still I was wondering if we can also move the other block into the utility.

```
testUtils.esmApp.exists({ appPath: opts.appPath })
testUtils.esmApp.exists({ dirname: opts.dirname })

```

abhilash-sivan: No reason, one is checking for the exact filename, and the other is checking for just any .mjs files. 
The logic can be combined, l will update

kirrg001 (approved):

PR #1611, opened 2025-03-04
ci: fix build name for manual esm (node-20-true)

kirrg001:
refs https://github.com/instana/nodejs/pull/1599


kirrg001:
I will just merge this and try it out on main. I don't want to change the Tekton config for this tiny change.

PR #1602, opened 2025-03-03
test: added allow root exit span apps for bull and amqp

abhilash-sivan:
This PR adds rootExitSpan tests for bull and amqp.

As per the current logic, `rootExitSpanApp.js` or by default `app.js` files were running in ESM test (logic in ProcessControls) as a replacement.

We are getting the flaky tests now because of the newly added ESM test for bull. As the bull test ran, the amqp rootExitSpanApp test was also running parallel causing some timing issues.

This weird issue was happening in the first place due to a combination of factors:

- the dynamic importing structure of ESM and 
- synchronous module loading of the commonjs app and 
- also the asynchronous behavior of message queues

So currently adding this ESM tests fixes this issue.

PR #1614, opened 2025-03-04
test: added short-term solution for multiple esm apps + describes

kirrg001:
- `allowRootExitSpan` app in bull test was executed although it did not have an esm implementation
- We are not sure what long term solution look like.
- This is accaptable for now.
- Note: We only have to this this if multiple mjs files are in a folder!

### DIFF
--- a/packages/collector/test/hooks.js
+++ b/packages/collector/test/hooks.js
@@ -34,6 +34,9 @@ exports.mochaHooks = {
   beforeEach() {
     const testFile = this.currentTest.file;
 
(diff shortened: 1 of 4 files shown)


abhilash-sivan (approved):
pre approving

kirrg001:
Closing in favor of https://github.com/instana/nodejs/pull/1602
We should keep it simple and say:

> As soon as you have more than one mjs, all tests need to support mjs in the folder.

@abhilash-sivan will update the CONTRIBUTING.md.

PR #1623, opened 2025-03-06
feat(serverless): added request id to instana debug logs

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-13498

Background: On Lambda for example requests will get frozen and wake up again if the handler is unfrozen.
We have to understand which log belongs to which request to understand flows and behaviors.

### DIFF
--- a/packages/serverless/src/backend_connector.js
+++ b/packages/serverless/src/backend_connector.js
@@ -5,6 +5,7 @@
 
 'use strict';
 
+const crypto = require('crypto');
 // eslint-disable-next-line import/order
 const environmentUtil = require('./environment');
 const uninstrumented = require('./uninstrumentedHttp');
@@ -27,6 +28,8 @@ let proxyAgent;
 
 const disableCaCheckEnvVar = 'INSTANA_DISABLE_CA_CHECK';
 const disableCaCheck = process.env[disableCaCheckEnvVar] === 'true';
+
+const getRequestId = () => crypto.randomBytes(16).toString('hex');
 
 let requestHasFailed = false;
 let warningsHaveBeenLogged = false;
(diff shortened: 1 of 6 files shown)


packages/serverless/src/backend_connector.js
```diff
@@ -28,6 +28,8 @@ let proxyAgent;
 const disableCaCheckEnvVar = 'INSTANA_DISABLE_CA_CHECK';
 const disableCaCheck = process.env[disableCaCheckEnvVar] === 'true';
 
+const getRequestId = () => Date.now().toString(36) + Math.random().toString(36).slice(2);
```
kirrg001, line 31: This is the fastest option we have. Short & unique.

aryamohanan: Should the request ID have a fixed or maximum length?

packages/serverless/src/backend_connector.js
```diff
@@ -114,25 +116,31 @@ exports.setLogger = function setLogger(_logger) {
  * the variable is always true.
  */
 exports.sendBundle = function sendBundle(bundle, finalLambdaRequest, callback) {
-  logger.debug(`Sending bundle to Instana (no. of spans: ${bundle?.spans?.length ?? 'unknown'})`);
-  send({ resourcePath: '/bundle', payload: bundle, finalLambdaRequest, callback });
+  const requestId = getRequestId();
```
aryamohanan, line 119: qs: We use the requestid only for tracking the logs of each request?

packages/serverless/src/backend_connector.js
kirrg001: Its fixed? Why do you think this has no fixed length? 🤔

packages/serverless/src/backend_connector.js
kirrg001: Yes exactly. Most of the logs are debug anyway.

packages/serverless/src/backend_connector.js
```diff
@@ -40,12 +38,31 @@ const defaults = {
...
+ * on serverless we do not want to log too much.
+ * If the debug mode is enabled, we want to add a request id to the instana
+ * logs because the AWS runtime freezes requests and they are waking up
+ * as soon as the next request is coming in. With the request id we can
+ * identify the logs of a single request.
+ *
+ * Due to performance reasons, we do not want to generate a request id
+ * in the non debug mode.
+ */
+const getRequestId = () => {
```
kirrg001, line 56: Added this to only generate the id if debug is really on,

packages/serverless/src/backend_connector.js
aryamohanan: I was wondering because Math.random().toString(36).slice(2) can produce varying lengths depending on the random value. I came across a similar issue before, and one way to ensure a fixed length is by using padding.
Would it make sense to apply that here for consistency?

kirrg001: Not important. Its only a lookup id.

aryamohanan (approved):
LGTM

kirrg001:
I will manually test tomorrow before merging

abhilash-sivan (approved):

kirrg001:
Manually tested. Looks normal

PR #1627, opened 2025-03-07
build: fixed package-lock diff

aryamohanan:


package-lock.json
```diff
@@ -53012,7 +53012,7 @@
       },
       "devDependencies": {
         "opentracing": "^0.14.5",
-        "typeorm": "^0.3.21"
```
kirrg001, line 53015: 🤔 

This looks wrong.
https://github.com/instana/nodejs/commits/main/

I feel like something else is broken, because the dep in collector pkg does not get updated.

aryamohanan: There was an issue in our `update-currencies` script, specifically in the step where we stage changes using `git add`. Initially, we used the command:  

```sh
git add *package.json ackage-lock.json
```   
However, the wildcard (`*package.json`) did not work as expected, likely due to the way the shell interprets it when expanding file paths. Since the wildcard wasn't enclosed in quotes, it may have failed to match the intended files.  

To resolve this, I modified the command by adding quotes around the wildcard, like this:  

```sh
git add '*package.json'
```  

After making this adjustment, the staging process worked correctly, and the script now behaves as expected.

In case of typeorm, the package.json was inside collector package, and while staging the changes package.json changes were not added, only lock changes added. This cause the issue.

kirrg001 (approved):
LGTM <3

abhilash-sivan (approved):

PR #1628, opened 2025-03-07
test: mocha after not called for aws fargate tests

kirrg001:
Omg this bug drove me crazy.

I had a failing aws-fargate test.
And I could not quit the test with CTRL+C.
Then I figured out the mocha after hook is not called. So the processes are not stopped.

:( 

1. No reporters for local testing
2. [FailureReporter](https://github.com/instana/nodejs/blob/main/packages/core/test/test_util/failureReporter.js) is broken. It radically destroys the process and then mocha after is not called!

### DIFF
--- a/packages/aws-fargate/package.json
+++ b/packages/aws-fargate/package.json
@@ -27,7 +27,7 @@
   "scripts": {
     "audit": "npm audit --omit=dev",
     "node_modules:exists": "mkdir -p node_modules",
-    "test": "mocha $npm_config_watch --config=test/.mocharc.js --reporter mocha-multi-reporters --reporter-options configFile=reporter-config.json --sort $(find test -iname '*test.js')",
+    "test": "mocha $npm_config_watch --config=test/.mocharc.js --sort $(find test -iname '*test.js')",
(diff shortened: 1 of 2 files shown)


kirrg001:
Raised https://jsw.ibm.com/browse/INSTA-30235

abhilash-sivan (approved):

packages/aws-fargate/package.json
```diff
@@ -27,7 +27,7 @@
   "scripts": {
     "audit": "npm audit --omit=dev",
     "node_modules:exists": "mkdir -p node_modules",
-    "test": "mocha $npm_config_watch --config=test/.mocharc.js --reporter mocha-multi-reporters --reporter-options configFile=reporter-config.json --sort $(find test -iname '*test.js')",
+    "test": "mocha $npm_config_watch --config=test/.mocharc.js --sort $(find test -iname '*test.js')",
```
aryamohanan, line 30: Why the changes only aplied for these two packages?

kirrg001: Yes. I could have shared this context 👍 Sry

PR #1629, opened 2025-03-11
refactor: applied batch endpoint filtering rules

aryamohanan:
Improved the `shouldIgnore` function to handle batch processing cases, particularly in `sendBatch` scenarios (e.g., Kafka). The updated logic ensures that spans are ignored only if **all** endpoints in a batch match the ignore configuration.

### Changes:
- Modified the filtering logic to check whether **every** endpoint in the span is listed in `ignoreconfig.endpoints`.
- Added detailed inline comments explaining the special case of batch processing and its effect on span filtering.

### Why This Change?
- Previously, spans containing multiple endpoints could be incorrectly ignored even if few endpoints configured to be traced.
- This enhancement ensures **partial matches do not cause unintended ignores**, improving visibility in distributed tracing.
- It is especially useful in Kafka batch produces, where multiple topics are involved in a single span.

### DIFF
--- a/packages/core/src/util/spanFilter.js
+++ b/packages/core/src/util/spanFilter.js
@@ -77,7 +77,7 @@ function matchEndpoints(span, ignoreconfig) {
    */
   return spanEndpoints.every((/** @type {string} */ endpoint) =>
     ignoreconfig.endpoints.some(
-      e => e?.toLowerCase() === endpoint?.toLowerCase() // Case-insensitive match for endpoint names
+      e => e?.toLowerCase() === endpoint?.toLowerCase()
     )
   );
 }
@@ -153,7 +153,8 @@ function applyFilter(span) {
 /**
  * Parses the endpoints field in span data.
  *
- * In Kafka batch produce scenarios, the `endpoints` field is a comma-separated string
+ * Use Case: In Kafka, when sendBatch is executed, the endpoints field contains a comma-separated string of topic names.
+ * We filtering we treat the endpoints as an array.
  * (e.g., `"test-topic-1,test-topic-2"`).
  *
  * @param {string} endpoints
--- a/packages/core/src/util/spanFilter.js
+++ b/packages/core/src/util/spanFilter.js
@@ -63,15 +63,20 @@ function matchEndpoints(span, ignoreconfig) {
     return true;
   }
   /**
-   * Batch Handling Logic:
-   * 1. If every endpoint in the span is present in the `ignoreconfig.endpoints`, the span is ignored.
+   * General rules:
+   * 1. If every endpoint in the span is present in `ignoreconfig.endpoints`, the span is ignored.
    * 2. If at least one endpoint is not in the ignored list, the span is traced.
-   * This ensures that partial matches do not cause unintended ignores in batch operations.
+   *    This ensures that partial matches do not cause unintended ignores in batch operations.
    *
-   * Case: - Batch Processing in kafka (sendBatch):
+   * Case 1: Single Endpoint (General Case)
+   * - Normally, a span contains only one endpoint, which is represented as an array with a single value.
+   * - If this endpoint exists in `ignoreconfig.endpoints`, the span is ignored.
+   * - Otherwise, the span is traced as usual.
+   *
+   * Case 2: Batch Processing in Kafka (`sendBatch`)
    * - In Kafka, the `sendBatch` operation allows sending messages to multiple topics in a single request.
-   * - We generate a single span for the batch operation and attach all topic names as a comma-separated list.
-   * - If all topics*in the batch appear in `ignoreconfig.endpoints`, the span is ignored.
+   * - A single span is generated for the batch operation, attaching all topic names as a comma-separated list.
+   * - If all topics in the batch appear in `ignoreconfig.endpoints`, the span is ignored.
    * - If at least one topic is not ignored, the span remains traced, ensuring visibility where needed.
    */
   return spanEndpoints.every((/** @type {string} */ endpoint) =>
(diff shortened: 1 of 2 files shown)


packages/core/src/util/spanFilter.js
```diff
@@ -52,14 +52,8 @@ const IGNORABLE_SPAN_TYPES = ['redis', 'dynamodb', 'kafka_placeholder'];
...
-  // TODO:
-  // Normalizing the endpoints will be addressed in future PRs.
-  // For Kafka span data, the endpoints correspond to the service field,
-  // which may contain multiple values.
-  const spanEndpoints = Array.isArray(span.data[span.n]?.endpoints)
-    ? span.data[span.n].endpoints
-    : [span.data[span.n]?.endpoints].filter(Boolean);
-
+  // Parse the endpoints from the span data.
+  const spanEndpoints = parseSpanEndpoints(span.data[span.n]?.endpoints);
```
kirrg001, line 56: +1 this is nice

packages/core/src/util/spanFilter.js
```diff
@@ -142,4 +150,25 @@ function applyFilter(span) {
   return span;
 }
 
+/**
+ * Parses the endpoints field in span data.
+ *
+ * Use Case: In Kafka, when sendBatch is executed, the endpoints field contains a comma-separated string of topic names.
```
kirrg001, line 156: I am missing two clear cases

case 1: Kafka uses single topics by default
case 2: Kafka uses multiple topics for "sendBatch"

packages/core/src/util/spanFilter.js
```diff
@@ -142,4 +150,25 @@ function applyFilter(span) {
...
+ * (e.g., `"test-topic-1,test-topic-2"`).
+ *
+ * @param {string} endpoints
+ */
+function parseSpanEndpoints(endpoints) {
+  if (Array.isArray(endpoints)) {
+    return endpoints;
+  }
+
+  if (typeof endpoints === 'string') {
```
kirrg001, line 167: This is for case1 & 2 or?

kirrg001: What is the case for "blabla,lalala"?
I thought this is an array in the instr

packages/core/src/util/spanFilter.js
```diff
@@ -142,4 +150,25 @@ function applyFilter(span) {
...
+ * Parses the endpoints field in span data.
+ *
+ * Use Case: In Kafka, when sendBatch is executed, the endpoints field contains a comma-separated string of topic names.
+ * For filtering we treat the endpoints as an array.
+ * (e.g., `"test-topic-1,test-topic-2"`).
+ *
+ * @param {string} endpoints
+ */
+function parseSpanEndpoints(endpoints) {
+  if (Array.isArray(endpoints)) {
```
kirrg001, line 163: This is case 2

packages/core/src/util/spanFilter.js
```diff
@@ -68,9 +62,23 @@ function matchEndpoints(span, ignoreconfig) {
   if (ignoreconfig.endpoints.includes('*')) {
     return true;
   }
-
-  return spanEndpoints.some((/** @type {string} */ endpoint) =>
-    ignoreconfig.endpoints.some(e => e?.toLowerCase() === endpoint?.toLowerCase())
+  /**
+   * Batch Handling Logic:
```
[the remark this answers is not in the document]
aryamohanan: updated

packages/core/src/util/spanFilter.js
aryamohanan: updated, may be in future cases arise

packages/core/src/util/spanFilter.js
```diff
@@ -142,4 +147,30 @@ function applyFilter(span) {
...
+ * Parses the endpoints field in span data.
+ *
+ * Case 1: In general, endpoints is a string representing a single endpoint in span.data.
+ *         We convert it into an array.
+ * Case 2: In some cases (e.g., Kafka batch produce), endpoints is a comma-separated string
+ *         (e.g., `"test-topic-1,test-topic-2"`). We split it into an array.
+ * Case 3: If endpoints is an array(future cases). In this case, we return it as-is
+ *         since filtering logic expects an array internally.
+ *
+ * For filtering endpoints is always treated as an array,
```
aryamohanan, line 160:
```suggestion
 * For filtering, endpoints in span always treated as an array,
```

packages/core/src/util/spanFilter.js
```diff
@@ -68,8 +62,19 @@ function matchEndpoints(span, ignoreconfig) {
   if (ignoreconfig.endpoints.includes('*')) {
     return true;
   }
-
-  return spanEndpoints.some((/** @type {string} */ endpoint) =>
+  /**
+   * Batch Handling Logic:
```
kirrg001, line 66: So "every" is only for batching?
We used "some" before. Why did we use "some"?

aryamohanan: Previously, the logic was different—if any topic from the list matched, it would be ignored. Now, we are making it stricter by requiring every endpoint to match.

> Normally we only have one entry in the array or?

Yes, but in case batching we have many.

packages/core/src/util/spanFilter.js
```diff
@@ -68,8 +62,19 @@ function matchEndpoints(span, ignoreconfig) {
...
+   * 1. If every endpoint in the span is present in the `ignoreconfig.endpoints`, the span is ignored.
+   * 2. If at least one endpoint is not in the ignored list, the span is traced.
+   * This ensures that partial matches do not cause unintended ignores in batch operations.
+   *
+   * Case: - Batch Processing in kafka (sendBatch):
```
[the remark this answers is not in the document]
aryamohanan: Updated the comments

kirrg001 (approved):

PR #1633, opened 2025-03-14
fix: resolved TypeError gcStats.on is not a function

abhilash-sivan:


packages/shared-metrics/src/gc.js
```diff
@@ -81,28 +81,30 @@ function actuallyActivate() {
...
-    } else if (type === 15) {
-      minorGcWindow.addPoint(1);
-      majorGcWindow.addPoint(1);
-      incrementalMarkingsWindow.addPoint(1);
-      processWeakCallbacksWindow.addPoint(1);
-    }
-    exports.currentPayload.usedHeapSizeAfterGc = stats.after.usedHeapSize;
-  });
-  startSensing();
+  if (gcStats && typeof gcStats.on === 'function') {
```
kirrg001, line 84: This needs to go to line 72 or?
Because hasBeenActivated should stay false.

abhilash-sivan: Yes makes sense. I will update accordingly. Thanks

kirrg001 (approved):

PR #1642, opened 2025-03-19
fix: improved log for precompiled binary loading failure

aryamohanan:
## Why?

Previously, the failure to load precompiled binaries was logged as a warning, which could give customers the impression that something critical had failed. Since a separate warning log already exists for this scenario, the detailed failure message has been moved to the debug level to avoid redundancy and unnecessary log precedence while still retaining the stack trace for troubleshooting.

Additionally, the log message has been updated to provide better insights into potential causes, such as restricted filesystem access or an unknown system error, improving diagnostic clarity.

Ref: [INSTA-31177](https://jsw.ibm.com/browse/INSTA-31177)

### DIFF
--- a/packages/shared-metrics/src/gc.js
+++ b/packages/shared-metrics/src/gc.js
@@ -46,7 +46,7 @@ exports.init = function init() {
     nativeModuleName: 'gcstats.js',
     moduleRoot: path.join(__dirname, '..'),
     message:
-      'Could not load gcstats.js. Garbage collection information will not be available in Instana for this ' +
+      'Could not load gcstats.js. Garbage collection information will not be available for this ' +
       'application. This typically occurs when native add-ons fail to build during module installation. ' +
       'However, general tracing functionality is unaffected. Enable debug logging for more details. ' +
       'For more information on native add-on, visit: ' +
--- a/packages/shared-metrics/src/gc.js
+++ b/packages/shared-metrics/src/gc.js
@@ -48,8 +48,8 @@ exports.init = function init() {
     message:
       'Could not load gcstats.js. Garbage collection information will not be available for this ' +
       'application. ' +
-      'However, general tracing functionality is unaffected. Enable debug logging for more details. ' +
-      'For more information on native add-on, visit: ' +
+      "However, general tracing functionality remains unaffected. Enable debug logs with 'INSTANA_DEBUG=true' " +
+      'for more details about the error. For more information on native add-ons, visit: ' +
       'https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-collector-installation#native-add-ons'
   });
 
--- a/packages/shared-metrics/src/gc.js
+++ b/packages/shared-metrics/src/gc.js
@@ -47,7 +47,7 @@ exports.init = function init() {
     moduleRoot: path.join(__dirname, '..'),
     message:
       'Could not load gcstats.js. Garbage collection information will not be available for this ' +
-      'application. This typically occurs when native add-ons fail to build during module installation. ' +
+      'application. ' +
       'However, general tracing functionality is unaffected. Enable debug logging for more details. ' +
       'For more information on native add-on, visit: ' +
       'https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-collector-installation#native-add-ons'
--- a/packages/shared-metrics/src/gc.js
+++ b/packages/shared-metrics/src/gc.js
@@ -46,10 +46,10 @@ exports.init = function init() {
     nativeModuleName: 'gcstats.js',
     moduleRoot: path.join(__dirname, '..'),
     message:
-      'Could not load gcstats.js. You will not be able to see GC information in Instana for this application. This ' +
-      'typically occurs when native addons could not be built during module installation (npm install/yarn) or when ' +
-      'npm install --no-optional or yarn --ignore-optional have been used to install dependencies. See the ' +
-      'instructions to learn more about the requirements of the collector: ' +
+      'Could not load gcstats.js. You will not be able to see GC information in Instana for this application. ' +
+      'This typically occurs when native add-ons could not be built during module installation (npm install/yarn) or ' +
+      'when npm install --no-optional, yarn --ignore-optional, or the AutoTrace webhook was used to install ' +
+      'dependencies. See the instructions to learn more about the requirements of the collector: ' +
       'https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-collector-installation#native-add-ons'
   });
 
--- a/packages/shared-metrics/src/gc.js
+++ b/packages/shared-metrics/src/gc.js
@@ -46,10 +46,10 @@ exports.init = function init() {
     nativeModuleName: 'gcstats.js',
     moduleRoot: path.join(__dirname, '..'),
     message:
-      'Could not load gcstats.js. You will not be able to see GC information in Instana for this application. ' +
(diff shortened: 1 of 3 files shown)


packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -219,9 +219,18 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
...
-            logger.warn(
-              `Copying the precompiled build for ${opts.nativeModuleName} ${label} failed. 
-              ${error?.message} ${error?.stack}`
+            // The log triggered when the Instana Node.js collector fails to load a precompiled native module.
+            // The collector first attempts to load the module directly; if that fails, it extracts a precompiled
+            // version from the instrumentation image, which is currently compiled only for Node.js v21.
+            // If the application runs a different Node.js version and the filesystem is read-only, extraction may fail,
+            // preventing collection of certain telemetry data (garbage collection and event loop stats).
+            // TODO: Known issue tracked under INSTA-6673.
+            logger.debug(
```
aryamohanan, line 228: This is how the log look like previously(without enabling debug):

> Copying the precompiled build for gcstats.js (linux/x64/glibc/ABI 115) failed. [Error: ENOENT: no such file or directory, lstat '/tmp/gcstats.js'] {
 errno: -2,
 code: 'ENOENT',
 syscall: 'lstat',
 path: '/tmp/gcstats.js'
}

> Could not load gcstats.js. You will not be able to see GC information in Instana for this application. This typically occurs when native addons could not be built during module installation (npm install/yarn) or when npm install --no-optional or yarn --ignore-optional have been used to install dependencies.

kirrg001: Can we please add two logs?

1. A small warning.

e.g. (Engl. not perfect, pls correct)

> Could not load gcstats.js. Metrics BLA wont be available. This does not affect general tracing. Please enable debug log to see more information.

2. A debug log

This will contain more information.

kirrg001 (commented):
Suggestion

> fix: improved log for precompiled binary loading failure #1642

packages/shared-metrics/src/util/nativeModuleRetry.js
```diff
@@ -219,9 +219,17 @@ function copyPrecompiled(opts, loaderEmitter, callback) {
...
+            // The log triggered when the Instana Node.js collector fails to load a precompiled native module.
+            // The collector first attempts to load the module directly; if that fails, it extracts a precompiled
+            // version from the instrumentation image, which is currently compiled only for Node.js v21.
+            // If the application runs a different Node.js version and the filesystem is read-only, extraction may fail,
+            // preventing collection of certain telemetry data (garbage collection and event loop stats).
+            // TODO: Fix the issue tracked under INSTA-6673. Ensure prebuilt binaries are available for the
+            // corresponding Node.js version.
+            logger.debug(
+              `Failed to load precompiled build for ${opts.nativeModuleName} ${label}. ` +
+                'Precompiled binary extraction was blocked due to a read-only filesystem and an unavailable ' +
```
kirrg001, line 231: I don't think its always caused by "read-only" filesystems or?

kirrg001: Can be any cause.

packages/shared-metrics/src/gc.js
```diff
@@ -46,10 +46,10 @@ exports.init = function init() {
     nativeModuleName: 'gcstats.js',
     moduleRoot: path.join(__dirname, '..'),
     message:
-      'Could not load gcstats.js. You will not be able to see GC information in Instana for this application. This ' +
-      'typically occurs when native addons could not be built during module installation (npm install/yarn) or when ' +
-      'npm install --no-optional or yarn --ignore-optional have been used to install dependencies. See the ' +
-      'instructions to learn more about the requirements of the collector: ' +
+      'Could not load gcstats.js. You will not be able to see GC information in Instana for this application. ' +
+      'This typically occurs when native add-ons could not be built during module installation (npm install/yarn) or ' +
```
kirrg001, line 50: Instead of these details, can we just say "Tracing is funcitoning as normal. Please enable debug log to see more information?

packages/shared-metrics/src/gc.js
```diff
@@ -46,10 +46,10 @@ exports.init = function init() {
     nativeModuleName: 'gcstats.js',
     moduleRoot: path.join(__dirname, '..'),
     message:
-      'Could not load gcstats.js. You will not be able to see GC information in Instana for this application. This ' +
-      'typically occurs when native addons could not be built during module installation (npm install/yarn) or when ' +
-      'npm install --no-optional or yarn --ignore-optional have been used to install dependencies. See the ' +
-      'instructions to learn more about the requirements of the collector: ' +
+      'Could not load gcstats.js. You will not be able to see GC information in Instana for this application. ' +
```
kirrg001, line 49: GC = Garbage. Lets be specific. Not everyone knows straight what GC is.

packages/shared-metrics/src/gc.js
aryamohanan: updated

packages/shared-metrics/src/gc.js
aryamohanan: agree

packages/shared-metrics/src/util/nativeModuleRetry.js
aryamohanan: updated the sentence

packages/shared-metrics/src/gc.js
```diff
@@ -46,10 +46,10 @@ exports.init = function init() {
     nativeModuleName: 'gcstats.js',
     moduleRoot: path.join(__dirname, '..'),
     message:
-      'Could not load gcstats.js. You will not be able to see GC information in Instana for this application. This ' +
-      'typically occurs when native addons could not be built during module installation (npm install/yarn) or when ' +
-      'npm install --no-optional or yarn --ignore-optional have been used to install dependencies. See the ' +
-      'instructions to learn more about the requirements of the collector: ' +
+      'Could not load gcstats.js. Garbage collection information will not be available in Instana for this ' +
+      'application. This typically occurs when native add-ons fail to build during module installation. ' +
```
kirrg001, line 50:
```suggestion
      'application. ' +
```

kirrg001 (approved):
Thanks a lot! <3

packages/shared-metrics/src/gc.js
```diff
@@ -46,10 +46,10 @@ exports.init = function init() {
     nativeModuleName: 'gcstats.js',
     moduleRoot: path.join(__dirname, '..'),
     message:
-      'Could not load gcstats.js. You will not be able to see GC information in Instana for this application. This ' +
-      'typically occurs when native addons could not be built during module installation (npm install/yarn) or when ' +
-      'npm install --no-optional or yarn --ignore-optional have been used to install dependencies. See the ' +
-      'instructions to learn more about the requirements of the collector: ' +
+      'Could not load gcstats.js. Garbage collection information will not be available for this ' +
+      'application. ' +
+      'However, general tracing functionality is unaffected. Enable debug logging for more details. ' +
```
kirrg001, line 51: Wondering if we should say

> Enable debug logs with `INSTANA_DEBUG=true` for more details about the error.

aryamohanan: Yes, specifying INSTANA_DEBUG=true explicitly will be more helpful

PR #1650, opened 2025-03-25
build: bumped express v5.0.0-beta.3 to v5.0.1

aryamohanan:
This PR is intended to add support for the latest express v5 version [5.0.1] (https://www.npmjs.com/package/express/v/5.0.1).

We have added support for Express v5-beta a while ago. Now that v5.0.1 is available, we have updated the test to run against v5 instead of v5-beta, while the test environment currently runs with v4. Once Express is officially released with the latest tag, we will need to update the test environment to use Express v5 as the default.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -89,7 +89,7 @@
         "eslint-plugin-mocha": "^10.1.0",
         "eslint-plugin-monorepo-cop": "^1.0.2",
         "express": "^4.21.2",
-        "express-v5-beta": "npm:express@^5.0.0-beta.3",
+        "express-v5": "npm:express@^5.0.1",
         "express-winston": "^4.2.0",
         "fastify": "^5.2.1",
         "fastify-v3": "npm:fastify@^3.21.5",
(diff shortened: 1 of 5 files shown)


kirrg001 (changes requested):
If you would like to release this as the official Express v5 support, you have to go through the whole [Express v5 changelog](https://github.com/expressjs/express/releases/tag/v5.0.0). We did not do that yet. We only added a minimal beta support for Express v5 in collector v3.

kirrg001:
So I'd suggest to do this in two steps. This PR is only a build update.

kirrg001 (approved):

kirrg001:
The currency report still points to v4. How much effort does it take to make Express v5 the main dependency in our pkg.json? Any alternative solutions?

aryamohanan:
> The currency report still points to v4. How much effort does it take to make Express v5 the main dependency in our pkg.json? Any alternative solutions?

IMO, once Express v5 is officially marked as the latest version, we can update package.json accordingly. In this way, express will point to v5 (latest), while express-v4 will explicitly reference Express v4 (npm:express@^4.2x.x), eliminating the need to update the currencies JSON.

kirrg001:
I can't follow. 

> we can update package.json accordingly. 

Yes, but it won't resolve what I said

> The currency report still points to v4. How much effort does it take to make Express v5 the main dependency in our pkg.json? Any alternative solutions?


> In this way, express will point to v5 (latest), while express-v4 will explicitly reference Express v4 (npm:express@^4.2x.x), eliminating the need to update the currencies JSON.

No. The package.json points to v4 because of the missing test env refactoring. The currency report won't be updated.

kirrg001:
@abhilash-sivan picks up

> No. The package.json points to v4 because of the missing test env refactoring. The currency report won't be updated.
> you have to go through the whole [Express v5 changelog](https://github.com/expressjs/express/releases/tag/v5.0.0).

PR #1652, opened 2025-03-26
feat(kafka): added option to disable downstream suppression for ignored endpoints

aryamohanan:
By default, suppression is enabled for ignored endpoints, preventing downstream traces from being recorded. However, in some cases, this behavior may need to be disabled. The environment variable `INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION` provides flexibility to control this suppression.

Since certain traces are already ignored, trace correlation is broken regardless. This makes suppression a configurable option without additional impact.

### Usage

To disable suppression for ignored endpoints, set the following environment variable:

```bash
export INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION=true
```

### DIFF
--- a/packages/collector/test/tracing/messaging/kafkajs/test.js
+++ b/packages/collector/test/tracing/messaging/kafkajs/test.js
@@ -1102,88 +1102,180 @@ mochaSuiteFn('tracing/kafkajs', function () {
     });
 
     describe('when downstream suppression is disabled via INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION', function () {
-      before(async () => {
-        consumerControls = new ProcessControls({
-          appPath: path.join(__dirname, 'consumer'),
-          useGlobalAgent: true
-        });
-
-        producerControls = new ProcessControls({
-          appPath: path.join(__dirname, 'producer'),
-          useGlobalAgent: true,
-          env: {
-            // basic ignoring config for send
-            INSTANA_IGNORE_ENDPOINTS: 'kafka:send',
-            INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION: true
-          }
-        });
-
-        await consumerControls.startAndWaitForAgentConnection();
-        await producerControls.startAndWaitForAgentConnection();
-      });
-
-      beforeEach(async () => {
-        await agentControls.clearReceivedTraceData();
-      });
-
-      after(async () => {
-        await producerControls.stop();
-        await consumerControls.stop();
-      });
-
-      it('should ignore only the kafka call send and should trace downstream calls', async () => {
-        const message = {
-          key: 'someKey',
-          value: 'someMessage'
-        };
-        await producerControls.sendRequest({
-          method: 'POST',
-          path: '/send-messages',
-          simple: true,
-          body: JSON.stringify(message),
-          headers: {
-            'Content-Type': 'application/json'
-          }
-        });
-
-        await consumerControls.sendRequest({
-          method: 'GET',
-          path: '/health',
-          simple: true
-        });
-
-        await delay(200);
-        await retry(async () => {
-          const spans = await agentControls.getSpans();
-
-          // 2 x HTTP server (1 x consumer, 1 x producer)
-          // 3 x HTTP client (1 x producer)(2 x consumer)
-          // 2 x Kafka consume span (consumer)
-          expect(spans).to.have.lengthOf(7);
-
-          // Flow: HTTP entry (producer) (traced)
-          //       ├── Kafka Produce (ignored)
-          //       │      └── Kafka Consume  (traced) → HTTP (traced)
-          //       └── HTTP exit (traced)
-          //
-          //      HTTP entry (consumer) (traced)
-          const kafkaConsumerSpan = spans.find(span => span.n === 'kafka' && span.k === 1);
-          const producerHttpSpan = spans.find(
-            span => span.n === 'node.http.server' && span.k === 1 && span.data.http.url === '/send-messages'
-          );
-          const producerHttpExitSpan = spans.find(span => span.n === 'node.http.client' && span.k === 2);
-          const consumerHttpSpan = spans.find(
-            span => span.n === 'node.http.server' && span.k === 1 && span.data.http.url === '/health'
-          );
-
-          expect(kafkaConsumerSpan).to.exist;
-          expect(kafkaConsumerSpan.data.kafka).to.include({
-            service: 'test-topic-1',
-            access: 'consume'
+      describe('when send method is ignored', function () {
+        before(async () => {
+          consumerControls = new ProcessControls({
+            appPath: path.join(__dirname, 'consumer'),
+            useGlobalAgent: true
           });
-          expect(producerHttpSpan).to.exist;
-          expect(producerHttpExitSpan).to.exist;
-          expect(consumerHttpSpan).to.exist;
+
+          producerControls = new ProcessControls({
+            appPath: path.join(__dirname, 'producer'),
+            useGlobalAgent: true,
+            env: {
+              // basic ignoring config for send
+              INSTANA_IGNORE_ENDPOINTS: 'kafka:send',
+              INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION: true
+            }
+          });
+
+          await consumerControls.startAndWaitForAgentConnection();
+          await producerControls.startAndWaitForAgentConnection();
+        });
+
+        beforeEach(async () => {
+          await agentControls.clearReceivedTraceData();
+        });
+
+        after(async () => {
+          await producerControls.stop();
+          await consumerControls.stop();
+        });
+
+        it('should ignore only the kafka call send and should trace downstream calls', async () => {
+          const message = {
+            key: 'someKey',
+            value: 'someMessage'
+          };
+          await producerControls.sendRequest({
+            method: 'POST',
+            path: '/send-messages',
+            simple: true,
+            body: JSON.stringify(message),
+            headers: {
+              'Content-Type': 'application/json'
+            }
+          });
+
+          await consumerControls.sendRequest({
+            method: 'GET',
+            path: '/health',
+            simple: true
+          });
+
+          await delay(200);
+          await retry(async () => {
+            const spans = await agentControls.getSpans();
+
+            // 2 x HTTP server (1 x consumer, 1 x producer)
+            // 3 x HTTP client (1 x producer)(2 x consumer)
+            // 2 x Kafka consume span (consumer)
+            expect(spans).to.have.lengthOf(7);
+
+            // Flow: HTTP entry (producer) (traced)
+            //       ├── Kafka Produce (ignored)
+            //       │      └── Kafka Consume  (traced) → HTTP (traced)
+            //       └── HTTP exit (traced)
+            //
+            //      HTTP entry (consumer) (traced)
+            const kafkaConsumerSpan = spans.find(span => span.n === 'kafka' && span.k === 1);
+            const producerHttpSpan = spans.find(
+              span => span.n === 'node.http.server' && span.k === 1 && span.data.http.url === '/send-messages'
+            );
+            const producerHttpExitSpan = spans.find(span => span.n === 'node.http.client' && span.k === 2);
+            const consumerHttpSpan = spans.find(
+              span => span.n === 'node.http.server' && span.k === 1 && span.data.http.url === '/health'
+            );
+
+            expect(kafkaConsumerSpan).to.exist;
+            expect(kafkaConsumerSpan.data.kafka).to.include({
+              service: 'test-topic-1',
+              access: 'consume'
+            });
+            expect(producerHttpSpan).to.exist;
+            expect(producerHttpExitSpan).to.exist;
+            expect(consumerHttpSpan).to.exist;
+          });
+        });
+      });
+
+      describe('when send and consume methods are ignored', function () {
+        before(async () => {
+          consumerControls = new ProcessControls({
+            appPath: path.join(__dirname, 'consumer'),
+            useGlobalAgent: true,
+            env: {
+              // basic ignoring config for consume
+              INSTANA_IGNORE_ENDPOINTS: 'kafka:consume',
+              INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION: true
+            }
+          });
+
+          producerControls = new ProcessControls({
+            appPath: path.join(__dirname, 'producer'),
+            useGlobalAgent: true,
+            env: {
+              // basic ignoring config for send
+              INSTANA_IGNORE_ENDPOINTS: 'kafka:send',
+              INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION: true
+            }
+          });
+
+          await consumerControls.startAndWaitForAgentConnection();
+          await producerControls.startAndWaitForAgentConnection();
+        });
+
+        beforeEach(async () => {
+          await agentControls.clearReceivedTraceData();
+        });
+
+        after(async () => {
+          await producerControls.stop();
+          await consumerControls.stop();
+        });
+
+        it('should ignore send and consume spans while trace downstream HTTP calls', async () => {
+          const message = {
+            key: 'someKey',
+            value: 'someMessage'
+          };
+          await producerControls.sendRequest({
+            method: 'POST',
+            path: '/send-messages',
+            simple: true,
+            body: JSON.stringify(message),
+            headers: {
+              'Content-Type': 'application/json'
+            }
+          });
+
+          await consumerControls.sendRequest({
+            method: 'GET',
+            path: '/health',
+            simple: true
+          });
+
+          await delay(200);
+          await retry(async () => {
+            const spans = await agentControls.getSpans();
+
+            // 2 x HTTP server (1 x consumer, 1 x producer)
+            // 3 x HTTP client (1 x producer)(2 x consumer)
+            expect(spans).to.have.lengthOf(5);
+
+            // Flow: HTTP entry (producer) (traced)
+            //       ├── Kafka Produce (ignored)
+            //       │      └── Kafka Consume  (ignored) → HTTP (traced)
+            //       └── HTTP exit (traced)
+            //
+            //      HTTP entry (consumer) (traced)
+            const kafkaConsumerSpan = spans.find(span => span.n === 'kafka' && span.k === 1);
+            const kafkaProducerSpan = spans.find(span => span.n === 'kafka' && span.k === 2);
+            const producerHttpSpan = spans.find(
+              span => span.n === 'node.http.server' && span.k === 1 && span.data.http.url === '/send-messages'
+            );
+            const producerHttpExitSpan = spans.find(span => span.n === 'node.http.client' && span.k === 2);
+            const consumerHttpSpan = spans.find(
+              span => span.n === 'node.http.server' && span.k === 1 && span.data.http.url === '/health'
+            );
+
+            expect(kafkaConsumerSpan).not.exist;
+            expect(kafkaProducerSpan).not.exist;
+
+            expect(producerHttpSpan).to.exist;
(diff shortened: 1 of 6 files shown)


packages/core/src/tracing/cls.js
```diff
@@ -142,9 +145,9 @@ class InstanaSpan {
       writable: true,
       enumerable: false
     });
-    // Indicates whether the span is ignored.
-    // If true, the span will be excluded from tracing and not propagated.
-    Object.defineProperty(this, 'isIgnored', {
+    // Indicates whether the downstream call should be suppressed.
+    // If true, suppression should be applied.
+    Object.defineProperty(this, 'shouldSupress', {
```
aryamohanan, line 150: Renamed the property to better reflect its purpose. This property is used in instrumentation to determine whether suppression should be applied for downstream calls. If set to `true`, the suppression header is added and propagated.

packages/core/src/util/normalizeConfig.js
```diff
@@ -745,3 +748,18 @@ function normalizeIgnoreEndpoints(config) {
     return;
   }
 }
+
+/**
+ * @param {InstanaConfig} config
+ */
+function normalizeDisableSupression(config) {
+  if (process.env['INSTANA_DISABLE_SUPRESSION'] === 'true') {
```
aryamohanan, line 756: Named the environment variable `INSTANA_DISABLE_SUPPRESSION` to clearly indicate its purpose. While it provides an option to disable downstream suppression specifically for ignoring endpoints, I chose a more generic name to keep it flexible. If you have any suggestions for a better name, feel free to share.

packages/core/src/tracing/cls.js
kirrg001: But from the instrumentation perspective, the attribute name is right.

Kafkajs
span = startSpan
Kafka does not know which span it is.

span.isIgnored "true" means its an ignored span (we added this to avoid class instance checks)
span.isIgnored "false" means its a normal span

aryamohanan: That also make sense!

packages/core/src/util/normalizeConfig.js
```diff
@@ -745,3 +748,18 @@ function normalizeIgnoreEndpoints(config) {
     return;
   }
 }
+
+/**
+ * @param {InstanaConfig} config
+ */
+function normalizeDisableSupression(config) {
+  if (process.env['INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION'] === 'true') {
```
abhilash-sivan, line 756: suggestion:

How about INSTANA_IGNORE_ENDPOINTS_DISABLE_DOWNSTREAM_SUPPRESSION ?

kirrg001: We discussed this yesterday and we think this env variable is simply too long.
Suppression in the context of ignoring endpoints is always downstream. We decided to drop the word "downstream".

packages/core/src/util/normalizeConfig.js
```diff
@@ -243,6 +245,7 @@ function normalizeTracingConfig(config) {
   normalizeTracingKafka(config);
   normalizeAllowRootExitSpan(config);
   normalizeIgnoreEndpoints(config);
+  normalizeDisableSupression(config);
```
kirrg001, line 248: This fn needs to be renamed as well.

packages/core/src/util/normalizeConfig.js
```diff
@@ -745,3 +748,18 @@ function normalizeIgnoreEndpoints(config) {
...
   }
 }
+
+/**
+ * @param {InstanaConfig} config
+ */
+function normalizeDisableSupression(config) {
+  if (process.env['INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION'] === 'true') {
+    logger.info(
+      'Disabling supression as it is explicitly disabled via environment variable INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION.'
```
kirrg001, line 758:
```suggestion
      'Disabling downstream suppression for ignoring endpoints feature as it is explicitly disabled via environment variable "INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION".'
```

Or similar.

packages/core/src/tracing/cls.js
```diff
@@ -241,7 +244,10 @@ class InstanaIgnoredSpan extends InstanaSpan {
    */
   constructor(name, data) {
     super(name, data);
-    this.isIgnored = true;
+
+    // By default, downstream suppression for ignoring endpoints is enabled.
+    // If the environment variable `INSTANA_DISABLE_SUPPRESSION` is set, should not suppress the downstream calls.
```
kirrg001, line 249:
```suggestion
    // If the environment variable `INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION` is set, should not suppress the downstream calls.
```

packages/core/src/tracing/cls.js
```diff
@@ -439,7 +445,9 @@ function setIgnoredSpan({ spanName, kind, traceId, parentId, data = {} }) {
 
     // For entry spans, we need to retain suppression information to ensure that
     // tracing is suppressed for all internal (!) subsequent outgoing (exit) calls.
-    setTracingLevel('0');
+    // By default, downstream suppression is enabled.
+    // If the environment variable `INSTANA_DISABLE_SUPPRESSION` is set, should not suppress the downstream calls.
```
kirrg001, line 449:
```suggestion
    // If the environment variable `INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION` is set, should not suppress the downstream calls.
```

packages/core/src/tracing/cls.js
```diff
@@ -241,7 +244,10 @@ class InstanaIgnoredSpan extends InstanaSpan {
    */
   constructor(name, data) {
     super(name, data);
-    this.isIgnored = true;
+
+    // By default, downstream suppression for ignoring endpoints is enabled.
+    // If the environment variable `INSTANA_DISABLE_SUPPRESSION` is set, should not suppress the downstream calls.
+    this.isIgnored = !ignoreEndpointsDisableSuppression;
```
kirrg001, line 250: issue: We cannot reuse this attribute. Please use a new attribute.

> this.isIgnored
> this.suppressDownstream

kirrg001: Its too confusing that 

s = new InstanaIgnoredSpan
s.isIgnored === suddenly true

The readability of the code is decreased with this change.

packages/core/src/tracing/cls.js
```diff
@@ -149,6 +152,14 @@ class InstanaSpan {
       writable: true,
       enumerable: false
     });
+    // Indicates whether downstream suppression should be propagated.
```
kirrg001, line 155: We need a bit more context here I think. And maybe some more thinking.

So any instantiation of `InstanaSpan` outputs

> instanaSpan.shouldSuppressDownstream = true

But right now the `InstanaSpan` does not suppress downstream, right?

packages/core/src/tracing/cls.js
aryamohanan: Discussed and updated

packages/core/src/tracing/cls.js
```diff
@@ -439,7 +460,10 @@ function setIgnoredSpan({ spanName, kind, traceId, parentId, data = {} }) {
 
     // For entry spans, we need to retain suppression information to ensure that
     // tracing is suppressed for all internal (!) subsequent outgoing (exit) calls.
-    setTracingLevel('0');
+    // By default, downstream suppression is enabled.
+    // If the environment variable `INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION is set,
+    // should not suppress the downstream calls.
+    if (!ignoreEndpointsDownStreamSuppression) setTracingLevel('0');
```
kirrg001, line 466: Why do we not use `span.shouldSuppressDownstream` here?

aryamohanan: Yes, I think we can use the it

packages/core/src/tracing/cls.js
```diff
@@ -29,6 +29,8 @@ let serviceName;
 let processIdentityProvider = null;
 /** @type {Boolean} */
 let allowRootExitSpan;
+/** @type {Boolean} */
+let ignoreEndpointsDownStreamSuppression = true;
```
kirrg001, line 33: The default value in normalizeConfig is false. Here it is true.
This looks wrong to me.

Why do we need a second default value at all?

packages/core/src/tracing/cls.js
```diff
@@ -149,6 +152,19 @@ class InstanaSpan {
       writable: true,
       enumerable: false
     });
+    // This property was introduced as part of the ignoring endpoint feature
+    // and determines whether suppression should be propagated downstream.
+    // By default, suppression is disabled.
+    // This property currently applicable only for ignoring endpoints.
```
kirrg001, line 158:
```suggestion
    // This property "shouldSuppressDownstream" currently applicable only for ignoring endpoints.
```

packages/core/src/tracing/cls.js
```diff
@@ -149,6 +152,19 @@ class InstanaSpan {
...
+    // When a span is ignored, suppression is automatically enabled (`true`),
```
kirrg001, line 159:
```suggestion
    // When a span should be ignored, downstream suppression is automatically enabled (`true`),
```

packages/core/src/tracing/cls.js
```diff
@@ -149,6 +152,19 @@ class InstanaSpan {
...
+    // propagating suppression headers downstream.
+    // However, if required, suppression propagation can be explicitly disabled
+    // by setting this property to `false`.
```
kirrg001, line 162:
```suggestion
    // via the env variable `INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION`.
```

packages/core/src/tracing/cls.js
kirrg001, line 159:
```suggestion
    // When a span is ignored, downstream suppression is automatically enabled.
```

packages/core/src/tracing/cls.js
```diff
@@ -149,6 +152,19 @@ class InstanaSpan {
       writable: true,
       enumerable: false
     });
+    // This property was introduced as part of the ignoring endpoint feature
+    // and determines whether suppression should be propagated downstream.
+    // By default, suppression is disabled.
+    // This property currently applicable only for ignoring endpoints.
+    // When a span is ignored, suppression is automatically enabled (`true`),
+    // propagating suppression headers downstream.
+    // However, if required, suppression propagation can be explicitly disabled
```
kirrg001, line 161:
```suggestion
    // However, if required, downstream suppression propagation can be explicitly disabled
```

packages/core/src/tracing/cls.js
```diff
@@ -439,7 +460,10 @@ function setIgnoredSpan({ spanName, kind, traceId, parentId, data = {} }) {
 
     // For entry spans, we need to retain suppression information to ensure that
     // tracing is suppressed for all internal (!) subsequent outgoing (exit) calls.
-    setTracingLevel('0');
+    // By default, downstream suppression is enabled.
```
kirrg001, line 463:
```suggestion
    // By default, downstream suppression for ignored spans is enabled.
```

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
```diff
@@ -153,7 +153,12 @@ function instrumentedProduce(ctx, originalProduce, originalArgs) {
 
     span.stack = tracingUtil.getStackTrace(instrumentedProduce, 1);
 
-    originalArgs[6] = setTraceHeaders({ headers: originalArgs[6], span, ignored: span.isIgnored });
+    originalArgs[6] = setTraceHeaders({
+      headers: originalArgs[6],
+      span,
+      ignored: span.isIgnored,
+      suppressDownstream: span.shouldSuppressDownstream
```
kirrg001, line 160: Why do we invent new attribute names?

> **ignored**: span.**isIgnored**,
> **suppressDownstream**: span.**shouldSuppressDownstream**

packages/collector/test/tracing/messaging/kafkajs/test.js
```diff
@@ -1100,6 +1100,92 @@ mochaSuiteFn('tracing/kafkajs', function () {
...
+        await delay(200);
+        await retry(async () => {
+          const spans = await agentControls.getSpans();
+          // 2 x HTTP server (1 x consumer, 1 x producer)
+          // 4 x HTTP client (1 producer)(2 consumer)
+          // 2 x Kafka consume span (consumer)
+          expect(spans).to.have.lengthOf(8);
+
+          // Flow: HTTP entry (producer) (traced)
+          //       ├── Kafka Produce (traced)
```
kirrg001, line 1164: This comment is wrong. Kafka produce is ignored.

packages/collector/test/tracing/messaging/kafkajs/test.js
```diff
@@ -1100,6 +1100,92 @@ mochaSuiteFn('tracing/kafkajs', function () {
...
+          //       │      └── Kafka Consume → HTTP (ignored)
```
kirrg001, line 1165: This comment is wrong. Consume is traced.

packages/collector/test/tracing/messaging/kafkajs/test.js
```diff
@@ -1100,6 +1100,92 @@ mochaSuiteFn('tracing/kafkajs', function () {
...
+          const kafkaProducerSpan = spans.find(span => span.n === 'kafka' && span.k === 2);
+          const producerHttpSpan = spans.find(
+            span => span.n === 'node.http.server' && span.k === 1 && span.data.http.url === '/send-messages'
+          );
+          const producerHttpExitSpan = spans.find(span => span.n === 'node.http.client' && span.k === 2);
+          const consumerHttpSpan = spans.find(
+            span => span.n === 'node.http.server' && span.k === 1 && span.data.http.url === '/health'
+          );
+
+          expect(kafkaProducerSpan).to.exist;
```
kirrg001, line 1178: Why?

> INSTANA_IGNORE_ENDPOINTS: 'kafka:send',

packages/collector/test/tracing/messaging/kafkajs/test.js
aryamohanan: updated

packages/collector/test/tracing/messaging/kafkajs/test.js
aryamohanan: updated

packages/collector/test/tracing/messaging/kafkajs/test.js
aryamohanan: updated the test

packages/core/src/tracing/cls.js
aryamohanan: Renamed the var and removed default value

packages/core/src/tracing/instrumentation/messaging/rdkafka.js
aryamohanan: `isIgnored` simply indicates whether the span is ignored, but it does not determine whether we should suppress the downstream call. The key decision here is based on `shouldSuppressDownstream`, which explicitly tells us whether to suppress the downstream call or not. That’s why we use `suppressDownstream` to represent it more clearly.

aryamohanan: I will omit `    ignored: span.isIgnored,` that is really not required.

packages/collector/test/tracing/messaging/kafkajs/test.js
```diff
@@ -1100,6 +1100,93 @@ mochaSuiteFn('tracing/kafkajs', function () {
...
+          }
+        });
+
+        await consumerControls.sendRequest({
+          method: 'GET',
+          path: '/health',
+          simple: true
+        });
+
+        await delay(200);
```
kirrg001, line 1155: Could you please increase this delay to avoid false positives?
You downgraded from 8 to 7 asserts without modifications or?

packages/collector/test/tracing/messaging/kafkajs/test.js
```diff
@@ -1100,6 +1100,93 @@ mochaSuiteFn('tracing/kafkajs', function () {
...
+          appPath: path.join(__dirname, 'consumer'),
+          useGlobalAgent: true
+        });
+
+        producerControls = new ProcessControls({
+          appPath: path.join(__dirname, 'producer'),
+          useGlobalAgent: true,
+          env: {
+            // basic ignoring config for send
+            INSTANA_IGNORE_ENDPOINTS: 'kafka:send',
```
kirrg001, line 1116: IMO we need a second test for "kafka:consume" single span ignoring. Because these are two different logic units. 

1. Consume receives no suppression and is not ignored.
2. Consume receives no suppression and is ignored as a single span. Should not trigger internal cls logic.

packages/collector/test/tracing/messaging/kafkajs/test.js
aryamohanan: There's no need to add the extra delay. Previously, I had configured the ignoring endpoints in the consumer, which resulted in 8 spans. However, I later corrected it by moving the configuration to the producer.

packages/collector/test/tracing/messaging/kafkajs/test.js
aryamohanan: Added one more test

packages/collector/test/tracing/messaging/kafkajs/test.js
```diff
@@ -1100,6 +1100,185 @@ mochaSuiteFn('tracing/kafkajs', function () {
...
+              INSTANA_IGNORE_ENDPOINTS_DISABLE_SUPPRESSION: true
+            }
+          });
+
+          producerControls = new ProcessControls({
+            appPath: path.join(__dirname, 'producer'),
+            useGlobalAgent: true,
+            env: {
+              // basic ignoring config for send
+              INSTANA_IGNORE_ENDPOINTS: 'kafka:send',
```
kirrg001, line 1209: This needs to be removed?

packages/collector/test/tracing/messaging/kafkajs/test.js
kirrg001: Oki

packages/collector/test/tracing/messaging/kafkajs/test.js
aryamohanan: you mean, the test case with only ignore consume calls + no down stream suppression?

kirrg001: Yes. Only the config in the consumer pls

kirrg001:
Commit suggestion

> feat(kafka): added ability to disable downstream suppression for ignoring endpoints

Or similar. We currently only support this for Kafka.
The env var belongs to the commit body and optionally to the Node.js docs.

packages/core/src/tracing/instrumentation/messaging/kafkaJs.js
```diff
@@ -165,7 +165,11 @@ function instrumentedSendBatch(ctx, originalSendBatch, originalArgs, topicMessag
 
     span.stack = tracingUtil.getStackTrace(instrumentedSend);
     topicMessages.forEach(topicMessage => {
-      setTraceHeaders({ messages: topicMessage.messages, span, ignored: span.isIgnored });
+      setTraceHeaders({
+        messages: topicMessage.messages,
+        span,
+        suppressDownstream: span.shouldSuppressDownstream
```
kirrg001, line 171: nitpik: Why do we set `span` and then `suppressDownstream` from `span`?

> span,
> suppressDownstream: span.shouldSuppressDownstream

kirrg001: > function setTraceHeaders({ messages, span }) {
> if (span.shouldSuppressDownstream) {

PR #1655, opened 2025-03-27
chore: update tracer report with express v5

abhilash-sivan:


abhilash-sivan:
We have an alternative solution in https://github.com/instana/nodejs/pull/1657

PR #1654, opened 2025-03-27
feat: added support for express v5

abhilash-sivan:
refs [INSTA-12542](https://jsw.ibm.com/browse/INSTA-12542)

This PR sets the new express v5(tagged `next` major release) as latest version for express in our SDK.
The plan is to adapt v5 for the tests upon release. So till then we will keep the tests to run against express v4.

Express released 5.1.0:

github:  https://github.com/expressjs/express/releases/tag/v5.1.0
see npm:  https://www.npmjs.com/package/express?activeTab=versions

PR #1657, opened 2025-03-28
chore: bump express version in currency report to v5

abhilash-sivan:
This PR bumps the latest supported version in currency report to v5.0.1 statically.
We need to keep an eye on [Express releases](https://www.npmjs.com/package/express?activeTab=versions) to keep this updated until v5 becomes `latest` in NPM.

This static change will ensure that we get the minor updates for v4(latest) in timely manner but the currency report will point to the latest major version


kirrg001 (commented):
I prefer to to finish #1654

PR #1664, opened 2025-04-02
chore: included prerelease version in currency report

abhilash-sivan:
beta flag in currencies.json will now include pre-release versions in currency report


kirrg001 (approved):

aryamohanan (approved):
Commit suggestion:

> chore: included prerelease version in currency report #1664

There is a typo in current git commit message

PR #1671, opened 2025-04-04
ci: added support for `npm-version` param in update-dependencies task 

abhilash-sivan:
This PR addresses an issue observed during the currency update where the typeorm package changes were only reflected in the package-lock.json, while package.json remained unchanged. This inconsistency may be caused by an npm bug when dependencies are installed from a non-root directory.

To resolve this, we've upgraded to npm v11, which behaves as expected locally and includes the necessary changes in both files.

Changes:
 - Adds support for an `npm-version` parameter in the `update-dependencies` task.

 - This parameter can now be specified via the Tekton pipeline to control the npm version used during dependency updates.

### DIFF
--- a/bin/currency/update-currencies.js
+++ b/bin/currency/update-currencies.js
@@ -33,6 +33,8 @@ const hasCommits = (branch, cwd) => {
   }
 };
 
+execSync('npm install npm@11', { cwd });
+
 if (!MAJOR_UPDATES_MODE) {
   console.log('Preparing patch/minor updates...');
   execSync('git checkout main', { cwd });
--- a/bin/currency/update-currencies.js
+++ b/bin/currency/update-currencies.js
@@ -33,8 +33,6 @@ const hasCommits = (branch, cwd) => {
   }
 };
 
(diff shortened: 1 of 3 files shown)


bin/currency/update-currencies.js
```diff
@@ -33,6 +33,8 @@ const hasCommits = (branch, cwd) => {
   }
 };
 
+execSync('npm install npm@11', { cwd });
```
aryamohanan, line 36: Why is this change implemented here? Shouldn't it be part of the `update-dependencies` task?  This seems wrong to me.

aryamohanan (changes requested):
You have to mention the params in [currency-bot-pipeline.yaml](https://github.com/instana/nodejs/blob/main/.tekton/pipeline/currency-bot-pipeline.yaml#L142) file as well.

kirrg001:
Please add the context/reason to the PR body 👍

kirrg001 (approved):

aryamohanan (approved):

PR #1675, opened 2025-04-08
feat: added support for @apollo/server

aryamohanan:
This PR removes the use of [apollo-server-express](https://www.npmjs.com/package/apollo-server-express) and migrates the tests to [@apollo/server](https://www.npmjs.com/package/@apollo/server), which is the officially supported package starting from Apollo Server v4.

### Changes Made

- Removed deprecated `apollo-server-express` package.
- Added `@apollo/server`, `@apollo/server/express4`.
- Replaced `server.applyMiddleware({ app })` with:
  ```js
  app.use('/graphql', cors(), expressMiddleware(server));
  ```

### Notes
For reference and detailed migration steps, see the official guide:  

🔗 https://www.apollographql.com/docs/apollo-server/migration#migrate-from-apollo-server-express

ref INSTA-780

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,7 +143,6 @@
     "chai-as-promised": "^7.1.1",
     "cls-hooked": "^4.2.2",
     "commitizen": "^4.2.4",
-    "cors": "^2.8.5",
     "cz-conventional-changelog": "^3.3.0",
     "depcheck": "^1.4.3",
     "dev-null": "^0.1.1",
--- a/package.json
+++ b/package.json
@@ -143,6 +143,7 @@
     "chai-as-promised": "^7.1.1",
     "cls-hooked": "^4.2.2",
     "commitizen": "^4.2.4",
+    "cors": "^2.8.5",
     "cz-conventional-changelog": "^3.3.0",
     "depcheck": "^1.4.3",
     "dev-null": "^0.1.1",
--- a/package.json
+++ b/package.json
@@ -260,10 +260,7 @@
     "prisma": "^6.6.0"
   },
   "overrides": {
-    "stream-shift": "1.0.2",
-    "apollo-server-express": {
-      "express": "^5.1.0"
-    }
+    "stream-shift": "1.0.2"
   },
   "config": {
     "commitizen": {
--- a/package.json
+++ b/package.json
@@ -130,7 +130,6 @@
     "admin-plugin-healthcheck": "^1.1.0",
     "amqplib": "^0.10.7",
     "amqplib-v0": "npm:amqplib@^0.8.0",
-    "apollo-server-express": "^3.13.0",
     "async": "^3.2.6",
     "aws-sdk": "^2.1692.0",
     "basic-auth-parser": "^0.0.2",
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
   },
   "devDependencies": {
     "@apollo/gateway": "^2.10.2",
+    "@apollo/server": "^4.12.0",
     "@apollo/subgraph": "^2.10.2",
     "@aws-sdk/client-dynamodb": "^3.796.0",
     "@aws-sdk/client-dynamodb-v3": "npm:@aws-sdk/client-dynamodb@3.20.0",
--- a/packages/collector/test/tracing/protocols/apollo_subgraph/gateway.js
+++ b/packages/collector/test/tracing/protocols/apollo_subgraph/gateway.js
@@ -14,12 +14,14 @@ require('@instana/core/test/test_util/loadExpressV4');
 
 require('../../../..')();
 
-const { ApolloServer } = require('apollo-server-express');
+const { ApolloServer } = require('@apollo/server');
 const { ApolloGateway, IntrospectAndCompose } = require('@apollo/gateway');
+const { expressMiddleware } = require('@apollo/server/express4');
 const bodyParser = require('body-parser');
 const express = require('express');
 const http = require('http');
 const morgan = require('morgan');
+const cors = require('cors');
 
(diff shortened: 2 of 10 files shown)


packages/collector/test/tracing/protocols/apollo_subgraph/gateway.js
```diff
@@ -47,15 +49,20 @@ const gateway = new ApolloGateway({
...
   })
 });
+
 app.get('/', (req, res) => {
   res.sendStatus(200);
 });
 
 (async () => {
-  const { schema, executor } = await gateway.load();
-  const server = new ApolloServer({ schema, executor });
```
aryamohanan, line 56: Apollo Server v4 handles this automatically with gateway option and no longer supports the executor option. More details [here](https://www.apollographql.com/docs/apollo-server/migration/#executor).

kirrg001:
Do we know why apollo-server-express is not on our [support page? ](https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-supported-libraries-frameworks) We do list Apollo Subgraph and Apollo Federation.

aryamohanan:
> Do we know why apollo-server-express is not on our [support page? ](https://www.ibm.com/docs/en/instana-observability/current?topic=nodejs-supported-libraries-frameworks) We do list Apollo Subgraph and Apollo Federation.


It's likely that it's mainly a helper library for integrating Apollo Server with Express, and not a core part of the architecture itself. Since we list Apollo Subgraph and Apollo Federation, which are the primary components used in our setup, apollo-server-express might have been considered too implementation-specific to include. This is what I think or it was intentionally omitted.

With the migration to Apollo Server v4—apollo no longer depends directly on Express or any specific HTTP framework. Instead, it introduces a core package—@apollo/server . I think it would make sense to update the support page to include @apollo/server  and also currencies.json.

kirrg001:
> I think it would make sense to update the support page to include @apollo/server and also currencies.json.

Agree

packages/collector/test/tracing/protocols/apollo_subgraph/gateway.js
```diff
@@ -47,15 +49,21 @@ const gateway = new ApolloGateway({
...
+  // Removed the manual gateway.load() call and schema/executor extraction - Apollo Server v4 handles this internally
+  const server = new ApolloServer({
+    gateway
+  });
+
   await server.start();
-  server.applyMiddleware({ app });
+
+  app.use('/graphql', cors(), expressMiddleware(server));
```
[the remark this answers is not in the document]
kirrg001: Why do we suddenly need cors module?

kirrg001:
> chore: replaced apollo-server-express with @apollo/server in tests #1675

feat: added support for @apollo/server

package.json
```diff
@@ -75,6 +75,7 @@
   },
   "devDependencies": {
     "@apollo/gateway": "^2.10.1",
+    "@apollo/server": "^4.12.0",
```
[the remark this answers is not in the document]
aryamohanan: Ah it was added after I made the changes. Removed

packages/collector/test/tracing/protocols/apollo_subgraph/gateway.js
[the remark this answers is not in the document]
aryamohanan: I have referenced the same syntax from the official Apollo server page and for our test scenario cors not required. Removed

packages/collector/test/tracing/protocols/apollo_subgraph/gateway.js
```diff
@@ -14,8 +14,9 @@ require('@instana/core/test/test_util/loadExpressV4');
 
 require('../../../..')();
 
-const { ApolloServer } = require('apollo-server-express');
+const { ApolloServer } = require('@apollo/server');
 const { ApolloGateway, IntrospectAndCompose } = require('@apollo/gateway');
+const { expressMiddleware } = require('@apollo/server/express4');
```
kirrg001, line 19: Did you see that there is the `startStandaloneServer`, who does exactly this ^ + http server?

https://github.com/apollographql/apollo-server/tree/8c6579e5b61276b62dc7e30e6fac9a4242e24daa?tab=readme-ov-file#getting-started-standalone-server

https://github.com/apollographql/apollo-server/blob/8c6579e5b61276b62dc7e30e6fac9a4242e24daa/packages/server/src/standalone/index.ts#L44

aryamohanan: Yes, I saw that. The reason for not using the standalone server in this case is due to the migration process from apollo-server-express in Apollo Server v3 to v4.

In Apollo Server v3, we were using the apollo-server-express package to integrate Apollo Server with Express. The standalone server, which is more aligned with how Apollo Server 3’s apollo-server package worked, does not have the same integration with Express as apollo-server-express did in v3.

Since we are migrating from apollo-server-express, we're using the expressMiddleware function. This ensures that we maintain the same functionality and behavior that we had with apollo-server-express, without having to overhaul the architecture to accommodate the standalone server, which is a different approach more suited to a non-Express setup.

See the migration guide https://www.apollographql.com/docs/apollo-server/migration#the-new-apolloserver-package

Do you have any other opinions?

aryamohanan: It seems they are internally using expressMiddleware

> This is not important. We can keep it as it is

Agree. We can continue using the existing design for test now, unless there is a breaking change in the future.

kirrg001 (approved):

abhilash-sivan (approved):

PR #1677, opened 2025-04-10
build: updated node-gyp and node-gyp-build

abhilash-sivan:
Updated the node-gyp and related packages to latest version for getting the latest changes.
see comment: https://github.com/instana/nodejs/pull/1677#discussion_r2036554551

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,7 +207,7 @@
     "nats-v1": "npm:nats@^1.4.12",
     "node-fetch": "^3.3.2",
     "node-fetch-v2": "npm:node-fetch@^2.7.0",
-    "node-gyp": "^9.4.1",
+    "node-gyp": "^11.2.0",
     "node-nats-streaming": "^0.3.2",
     "pg": "^8.15.6",
     "pino": "^9.6.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -140,7 +140,7 @@
         "nats-v1": "npm:nats@^1.4.12",
         "node-fetch": "^3.3.2",
         "node-fetch-v2": "npm:node-fetch@^2.7.0",
-        "node-gyp": "^9.4.1",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -203,16 +203,16 @@
     "nats-v1": "npm:nats@^1.4.12",
     "node-fetch": "^3.3.2",
     "node-fetch-v2": "npm:node-fetch@^2.7.0",
-    "node-gyp": "^9.4.1",
```
abhilash-sivan, line 206: The node-gyp version was kept to `v9.4.1` to support the node v14. 
see: https://github.com/instana/nodejs/commit/210e33366b384f5baee3140858f6ae4d21a1f936

From collector version 4, our minimun node version is 18. So this can be updated to the latest version which supports v18. see: https://github.com/nodejs/node-gyp/blob/8f3cd8b3a157bccd8d7110e7d46a27c2926625cd/package.json#L37

kirrg001: Wondering why they pinned 18.17.0 🤔 

But yes, looks like we can upgrade to v11 👍

package.json
```diff
@@ -203,16 +203,16 @@
...
     "node-nats-streaming": "^0.3.2",
     "pg": "^8.14.1",
     "pino": "^9.6.0",
     "pino-http": "^9.0.0",
     "pino-v8": "npm:pino@^8.21.0",
     "pkg-up": "^3.1.0",
     "portfinder": "^1.0.32",
-    "prebuildify": "^5.0.1",
-    "prebuildify-cross": "^5.1.0",
+    "prebuildify": "^6.0.1",
```
kirrg001, line 214: This needs to be reverted as discussed 👍

kirrg001 (approved):
Commit msg suggestion:

> build: updated node-gyp and node-gyp-build

PR #1680, opened 2025-04-11
ci(serverless): replaced AWS access key

kirrg001:
- replaced access key of user "ci-manage-lambda-layers" on 410797082306
- every Team has its own access key to publish
- background: key rotation, security, decoupling
- cn key coming later


aryamohanan (approved):
LGTM

PR #1684, opened 2025-04-16
build: removed event-loop-stats and gcstats from dev deps

abhilash-sivan:
In the prerelease pipeline, the npm install step failed for multiple reasons. One reason is the failure to install event-loop-stats and gcstats.

The reason is that the C++ builds are never available for rc candidates and are also usually never immediately available when a new release comes out.
The c++ libraries need to first ship the support for the next node version.
We have decided to revert the changes we made in
https://github.com/instana/nodejs/commit/a2fcb0ffe17a468ace5150f97a3839b507b2348f
https://github.com/instana/nodejs/commit/605da7bb9d0a8ce1363a0988a9c03c3c474ebdbd
and to wait till the builds are randomly failing again to search for a better long-term solution

Step by step, resolving each issue resolved in https://github.com/instana/nodejs/pull/1419 to see the action in pipeline


kirrg001:
> chore: remove native modules from devDependencies #1684

build: removed event-loop-stats and gcstats from dev deps

PR #1685, opened 2025-04-16
ci: udpate npm install step for prerelease pipeline

abhilash-sivan:
In the CI, npm install step for prerelease pipeline was giving a false positive.
Updated the logic and order for resolving this issue


kirrg001:
This would require a force of "skip-cache: true" flag in .tekton/listeners/node-pre-release.yaml

kirrg001 (approved):

PR #1686, opened 2025-04-16
chore: fixed npm install failure for autorpofile in prerelease pipelines

abhilash-sivan:
Skipping the installation of autoprofile if it fails because this is an optional dependency and it keeps failing on prerelease pipeline because of the native c++ modules incompatibility


kirrg001:
Autoprofiler installation can **only** fail e.g. when using a Node.js prerelease version or an unsupported Node.js version or if the target arch is not supported. Its okay to add `|| true` because

> this is an optional dependency

anyway.

kirrg001 (approved):

PR #1688, opened 2025-04-16
chore: fixed npm uninstall usages

abhilash-sivan:
`npm uninstall -D` command is removed from npm v7 and above. 
Also `npm uninstall` with `--production` falg is also not valid anymore

see: https://docs.npmjs.com/cli/v11/commands/npm-uninstall

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -185,11 +185,8 @@ The following sections describe how to manage dependencies in practice.
 
 ### Removing A Package Dependency
 
-`npm uninstall -D ${dependency-name}`: Removes a dev dependency from the root package.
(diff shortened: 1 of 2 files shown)


kirrg001:
There are some more usages of `npm uninstall`. Can we fix them as well? 

Commit suggestion:
> chore: fixed npm uninstall usages

abhilash-sivan:
> There are some more usages of npm uninstall

Do you mean usage in packages/aws-lambda/lambdas/bin/create-zip-util ?

We don't have any more usages for `npm uninstall -D` in our codebase. or do we ?

kirrg001:
I saw a usage with --production

kirrg001 (approved):

PR #1690, opened 2025-04-16
chore: fixed native module installation in node v24-rc 

abhilash-sivan:
I noticed that for some local working environments, I was able to install `gcstats` and `event-loop-stats` with the Node v24 rc version.
(Initially, it was failing for me due to a Python version issue.)

So I tried removing package-lock.json and tried to install the said dependencies in our pkg. It was successful. So I did a deep debugging and identified it was failing because of a sub dependency `nan` in our package.json, (which is a [dependency of node-rd-kafka](https://github.com/Blizzard/node-rdkafka/blob/master/package.json))

So I updated it to the latest version, which seems to be a fix for this issue. 

nan changelog: https://github.com/nodejs/nan/blob/main/CHANGELOG.md

Also reverting the changes made to make the prerelease pipeline green here. 

checks:

- [x] Run this branch in prerelease pipeline and verify it works fine


packages/shared-metrics/package.json
```diff
@@ -66,9 +66,7 @@
     "tar": "^6.2.1"
   },
   "devDependencies": {
-    "@types/tar": "^6.1.6"
-  },
-  "optionalDependencies": {
```
abhilash-sivan, line 71: I don't see a use case for keeping this in both devDependencies and optional dependencies!

kirrg001: Why did you move the packages to dev deps? That is not right. These are optional dependencies for our customers.

package.json
```diff
@@ -202,7 +202,7 @@
     "mysql": "^2.18.1",
     "mysql2": "^3.14.0",
     "mysql2-v3114": "npm:mysql2@3.11.4",
-    "nan": "^2.22.0",
+    "nan": "^2.22.2",
```
kirrg001, line 205: > So I updated it to the latest version, which seems to be a fix for this issue.

So the `nan` update fixes what exactly?

abhilash-sivan: It fixed the installation issue with gcstats and event-loop-stats in the node v24

packages/shared-metrics/package.json
abhilash-sivan: Okay, now I understand the use case for adding the packages in both dev and optional dependencies.
It optionally does the installation in NODE_ENV=production (client application) and forces the installation in other environments (development/test).

kirrg001 (approved):

PR #1691, opened 2025-04-16
feat: added support for @elastic/elasticsearch v9

instanacd:
refs https://jsw.ibm.com/browse/INSTA-34379

### DIFF
--- a/.tekton/tasks/test-groups/test-ci-collector-general-task.yaml
+++ b/.tekton/tasks/test-groups/test-ci-collector-general-task.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   sidecars:
     - name: mysql
-      image: mysql:8.0.1
+      image: mysql:8.0.26
       args:
           - "--default-authentication-plugin=mysql_native_password"
       env:
--- a/.tekton/tasks/test-groups/test-ci-collector-tracing-database-task.yaml
+++ b/.tekton/tasks/test-groups/test-ci-collector-tracing-database-task.yaml
@@ -29,7 +29,7 @@ spec:
           periodSeconds: 2
           timeoutSeconds: 60
     - name: mysql
-      image: mysql:8.0.1
+      image: mysql:8.0.26
       args:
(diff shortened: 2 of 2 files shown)


kirrg001:
No priority. 45 days policy. JIRA should come tomorrow.

package.json
```diff
@@ -95,7 +95,7 @@
     "@azure/storage-blob": "^12.27.0",
     "@commitlint/cli": "^14.1.0",
     "@commitlint/config-conventional": "^14.1.0",
-    "@elastic/elasticsearch": "^8.17.1",
+    "@elastic/elasticsearch": "^9.0.0",
```
aryamohanan, line 98: We're running tests against versions 7.17, 7.9, and 9. Is there a reason why version 8 is being skipped?

aryamohanan: > Breaking changes between 8x and 9x are not big enough to run the tests for both v8 and v9

Ah this. But it still confusing

packages/collector/test/tracing/database/elasticsearch/test.js
```diff
@@ -31,6 +33,9 @@ mochaSuiteFn('tracing/elasticsearch', function () {
    *
    * The words "transport" and "api" try to describe the different
    * mechanismn we use in core/src/tracing/instrumentation/database/elasticsearch.js
+   *
+   * Breaking changes between 8x and 9x are not big enough to run the tests for both v8 and v9:
```
kirrg001, line 37: @aryamohanan See ^

aryamohanan (approved):
LGTM

package.json
kirrg001: See https://github.com/instana/nodejs/pull/1691/files#diff-41f67a44abc56f32440f76e4d38cac6e414926a492a11b30460c391e4b2d2903L36-L50

They were large breaking changes in v7.9 and v7.17. Thats why we still test against them. We can drop support for v7 in next major to get rid of it.

Issue #1701, opened 2025-04-23
[Bug]: Cannot read properties of undefined (reading 'debug')


LeTarzan:
### Problem Description

Hi everyone,

After a deploy, ours AWS ECS tasks crashed with status 0 and this error:
```
logger.debug(`Method ${origMethod} of ${origObject} was already wrapped. Unwrapping.`);
^
TypeError: Cannot read properties of undefined (reading 'debug')
```

After some tests, I found out the problem was the instana version. 
We're using the latest version, I tried to use [v4.9.0](https://github.com/instana/nodejs/releases/tag/v4.9.0), but only [v4.6.1](https://github.com/instana/nodejs/releases/tag/v4.6.1) solve the problem.
This may have arisen after the [v4.6.2](https://github.com/instana/nodejs/releases/tag/v4.6.2) update that changed something on log.

### Instana Node.js Tracer Version

latest

### Node.js Version

v18.20.0

### Short, Self Contained Example

We don't use npm/collector directly, only by the Dockerfile.

// summary Dockerfile
```
FROM node:18.20.0-alpine3.19 as production

COPY --from=icr.io/instana/aws-fargate-nodejs:4.6.1 /instana /instana
RUN /instana/setup.sh

ENV NODE_OPTIONS="--require /instana/node_modules/@instana/aws-fargate"
```

// index.ts
```
import {
  SendMessageCommand,
  SendMessageCommandInput,
} from '@aws-sdk/client-sqs';

const buildMessageCommandInput = (
  event,
  queueUrl,
  delaySeconds = Application.DELAY_SECONDS_METRICS_ERROR_QUEUE,
): SendMessageCommandInput => ({
  MessageBody: JSON.stringify(event),
  DelaySeconds: delaySeconds,
  QueueUrl: queueUrl,
});

const buildMessageCommand = (
  input: SendMessageCommandInput,
): SendMessageCommand => new SendMessageCommand(input);

buildMessageCommand(buildMessageCommandInput({ any: 'any'}, 'queueUrl'))
```

### Logs & Error Messages

```
logger.debug(`Method ${origMethod} of ${origObject} was already wrapped. Unwrapping.`);
^
TypeError: Cannot read properties of undefined (reading 'debug')
at exports.wrap (/instana/node_modules/@instana/core/src/tracing/shimmer.js:58:12)
at Object.fn (/instana/node_modules/@instana/core/src/tracing/instrumentation/cloud/aws-sdk/v3/sqs.js:74:15)
[...]

PR #1702, opened 2025-04-23
fix: resolved "Cannot read properties of undefined (reading 'debug')"

kirrg001:
refs #1701

On e.g. AWS Fargate the `preInit` fn is called as early as possible to init the instrumentations.
If **one** instrumentation (any) had a method which was already wrapped, we unwrap it and log a warning. This code would have been executed:
https://github.com/instana/nodejs/blob/v4.11.0/packages/core/src/tracing/shimmer.js#L51-L59

But the logger for `preInit` case was always undefined. This PR fixes it.
This fix is not just for Fargate.

### DIFF
--- a/packages/core/test/tracing/index_test.js
+++ b/packages/core/test/tracing/index_test.js
@@ -81,80 +81,95 @@ mochaSuiteFn('[UNIT] tracing/index', function () {
     sinon.restore();
   });
 
-  it('deactivate instrumentation via config', () => {
-    initAndActivate({ tracing: { disabledTracers: ['grpc', 'kafkajs', 'aws-sdk/v2'] } });
+  describe('preInit', function () {
+    it('should trace on preinit', () => {
+      const logger = testUtils.createFakeLogger();
+      const config = normalizeConfig({}, logger);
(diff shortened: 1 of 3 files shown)


packages/core/test/tracing/index_test.js
```diff
@@ -81,80 +81,95 @@ mochaSuiteFn('[UNIT] tracing/index', function () {
...
-
-    expect(activateStubGrpcJs).to.have.been.called;
-    expect(activateStubKafkaJs).to.have.been.called;
-    expect(activateStubRdKafka).to.have.been.called;
-    expect(activateAwsSdkv2).to.have.been.called;
+  describe('preInit', function () {
+    it('should trace on preinit', () => {
+      const logger = testUtils.createFakeLogger();
+      const config = normalizeConfig({}, logger);
+      tracing.preInit(config);
```
kirrg001, line 88: This has thrown the error, which was reported in #1701

aryamohanan (approved):
LGTM. Kudos for the quick fix❤️

abhilash-sivan (approved):

PR #1707, opened 2025-04-29
chore: updated days behind calculation logic during currency generation

aryamohanan:
Resolved an issue where a lower version (e.g., from an older major) released after the installed version was incorrectly considered the latest.

For example, if 16.10.0 is installed and 15.10.0 is released after it, the logic wrongly chose 15.10.0 instead of skipping to 16.11.0.
This fix ensures we skip such mistakenly published versions based on major version and semver comparison.

See grapgql release pattern for reference: https://www.npmjs.com/package/graphql?activeTab=versions

### DIFF
--- a/bin/currency/utils.js
+++ b/bin/currency/utils.js
@@ -122,10 +122,11 @@ exports.getLatestVersion = ({ pkgName, installedVersion, isBeta }) => {
 };
 
 function filterStableReleases(releaseList) {
+  const unstableReleaseKeyWords = ['alpha', 'beta', 'canary', 'dev', 'experimental', 'rc'];
+
   return Object.fromEntries(
     Object.entries(releaseList).filter(
-      ([version]) =>
-        !version.includes('alpha') && !version.includes('beta') && !version.includes('dev') && !version.includes('rc')
+      ([version]) => !unstableReleaseKeyWords.some(keyword => version.includes(keyword))
     )
   );
 }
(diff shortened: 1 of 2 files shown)


bin/currency/utils.js
```diff
@@ -122,10 +122,11 @@ exports.getLatestVersion = ({ pkgName, installedVersion, isBeta }) => {
 };
 
 function filterStableReleases(releaseList) {
+  const unstableReleaseKeyWords = ['alpha', 'beta', 'canary', 'dev', 'experimental', 'rc'];
```
aryamohanan, line 125: Added support for filtering out canary and experimental versions.These were discovered during the testing.

bin/currency/utils.js
```diff
@@ -136,11 +137,18 @@ function calculateDaysDifference(date1, date2) {
...
-  // CASE: library releases a lower version by mistake
-  if (semver.major(installedVersion) === semver.major(nextVersion) && semver.lt(nextVersion, installedVersion)) {
-    return getNextVersion(versions, installedVersionIndex + 1, nextVersion);
+  const nextIndex = installedVersionIndex + 1;
+  const nextVersion = versions[nextIndex];
+
+  // CASE: Check if the next version is invalid based on two conditions:
+  // 1. The next version is from an older major version.
+  // 2. The next version has the same major version but is lower than the installed version.
+  if (
```
aryamohanan, line 146: Here is the fix, previously, the logic did not check whether the next version was from a older major (e.g., 15.x after 16.x).
This caused incorrect calculations in cases of GraphQL, where 16.10.0 was followed by 15.10.0.

abhilash-sivan (approved):

PR #1710, opened 2025-04-30
feat: added support for redis v5

instanacd:
### Redis v5: Changes in this PR

1. **Connection Termination**
   In Redis v5, the method `client.quit()` (or `client.QUIT`) has been deprecated and replaced by `client.close()`. 

2. **Scan Iterator Behavior**
   * **v4:** Iterators yielded individual elements.
   * **v5:** Iterators now yield entire collections.

3. **Cluster Command Signature Change**
   In Redis v5, the method `cluster.multi().addCommand()` has an updated signature.
   * The second argument is now a boolean (`isReadonly`), and the actual command array is passed as the third argument.

4. **Connection Pool**
   Redis v5 introduces a dedicated `RedisClientPool` class for managing connection pools.
   * **v4:** Pooling was handled internally by the `RedisClient` using an "Isolation Pool".
   * **v5:** Pooling responsibilities are now abstracted into a separate class.

5. **Internal File Structure Changes**
   Redis v5 introduces a new internal structure for the package, requiring updates to how we reference and mock internal modules.

6. **Package Versioning Alignment**
   The `@redis/client` package versioning now aligns with the Redis server version:
   * **v4:** `@redis/client` was versioned as 1.x.
   * **v5:** `@redis/client` is now versioned as 5.x.

### Additional Update

* The Redis Docker image has been updated to the latest version.

### Tasks

- [x] Replace client.quit() with client.close().
- [x] Update  test to support scan iterator handling.
- [x] Adjust cluster.multi().addCommand() to new signature.
- [x] Docker image update to latest
- [x] Update internal imports and mocks for new file structure.
- [x] Support RedisClientPool https://github.com/redis/node-redis/blob/master/docs/pool.md
- [x] Add support for createSentinel() https://github.com/redis/node-redis/blob/master/docs/sentinel.md : see https://github.com/instana/nodejs/pull/1737

### References

Release notes: https://github.com/redis/node-redis/releases/tag/redis%405.0.0
Migration guide: https://github.com/redis/node-redis/blob/master/docs/v4-to-v5.md

Ref: https://jsw.ibm.com/browse/INSTA-35744

### DIFF
--- a/docker-compose-base.yaml
+++ b/docker-compose-base.yaml
@@ -1,7 +1,7 @@
 version: '2'
 services:
   redis:
-    image: redis:7.4
+    image: redis:7.4.3
     ports:
       - 6379:6379
 
--- a/docker-compose-base.yaml
+++ b/docker-compose-base.yaml
@@ -1,7 +1,7 @@
 version: '2'
 services:
   redis:
-    image: redis:5.0.14
+    image: redis:7.4
     ports:
       - 6379:6379
 
--- a/package.json
+++ b/package.json
@@ -226,6 +226,7 @@
     "recursive-copy": "^2.0.14",
     "redis": "^5.0.0",
     "redis-v3": "npm:redis@^3.1.2",
+    "redis-v4": "npm:redis@^4.7.0",
     "restify": "^11.1.0",
     "rimraf": "^3.0.2",
     "semver": "^7.5.4",
--- a/package.json
+++ b/package.json
@@ -111,7 +111,6 @@
     "@opentelemetry/instrumentation": "0.50.0",
     "@opentelemetry/sdk-node": "^0.49.1",
     "@opentelemetry/sdk-trace-base": "^1.22.0",
-    "@redis/client": "^5.0.1",
     "@redis/client-v4": "npm:@redis/client@1.6.0",
     "@types/chai": "^4.3.0",
     "@types/detect-libc": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
     "@opentelemetry/instrumentation": "0.50.0",
     "@opentelemetry/sdk-node": "^0.49.1",
     "@opentelemetry/sdk-trace-base": "^1.22.0",
+    "@redis/client": "^5.0.1",
     "@redis/client-v4": "npm:@redis/client@1.6.0",
     "@types/chai": "^4.3.0",
     "@types/detect-libc": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -224,7 +224,7 @@
     "proxyquire": "^2.1.3",
     "read-pkg-up": "^7.0.1",
     "recursive-copy": "^2.0.14",
-    "redis": "^4.7.0",
+    "redis": "^5.0.0",
     "redis-v3": "npm:redis@^3.1.2",
     "restify": "^11.1.0",
     "rimraf": "^3.0.2",
--- a/package.json
+++ b/package.json
@@ -111,6 +111,8 @@
     "@opentelemetry/instrumentation": "0.50.0",
     "@opentelemetry/sdk-node": "^0.49.1",
     "@opentelemetry/sdk-trace-base": "^1.22.0",
+    "@redis/client": "^5.0.1",
+    "@redis/client-v4": "npm:@redis/client@1.6.0",
     "@types/chai": "^4.3.0",
     "@types/detect-libc": "^1.0.0",
     "@types/event-loop-lag": "^1.0.30",
--- a/package.json
+++ b/package.json
@@ -226,7 +226,7 @@
     "proxyquire": "^2.1.3",
     "read-pkg-up": "^7.0.1",
     "recursive-copy": "^2.0.14",
-    "redis": "^5.0.0",
+    "redis": "^5.0.1",
     "redis-v3": "npm:redis@^3.1.2",
     "redis-v4": "npm:redis@^4.7.0",
     "restify": "^11.1.0",
--- a/packages/collector/test/tracing/database/redis/test.js
+++ b/packages/collector/test/tracing/database/redis/test.js
@@ -749,7 +749,7 @@ const legacyVersion = 'v3';
               // scanIterator not available on cluster.
               if (setupType !== 'cluster') {
                 // eslint-disable-next-line mocha/no-exclusive-tests
-                it.only('must trace scan iterator usage', () =>
+                it('must trace scan iterator usage', () =>
                   controls
                     .sendRequest({
                       method: 'GET',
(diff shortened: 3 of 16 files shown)


aryamohanan:
Let's wait for the  currencyJira card to be created.

aryamohanan:
In v5, Redis moved “Isolation Pool” into RedisClientPool with a createClientPool export—but importing createClientPool throws “Not Found,” so the docs don’t match the package. Opened an issue in redis: https://github.com/redis/node-redis/issues/2936

aryamohanan:
The new sentinel API is not working for me created an issue: https://github.com/redis/node-redis/issues/2962

package.json
```diff
@@ -111,6 +111,7 @@
     "@opentelemetry/instrumentation": "0.50.0",
     "@opentelemetry/sdk-node": "^0.49.1",
     "@opentelemetry/sdk-trace-base": "^1.22.0",
+    "@redis/client-v4": "npm:@redis/client@1.6.0",
```
aryamohanan, line 114: The @redis/client package was introduced with Redis v4 (@redis/client v1). In Redis v5, the versioning of @redis/client was synchronized with Redis, meaning @redis/client v5 corresponds to Redis v5. 

To improve the testing setup, I’ve specified @redis/client v4 (corresponding to Redis v4) instead of v1.

package.json
```diff
@@ -224,8 +225,9 @@
     "proxyquire": "^2.1.3",
     "read-pkg-up": "^7.0.1",
     "recursive-copy": "^2.0.14",
-    "redis": "^4.7.0",
+    "redis": "^5.0.1",
     "redis-v3": "npm:redis@^3.1.2",
+    "redis-v4": "npm:redis@^4.7.0",
```
aryamohanan, line 230: I maintain tests for three versions of Redis, and each of these tests involves changes.

aryamohanan:
As discussed the sentinel support will added in a separate PR

docker-compose-base.yaml
```diff
@@ -1,7 +1,7 @@
 version: '2'
 services:
   redis:
-    image: redis:5.0.14
+    image: redis:7.4.3
```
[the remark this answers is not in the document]
kirrg001: MUA

packages/collector/test/tracing/database/redis/test.js
```diff
@@ -721,8 +771,12 @@ const globalAgent = require('../../../globalAgent');
                             span => expect(span.n).to.equal('node.http.server'),
                             span => expect(span.data.http.method).to.equal('GET')
                           ]);
+                          // NOTE: v5 SCAN iterators yield batches of keys, enabling multi-key commands like MGET.
+                          // See: https://github.com/redis/node-redis/blob/master/docs/v4-to-v5.md#scan-iterators
+                          const expectedSpanCount = redisVersion === 'latest' ? 1 : 4;
```
kirrg001, line 776: QS: Could you please add an assertion for the batch property `span.b`

package.json
kirrg001: >     "@redis/client-v4": "npm:@redis/client@1.6.0",

v4 = 1.6.0?

packages/collector/test/tracing/database/redis/test.js
```diff
@@ -85,6 +87,52 @@ const globalAgent = require('../../../globalAgent');
...
+            if (redisVersion === 'latest') {
+              mochaSuiteFn('When connected via clientpool', function () {
+                globalAgent.setUpCleanUpHooks();
+                let controls;
+
+                before(async () => {
+                  controls = new ProcessControls({
+                    useGlobalAgent: true,
+                    appPath:
+                      redisVersion === legacyVersion
```
kirrg001, line 101: I guess this can be removed?

packages/collector/test/tracing/database/redis/test.js
```diff
@@ -85,6 +87,52 @@ const globalAgent = require('../../../globalAgent');
...
+
+                after(async () => {
+                  await controls.stop();
+                });
+
+                afterEach(async () => {
+                  await controls.clearIpcMessages();
+                });
+
+                it('should trace blocking commands', () => testBlockingCommand(controls, setupType));
```
kirrg001, line 133: I don't understand this test. Blocking commands? 

>                    method: 'POST',
                    path: '/clearkeys'

package.json
aryamohanan: > v4 = 1.6.0?

In v4, the `redis/client` was at version 1.6.x (see [here](https://github.com/redis/node-redis/blob/redis%404.7.1/package.json#L27)). With v5, the client version was updated to v5.0 to align its versioning with Redis itself. For simplicity in testing, it is  labeled as v4, and I have added a comment in the test file.

Additionally, for `redis/client`, after version 1.6, the major version jumped directly to v5, with no official stable releases in between—only prerelease versions were published. See: https://www.npmjs.com/package/@redis/client?activeTab=versions

kirrg001: I can't follow why we have to install this package? Where is it used?

kirrg001: Don't we have to add 

@redis/client: 5.x
@redis/client-4: v1

?

So we test against all legacy package versions and the new releases?

aryamohanan: Yea. Make sense!. @redis/client is a direct dependency of the redis package, so it gets installed automatically along with it. However, for better readability and maintainability, I can explicitly include it as well.

> @redis/client: v5.x
> @redis/client-4: v1.x

kirrg001: We also need it for currency?

packages/core/src/tracing/instrumentation/database/redis.js
```diff
@@ -147,8 +165,29 @@ function instrument(redis) {
...
+      if (poolDescriptor?.configurable) {
+        if (typeof poolDescriptor.get === 'function') {
+          Object.defineProperty(redis, 'createClientPool', {
+            configurable: true,
+            enumerable: true,
+            get() {
+              return creatPoolWrap(poolDescriptor.get.call(this));
+            }
+          });
+        } else if (typeof poolDescriptor.value === 'function') {
```
kirrg001, line 182: I assume this line is only needed to know that its v5?
I was wondering if we need this check at all

aryamohanan: removed

package.json
aryamohanan: added

packages/collector/test/tracing/database/redis/test.js
aryamohanan: In case of scanIterators. It's not a batched span. 'Multi-key' means we're passing multiple keys to a single Redis command.

kirrg001 (approved):
💁‍♀️❤️

PR #1715, opened 2025-05-05
fix(aws-lambda): resolved TypeError when memory is too low

kirrg001:
```{
  "errorType": "TypeError",
  "errorMessage": "Cannot read properties of undefined (reading 'logger')",
  "trace": [
    "TypeError: Cannot read properties of undefined (reading 'logger')",
    "    at warn (/opt/nodejs/node_modules/@instana/serverless/src/logger.js:74:15)",
    "    at shouldUseLambdaExtension (/opt/nodejs/node_modules/@instana/aws-lambda/src/wrapper.js:383:7)",
    "    at init (/opt/nodejs/node_modules/@instana/aws-lambda/src/wrapper.js:259:30)",
    "    at shimmedHandler (/opt/nodejs/node_modules/@instana/aws-lambda/src/wrapper.js:71:26)",
    "    at handler1 (/opt/nodejs/node_modules/@instana/aws-lambda/src/wrapper.js:52:16)",
    "    at Runtime.instanaAutowrapHandler [as handler] (/opt/nodejs/node_modules/instana-aws-lambda-auto-wrap-esm/index.js:24:10)"
  ]
}

### DIFF
--- a/packages/aws-lambda/test/using_api/test.js
+++ b/packages/aws-lambda/test/using_api/test.js
@@ -30,6 +30,9 @@ function prelude(opts) {
   if (opts.instanaEndpointUrlMissing) {
     env.INSTANA_ENDPOINT_URL = '';
   }
+  if (opts.instanaAgentKeyMissing) {
+    env.INSTANA_AGENT_KEY = '';
+  }
   if (opts.instanaAgentKey) {
     env.INSTANA_AGENT_KEY = opts.instanaAgentKey;
   }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/aws-lambda/test/using_api/test.js [hunk 1]
```diff
@@ -354,6 +357,41 @@ describe('Using the API', function () {
     });
   });
 
+  describe('when INSTANA_AGENT_KEY is missing', function () {
```
kirrg001, line 360: FYI added this test as well, because it was not covered.

packages/serverless/src/logger.js
```diff
@@ -62,45 +62,15 @@ class InstanaServerlessLogger {
...
-    }
-
-    this.logger.info.apply(this.logger, arguments);
-  }
-
-  warn() {
-    if (!this.logger.warn) {
-      return;
-    }
+  warn = (...args) => this.logger?.warn?.(...args);
```
kirrg001, line 65: See https://github.com/instana/nodejs/blob/v4.11.1/packages/aws-lambda/src/wrapper.js#L373-L392

When copying a fn for a class, the class itself needs to bind the this context (or the caller).

> logFn = logger.warn

Using the arrow fn here auto-resolves it. It allows to copy the fn to a new variable without lossing the binding.

kirrg001: I will create a new PR for the collector logger soon. Its using a similar logic.

aryamohanan (approved):

PR #1716, opened 2025-05-05
chore: fixed logger class reference

kirrg001:


packages/serverless/src/logger.js
```diff
@@ -45,18 +45,18 @@ class InstanaServerlessLogger {
   }
 
   isInDebugMode() {
-    if (!instanaServerlessLogger.logger) return false;
```
kirrg001, line 48: I dunno what I have smoked back then.

aryamohanan (approved):
LGTM

PR #1724, opened 2025-05-08
ci: fixed the issue with publishing the image for the aws lambda layer in each task

aryamohanan:


kirrg001 (approved):
LGTM

How about we add a tiny comment in the publish-layer script

> NOTE: We currently only publish to x86.....blablabla...
> if [[ -z $SKIP_DOCKER_IMAGE ]]; then

abhilash-sivan (approved):

PR #1725, opened 2025-05-08
build: migrated Kafka from v2 to v3

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-36044

### DIFF
--- a/.tekton/assets/sidecars.json
+++ b/.tekton/assets/sidecars.json
@@ -35,7 +35,7 @@
     },
     {
       "name": "zookeeper",
-      "image": "zookeeper:3.9.1",
+      "image": "confluentinc/cp-zookeeper:7.9.0",
       "readinessProbe": {
         "tcpSocket": {
           "port": 2181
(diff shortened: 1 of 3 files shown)


kirrg001:
kafka-node is not working with Kafka v3.
There was no release since 5y anyway.

packages/collector/test/tracing/messaging/node-rdkafka/consumer.js
```diff
@@ -99,67 +101,66 @@ function setupConsumer() {
...
+      });
+    };
+
+    connect();
 
     _consumer.on('ready', () => {
+      log('Consumer received ready event');
+
       function startConsuming() {
-        _consumer.getMetadata(null, (err, metadata) => {
```
kirrg001, line 106: We use KAFKA_AUTO_CREATE_TOPICS_ENABLE now. No need to check if the topic was created. Only removed this logic thats why the diff is hard to review.

abhilash-sivan: Yes the first test fails

kirrg001: Yeah its a flakyness, which exists on main branch as well. I don't know what this is. 
I remember we disabled the rdkafka tests because of that in the past. This is not related to this branch

kirrg001: There is an issue with `KAFKA_AUTO_CREATE_TOPICS_ENABLE`, it seems its only creating the topic when producing a message. I will look into this. I guess the topic on main is just another random error.

kirrg001: Okay added `kafka-topics` container. There is no other way.
Adding the logic to each test to create the topic is too much effort + will create raise conditions between producer and consumer.

PR #1726, opened 2025-05-08
chore: update precompiled addons script

abhilash-sivan:
All the add-ons were removed, even if we tried to build only for a single version. 
We can manually remove the builds if we want to regenerate anything. Usually we only add new builds so that the removal can be handled manually

### DIFF
--- a/native-dep-packs/rebuild-precompiled-addons.sh
+++ b/native-dep-packs/rebuild-precompiled-addons.sh
@@ -8,7 +8,7 @@ set -eo pipefail
 
 cd `dirname $BASH_SOURCE`
 
-declare -a ABI_VERSIONS=( \
+declare -A ABI_VERSIONS=( \
   ["108"]="18.18.2" \
   ["115"]="20.3.0" \
   ["120"]="21.2.0" \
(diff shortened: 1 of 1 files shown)


native-dep-packs/rebuild-precompiled-addons.sh
```diff
@@ -8,9 +8,7 @@ set -eo pipefail
 
 cd `dirname $BASH_SOURCE`
 
-# Maintenance Note:
-# This should be kept in sync with packages/autoprofile/precompile/build-all-addons.js-> ABI_VERSIONS.
-declare -A ABI_VERSIONS=( \
+declare -a ABI_VERSIONS=( \
   ["108"]="18.18.2" \
```
kirrg001, line 12: Would be really nice if we can add an argument to build a single prebuild version. Must not be part of this PR.

native-dep-packs/rebuild-precompiled-addons.sh
```diff
@@ -31,8 +29,6 @@ LIBC_VARIANTS=( \
 if [[ -z "$BUILD_FOR_MACOS" ]]; then
   source ./build-and-copy-node-modules-linux
 
-  rm -rf ../packages/shared-metrics/addons/linux
```
kirrg001, line 34: I guess its okay to just remove this line because it will override the prebuild anyway.
I was wondering why the developer has used this line in the first place.

abhilash-sivan: Yeah, it overrides anyway.

kirrg001 (approved):

PR #1728, opened 2025-05-09
ci: updated aws Lambda base image to node:20

aryamohanan:
We already dropped support for Node.js 16 aws lambda runtime in the last major release, we missed to update the dockerfile.

Updated the base image in the Dockerfile to public.ecr.aws/lambda/nodejs:22 to align with the latest Node.js version for AWS Lambda.

### DIFF
--- a/packages/aws-lambda/layer/bin/publish-layer.sh
+++ b/packages/aws-lambda/layer/bin/publish-layer.sh
@@ -165,7 +165,15 @@ SUPPORTED_RUNTIMES="nodejs18.x nodejs20.x nodejs22.x"
 # The Node.js version to use for building the Docker image.
 # This should be aligned with one of the supported runtimes above.
 # We're using a development version.
-NODE_VERSION_TO_BUILD="20"
+ROOT_DIR=$(git rev-parse --show-toplevel 2>/dev/null || echo "../../..")
+NVMRC_PATH="$ROOT_DIR/.nvmrc"
+if [[ -f "$NVMRC_PATH" ]]; then
+  NODEJS_DEV_VERSION=$(cat "$NVMRC_PATH")
+  echo "Using Node.js version $NODEJS_DEV_VERSION from .nvmrc for Docker build"
+else
+  echo "Warning: .nvmrc file not found at $NVMRC_PATH, falling back to default Node.js version 20"
+  NODEJS_DEV_VERSION=20 
+fi
 
 echo Will build Lambda layer with name \"$LAYER_NAME\".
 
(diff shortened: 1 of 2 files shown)


kirrg001 (commented):
For what are we using this Dockerfile?
And where do we reference this file?

kirrg001:
Yeah I saw this too, but I wasn't sure. But I can see a `cd ..`, which navigates up to the Dockerfile 👍 

I think we should pass the Node version as an argument.

>   docker build --build-arg NODEJS_VERSION=$NODEJS_VERSION . -t "$DOCKER_IMAGE_NAME:$VERSION"

Dockerfile

```
# see publish-layer.sh
ARG NODEJS_VERSION
FROM public.ecr.aws/lambda/nodejs:${NODEJS_VERSION}
```

This `NODEJS_VERSION` needs to be the version which is supported by our layer

>  --compatible-runtimes nodejs18.x nodejs20.x nodejs22.x \

Either we use our development version or define the version in the publish-layer script.

---

In general: It seems there is a poor support for Lambda layers in a Docker container. Because we only build the image with a specific AWS Lambda Node.js version always for the latest version. Our docs may need to make that clear. See https://www.ibm.com/docs/de/instana-observability/1.0.293?topic=lambda-aws-native-tracing-nodejs#instana-lambda-layer-for-container-based-functions

kirrg001:
I guess its good to do

```sh
supportedVersions="--compatible-runtimes nodejs18.x nodejs20.x nodejs22.x"
dockerNodeVersion="grab the lowest or highest supported node version from $supportedVersions"
```

abhilash-sivan:
+1 for using `development version`

aryamohanan:
As recommended, I have added the development version for the Docker image. Since v18 has already reached its end of life, and v22 officially added AWS support in November 2024, this update aligns with the current standards. Please share your thoughts

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -159,6 +159,14 @@ if [[ -z $NO_PROMPT ]]; then
   done
 fi
 
+# List of supported AWS Lambda Node.js runtimes
+SUPPORTED_RUNTIMES="nodejs18.x nodejs20.x nodejs22.x"
+
+# The Node.js version to use for building the Docker image.
+# This should be aligned with one of the supported runtimes above.
+# We're using a development version.
+NODE_VERSION_TO_BUILD="20"
```
abhilash-sivan, line 168: Can we define the development version in a single constants file and reference it across all packages?
This would help improve maintainability since the version is repeated in multiple places.

P.S.: This might be out of scope for this PR.

aryamohanan: Good suggestion, we can certainly consider doing that.

kirrg001, line 168: Any reason we are not using

>NVMRC_PATH="$ROOT_DIR/.nvmrc"
>NODEJS_VERSION=$(cat "$NVMRC_PATH")

?

aryamohanan: Good suggestion. i never think about it, i was only considering a generic gfile where we can define the version and use every where. reading from .nvmrc file make sense. as we are using dev version of node for images

kirrg001 (approved):
pre-approving

PR #1729, opened 2025-05-09
chore: replaced $npm_config_watch for npm v11

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-36236

- no longer supported: https://github.com/npm/cli/pull/8071


abhilash-sivan (approved):

abhilash-sivan:
suggestion: 

The commit message is a bit confusing - we're actually reducing warn logs by adopting the new -- syntax.

aryamohanan (commented):
The current commit message is a bit unclear. Consider something like: 

> chore: updated watch mode configuration to eliminate warnings in npm v11

or something similiar

aryamohanan (commented):
qs: Should we update [this](https://github.com/instana/nodejs/blob/main/bin/add-test-aliases.js#L51) as well? Currently, we're checking for --watch, so there won't be any warnings ([see](https://github.com/instana/nodejs/blob/main/bin/run-tests.sh#L14)), but it might be good to update this for consistency if needed.

aryamohanan (approved):

kirrg001:
> qs: Should we update [this](https://github.com/instana/nodejs/blob/main/bin/add-test-aliases.js#L51) as well? Currently, we're checking for --watch, so there won't be any warnings ([see](https://github.com/instana/nodejs/blob/main/bin/run-tests.sh#L14)), but it might be good to update this for consistency if needed.

No this is the input argument for our test script.

PR #1731, opened 2025-05-09
feat: added support for  sqs-consumer v12

instanacd:
ref https://jsw.ibm.com/browse/INSTA-36704

### DIFF
--- a/package.json
+++ b/package.json
@@ -234,7 +234,7 @@
     "sinon-chai": "^3.7.0",
     "socket.io": "^4.8.1",
     "socket.io-client": "^4.8.1",
-    "sqs-consumer": "^11.6.0",
+    "sqs-consumer": "^12.0.0",
     "sqs-consumer-v5": "npm:sqs-consumer@^5.7.0",
     "square-calc": "3.2.1",
     "square-calc-v2": "npm:square-calc@2.4.0",
--- a/packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/test.js
+++ b/packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/test.js
@@ -48,6 +48,12 @@ if (!supportedVersion(process.versions.node)) {
   mochaSuiteFn = describe;
 }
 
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -234,8 +234,7 @@
     "sinon-chai": "^3.7.0",
     "socket.io": "^4.8.1",
     "socket.io-client": "^4.8.1",
-    "sqs-consumer": "^11.6.0",
-    "sqs-consumer-v5": "npm:sqs-consumer@^5.7.0",
```
aryamohanan, line 238: I'm not sure why we were testing against this very old version v5. I've removed it, since it was released on February 2019, and I couldn't find any justification for including it in our tests. Additionally, I'm not testing against the previous version (v11) either, as there are no significant changes between v11 and v12—aside from dropping support for Node.js v18.

aryamohanan:
Let's proceed once the currency jira ticket for sqs-consumer has been created

package.json
kirrg001: Its because of aws v2 and aws v3.
v2 uses sqs-consumer-v5

aryamohanan: Yes, I just noticed that starting from version 6, it uses AWS SDK v3. Since AWS SDK v2 will reach end-of-support in September 2025, we can remove it after that. For now, I'll stick with version 5 and will add a comment

packages/collector/test/tracing/cloud/aws-sdk/v2/sqs/test.js
```diff
@@ -48,6 +48,12 @@ if (!supportedVersion(process.versions.node)) {
   mochaSuiteFn = describe;
 }
 
+// v12, the `sqs-consumer` library has dropped support for Node.js versions earlier than v20.
+// The minimum supported Node.js version is now v20.
+// Reference: https://github.com/bbc/sqs-consumer/blob/main/package.json#L21
+const runSqsConsumerAPI =
+  supportedVersion(process.versions.node) && semver.gte(process.versions.node, '20.0.0') ? describe : describe.skip;
```
kirrg001, line 55: This is the v2 test. Its not using v12?

PR #1737, opened 2025-05-16
feat: added support for redis sentinel

aryamohanan:


.tekton/assets/sidecars.json
```diff
@@ -102,6 +102,25 @@
       "name": "redis",
       "image": "redis:7.4.3"
     },
+    {
+      "name": "redis-sentinel-master",
```
aryamohanan, line 106: I’ve added a minimal Sentinel setup for testing purposes. For more realistic, production-like testing, the number of Sentinel instances should be increased. However, for our current use case, a single Sentinel is sufficient.
see: https://redis.io/docs/latest/operate/oss_and_stack/management/sentinel/

.tekton/assets/sidecars.json
```diff
@@ -102,6 +102,20 @@
       "name": "redis",
       "image": "redis:7.4.3"
     },
+    {
```
aryamohanan, line 105: I'm using a minimal Sentinel setup here, with exsting Redis instance as the master, along with one slave and one Sentinel instance. While a production like system would require additional Sentinel instances for high availability, this simple configuration is sufficient for our integration testing purposes.

.tekton/tasks/test-groups/test-ci-collector-tracing-database-task.yaml
```diff
@@ -19,6 +19,24 @@ spec:
             value: "false"
     - name: redis
       image: redis:7.4.3
+ 
+    - name: redis-slave
```
kirrg001, line 23: qs: How did you add that? 🤔 

The sidecars are attached via .tekton/generate-test-files.js



>   'test:ci:collector:tracing:database': {
    sidecars: ['memcached', 'mongodb', 'elasticsearch', 'redis', 'couchbase', 'mysql', 'postgres', 'mssql'],
    condition: ' && ! echo "$MODIFIED_FILES" | grep -q "packages/core/src/tracing/instrumentation/database"'
  },

aryamohanan: Ooops, I haven't used the template, manually added to the database task. I will add to the generate-test-files.

packages/core/src/tracing/instrumentation/database/redis.js
```diff
@@ -116,6 +116,21 @@ function instrument(redis) {
       };
     };
 
+    const createSentinelWrap = originalCreateSentinelFn => {
+      return function instrumentedCreateSentinelInstana(createRedisOpts) {
+        const redisSentinel = originalCreateSentinelFn.apply(this, arguments);
+        const addressUrl = createRedisOpts?.sentinelRootNodes.map(node => node?.host).join(', ');
```
kirrg001, line 122: qs: Is `sentinelRootNodes` optional? Shoud we use the optional chaining operator?

packages/collector/test/tracing/database/redis/test.js
```diff
@@ -22,18 +22,32 @@ const {
...
 
-// Please run this command on the root folder to start the redis instance:
-// node bin/start-test-containers.js --redis
-//
-// Please set the environment variables to run the tests against azure redis cluster:
-//    export AZURE_REDIS_CLUSTER=team-nodejs-redis-cluster-tekton.redis.cache.windows.net:6380
-//    export AZURE_REDIS_CLUSTER_PWD=
-['default', 'cluster'].forEach(setupType => {
+/**
+ * Supported Redis setups for local testing:
```
kirrg001, line 33: qs: I was wondering if we should add "runSingleType" because I assume you do not start all types at once? 

```js
let setupTypes = ['default', 'cluster', 'sentinel']
const singleSetupType = false; 
setupTypes = [singleSetupType] || setupTypes

setupTypes.forEach

packages/core/src/tracing/instrumentation/database/redis.js
aryamohanan: not optional, removed the condition check

.tekton/tasks/test-groups/test-ci-collector-tracing-database-task.yaml
aryamohanan: done

PR #1738, opened 2025-05-19
ci: updated npm installation logic in tekton pipelines

aryamohanan:
Previously, the logic was simple: if a specific npm-version was provided, it would install that version; otherwise, it would always install the latest npm version globally.

Now, the updated logic improves this by first checking if npm-version is provided—if yes, it installs that version. If not, it uses the npm is already available in the environment (which it usually is, since it's bundled with the Node.js image used in the tasks). .

This change helps avoid unnecessary global installations and makes better use of the preinstalled npm that comes with the Node.js base image.


abhilash-sivan (approved):
nice catch ❤️

PR #1742, opened 2025-05-25

instanacd:
Tada!


package-lock.json
```diff
@@ -53629,6 +53698,26 @@
         "node": ">=16 || 14 >=14.17"
       }
     },
+    "packages/collector/node_modules/redis": {
```
aryamohanan, line 53701: I'm not sure why redis v4.7.1 is installed within the collector module. I suspect this because of typeorm dependency. I also noticed that typeorm appears to be installed inside the collector package's node_modules, as seen in [this commit](https://github.com/instana/nodejs/pull/1722/commits/f79f8c1796a8ebca39d0b648585df8cf7b2ef2b4#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519).

aryamohanan (approved):

PR #1743, opened 2025-05-26
fix: updated base images for serverless with fedramp compliant dreadn…

abhilash-sivan:
…aught images

### DIFF
--- a/packages/aws-fargate/images/instana-aws-fargate/Dockerfile-npm
+++ b/packages/aws-fargate/images/instana-aws-fargate/Dockerfile-npm
@@ -1,5 +1,6 @@
 ARG NODEJS_VERSION
-FROM node:${NODEJS_VERSION} AS instana-aws-fargate-build-nodejs
+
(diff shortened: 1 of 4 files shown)


packages/aws-fargate/images/instana-aws-fargate/Dockerfile-npm
```diff
@@ -1,5 +1,6 @@
 ARG NODEJS_VERSION
-FROM node:${NODEJS_VERSION} AS instana-aws-fargate-build-nodejs
+
+FROM registry.access.redhat.com/ubi9/nodejs-${NODEJS_VERSION}-minimal:latest AS instana-aws-fargate-build-nodejs
```
abhilash-sivan, line 3: we have to use [delivery.instana.io/int-docker-base-images-local/base-images/nodejs20-builder:latest](http://delivery.instana.io/int-docker-base-images-local/base-images/nodejs20-builder:latest) image actually.

PR #1750, opened 2025-06-02
fix(aws-lambda): gracefully handled non-string path values

aryamohanan:
**Issue**

When the AWS Lambda invocation event includes a `path` field with an empty object (e.g., `"path": {}`), the trace fails to appear in the Instana UI.
In contrast, if the `path` field is completely omitted, the trace is displayed as expected.

 **Fix**

Introduced a `normalizePath()` utility function to ensure that only string values are accepted for the `path` field. If the value is not a string, it is gracefully converted to `null`. This prevents invalid data types from being sent to the backend and breaking trace visibility.

**Why This Fix**

The [BE](https://github.ibm.com/instana/backend/blob/develop/forge/src/main/java/com/instana/forge/connection/serverless/aws/lambda/AwsLambdaEntrySpan.java#L170) expects the `path` value to be a string, and the AWS Lambda event input format documentation also specifies it as such. When an unexpected type is provided, it leads to broken or missing traces. This fix enforces type safety and ensures handling of such cases within the tracer.

Reference:
[AWS gateway input format](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format)

jira [INSTA-39103](https://jsw.ibm.com/browse/INSTA-39103)

### DIFF
--- a/packages/aws-lambda/src/triggers.js
+++ b/packages/aws-lambda/src/triggers.js
@@ -20,6 +20,11 @@ const maxDynamoDbRecords = 3;
 const maxS3ObjectKeyLength = 200;
 const maxSQSRecords = 3;
 const awsLambdaFunctionUrlHostRegex = /^.*\.lambda-url\..*\.on\.aws$/i;
+let logger;
+
+module.exports.init = config => {
+  logger = config.logger;
+};
 
 exports.enrichSpanWithTriggerData = function enrichSpanWithTriggerData(event, context, span) {
   if (isApiGatewayProxyTrigger(event)) {
@@ -128,7 +133,18 @@ function extractHttpFromApiGatewwayProxyEvent(event, span) {
 //  https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format
 //  https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html#urls-payloads
 function normalizePath(value) {
-  return typeof value === 'string' ? value : null;
+  if (typeof value === 'string') {
+    return value;
+  }
+  logger.debug(
+    `Received a non-string value for the "path" field in the incoming request event: ${JSON.stringify(value)}. ` +
+      'Expected a string, as per AWS API Gateway event structure. See: ' +
+      // eslint-disable-next-line max-len
+      'https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format. ' +
+      'This likely indicates a misconfiguration on the AWS side and may result in incomplete or broken traces ' +
+      'in the Instana UI.'
+  );
+  return null;
 }
 
 function readHttpQueryParams(event) {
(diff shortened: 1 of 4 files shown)


packages/aws-lambda/src/triggers.js
```diff
@@ -105,22 +105,32 @@ function extractHttpFromApiGatewwayProxyEvent(event, span) {
...
 }
 
+//  Ensures the `path` value is always a string. The backend expects `path` to be a string,
+//  so we convert any non-string (for example, an empty object) to null.
+//  For reference, see AWS documentation on the Lambda event input format for API Gateway and Function URLs:
+//  eslint-disable-next-line max-len
+//  https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format
+//  https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html#urls-payloads
+function normalizePath(value) {
+  return typeof value === 'string' ? value : null;
```
abhilash-sivan, line 131: so even if we set the value to null, the traces will pop up, and it will contain all the info right ?

aryamohanan: Yes, the trace will still appear in the UI even if the path is not explicitly set or if null is passed. However, the backend expects the path to be a string. If a value of an invalid type is provided, the trace may not show up as expected. 

I’ve manually tested and verified this behavior.

packages/aws-lambda/test/integration_test/test_definition.js
```diff
@@ -2138,6 +2138,75 @@ function registerTests(handlerDefinitionPath, reduced) {
...
+
+        const spans = await control.getSpans();
+
+        expectExactlyOneMatching(spans, [
+          span => expect(span.n).to.equal('aws.lambda.entry'),
+          span => expect(span.k).to.equal(constants.ENTRY),
+          span => expect(span.data.http).to.be.an('object'),
+          span => expect(span.data.http.method).to.equal('POST'),
+          span => expect(span.data.http.url).to.equal(null),
+          span => expect(span.data.http.path_tpl).to.equal(null),
```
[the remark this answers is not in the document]
kirrg001: Saw this: https://github.com/instana/nodejs/pull/1750/files#r2120318354
WIll continue there.

kirrg001 (approved):

packages/aws-lambda/src/triggers.js
```diff
@@ -102,25 +107,43 @@ function extractHttpFromApiGatewwayProxyEvent(event, span) {
...
+//  eslint-disable-next-line max-len
+//  https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format
+//  https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html#urls-payloads
+function normalizePath(value) {
+  if (typeof value === 'string') {
+    return value;
+  }
+  logger.debug(
+    `Received a non-string value for the "path" field from the incoming request event: ${JSON.stringify(value)}. ` +
+      'Expected a string. This may result in incomplete or broken traces in the Instana UI.'
```
kirrg001, line 142: Can we please add an official aws link? That its clear that its not our requirement.

kirrg001: We need to make clear that hit is an AWS misconfiguration.

PR #1761, opened 2025-06-10
fix(aws-lambda): upgraded go version to 1.24.0

aryamohanan:
The Lambda extension is affected by CVE-2024-24790, a vulnerability in Go's net/netip standard library. Since the extension doesn't directly use this dependency, we're mitigating the risk by updating to the latest Go version.

The CVE-2024-24790 is resolved in https://go-review.googlesource.com/c/go/+/590316

ref https://jsw.ibm.com/browse/INSTA-39904


kirrg001 (approved):
pre approving

PR #1768, opened 2025-06-12
test: improved http vs https handling

kirrg001:
Problems:

- NODE_TLS_REJECT_UNAUTHORIZED **can create a native console log span** (the warning comes from Nodejs as console log on runtime). HTTPS without certification -> creates a node warning. This can create real problems if you only assert the length of spans to expect
- As part of fixing a bug in https://github.com/instana/nodejs/pull/1767, I wanted to fix a test, which suddenly showed me this console span and it was part of a span assertion lengths!
- We switch to http default. There is no reason to run tests on https by default. As long as https is covered.
- Improved naming of variables. It was really chaotic.

### DIFF
--- a/packages/aws-lambda/test/integration_test/test_definition.js
+++ b/packages/aws-lambda/test/integration_test/test_definition.js
@@ -1661,6 +1661,7 @@ function registerTests(handlerDefinitionPath, reduced) {
         faasRuntimePath: path.join(__dirname, '../runtime_mock'),
         handlerDefinitionPath,
         startBackend: true,
+        backendUsesHttps: true,
         startProxy: false,
         proxyPort,
         env


packages/serverless/test/util/AbstractServerlessControl.js
```diff
@@ -16,10 +16,6 @@ const retry = require('@instana/core/test/test_util/retry');
 // const delay = require('@instana/core/test/test_util/delay');
 const config = require('@instana/core/test/config');
 
-// To address the certificate authorization issue with node-fetch, process.env.NODE_TLS_REJECT_UNAUTHORIZED
-// was set to '0'. Refer to the problem discussed in https://github.com/node-fetch/node-fetch/issues/19
-process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
kirrg001, line 21: If we have to set this env, then we set it in the mocha test config file. Not here.

packages/aws-fargate/test/Control.js
```diff
@@ -19,10 +19,21 @@ let execArg;
...
   this.port = opts.port || portfinder();
+
+  this.backendUsesHttps = 'backendUsesHttps' in opts ? opts.backendUsesHttps : false;
+  const backendProtocol = this.backendUsesHttps ? 'https' : 'http';
+
+  // We execute requests from THIS process. If any url runs on https,
+  // we need to disable the TLS certificate check, otherwise the request will fail.
+  // Refer to the problem discussed in https://github.com/node-fetch/node-fetch/issues/1
+  if (this.backendUsesHttps) {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
kirrg001, line 30: We do not set NODE_TLS_REJECT_UNAUTHORIZED globally for a whole test env.
Only if the backend / app runs in https and we use self signed certs.

kirrg001:
Tests only failing because we currently cannot connect to ioredis cluster.

PR #1782, opened 2025-06-19
chore: exclude markdown files from prettier formatting

abhilash-sivan:
We don't want the formatter for markdown files, as formatting can alter the intended structure or layout.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,3 @@
 *.yml
 *.yaml
+*.md


.prettierignore
```diff
@@ -1,2 +1,3 @@
 *.yml
-*.yaml
\ No newline at end of file
+*.yaml
+*.md
```
aryamohanan, line 3: qs: What are the advantages of excluding .md files from Prettier formatting?

abhilash-sivan: Currently, when we make changes to the README.md or CONTRIBUTING.md files, Prettier reformats the entire file, which unintentionally alters the existing formatting and applies syntax changes.

I don't think this is necessary, as we already format the markdown content intentionally within these files.

abhilash-sivan: Prettier reformats the entire Markdown file even for small changes, which ends up changing a lot of lines unnecessarily. Since these are documentation files, we usually format them intentionally, so auto-formatting isn’t really needed imo

kirrg001: Not for me 🤔

kirrg001: Ah yes, I do not allow markdown editing

>   "[markdown]": {
    "editor.formatOnSave": false
  },
  "[md]": {
    "editor.formatOnSave": false
  },

kirrg001 (approved):

PR #1795, opened 2025-06-26
feat: added support for disabling config from agent

aryamohanan:
Added support for disabling instrumentations and groups via the agent configuration.

#### Agent Config (YAML):

Example agent's `configuration.yaml` file:

```yaml
tracing:
  disable:
    redis: true         # disable redis
    console: false      # enable console
    logging: true       # disable logging grp
```

####  Internally Normalized As:

```js
tracing.disable = {
  instrumentations: ['redis', '!console'],
  groups: ['logging']
}
```

#### Internal Notes:

* When a package or group is explicitly enabled (false in config), we internally normalize it as a negated entry using ! (e.g., !console).
* Instrumentation level disables take precedence over group enables.
* Group disables apply to all matching instrumentations unless explicitly enabled.

Ref: https://github.ibm.com/instana/technical-documentation/pull/344/files

### DIFF
--- a/packages/collector/src/announceCycle/unannounced.js
+++ b/packages/collector/src/announceCycle/unannounced.js
@@ -44,6 +44,7 @@ const maxRetryDelay = 60 * 1000; // one minute
  * @property {KafkaTracingConfig} [kafka]
  * @property {import('@instana/core/src/tracing').IgnoreEndpoints} [ignore-endpoints]
  * @property {boolean} [span-batching-enabled]
+ * @property {Object} [disable]
  */
 
 /**
@@ -119,6 +120,7 @@ function applyAgentConfiguration(agentResponse) {
   applyKafkaTracingConfiguration(agentResponse);
   applySpanBatchingConfiguration(agentResponse);
   applyIgnoreEndpointsConfiguration(agentResponse);
+  applyDisableConfiguration(agentResponse);
 }
 
 /**
@@ -237,6 +239,21 @@ function applyIgnoreEndpointsConfiguration(agentResponse) {
   agentOpts.config.tracing.ignoreEndpoints = configNormalizers.ignoreEndpoints.normalizeConfig(ignoreEndpointsConfig);
 }
 
+/**
+ * The incoming agent configuration include  `disable` object that include
+ * which instrumentation/categories should be disabled. For example: { logging: true, console: false }
+ * This will be normalized into an array of strings, such as: ['logging', '!console']
+ * For more details on the design, refer to:https://github.ibm.com/instana/technical-documentation/pull/344
+ *
+ * @param {AgentAnnounceResponse} agentResponse
+ */
+function applyDisableConfiguration(agentResponse) {
+  const disablingConfig = agentResponse?.tracing?.disable;
+  if (!disablingConfig) return;
+
+  ensureNestedObjectExists(agentOpts.config, ['tracing', 'disable']);
+  agentOpts.config.tracing.disable = configNormalizers.disable.normalize({ tracing: { disable: disablingConfig } });
+}
 module.exports = {
   init,
 
--- a/packages/collector/src/announceCycle/unannounced.js
+++ b/packages/collector/src/announceCycle/unannounced.js
@@ -44,7 +44,7 @@ const maxRetryDelay = 60 * 1000; // one minute
  * @property {KafkaTracingConfig} [kafka]
  * @property {import('@instana/core/src/tracing').IgnoreEndpoints} [ignore-endpoints]
  * @property {boolean} [span-batching-enabled]
- * @property {Object} [disable]
+ * @property {import('@instana/core/src/tracing').Disable} [disable]
  */
 
 /**
--- a/packages/collector/src/announceCycle/unannounced.js
+++ b/packages/collector/src/announceCycle/unannounced.js
@@ -252,7 +252,9 @@ function applyDisableConfiguration(agentResponse) {
   if (!disablingConfig) return;
 
   ensureNestedObjectExists(agentOpts.config, ['tracing', 'disable']);
-  agentOpts.config.tracing.disable = configNormalizers.disable.normalize({ tracing: { disable: disablingConfig } });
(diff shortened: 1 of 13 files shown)


packages/core/src/util/configNormalizers/disable.js
```diff
@@ -150,19 +156,41 @@ function categorizeDisableEntries(rawEntries) {
...
+ */
+function flattenDisableConfigs(disableConfig) {
+  // Converts a config with boolean values into a flat list of strings.
+  // Each key is a instrumentation or group.
+  // - true - disable
+  // - false - enable, we internally formated by adding ! in prefix
+  // Example: { logging: true, console: false } => ['logging', '!console']
+  return Object.entries(disableConfig).flatMap(([entryName, shouldDisable]) => {
+    if (shouldDisable === true) return [entryName];
+    if (shouldDisable === false) return [`!${entryName}`];
```
aryamohanan, line 193: Using ! for negation internally is a simple and clean way. It keeps things easy to manage and avoids adding new config fields like tracing.enable. Unless we really need more detailed control later, I think this approach is good enough.

Let me know if you have a better idea or see any issues with this.

packages/core/src/util/disableInstrumentation.js
```diff
@@ -64,19 +64,37 @@ function shouldDisable(cfg, { moduleName, instrumentationName, group } = {}) {
     return false;
   }
 
-  // Case 1: Check if module or instrumentation is disabled
+  // Case 1: Check if module or instrumentation is explicitly enabled
+  if (moduleName) {
```
aryamohanan, line 68: enable always have higher precedence. Currently supported enable is via agent config

packages/core/src/util/configNormalizers/disable.js
```diff
@@ -68,6 +67,23 @@ exports.normalize = function normalize(config) {
   }
 };
 
+/**
+ * Handles config from agent.
+ * @param {import('../../util/normalizeConfig').InstanaConfig} config
+ */
+exports.normalizeExternalConfig = function normalizeExternalConfig(config) {
```
aryamohanan, line 74: Created a separate function to handle agent configuration. Integrating it directly into the existing normalize function could impact performance, as normalize includes several other checks.

packages/collector/test/tracing/logging/console/test.js
```diff
@@ -131,6 +131,223 @@ mochaSuiteFn('tracing/logging/console', function () {
       ));
   });
 
+  describe('when logging is disabled', () => {
```
aryamohanan, line 134: Added one integration test

kirrg001: in-code for agent? I do not understand the poll :D

packages/collector/src/announceCycle/unannounced.js
```diff
@@ -44,6 +44,7 @@ const maxRetryDelay = 60 * 1000; // one minute
  * @property {KafkaTracingConfig} [kafka]
  * @property {import('@instana/core/src/tracing').IgnoreEndpoints} [ignore-endpoints]
  * @property {boolean} [span-batching-enabled]
+ * @property {Object} [disable]
```
kirrg001, line 47: I think we have to use TS object - we also do that for ignore-endpoints line 45

packages/core/src/util/configNormalizers/disable.js
```diff
@@ -40,18 +40,17 @@ exports.normalize = function normalize(config) {
       delete config.tracing.disabledTracers;
     }
 
-    let disableConfig = config.tracing.disable;
-
     // Fallback to environment variables if `disable` is not configured
+    const disableConfig = config.tracing?.disable || getDisableFromEnv();
```
kirrg001, line 44: 🤔 Isn't that an empty object for getDisableFromEnv? If yes, it won't work

> if (!disableConfig) {

aryamohanan: Normally, getDisableFromEnv returns null for non-applicable or invalid cases.

packages/core/src/util/configNormalizers/disable.js
```diff
@@ -64,14 +63,36 @@ exports.normalize = function normalize(config) {
...
+  try {
+    if (typeof config.tracing.disable === 'object') {
+      const flattenedEntries = flattenDisableConfigs(config.tracing.disable);
+      return categorizeDisableEntries(flattenedEntries);
+    }
+  } catch (error) {
+    logger?.debug(`Error while normalizing external tracing.disable config: ${error?.message} ${error?.stack}`);
+  }
+
+  return {};
```
kirrg001, line 85: I guess you return {} because this is the default value for disable or?
Maybe we can reuse somehow

packages/core/src/util/normalizeConfig.js
```diff
@@ -88,6 +88,7 @@ const allowedSecretMatchers = ['equals', 'equals-ignore-case', 'contains', 'cont
  * @property {AgentTracingKafkaConfig} [kafka]
  * @property {boolean|string} [spanBatchingEnabled]
  * @property {import('../tracing').IgnoreEndpoints} [ignoreEndpoints]
+ * @property {TracingDisableOptions} [disable]
```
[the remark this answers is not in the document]
aryamohanan: updated

packages/core/src/util/configNormalizers/disable.js
aryamohanan: Yes. This is thedefault value for disable

packages/collector/test/tracing/logging/console/test.js
abhilash-sivan: Maybe we can add that also. Just to cover the case

abhilash-sivan (approved):

packages/collector/test/tracing/logging/console/test.js
abhilash-sivan: Maybe in a separate PR

PR #1796, opened 2025-06-26
fix: depreacted INSTANA_DISABLE_TRACING env variable

aryamohanan:
#### 1. **Disabling All Tracing via Environment Variable**

Previously, full tracing could be disabled using:

```bash
INSTANA_DISABLE_TRACING=true
```

This environment variable has now been deprecated and  **renamed** to:

```bash
INSTANA_TRACING_DISABLE=true
```

#### 2. **Disabling All Tracing via Code Configuration**

You can now disable tracing directly in code using the updated configuration structure:

```js
require('@instana/collector')({
  tracing: {
    disable: true
  }
});
```


packages/core/src/util/normalizeConfig.js
```diff
@@ -242,12 +242,12 @@ function normalizeTracingConfig(config) {
   }
   normalizeTracingEnabled(config);
   normalizeUseOpentelemetry(config);
+  normalizeDisableTracing(config);
```
aryamohanan, line 245: Moved this config check earlier in the flow to ensure full trace disable  is respected before any auto-tracing checks runs. When tracing is fully disabled, auto-tracing should implicitly be considered enabled:false.

packages/core/test/util/normalizeConfig_test.js
```diff
@@ -20,6 +20,7 @@ describe('util.normalizeConfig', () => {
     delete process.env.INSTANA_DISABLE_AUTO_INSTR;
     delete process.env.INSTANA_DISABLE_TRACING;
```
kirrg001, line 21:
```suggestion
    // deprecated
    delete process.env.INSTANA_DISABLE_TRACING;
```

PR #1806, opened 2025-07-03
chore: added dependabot configuration

abhilash-sivan:
refs [INSTA-38933](https://jsw.ibm.com/browse/INSTA-38933)

The current logic enables Dependabot and does the following:

- Creates a single PR for all packages when there are minor or patch updates
- Creates individual PRs for each package with major updates

This behavior aligns with the current PR logic used by our currency bot.

You can see a sample PR structure here: https://github.com/abhilash-sivan/nodejs/pulls

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/" 
+    schedule:
+      interval: "weekly"
+    allow:
+      - dependency-type: "production"
+    groups:
+      minor-and-patch:
+        update-types:
+          - "minor"
+          - "patch"
+        patterns: ["*"]


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/" 
+    schedule:
+      interval: "monthly"
+    allow:
+      - dependency-type: "production"
+    open-pull-requests-limit: 30
```
aryamohanan, line 9: suggestion:  We can set rebase-strategy to disabled or none. By default, Dependabot rebases its pull requests to stay updated with the main branch. Instead, we can rebase manually. This will help avoid unnecessary builds in the Tekton pipeline every time the main branch changes.

aryamohanan (approved):
Pre-approving..

.github/dependabot.yml
abhilash-sivan: rebase-strategy: "none" is removed, and "disabled" is the new value

PR #1810, opened 2025-07-08
chore: switched development version to v22

aryamohanan:
[v22](https://github.com/nodejs/Release) is LTS.

- [x] Update v20 references
- [x] switch pipelines to use v22

ref: https://jsw.ibm.com/browse/INSTA-36125


.nvmrc
```diff
@@ -1 +1 @@
-20.19
+22.17
```
aryamohanan, line 1: used latest version of v22

kirrg001 (commented):
I could not find any hint/solution regarding Node v18.
AFAIK Node v22 ships with NPM 11 and does not support Node v18.

aryamohanan:
> I could not find any hint/solution regarding Node v18.
> AFAIK Node v22 ships with NPM 11 and does not support Node v18.

In our case, Node.js is used for Lerna and npm within the release workflow, so it doesn't impact the codebase. 
We're already pinning the npm version for Node v18 in our Tekton builds, so this shouldn't pose any issues.
For the serverless images we already using development version.

I don't find any other cases.

packages/aws-lambda/layer/bin/publish-layer.sh
```diff
@@ -167,13 +167,8 @@ SUPPORTED_RUNTIMES="nodejs18.x nodejs20.x nodejs22.x"
 # We're using a development version.
 ROOT_DIR=$(git rev-parse --show-toplevel 2>/dev/null || echo "../../..")
 NVMRC_PATH="$ROOT_DIR/.nvmrc"
-if [[ -f "$NVMRC_PATH" ]]; then
-  NODEJS_DEV_VERSION=$(cut -d '.' -f 1 "$NVMRC_PATH")
-  echo "Using Node.js version $NODEJS_DEV_VERSION from .nvmrc for Docker build"
-else
-  echo "Warning: .nvmrc file not found at $NVMRC_PATH, falling back to default Node.js version 20"
-  NODEJS_DEV_VERSION=20
```
kirrg001, line 175: 👍 Great cleanup

packages/google-cloud-run/images/inspector/build.sh
```diff
@@ -23,7 +23,11 @@ docker rm -f $containername || true
 echo "Removing image $imagetag"
 docker rmi -f $imagetag
 
-echo "Building $dockerfile -> $imagetag"
-docker build -f $dockerfile -t $imagetag -t $gcr_repository/$imagetag .
+ROOT_DIR=$(git rev-parse --show-toplevel)
```
abhilash-sivan, line 26: 👏 Good catch here

abhilash-sivan (approved):

PR #1830, opened 2025-07-09
fix: added eol event for Node v18, v21 and v23

aryamohanan:
updated eol logic and events to treat node <20, 21.x, and 23.x as eol

Node.js versions below 20, as well as versions 21.x and 23.x, have reached end-of-life.
this update ensures the node.js collector triggers eol events for these versions to inform users of security and support risks.

reference: https://nodejs.org/en/about/previous-releases

- node 18: eol on mar 27, 2025
- node 21: eol since apr 10, 2024
- node 23: eol since may 14, 2025

### DIFF
--- a/packages/collector/src/util/eol.js
+++ b/packages/collector/src/util/eol.js
@@ -14,5 +14,5 @@ const { satisfies } = require('semver');
  * @returns {boolean}
  */
 exports.isNodeVersionEOL = function () {
-  return satisfies(process.versions.node, '<18 || 19');
+  return satisfies(process.versions.node, '<20 || 21 || 23');
 };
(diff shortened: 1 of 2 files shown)


packages/collector/src/util/eol.js
```diff
@@ -14,5 +14,5 @@ const { satisfies } = require('semver');
  * @returns {boolean}
  */
 exports.isNodeVersionEOL = function () {
-  return satisfies(process.versions.node, '<18 || 19');
+  return satisfies(process.versions.node, '<20 || 21 || 23');
```
aryamohanan, line 17: https://nodejs.org/en/about/previous-releases

kirrg001 (approved):
> fix: marked node versions <20, 21.x, and 23.x as eol to trigger eol events #1830

Please use

fix: added eol event for Node v18 & v21

PR #1842, opened 2025-07-11
test: updated bull test to fix the esm build

aryamohanan:
ESM build was failing due to a false positive case, the case was using Process that doesn't support the ESM. 
https://github.com/OptimalBits/bull/issues/924

### DIFF
--- a/packages/collector/test/tracing/messaging/bull/test.js
+++ b/packages/collector/test/tracing/messaging/bull/test.js
@@ -718,58 +718,28 @@ mochaSuiteFn('tracing/messaging/bull', function () {
     }
   });
 
-  describe('tracing disabled', function () {
-    this.timeout(config.getTestTimeout() * 2);
+  // NOTE: Bull doesn't officially support Process API when using processor with ES module syntax
+  // See: https://github.com/OptimalBits/bull/blob/489c6ab8466c1db122f92af3ddef12eacc54179e/lib/queue.js#L712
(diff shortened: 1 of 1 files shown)


packages/collector/test/tracing/messaging/bull/test.js
```diff
@@ -718,100 +718,108 @@ mochaSuiteFn('tracing/messaging/bull', function () {
     }
   });
 
-  describe('tracing disabled', function () {
-    this.timeout(config.getTestTimeout() * 2);
+  // NOTE: Bull doesn't officially support Process API when using processor with ES module syntax
+  // See: https://github.com/OptimalBits/bull/blob/489c6ab8466c1db122f92af3ddef12eacc54179e/lib/queue.js#L712
+  // Related issue: https://github.com/OptimalBits/bull/issues/924
+  if (!process.env.RUN_ESM) {
```
aryamohanan, line 724: Added this ESM check, the test uses Process, currently Process does not support ESM, the case is a false positive.

abhilash-sivan: Oh. It was false positive for node v20. Interesting

abhilash-sivan (approved):
Thanks ❤️

packages/collector/test/tracing/messaging/bull/test.js
aryamohanan: Yes. I will do a detailed test next week and fix the tests. For now, I'm skipping the test case.

PR #1849, opened 2025-07-15
chore: added production dependency update bot

abhilash-sivan:
- refs [INSTA-38933](https://jsw.ibm.com/browse/INSTA-38933)

Solution:

- added a new bot for production dependency updates
- This is scheduled to run weekly with a maximum of 5 (customisable) PRs

Reasons for not combining with currency bot:
- We need to see both pipelines and be able run them independently on certain events

TODO:
- [ ] enable [listener_dependency_prod_bot_weekly](https://cloud.ibm.com/devops/pipelines/tekton/c2cd6a8d-ea5a-47b0-913e-cd172d63833f/runs?env_id=ibm:yp:eu-de&triggerName=listener_dependency_prod_bot_weekly) before merging


bin/currency/update-prod-dependencies.js
```diff
@@ -0,0 +1,104 @@
...
+      const targetName = pkgJson.name || pkgDir;
+      console.log(`npm i ${dep}@${latestVersion} -w ${targetName}`);
+      execSync(`npm i ${dep}@${latestVersion} -w ${targetName} --no-audit`, {
+        stdio: 'inherit',
+        cwd
+      });
+    });
+
+    execSync("git add '*package.json' package-lock.json", { cwd });
+    execSync(`git commit -m "build: bumped ${dep} from ${currentVersion} to ${latestVersion}"`, { cwd });
```
aryamohanan, line 83: IMO, this should be different from the dev dependency update commit message.

abhilash-sivan: Any suggestions ?

bin/currency/update-prod-dependencies.js
```diff
@@ -0,0 +1,104 @@
+/*
```
kirrg001, line 1: Putting it here: Naming problem

bin/currency/update-currencies.js
bin/currency/update-prod-dependencies.js

Please come up with a better strukture.

bin/currency/update-prod-dependencies.js
```diff
@@ -0,0 +1,104 @@
...
+const { execSync } = require('child_process');
+const utils = require('./utils');
+const BRANCH = process.env.BRANCH;
+const SKIP_PUSH = process.env.SKIP_PUSH === 'true';
+const MAX_PROD_PR_LIMIT = process.env.MAX_PROD_PR_LIMIT || 5;
+const cwd = path.join(__dirname, '..', '..');
+
+if (!BRANCH) throw new Error('Please set env variable "BRANCH".');
+
+console.log(`BRANCH: ${BRANCH}`);
```
kirrg001, line 17: Putting it here:

IMO both procedures (currency updates + prod updates) run through the same steps:

1. source of dependencies (currencies.json vs prod deps in package.json)
2. update phase (IMO is equal -> git operations, slighlty different config e.g. PR limit or minor vs major handling)

It would be good to extract more common logic into the utils.js.

.github/workflows/cleanup-branches.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
 
           git fetch --prune --all
 
-          for branch in $(git branch -r | grep -E 'origin/(chore-|docs-|fix-|feat-|test-|refactor-|ci-|build-)'); do
+          for branch in $(git branch -r | grep -E 'origin/(chore-|docs-|fix-|feat-|test-|refactor-|ci-|build-|prod-dependency-bot-)'); do
```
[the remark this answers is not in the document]
abhilash-sivan: FYI, for currency bot we are using branch name like `currency-bot-` 

see https://github.com/instana/nodejs/blob/17874a91b321855241902f3f0090575879bc94e0/.github/workflows/cleanup-branches.yaml#L50

**build-package-name** is also a good suggestion for prod dependency bot updates.

bin/dependencies/production/update-prod-dependencies.js
```diff
@@ -0,0 +1,93 @@
+/*
```
aryamohanan, line 1: qs: Is there a way to configure certain libraries to be excluded from updates? For example, I’d like to prevent updates for the package abc since it no longer supports Node.js v18, which we still need to support.

abhilash-sivan: Hmm... good catch. that's a case to consider, but we can implement that as an improvement to the current feature.

kirrg001 (approved):

aryamohanan (approved):
Awesome work ❤️

PR #1856, opened 2025-07-16
chore: udpated currency bot branch name for major updates

abhilash-sivan:
- added currency bot branches to cleanup list

This was the structure of branch name:
`currency-bot-major-redisclient`

From now on it will be:
`currency-bot-major-redisclient-560` where the 560 will mean that the update of the dependency is to version 5.6.0


aryamohanan (approved):
Thanks ❤️

PR #1862, opened 2025-07-18
fix(serverless): removed trailing slashes from instana endpoint url

aryamohanan:
Trailing slashes in the Instana endpoint were causing double slashes in request URLs (e.g., https://example.com//traces), leading to trace delivery failures. 
This fix trims the trailing slash  to ensure clean and valid URLs IN the AWS Lambda extension.

The fix is only on the AWS lambda extension in case of directly sending to the backend, already handled in our [backend connector](https://github.com/instana/nodejs/blob/main/packages/serverless/src/backend_connector.js#L299).

See: https://github.ibm.com/instana/lambda-extension/pull/24.

ref: https://jsw.ibm.com/browse/INSTA-43618

### DIFF
--- a/packages/serverless/src/environment.js
+++ b/packages/serverless/src/environment.js
@@ -86,7 +86,7 @@ function _validate(instanaEndpointUrl, _instanaAgentKey, validateInstanaAgentKey
     backendPort = '443';
   }
 
-  backendPath = parsedUrl.pathname;
+  backendPath = sanitizePath(parsedUrl.pathname);
 
   instanaAgentKey = _instanaAgentKey;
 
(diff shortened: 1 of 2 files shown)


abhilash-sivan (commented):
commit suggestion:

fix(aws-lambda): corrected INSTANA_ENDPOINT_URL handling when provided with a trailing slash
or
fix(aws-lambda): removed trailing slash from INSTANA_ENDPOINT_URL to ensure correct endpoint formatting

abhilash-sivan (approved):

kirrg001 (approved):
LGTM

Suggestions:

fix(serverless): improved sanitization of instana endpoint url
fix(serverless): removed trailing slashes from instana endpoint url

PR #1866, opened 2025-07-21
ci: updated sonar scan command in coverage task

aryamohanan:
We're using npx so we don't have to install the Sonar scanner globally. This makes the script simpler and more reliable, especially when running in tekton CI pipelines.

https://github.com/SonarSource/sonar-scanner-npm?tab=readme-ov-file#getting-started


aryamohanan:
Closing this now, if needed, will reopen

PR #1868, opened 2025-07-22
fix(autoprofile): reduced package size

kirrg001:
- removed darwin rebuilds from production release
- there is no use case to ship them
- reduces size

### DIFF
--- a/packages/autoprofile/package.json
+++ b/packages/autoprofile/package.json
@@ -62,7 +62,10 @@
   "files": [
     "src",
     "lib",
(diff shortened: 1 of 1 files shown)


packages/autoprofile/package.json
```diff
@@ -62,7 +62,10 @@
   "files": [
     "src",
     "lib",
-    "prebuilds",
```
abhilash-sivan, line 65: How about Darwin AMD64? Aren't there any customers who use that or?

PR #1875, opened 2025-07-24
feat: added support for ignoring http entry calls

aryamohanan:
**Overview**

This update introduces support for ignoring **HTTP calls** using configurable rules. It allows users to ignore specific incoming HTTP requests from tracing, based on request method, endpoint path, or host.

**Supported Filtering Fields**

The following fields are available for HTTP span filtering:

* `method`: HTTP method (`get`, `post`, etc.)
* `url`: URL path (e.g., `/health`, `/api/v1/user`)
* `host`: Host and port (e.g., `localhost:3000`)

All values are case-insensitive. Wildcards (`*`) are supported.

> Note: Only **HTTP entry spans** can be filtered. **HTTP exit spans are not supported**.

**Configuration Options**

1. **YAML-based configuration via `INSTANA_IGNORE_ENDPOINTS_PATH`**

Set the environment variable:

```
INSTANA_IGNORE_ENDPOINTS_PATH=/absolute/path/to/config.yaml
```

Example `config.yaml`:

```yaml
tracing:
  ignore-endpoints:
    http:
      - methods: ["get"]
        endpoints: ["/status"]

      - methods: ["get", "send"]
        endpoints: ["/status", "/v1/status"]

      - methods: ["*"]
        connection: ["localhost:3000"]
```

2. **Simple environment variable: `INSTANA_IGNORE_ENDPOINTS`**

For basic filtering by method:

```
INSTANA_IGNORE_ENDPOINTS=http:get
```

This example ignores all HTTP `GET` requests regardless of the path.

3. **In-code configuration**

You can also configure filters when initializing the Instana collector:

```js
require('@instana/collector')({
  tracing: {
    ignoreEndpoints: {
      http: [
        {
          methods: ['get'],
          endpoints: ['/status']
        }
      ]
    }
  }
});
```

3. **Agent configuration via `configuration.yaml`**

Example `configuration.yaml`:

```yaml
com.instana.tracing:
  ignore-endpoints:
    http:
      - methods: ["get"]
        endpoints: ["/status"]

      - methods: ["get", "send"]
        endpoints: ["/status", "/v1/status"]

      - methods: ["*"]
 ```      
---

**TODO**

- [x] Update instrumentation to collect span data before creation https://github.com/instana/nodejs/pull/1903
- [x] Implement entry span ignoring logic
- [x] Suppression, apply suppression flags correctly to indicate ignored spans and prevent propagation downstream if necessary.
- [x] Add unit tests + integration tests
- [ ] Public doc update

ref: https://jsw.ibm.com/browse/INSTA-46689

### DIFF
--- a/packages/core/src/tracing/instrumentation/protocols/http2Client.js
+++ b/packages/core/src/tracing/instrumentation/protocols/http2Client.js
@@ -86,31 +86,13 @@ function instrumentClientHttp2Session(clientHttp2Session) {
     for (let i = 0; i < arguments.length; i++) {
       originalArgs[i] = arguments[i];
     }
-    let method;
-    let path;
-    method = method || 'GET';
-    path = path || '/';
 
-    /**
-     * The HTTP method and other fields are captured later during request processing,
-     * but we must initialize the span data early to ensure correct async context propagation.
-     *
-     * NOTE: HTTP span ignoring feature currently applies only to entry spans (phase 1),
-     * so setting exit span data here is safe. This will be refined in future iterations.
-     */
-    const spanData = {
-      http: {
-        operation: method,
-        endpoints: ''
-      }
-    };
     return cls.ns.runAndReturn(() => {
       const span = cls.startSpan({
         spanName: 'node.http.client',
         kind: constants.EXIT,
         traceId: parentSpan?.t,
-        parentSpanId: parentSpan?.s,
-        spanData
+        parentSpanId: parentSpan?.s
       });
 
       // startSpan updates the W3C trace context and writes it back to CLS, so we have to refetch the updated context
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: packages/core/src/tracing/instrumentation/protocols/httpClient.js [hunk 1]
```diff
@@ -419,6 +419,14 @@ function setHeadersOnRequest(clientRequest, span, w3cTraceContext) {
   if (!isItSafeToModifiyHeadersForRequest(clientRequest)) {
     return;
   }
+
+  if (span.shouldSuppressDownstream) {
```
aryamohanan, line 423: The suppression flag is set in cls in case of entry spans, see: https://github.com/instana/nodejs/blob/main/packages/core/src/tracing/cls.js#L462

packages/core/src/tracing/instrumentation/protocols/http2Client.js
```diff
@@ -164,6 +164,11 @@ function addHeaders(headers, span, w3cTraceContext) {
   if (!headers) {
     return;
   }
+  if (span.shouldSuppressDownstream) {
```
kirrg001, line 167: I had to remind myself for a minute: `shouldSuppressDownstream` is only true if the class instance is `InstanaIgnoredSpan` AND the config is ON.

I think the readability here is not ideal.

I'd suggest to raise a PR after this PR and change the 3 occurances in our code to:

```
if (span.isIgnored AND config flag is ON)
```

packages/core/src/util/spanFilter.js
```diff
@@ -133,8 +135,13 @@ function shouldIgnore(span, ignoreEndpointsConfig) {
     return false;
   }
 
-  const ignoreConfigs = ignoreEndpointsConfig[span.n];
-  if (!ignoreConfigs) {
+  spanTypeKey = normalizeSpanDataTypeKey(span.n);
```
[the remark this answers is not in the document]
aryamohanan: updated

kirrg001 (approved):

PR #1884, opened 2025-07-29
feat: added support for @azure/storage-blob v12.28.0

aryamohanan:
* **Breaking Changes in [@azure/storage-blob](https://www.npmjs.com/package/@azure/storage-blob) v12.28.0**
  * It has dropped support for end-of-life (EOL) versions of Node.js. 
      See: https://github.com/Azure/azure-sdk-for-js/commit/b8ff66f14b2fbea14aabf798b46a0625f5ce7650
  * The package has been migrated to ESM.
      See: https://github.com/Azure/azure-sdk-for-js/pull/33329/files

* **Impact**

  * Due to the migration to ESM, our ESM builds started failing.
  * ESM-based applications were not being traced.

* **Resolution**

  * To address this issue, we enabled the **IITM hook**, which was introduced a year ago.
  * Previously, the IITM hook was not really used because there were no instrumented ESM packages at that time.

  * ESM applications are now successfully traced using the IITM hook.
  * CJS applications continue to be traced using the traditional `require` hook.

ref:  https://jsw.ibm.com/browse/INSTA-47737

### DIFF
--- a/packages/collector/test/tracing/cloud/azure/blob/test.js
+++ b/packages/collector/test/tracing/cloud/azure/blob/test.js
@@ -56,7 +56,7 @@ if (!storageAccount || !accountKey) {
 } else {
   mochaSuiteFn('tracing/cloud/azure/blob', function () {
     ['latest', 'v1227'].forEach(version => {
-      // NOTE: require-mock is not working with esm apps.
+      // NOTE: require-mock is not working with ESM apps.
       // TODO: Support for mocking `import` in ESM apps is planned under INSTA-788.
       if (process.env.RUN_ESM && version !== 'latest') return;
 
--- a/packages/collector/test/tracing/cloud/azure/blob/test.js
+++ b/packages/collector/test/tracing/cloud/azure/blob/test.js
@@ -61,6 +61,9 @@ if (!storageAccount || !accountKey) {
         mochaSuiteFn = describe.skip;
       }
 
+      // NOTE: require-mock is not working with esm apps. There is also no need to run the ESM APP for all versions.
+      if (process.env.RUN_ESM && version !== 'latest') return;
+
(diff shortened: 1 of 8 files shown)


packages/collector/test/tracing/cloud/azure/blob/test.js
```diff
@@ -55,454 +55,466 @@ if (!storageAccount || !accountKey) {
   });
 } else {
   mochaSuiteFn('tracing/cloud/azure/blob', function () {
-    this.timeout(config.getTestTimeout());
-    globalAgent.setUpCleanUpHooks();
-    const agentControls = globalAgent.instance;
+    ['latest', 'v1227'].forEach(version => {
```
aryamohanan, line 58: Adding the previous version to the test, no other changes in the test file.

packages/core/src/tracing/instrumentation/cloud/azure/blob.js
```diff
@@ -16,7 +16,10 @@ let isActive = false;
 exports.spanName = 'azstorage';
 
 exports.init = function init() {
-  hook.onModuleLoad('@azure/storage-blob', instrumentBlob);
+  // Starting from v12.28.0, the package has been migrated to ESM.
+  // To support ESM-based applications, we now utilize the IITM hook.
+  // Reference: https://github.com/Azure/azure-sdk-for-js/pull/33329
+  hook.onModuleLoad('@azure/storage-blob', instrumentBlob, { nativeEsm: true });
```
aryamohanan, line 22: Here is the fix. ESM applications are now successfully traced using the IITM hook.

Ref: https://github.com/Azure/azure-sdk-for-js/pull/33329/files

abhilash-sivan (approved):
LGTM

PR #1885, opened 2025-07-30
fix(collector): reduced memory consumption in yaml reader component

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-35539


packages/core/src/util/yamlReader.js
```diff
@@ -35,12 +34,21 @@ exports.init = function init(config) {
  * @param {string} filePath - The absolute path to the YAML file.
  */
 exports.read = function read(filePath) {
+  // increases memory by ~4mb
```
kirrg001, line 37:
```suggestion
```

aryamohanan (approved):
Nice catch

PR #1888, opened 2025-07-31
test: limited ESM tests to run only latest version of the package

aryamohanan:
ESM  tests currently lack support for mocking imported modules. Because of this limitation, there’s no advantage in running ESM tests across multiple versions of a package.

Until proper mocking support is implemented (see ticket [INSTA-788](https://jsw.ibm.com/browse/INSTA-788)), ESM tests are intentionally restricted to the latest version only.

We've observed that some ESM tests are still running redundantly across the same version of a package. This logic is already correctly enforced in some other apps, and should be consistently applied here as well.


abhilash-sivan (approved):

kirrg001 (approved):
Thanks for adding the missing checks. Its not an ideal solution though in general. We should think about how to solve this differently as part of https://jsw.ibm.com/browse/INSTA-788

PR #1898, opened 2025-08-03
feat: added support for koa-router v14

instanacd:
ref https://jsw.ibm.com/browse/INSTA-48714

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -119,7 +119,7 @@
         "koa": "3.0.1",
         "koa-bodyparser": "4.4.1",
         "koa-morgan": "1.0.1",
-        "koa-router": "13.1.1",
+        "koa-router": "14.0.0",
         "lerna": "8.1.8",
         "lodash": "4.17.21",
         "log4js": "6.9.1",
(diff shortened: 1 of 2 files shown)


abhilash-sivan (approved):

kirrg001 (approved):

aryamohanan (approved):

kirrg001:
@abhilash-sivan 

I just noticed

>  npm warn deprecated koa-router@14.0.0: Please use @koa/router instead, starting from v9!

Could you please share your thoughts regarding this warning?

abhilash-sivan:
The effort required to add support for @koa/router is minimal. Since some customers may still be using the older package, we can either support both or decide internally which one to prioritise. The instrumentation is essentially the same and may only need minor changes. I’ve tested it locally, and the tests are also passing with @koa/router.

kirrg001:
The problem is a lot of customers will already use the new package, because koa-router is [deprecated](https://www.npmjs.com/package/koa-router) - since v9!

> The instrumentation is essentially the same and may only need minor changes.

Yes 👍 

> I’ve tested it locally, and the tests are also passing with @koa/router.

Please open this as a PR if you have modified it already.

We can finish it up as part of a JIRA ticket (committed label):

- Keep both koa-router and @koa/router in the instrumentation
- Only test against @koa/router

PR #1902, opened 2025-08-04
chore: updated version in .nvmrc to 22.18.0

instanacd:
Automatic weekly update to the latest **Node.js 22.18.0** release.


kirrg001 (approved):

.nvmrc
```diff
@@ -1 +1 @@
-22.17.1
+22.18.0
```
aryamohanan, line 1: Docker images not yet available https://hub.docker.com/_/node.
Since we're using this version as a base image across multiple places, let’s hold off on merging until the official image becomes available.

kirrg001: Hm this is not an ideal process. We have to either check that in the GH action or delay the PR in a way.

kirrg001:
> We have to either check that in the GH action

Closing PR. Will try to add an addition for that.

PR #1908, opened 2025-08-04
fix(collector): cleaned up request objects after agent host lookup

kirrg001:
Found while investigating #1886 

refs https://jsw.ibm.com/browse/INSTA-35539


aryamohanan (changes requested):
The tests appear to be failing on tekton because `req.destroy()` is being called from test, but this method does not exist on the `FakeRequest` class.

PR #1931, opened 2025-08-06
fix: handled spans with multiple data keys in backend mapper

aryamohanan:
Improved span data transformation logic to support cases where span.data contains multiple keys or where the span name (span.n) does not directly match any key in span.data. This ensures correct mapping for spans like 'node.http.server', 'node.http.client', and 'graphql.server', which may include nested structures such as span.data.http and span.data.graphql.


kirrg001 (approved):
Pre-approving (more tests required).

Commit suggestion:

chore: fixed filtering for multiple span keys #1931

Internal commit, because it has no affect on customers yet.

PR #1941, opened 2025-08-11
fix: handled ignoring endpoint configs with connections only

aryamohanan:
Previously, ignore configurations specifying only `connections` were not handled correctly, causing such filters to be ignored.

This fix updates the filtering check to include `connections` along with `methods` and `endpoints`, ensuring spans matching connection filters are properly ignored.

### DIFF
--- a/packages/core/src/util/spanFilter.js
+++ b/packages/core/src/util/spanFilter.js
@@ -139,28 +139,27 @@ function shouldIgnore(span, ignoreEndpointsConfig) {
   }
 
   return ignoreConfigs.some(ignoreconfig => {
-    // Advanced filtering:
-    // For e.g., for a Kafka, the ignoreconfig like { methods: ['consume'], endpoints: ['t1'] }.
-    if (typeof ignoreconfig === 'object') {
-      // Case where ignoreconfig does not specify any filtering criteria.
-      if (!ignoreconfig.methods && !ignoreconfig.endpoints && !ignoreconfig.connections) {
-        return false;
-      }
+    if (typeof ignoreconfig !== 'object') return false;
 
-      if (ignoreconfig.methods && !matchMethods(span, ignoreconfig)) {
-        return false;
-      }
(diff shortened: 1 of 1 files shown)


packages/core/src/util/spanFilter.js
```diff
@@ -143,7 +143,7 @@ function shouldIgnore(span, ignoreEndpointsConfig) {
     // For e.g., for a Kafka, the ignoreconfig like { methods: ['consume'], endpoints: ['t1'] }.
     if (typeof ignoreconfig === 'object') {
       // Case where ignoreconfig does not specify any filtering criteria.
-      if (!ignoreconfig.methods && !ignoreconfig.endpoints) {
+      if (!ignoreconfig.methods && !ignoreconfig.endpoints && !ignoreconfig.connections) {
```
kirrg001, line 146: This condition is dangerous, because we will always forget to extend it.
Why do we even need this?
The conditions below already check the presence of the config key?

packages/core/src/util/spanFilter.js
```diff
@@ -139,28 +139,27 @@ function shouldIgnore(span, ignoreEndpointsConfig) {
...
-
-      // extend more filtering cases in future
-      return true;
+    if (typeof ignoreconfig !== 'object') return false;
+
+    let matched = false;
+
+    if (ignoreconfig.methods) {
+      if (!matchMethods(span, ignoreconfig)) return false;
+      matched = true;
```
kirrg001, line 148: suggestion: 

if (matchMethods(span, ignoreconfig)) {
      matched = true;
}

kirrg001 (approved):

packages/core/src/util/spanFilter.js
aryamohanan: In this case, we intentionally keep the if (!match...) return false; pattern because when a config property is specified (e.g., methods), a mismatch should immediately stop further checks. We’re following an AND logic here; all specified conditions must match, so any single failure means no match.

PR #1958, opened 2025-08-22
test: reworked collector dependency for opentelemetry integration

kirrg001:
I had trouble in #1828, because there was again different npm versions from Otel dependencies. This approach is less error-prone. We install our latest collector as npm dependency and require it from the local folder. By that it will take the deps from the local node_modules folder.

Extracted from https://github.com/instana/nodejs/pull/1828


packages/collector/test/tracing/opentelemetry/fs-app.js
```diff
@@ -12,9 +12,17 @@ process.on('SIGTERM', () => {
...
+const expect = require('chai').expect;
+
 require('@instana/core/test/test_util/loadExpressV4');
 
-require('../../../src')();
+/**
+ * We install the latest version of the collector here locally.
+ * This ensures we are using the Opentelemetry production dependencies.
+ */
+expect(require.resolve('@instana/collector')).to.contain('opentelemetry/node_modules/@instana/collector');
```
kirrg001, line 23: This ensures that we require the local dependency in packages/collector/test/tracing/opentelemetry/node_modules.

kirrg001:
@instana/eng-nodejs 

Please review & merge this.
Then rebase https://github.com/instana/nodejs/pull/1828 against main.

aryamohanan (approved):
LGTM

PR #1977, opened 2025-09-01
chore: restrict OpenTelemetry package updates and allow other dependency updates

abhilash-sivan:
Otel packages can be updated together in a separate PR.  
Right now they block other dependency updates, so we’ll handle them separately later.


bin/dependencies/production/update-prod-dependencies.js
```diff
@@ -27,8 +27,8 @@ pkgPaths.forEach(obj => {
   const pkgJson = require(obj.pkgJsonAbsPath);
   const deps = pkgJson.dependencies || {};
   Object.entries(deps).forEach(([dep, version]) => {
-    // Exclude internal libraries
-    if (dep.startsWith('@instana')) return;
+    // Exclude internal libraries and OpenTelemetry packages
+    if (dep.startsWith('@instana') || dep.startsWith('@opentelemetry')) return;
```
aryamohanan, line 31: How do we update the otel packages then?

abhilash-sivan: I plan to introduce a new approach where all OpenTelemetry package updates will be grouped into a single PR using the production dependency bot. Currently, these Otel updates are blocking other dependency updates due to the weekly package update limit. see todo: https://jsw.ibm.com/browse/INSTA-48343

PR #1985, opened 2025-09-08
fix(collector): resolved PayloadTooLargeError

kirrg001:
> Request payload is too large. Will not send data to agent. (POST /com.instana.plugin.nodejs/profiles.2264580) PayloadTooLargeError: Request payload is too large. Will not send data to agent. (POST /com.instana.plugin.nodejs/profiles.2264580)

Investigating this error.

The payload size check was added in 2019:
https://github.com/instana/nodejs/commit/af2618b093b1d8ca76f9ccb5eb79a64a999f9017

[Profiles sensor](https://github.ibm.com/instana/sensors/blob/master/nodejs-profile-parent/sensor-nodejs-profile/src/main/java/com/instana/agent/nodejsprofile/sensor/NodejsProfile.java)

The payload agent/sensor size is: 49mb.

### DIFF
--- a/packages/collector/test/agentConnection_test.js
+++ b/packages/collector/test/agentConnection_test.js
@@ -121,7 +121,7 @@ mochaSuiteFn('agent connection', function () {
   });
 
   it('should throw error if size of payload exceeds maxContentLength', done => {
-    const bigData = 'a'.repeat(1024 * 1024 * 6);
+    const bigData = 'a'.repeat(1024 * 1024 * 21);
     const bigSpan = {
       n: 'span.exit',
       t: 'trace-id',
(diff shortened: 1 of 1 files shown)


packages/collector/src/agentConnection.js
```diff
@@ -384,7 +384,7 @@ exports.sendTracingMetricsToAgent = function sendTracingMetricsToAgent(tracingMe
  * @param {boolean} [ignore404]
  * @returns
  */
-function sendData(path, data, cb, ignore404 = false) {
+function sendData(path, data, cb, ignore404 = false, ignorePayloadSizeCheck = false) {
```
kirrg001, line 387: Its too much effort to refactor the function arguments. I continued with adding the parameter after the callback.
We should refactor that at one point.

PR #1997, opened 2025-09-15
refactor(core): added init and normalize for core config module

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-54638

- init provides the dependencies
- normalize is the actual action


packages/core/src/config/index.js
```diff
@@ -182,8 +175,8 @@ module.exports.init = (userConfig, _logger, defaultsOverride = {}) => {
     targetConfig = Object.assign({}, userConfig);
   }
 
+  // TODO: remove this and forward the logger via init fn.
```
kirrg001, line 178: This will be the next PR. We don't want to bind the logger to the config object.

PR #2011, opened 2025-09-17
refactor: moved agent config types to collector

kirrg001:
CollectorConfig should include AgentConfig. 
AgentOpts will be removed and moved into CollectorConfig.
This PR just moves the Types to Collector.


packages/collector/src/types/collector.d.ts
```diff
@@ -1,4 +1,21 @@
 import { GenericLogger } from '@instana/core/src/core';
+import { IgnoreEndpoints, Disable } from '@instana/core/src/config/types';
```
kirrg001, line 2: Moving AgentConfig into a ts declaration file requires to move its dependencies to other types too. But its so much better to use the types files anyway.

packages/core/src/config/types.d.ts
```diff
@@ -0,0 +1,16 @@
+export interface IgnoreEndpoints {
```
kirrg001, line 1: This is our new config types file. A good beginning for a better config module

PR #2012, opened 2025-09-17
refactor: removed agent/log.js and move the fn to agent connection

kirrg001:
From #2010


packages/collector/src/agent/log.js
```diff
@@ -1,54 +0,0 @@
...
-  const payloadObject = {
-    m: message.trim()
-  };
-  if (stackTrace) {
-    payloadObject.st = stackTrace.trim();
-  }
-
-  const payload = Buffer.from(JSON.stringify(payloadObject), 'utf8');
-
-  const req = http.request(
```
kirrg001, line 27: It makes 0 sense to have another request handler for the agent requests.
The fn belongs to agentConnection downstreamConnection obj.
This instance is already aware of the agent config host etc.

kirrg001:
Added test - the functionality was not covered at all

PR #2014, opened 2025-09-18
fix: resolved pino compatibility issues for v18

aryamohanan:
This PR pins `pino` to version **9.9.5**  to avoid compatibility issues in `v9.10.0`.

**Issue:**

The pipeline v18.18 ESM is failing due to the latest Pino update.

 It throws 
```
TypeError: diagChan.tracingChannel is not a function

```
 because tracingChannel() is not available in Node.js v18.18.
 
This method was added in the latest Pino release, [pino#2281](https://github.com/pinojs/pino/pull/2281) and diagnostics_channel.tracingChannel was introduced in Node.js starting from v18.19 (see [Node.js docs](https://nodejs.org/api/diagnostics_channel.html#diagnostics_channeltracingchannelnameorchannels)).

An [issue](https://github.com/pinojs/pino/issues/2289) has been created in the Pino repo regarding this incompatibility.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49178,7 +49178,7 @@
       "dependencies": {
         "@instana/core": "4.24.0",
         "@instana/shared-metrics": "4.24.0",
-        "pino": "^9.9.5",
+        "pino": "9.9.5",
         "semver": "^7.7.2",
         "serialize-error": "^8.1.0"
       },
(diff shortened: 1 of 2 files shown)


kirrg001 (approved):
LGTM

We need a blocker JIRA issue to revert this change 👍 

fix: resolved pino compatibility issues for v18

aryamohanan:
created card to revert the change https://jsw.ibm.com/browse/INSTA-55339

PR #2015, opened 2025-09-19
fix(http): correctly handle default ports for HTTP/HTTPS exit spans

aryamohanan:
**Issue:** Outgoing HTTPS requests without an explicit port were showing `:80` in the exit span URL. HTTP requests were not affected.

**Cause:** The span URL construction added the port unconditionally, without checking if it matched the default for the protocol.

 **Fix:** Now, the port is only appended if it is not the default: `443` for HTTPS, `80` for HTTP.

### DIFF
--- a/packages/core/src/tracing/instrumentation/protocols/httpClient.js
+++ b/packages/core/src/tracing/instrumentation/protocols/httpClient.js
@@ -336,16 +336,27 @@ function constructFromUrlOpts(options, self, forceHttps) {
 
   try {
     const agent = options.agent || self.agent;
-    const port = options.port || options.defaultPort || (agent && agent.defaultPort) || 80;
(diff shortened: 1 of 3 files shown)


packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -336,16 +336,27 @@ function constructFromUrlOpts(options, self, forceHttps) {
 
   try {
     const agent = options.agent || self.agent;
-    const port = options.port || options.defaultPort || (agent && agent.defaultPort) || 80;
+    const isHttps = forceHttps || options.protocol === 'https:' || (agent && agent.protocol === 'https:');
```
abhilash-sivan, line 339:
```suggestion
    const isHttps = forceHttps || options.protocol === 'https:' || agent?.protocol === 'https:';
```

We should use optional chaining as it is the new syntax!

PR #2022, opened 2025-09-22
test: replaced node-fetch with native fetch in test setup

aryamohanan:


packages/collector/test/tracing/logging/console/test.js
```diff
@@ -97,12 +97,7 @@ mochaSuiteFn('tracing/logging/console', function () {
     it('must trace error', () => runAndTrace('error', true, 'console.error - should be traced'));
     it('must trace timeout', () => runAndTrace('timeout', true, 'console.error - should be traced'));
     it('must trace exit span', () =>
-      runAndTrace(
-        'exit-span',
-        true,
-        'request to http://127.0.0.1:65212/ failed, reason: connect ' +
-          'ECONNREFUSED 127.0.0.1:65212 -- console.error - should be traced'
-      ));
+      runAndTrace('exit-span', true, 'fetch failed -- console.error - should be traced'));
```
aryamohanan: Native fetch and node-fetch represent the same failure; they are just surfaced differently.

**Native fetch:**
  - Throws a TypeError: fetch failed
  - The real cause is nested under error.cause

**node-fetch:**
  - Throws a FetchError
  - The connection error is exposed directly

packages/collector/test/tracing/misc/stack_trace/test.js
```diff
@@ -135,8 +135,8 @@ const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : descri
                   span => expect(span.n).to.equal('node.http.client'),
                   span => expect(span.k).to.equal(constants.EXIT),
                   span => expect(span.data.http.status).to.equal(201),
-                  span => expect(span.stack[2].m).to.equal('fetch'),
-                  span => expect(span.stack[2].c).to.contains('node-fetch')
+                  span => expect(span.stack).to.be.an('array'),
+                  span => expect(span.stack.length).to.be.greaterThan(0)
```
aryamohanan: The difference in the collected stack is that **native `fetch` is built into Node.js (via Undici, implemented in C++)**, so there are **no JavaScript frames to capture**, whereas **`node-fetch` is an external JavaScript library**, which adds extra frames to the stack.

aryamohanan:
> I believe this also needs to be removed!
> 
> https://github.com/instana/nodejs/blob/f1c571fa29d55c7b5d3aabd0801fae9727915a96/packages/collector/test/metrics/app/package.json#L21

@abhilash-sivan IMO, this is not part of the test setup code. It is a minimal test application (metrics-test-app) that exists solely to be test the metrics collection.

The purpose of the test is to verify that when this application runs with node-fetch as a dependency, the metrics system reports it correctly. Therefore, this dependency is not related to test setup and does not need to be replaced or removed the test explicitly requires an external dependency to validate the behavior.

packages/aws-lambda/test/lambdas/async.js
```diff
@@ -65,11 +63,13 @@ const handler = async event => {
 
   if (process.env.DETACHED_REQUEST) {
     setTimeout(async () => {
-      await fetch(downstreamDummyUrl, { headers: { 'X-Downstream-Header': 'yes' } });
+      const fetchResponse = await fetch(downstreamDummyUrl, { headers: { 'X-Downstream-Header': 'yes' } });
```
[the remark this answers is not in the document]
kirrg001: > Native fetch resolves when response headers arrive, not when the body finishes. The response body is streamed, and if it isn’t consumed, the tracing span is finalized too early and no trace is emitted.

Don't we have to add `await response.json();` to all apps?

We usually only do

```
fetch(`http://127.0.0.1:${agentPort}/ping`)
                .then(() => {
                  httpSuccess(res, data);
                })
```

abhilash-sivan (approved):
LGTM

packages/aws-lambda/test/lambdas/async.js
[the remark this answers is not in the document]
kirrg001: There is a bug/misbehavior in nativeFetch.

`finally` is triggered **with a delay** when using without `.text()` or `.json()`. This is not right IMO.

We already receive the `response.headers` in the `then` method.
I am quiet sure that the one who added `finally` intend to handle success or error in one place.

Similar to httpClient (please compare):
- we wait for response headers
- span.d + transmit is executed after the headers (not after the body)

packages/collector/test/tracing/misc/node-fetch/app.js
```diff
@@ -20,13 +20,18 @@ const express = require('express');
 const morgan = require('morgan');
 const bodyParser = require('body-parser');
 
+// From v3 onwards, node-fetch is a pure ESM module and does not support require().
+// To use it in a CommonJS environment, we load it via a dynamic import and
+// forward all arguments to the default export.
+// eslint-disable-next-line no-shadow
+const fetch = (...args) => import('node-fetch').then(({ default: fetch }) => fetch(...args));
```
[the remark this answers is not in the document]
aryamohanan: I’ve created a separate PR to clarify and extend the tests: #2264.

package.json
```diff
@@ -207,7 +207,7 @@
     "nats": "2.29.3",
     "nats-v1": "npm:nats@1.4.12",
     "node-fetch": "3.3.2",
-    "node-fetch-v2": "npm:node-fetch@2.7.0",
+    "node-fetch-v2.7.0": "npm:node-fetch@2.7.0",
```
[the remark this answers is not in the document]
aryamohanan: done

packages/collector/test/tracing/protocols/http/client/clientApp.js
```diff
@@ -146,15 +146,27 @@ app.get('/get-options-only', (req, res) => {
 });
 
 app.get('/timeout', (req, res) => {
-  fetch(`${baseUrl}/timeout`, {
+  // Native fetch doesn't support credentials in URL, we use Authorization header
+  const urlWithoutCreds = `${protocol}://localhost:${process.env.SERVER_PORT}/timeout`;
+  const auth = Buffer.from('user:password').toString('base64');
+
+  // Native fetch doesn't support timeout option, so we use AbortController
```
kirrg001: There is also `dispatcher` https://undici.nodejs.org/#/?id=undicifetchinput-init-promise

```
dispatcher: new Agent({
    keepAliveTimeout: 10,
    keepAliveMaxTimeout: 10
  })
```

packages/collector/test/tracing/logging/console/test.js
aryamohanan: Yea, its there

> TypeError: fetch failed
    at node:internal/deps/undici/undici:14900:13
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  [cause]: Error: connect ECONNREFUSED 127.0.0.1:65212
      at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1637:16)
      at TCPConnectWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
    errno: -61,
    code: 'ECONNREFUSED',
    syscall: 'connect',
    address: '127.0.0.1',
    port: 65212
  }
} console.error - should be traced

kirrg001: Then I guess there is a bug in the instrumentation?
See also https://github.com/instana/nodejs/pull/2022/files#r2675869932

packages/opentelemetry-exporter/test/test.js
```diff
@@ -207,15 +207,18 @@ mochaSuiteFn('Instana OpenTelemetry Exporter', function () {
 
 function verifySpans(spans, appControls) {
   // 1 x ENTRY /otel-test
-  // 1 x EXIT /internal-endpoint
   // 1 x tcp connect
   // 1 x ENTRY /internal-endpoint because this is the same server
+  // NOTE: Native fetch with OpenTelemetry does not create an EXIT span for the HTTP client call
```
kirrg001: https://github.com/instana/nodejs/pull/2278

packages/collector/test/tracing/protocols/http/server/test.js
```diff
@@ -471,7 +471,10 @@ function registerTests(agentControls, appUsesHttps, useHttp2CompatApi) {
         )
       ));
 
-  it(`must not collect credentials embedded in URLs (HTTPS: ${appUsesHttps})`, () =>
+  // Skipped this test because native fetch does not allow URLs with embedded credentials
+  // Error: "Request cannot be constructed from a URL that includes credentials"
+  // This is a security feature of the Fetch API standard
+  it.skip(`must not collect credentials embedded in URLs (HTTPS: ${appUsesHttps})`, () =>
```
kirrg001: We have a unit test for url.

>     it('must redact embedded credentials', () => {
      expect(sanitizeUrl('http://user:password@example.org/route')).to.equal(
        'http://<redacted>:<redacted>@example.org/route'
      );
    });
    
You can remove this test.
In general, I feel like native fetch does not offer a lot of features. This sucks a little bit.

kirrg001: I am not 100% sure we should replace the request library in ProcessControls as native fetch is poor with features.

packages/collector/test/tracing/protocols/http/proxy/test.js
```diff
@@ -197,13 +197,23 @@ const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : descri
           .then(() =>
             retry(() =>
               agentControls.getSpans().then(spans => {
-                expect(spans).to.have.lengthOf(1);
+                // Note: With native fetch, a client span is created even for malformed URLs
```
kirrg001: Thats nice. I think this test belongs to the nativeFetch test?
And node-fetch also should get this test. Both show different behaviors.

This is the proxy test - not sure why this test is here :D

aryamohanan: https://github.com/instana/nodejs/pull/2281

packages/collector/test/tracing/logging/console/test.js
aryamohanan: fixed

PR #2028, opened 2025-09-24

instanacd:
Tada!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "@aws-sdk/client-sns": "3.895.0",
         "@aws-sdk/client-sqs": "3.895.0",
         "@aws-sdk/client-sqs-v3": "npm:@aws-sdk/client-sqs@3.24.0",
-        "@aws-sdk/client-ssm": "3.893.0",
+        "@aws-sdk/client-ssm": "3.895.0",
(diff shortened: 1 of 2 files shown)


packages/collector/test/tracing/databases/prisma/test.js
```diff
@@ -32,7 +32,8 @@ describe('tracing/prisma', function () {
 
       // NOTE: require-mock is not working with esm apps. There is also no need to run the ESM APP for all versions.
       // TODO: Support for mocking `import` in ESM apps is planned under INSTA-788.
-      if (process.env.RUN_ESM && version !== 'latest') return;
+      // TODO: enable ESM tests against latest version
+      if (process.env.RUN_ESM) return;
```
aryamohanan, line 36: Please explain in a comment why this was skipped

abhilash-sivan: I considered skipping it and re-enabling it after the test is fixed. However, after verifying with a sample app, I found that the same app which works with ESM in version 6.15 does not work in 6.16. So, this approach doesn’t seem right. I’m therefore skipping the Prisma update for now and will address it as part of https://jsw.ibm.com/browse/INSTA-54386

PR #2040, opened 2025-10-03
feat: added support for pino v10

instanacd:
ref https://jsw.ibm.com/browse/INSTA-58047


aryamohanan (commented):
Let's wait for the currency jira card to be created

kirrg001:
It seems to me they dropped Node v18 without sharing that on the release notes.

https://github.com/pinojs/pino/compare/v9.13.1...v10.0.0

PR #2054, opened 2025-10-08
fix: OpenTelemetry instrumentation packages updates

instanacd:
**Changes**

- Load @opentelemetry/api from the app root to ensure all instrumentations and monkey patches share the same global API instance.

- Add peer dependency with bounded version range (>=1.3.0 <1.10.0) to enforce compatibility and avoid runtime issues like NonRecordingSpan.

- Updated all @opentelemetry instrumentations versions to latest

What This PR resolves

This PR addresses the multiple API instance issue that occurred when @opentelemetry/api was also installed at the application root. In such cases, our openTelemetry integretion was not functioning correctly before.

When the OTEL SDK and our collector are used together within the same application:

- If the OTEL SDK is initialized first, it now correctly captures all our instrumented libraries (including our OTEL). Previously, this was not happening. OTEL also works.

- If our collector is initialized first, it currently blocks OTEL tracing - this is a known issue and will be addressed separately.

This PR resolves the multiple instance problem. Tests demonstrating OTEL SDK and our collector behaviors will be included in a separate PR.

There is still issue when our collector initialized first.

References:

h[ttps://github.com/open-telemetry/opentelemetry-js/issues/4832](url)
https://github.com/open-telemetry/opentelemetry-js/issues/5454
https://jsw.ibm.com/browse/INSTA-59002

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -63,7 +63,7 @@
     "@opentelemetry/instrumentation-oracledb": "0.33.0",
     "@opentelemetry/instrumentation-restify": "0.53.0",
     "@opentelemetry/instrumentation-socket.io": "0.54.0",
-    "@opentelemetry/instrumentation-tedious": "0.13.0",
+    "@opentelemetry/instrumentation-tedious": "0.26.0",
     "@opentelemetry/sdk-trace-base": "1.25.0",
     "cls-bluebird": "^2.1.0",
     "import-in-the-middle": "2.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -11979,6 +11979,22 @@
         "@opentelemetry/api": "^1.3.0"
       }
     },
+    "node_modules/@opentelemetry/instrumentation-tedious": {
+      "version": "0.26.0",
+      "resolved": "https://registry.npmjs.org/@opentelemetry/instrumentation-tedious/-/instrumentation-tedious-0.26.0.tgz",
+      "integrity": "sha512-Z+r5EMhJqIqb6sWp0+zTI77HJ10yZOv6RUdh3U67D4KSIvpIU/WvhMa9OSoLwJSlNTXBPHJwhBpS88QVuV4OIg==",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@opentelemetry/instrumentation": "^0.207.0",
+        "@types/tedious": "^4.0.14"
+      },
+      "engines": {
+        "node": "^18.19.0 || >=20.6.0"
+      },
+      "peerDependencies": {
+        "@opentelemetry/api": "^1.3.0"
+      }
+    },
     "node_modules/@opentelemetry/instrumentation/node_modules/@opentelemetry/api-logs": {
       "version": "0.207.0",
       "resolved": "https://registry.npmjs.org/@opentelemetry/api-logs/-/api-logs-0.207.0.tgz",
@@ -16180,6 +16196,7 @@
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/@types/shimmer/-/shimmer-1.0.2.tgz",
       "integrity": "sha512-dKkr1bTxbEsFlh2ARpKzcaAmsYixqt9UyCdoEZk8rHyE4iQYcDCyvSjDSf7JUWJHlJiTtbIoQjxKh6ViywqDAg==",
+      "dev": true,
       "license": "MIT"
     },
     "node_modules/@types/sinon": {
@@ -16219,10 +16236,10 @@
       }
     },
     "node_modules/@types/tedious": {
-      "version": "4.0.12",
-      "resolved": "https://registry.npmjs.org/@types/tedious/-/tedious-4.0.12.tgz",
-      "integrity": "sha512-5NBYCLmidyXG3zxiBmR0beORRQcJOBoTKVL+9WaHQbX0E386UFXw6TSlY9/oxZDYqUWlBC98Funb83eJQt1aMw==",
-      "dev": true,
+      "version": "4.0.14",
+      "resolved": "https://registry.npmjs.org/@types/tedious/-/tedious-4.0.14.tgz",
+      "integrity": "sha512-KHPsfX/FoVbUGbyYvk1q9MMQHLPeRZhRJZdO45Q4YjvFkv4hMNghCWTvy7rdKessBsmtz4euWCWAB6/tVpI1Iw==",
+      "license": "MIT",
       "dependencies": {
         "@types/node": "*"
       }
@@ -26243,6 +26260,7 @@
       "version": "1.1.2",
       "resolved": "https://registry.npmjs.org/function-bind/-/function-bind-1.1.2.tgz",
       "integrity": "sha512-7XHNxH7qX9xG5mIwxkhumTox/MIRNcOgDrxWsMt2pAr23WHp6MrRlN7FBSFpCpr+oVO0F744iUgR82nJMfG2SA==",
+      "devOptional": true,
       "funding": {
         "url": "https://github.com/sponsors/ljharb"
       }
@@ -27591,6 +27609,7 @@
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/has/-/has-1.0.3.tgz",
       "integrity": "sha512-f2dvO0VU6Oej7RkWJGrehjbzMAjFp5/VKPp5tTpWIV4JHHZK1/BxbFRtf/siA2SWTe09caDmVtYYzWEIbBS4zw==",
+      "dev": true,
       "dependencies": {
         "function-bind": "^1.1.1"
       },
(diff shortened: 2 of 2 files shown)


packages/core/package.json
```diff
@@ -76,5 +75,8 @@
   },
   "overrides": {
     "debug": "^4.4.3"
+  },
+  "peerDependencies": {
```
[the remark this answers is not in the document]
aryamohanan: By default, npm will pick the latest version below 1.10.0, currently 1.9.0 and install it.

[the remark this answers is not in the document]
aryamohanan: > Customer sets "@opentelemetry/api: 1.2.0" and is using an older Opentelemetry setup. Our peerDependency is "> 1.3.0". Could you please check what will happen? Our SDK should not make the customers installation fail.

If a customer is using an older version @opentelemetry/api@1.2.0, npm  will block the installation by default and show a dependency resolution error, since 1.2.0 does not satisfy the required version range, but If we want to ensure installation does not fail hard, we could include @opentelemetry/api as a dependency as well 👍 

> How do you wanne manage the upper bound rule? (<1.10.0)

We can manage this by updating the version range whenever a new minor or major release of @opentelemetry/api is published, or earlier if a user reports an incompatibility or feature request.  they don't release api so often.

I referred this: https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/package.json#L87

packages/core/src/tracing/opentelemetry-instrumentations/wrap.js
```diff
@@ -93,25 +93,35 @@ module.exports.init = (_config, cls) => {
...
+   *
+   * Reference:
+   * https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/internal/global-utils.ts#L37
    */
+
+  const provider = new BasicTracerProvider();
+  api.trace.setGlobalTracerProvider(provider);
+
+  const contextManager = new AsyncHooksContextManager();
+  contextManager.enable();
```
aryamohanan: Reference: https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/context.md

aryamohanan:
This PR addresses the multiple API instance issue that occurred when @opentelemetry/api was also installed at the application root. In such cases, our openTelemetry integretion was not functioning correctly before.

When the OTEL SDK and our collector are used together within the same application:

- If the OTEL SDK is initialized first, it now correctly captures all our instrumented libraries (including our OTEL). Previously, this was not happening. OTEL also works.

- If our collector is initialized first, it currently blocks OTEL tracing - this is a known issue and needs to addressed separately.

This PR resolves the multiple  api instance problem. Tests demonstrating OTEL SDK and our collector behaviors will be included in a separate PR.

packages/collector/test/tracing/opentelemetry/fs-app.js
```diff
@@ -12,14 +12,7 @@ process.on('SIGTERM', () => {
...
 
-const expect = require('chai').expect;
-
-/**
- * We install the latest version of the collector here locally.
- * This ensures we are using the Opentelemetry production dependencies.
- */
-expect(require.resolve('@instana/collector')).to.contain('opentelemetry/node_modules/@instana/collector');
-require('@instana/collector')();
+require('../../../src')();
```
kirrg001: As far as I can see, we now no longer test against our core otel api version, because you now only load the api version from the root or?

What we want to keep IMO is:
- local test app installation of otel api (root) + collector (no root)
- local test app installation of collector only

IMO loading 

> require('../../../src')();

Has no real benefit for a production use case or?

packages/core/package.json
```diff
@@ -57,13 +57,13 @@
   },
   "homepage": "https://github.com/instana/nodejs/blob/main/packages/core/README.md",
   "dependencies": {
-    "@opentelemetry/api": "1.9.0",
+    "@opentelemetry/api": ">=1.3.0 <1.10.0",
```
kirrg001: Why do we need that? 🤔 
I think we should continue with a pinned version 1.9.0.

aryamohanan: Another point to note is that since we are always requiring the @opentelemetry/api package from the root, there’s no need to install it again within the Instana folder.

For example, if the root already has @opentelemetry/api version 1.8.0, it won’t install version 1.9.0 inside the Instana folder, as the dependency is already satisfied at the root level.

packages/core/src/tracing/opentelemetry-instrumentations/opentelemetryApi.js
```diff
@@ -0,0 +1,31 @@
...
+* Example structure:
+ * ├── node_modules/@opentelemetry/api/      # root (preferred)
+ * └── packages/service/node_modules/@opentelemetry/api/  # fallback
+ */
+let api;
+
+try {
+  api = require(require.resolve('@opentelemetry/api', { paths: [process.cwd()] }));
+} catch {
+  api = require('@opentelemetry/api');
```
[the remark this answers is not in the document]
aryamohanan: Yes, that was missing, I will extend the comment. ❤️

packages/core/package.json
kirrg001: Okay thank you. I understand the idea.

The initial idea of pinning exact version was to keep control for the integration. We only test and use specific versions of the otel dependencies to ensure the instrumentations work as we expect. 

So if npm only installs lets say 1.4.0 and we have to use this version, how do we know that our **integration** works? We only test against one single version (1.9.0).

If we test a matrix, we can do that 👍

packages/core/src/tracing/opentelemetry-instrumentations/wrap.js
aryamohanan: I’ve removed it from the change, but I’m not entirely sure the `AsyncHooksContextManager` actually worked in our case, it’s not enabled by default, and I’m not certain it was ever really used. 

We might want to consider switching to `AsyncLocalStorageContextManager` or remove the usage, though that needs further investigation and isn’t part of this PR.

Also, there’s an ongoing discussion in OpenTelemetry about deprecating `AsyncHooksContextManager`: https://github.com/open-telemetry/opentelemetry-js/issues/1405#issuecomment-1986106170

packages/core/package.json
aryamohanan: I think we can test both the lower and upper versions (1.3 and 1.9), since they’re in the same major version and shouldn’t have big breaking changes.

For compatibility:

* Supported versions are 1.3.0 to 1.9.x.
* If an unsupported version is at the root, it installs the closest compatible one in `core/node_modules`, but tracing might break because of multiple API instances.

So tracing will work fine as long as the root uses a supported `@opentelemetry/api` version.

packages/collector/test/tracing/opentelemetry/test.js
```diff
@@ -53,285 +53,686 @@ mochaSuiteFn('opentelemetry/instrumentations', function () {
...
+  // We run tests against multiple @opentelemetry/api versions to ensure compatibility.
+  //
+  // The core package declares a peer dependency range: ">=1.3.0 <1.10.0".
+  // This allows reusing the customer's existing API version if it falls within that range,
+  // instead of always installing our own copy. It helps avoid multiple conflicting instances.
+  //
+  // To verify stability across the supported range, we test with both v1.3.0 and v1.9.0:
+  // - v1.3.0 represents the lowest supported version.
+  // - latest is the latest verified release.
+  ['latest', 'v130'].forEach(version => {
```
aryamohanan: **Changes in Tests**

* Added a version range `['latest', 'v130']` so tests now run against both versions.
* Introduced an additional test to cover the OpenTelemetry SDK + Instana integration.
* Added a comment describing a known issue that occurs when using the SDK together with Instana tracing.
* Updated one span assertion in the Socket.IO tests — the span name was changed (see [opentelemetry-js-contrib#2495](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2495/files) and [instana/nodejs#2054](https://github.com/instana/nodejs/pull/2054#discussion_r2458959053)
* Other changes are related to linting.

aryamohanan:
### **Known Issues**

* **OpenTelemetry API version compatibility:**
  If a customer installs `@opentelemetry/api` with a version below `v1.3.0`, tracing will not function correctly.
  In such cases, installing `v1.3.0` at the project root and `v1.9.0` (latest) in our directory, the issue due to mismatched API instances and incomparable version
  Generally, this is not a valid setup since versions below `v1.3.0` are incompatible with most OpenTelemetry instrumentations.

* **OpenTelemetry SDK + Instana initialization order:**
  When both OpenTelemetry SDK and Instana are used together, and our collector is initialized first, tracing fails to work.
  This is a known limitation and needs to be addressed separately.

* **AsyncHooksContextManager usage:**
  The current use of `AsyncHooksContextManager` should be revisited. It may not be applied correctly or could be irrelevant in the existing setup.

packages/collector/test/tracing/opentelemetry/test.js
```diff
@@ -343,211 +744,113 @@ mochaSuiteFn('opentelemetry/instrumentations', function () {
...
     });
+  });
 
-    after(async () => {
-      await serverControls.stop();
-      await clientControls.stop();
+  describe('when otel sdk and instana is enabled', function () {
+    this.timeout(config.getTestTimeout() * 2);
+    before(async () => {
+      execSync('npm install --no-save --no-package-lock', {
```
kirrg001: Can you please add a `rm -rf node_modules..` execSync before?
Otherwise its hard to test this - core.tgz and collector.tgz should never be cached locally.
We always want a clean state.

kirrg001 (approved):
Super sweet

PR #2057, opened 2025-10-08
build: added an option to skip deps from prod update bot

aryamohanan:
Some prod dependencies (e.g., lru-cache, pino, etc.) require Node.js v20 or higher. Since we still support Node.js v18, these packages will be skipped from the production dependency update process.

We plan to drop Node.js v18 support in version 6 (2026). Until then, there’sno need to create bump PRs for these dependencies.

The list of skipped dependencies can be easily managed through theSKIP_DEPS variable in tekton. Update this list as needed to control which packages are excluded from automatic updates.

### DIFF
--- a/.tekton/listeners/prod-dependency-bot.yaml
+++ b/.tekton/listeners/prod-dependency-bot.yaml
@@ -28,6 +28,8 @@ spec:
       value: $(params.prod-deps-pr-limit)
     - name: org-pr-limit
       value: $(params.org-pr-limit)
+    - name: skip-deps
+      value: $(params.skip-deps)
     - name: pl-name
       value: "prod-dependency-bot-$(params.node-version)"
     - name: pipeline-ref
--- a/.tekton/pipeline/trigger-template.yaml
+++ b/.tekton/pipeline/trigger-template.yaml
@@ -58,6 +58,8 @@ spec:
       value: $(params.pr-payload)
     - name: org-pr-limit
       value: $(params.org-pr-limit)
+    - name: skip-deps
+      value: $(params.skip-deps)
 
   resourcetemplates:
     - apiVersion: v1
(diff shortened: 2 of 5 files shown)


.tekton/listeners/prod-dependency-bot.yaml
```diff
@@ -28,6 +28,8 @@ spec:
       value: $(params.prod-deps-pr-limit)
     - name: org-pr-limit
       value: $(params.org-pr-limit)
+    - name: skip-deps
```
kirrg001: The name is not really strong. We already have "prod-deps-pr-limit".

How about "prod-deps-skip"?

.tekton/pipeline/trigger-template.yaml
```diff
@@ -139,4 +141,6 @@ spec:
           - name: pr-payload
             value: $(params.pr-payload)
           - name: org-pr-limit
-            value: $(params.org-pr-limit)
\ No newline at end of file
+            value: $(params.org-pr-limit)
```
kirrg001: This should also be "prod-deps-org-pr-limit". Not part of the PR, but good to consider.

.tekton/listeners/prod-dependency-bot.yaml
aryamohanan: good suggestion. I will update

.tekton/pipeline/trigger-template.yaml
aryamohanan: Noted. Will update

kirrg001 (approved):
Pre-approving. The logic should work

bin/dependencies/production/update-prod-dependencies.js
```diff
@@ -40,6 +42,10 @@ pkgPaths.forEach(obj => {
 });
 
 Object.entries(dependencyMap).some(([dep, usageList]) => {
+  if (SKIP_DEPS.includes(dep)) {
```
kirrg001: I will try to add after this PR:

Support this format
"@opentelemetry/context-bla@2,x,y"

aryamohanan: I overlooked this PR. Will merge asap

Issue #2066, opened 2025-10-13
[Bug]: setLogger does not configure new Pino instance correctly


luchsamapparat:
### Problem Description

Hi,

We configured Pino to use the ECS format for logging. When I pass a Pino instance with that configuration to the Instana Collector, the configuration is not applied correctly.

The issue can be reproduced by running the example below in Node.js: `node ./instana.mjs`

As you can see from the log output below, my own log messages are logged with `"log.level":"info"`, while Instana's messages still use a numeric log level after changing the logger instance.

I debugged the issue and it seems this is where the configuration gets lost: [collector/src/logger.js:121](https://github.com/instana/nodejs/blob/main/packages/collector/src/logger.js#L121)

```
      parentLogger = uninstrumentedLogger(
        {
          ...parentLogger.levels,
          level: parentLogger.level || 'info',
          base: parentLogger.bindings(),
          timestamp: () => `,"time":"${new Date().toISOString()}"`
        },
        multiStream
      );
```

Only `levels`, `level` and `bindings` are taken from the provided Pino instance. However, the ECS format package configures more than only those properties.

### Instana Node.js Tracer Version

v4.26.0

### Node.js Version

v22.17.1

### Short, Self Contained Example

```js
import instanaCollector from '@instana/collector';

const instana = instanaCollector({
  level: 'debug'
});

import { ecsFormat } from '@elastic/ecs-pino-format';
import { pino } from 'pino';

const format = ecsFormat();
const logger = pino(format);

logger.info('Instana initialized');

instana.setLogger(logger);

logger.info('Instana logger configured');
```

### Logs & Error Messages

```
{"level":30,"time":"2025-10-13T09:55:45.452Z","threadId":0,"name":"@instana/collector","msg":"PID Store starting with pid 20652"}
{"log.level":"info","@timestamp":"2025-10-13T09:55:45.697Z","process.pid":20652,"host.hostname":"...","ecs.version":"8.10.0","message":"Instana initialized"}
{"level":30,"time":"2025-10-13T09:55:45.694Z","threadId":0,"name":"@instana/collector","msg":"@instana/collector module version: 4.26.0"}
...
{"log.level":"info","@timestamp":"2025-10-13T09:55:45.698Z","process.pid":20652,"host.hostname":"...","ecs.version":"8.10.0","message":"Instana logger configured"}
[...]

PR #2071, opened 2025-10-14
fix(collector): resolved extended formatting options for pino

kirrg001:
refs #2066 

```
{"log.level":"info","@timestamp":"2025-10-14T14:13:12.640Z","process.pid":56495,"host.hostname":"Mac.fritz.box","ecs.version":"8.10.0","message":"Instana initialized"}
{"log.level":"info","@timestamp":"2025-10-14T14:13:12.641Z","process.pid":56495,"host.hostname":"Mac.fritz.box","ecs.version":"8.10.0","message":"Instana logger configured"}
{"log.level":"warn","@timestamp":"2025-10-14T14:13:12.930Z","process.pid":56495,"host.hostname":"Mac.fritz.box","ecs.version":"8.10.0","module":"instana-nodejs-logger-parent","threadId":0,"message":"The Instana host agent cannot be reached via 127.0.0.1:42699 and the default gateway cannot be determined. Details: Error for the connection attempt: The attempt to connect to the Instana host agent on 127.0.0.1:42699 has failed with the following error: connect ECONNREFUSED 127.0.0.1:42699; error for determining the gateway: Failed to determine the default gateway: The file /proc/self/net/route does not exist. The Instana host agent might not be ready yet, scheduling another attempt to establish a connection in 60000 ms."}
```

It's recommended to configure the instana tracer with level error to avoid getting the first instana logs without the custom logger. This is a known issue -  see https://jsw.ibm.com/browse/INSTA-24679

```
const instana = instanaCollector({
  level: 'error'
});
```

### DIFF
--- a/packages/collector/src/logger.js
+++ b/packages/collector/src/logger.js
@@ -152,6 +152,8 @@ exports.init = function init(userConfig = {}) {
 
       parentLogger = uninstrumentedLogger(pinoOpts, multiStream);
     } catch (error) {
+      // TODO: If any error occurs, we don't forward any logs to the agent. Why?
+      //       https://jsw.ibm.com/browse/INSTA-59515
       parentLogger.debug(`An issue occurred while modifying the current logger: ${error.message}`);
     }
   } else if (parentLogger && parentLogger.addStream) {


packages/collector/test/tracing/logging/misc/controls.js
```diff
@@ -15,79 +15,125 @@ const config = require('../../../../../core/test/config');
...
-        break;
-      case 'receives-winston-logger':
-        appName = 'app-receives-winston-logger.js';
-        break;
-      case 'receives-log4js-logger':
-        appName = 'app-receives-log4js-logger.js';
-        break;
-      default:
-        throw new Error(`Unknown app: ${opts.instanaLoggingMode}`);
+class AppControls {
```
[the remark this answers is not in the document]
kirrg001: Follows syntax in ProcessControls.js.

packages/collector/test/tracing/logging/misc/test.js
```diff
@@ -19,74 +19,200 @@ mochaSuiteFn('tracing/logging/misc', function () {
...
+        await appControls.start();
+      });
+
+      afterEach(() => {
+        return appControls.stop();
       });
 
-      it('log calls are not traced', () => verifyInstanaLoggingIsNotTraced());
+      it('log calls are not traced', () => {
+        verifyInstanaLoggingIsNotTraced(appControls, { expectedAgentLogCount: 1 });
```
abhilash-sivan: The function is being called, but the test case completes before the result is actually returned, which causes a false positive (the test appears to pass even though it hasn’t verified the outcome).

Please ensure the function’s result is returned so that Mocha properly waits for the asynchronous operation to complete. For example:

```suggestion
        return verifyInstanaLoggingIsNotTraced(appControls, { expectedAgentLogCount: 1 });
```

It’s also possible that other test cases have a similar issue, so please review them as well.

packages/collector/test/tracing/logging/misc/test.js
```diff
@@ -19,74 +19,200 @@ mochaSuiteFn('tracing/logging/misc', function () {
...
+
+          await appControls.start();
+        });
+
+        afterEach(() => {
+          return appControls.stop();
+        });
+
+        it('log calls are traced', () => {
+          verifyLogSpansAreCreated(appControls, {
```
abhilash-sivan: Here also the same issue https://github.com/instana/nodejs/pull/2071/files#r2431281270

packages/collector/test/tracing/logging/misc/test.js
kirrg001: 👍

packages/collector/src/logger.js
```diff
@@ -118,16 +118,42 @@ exports.init = function init(userConfig = {}) {
...
+        const formatters = instance[formattersSym];
+
+        if (formatters) {
+          pinoOpts.formatters = formatters;
+        }
+      }
+
+      parentLogger = uninstrumentedLogger(pinoOpts, multiStream);
     } catch (error) {
+      // TODO: If any error occurs, we don't forward any logs to the agent. Why?
```
kirrg001: @instana/eng-nodejs FYI

Discovered this while working on this PR a lot. Added a ticket

PR #2072, opened 2025-10-15
fix: added single opentelemetry api instance for otel instrumentation

aryamohanan:
**Fix OpenTelemetry API duplication issues**

**Problem:**
Previously, our package was loading a different instance of `@opentelemetry/api` than the one used by the application or instrumentations.
This caused issues such as:

* `NonRecordingSpan` being returned unexpectedly.
* Tracer provider registration not affecting all instrumentations.
* Monkey-patched functions like `api.trace.setSpan` not being invoked correctly.

**Solution:**

1. Added `@opentelemetry/api` as a **peer dependency** in `package.json` with a bounded version range:

   ```json
   "peerDependencies": {
     "@opentelemetry/api": ">=1.3.0 <1.10.0"
   }
   ```

   This ensures the package relies on the app’s single global API instance.

2. Updated runtime imports to **explicitly resolve the root API instance** using `require.resolve` from the main module, guaranteeing that the monkey patch applies to the correct global singleton.

### DIFF
--- a/packages/core/src/tracing/opentelemetry-instrumentations/wrap.js
+++ b/packages/core/src/tracing/opentelemetry-instrumentations/wrap.js
@@ -111,12 +111,12 @@ module.exports.init = (_config, cls) => {
    */
   const uniqueApis = Array.from(new Set(Object.values(apiInstances)));
 
-  uniqueApis.forEach(apiInstance => {
-    const tracerProvider = new BasicTracerProvider();
-    const contextManager = new AsyncHooksContextManager();
+  const sharedProvider = new BasicTracerProvider();
+  const sharedContextManager = new AsyncHooksContextManager().enable();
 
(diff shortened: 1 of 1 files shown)


packages/core/package.json
```diff
@@ -76,5 +75,8 @@
   },
   "overrides": {
     "debug": "^4.4.3"
+  },
+  "peerDependencies": {
```
aryamohanan: Added @opentelemetry/api as a peer dependency with a bounded version range (>=1.3.0 <1.10.0) to ensure our package uses the same global API instance as the application and instrumentations. This prevents duplicate API instances, which could otherwise result in NonRecordingSpan and make our monkey patch fail. 

The version bounds also guarantee compatibility with the APIs our patch relies on while avoiding breaking changes from newer versions.

PR #2078, opened 2025-10-19
[Prod Dependency Bot] Bumped pino from 9.13.0 to 10.1.0

instanacd:
Tada!


kirrg001:
We can't update pino currently. See https://github.com/instana/nodejs/pull/2042
See https://jsw.ibm.com/browse/INSTA-48343 to ignore deps. Not implemented

PR #2083, opened 2025-10-20
fix(shared-metrics): re-generated prebuild for linux x64 glibc 115

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-59265
refs https://jsw.ibm.com/browse/INSTA-60126

Somehow this prebuild contained a python3 executable.
I checked the other Node.js versions for x64 glibc. Only 115 was affected.

- Unpack the prebuild for linux x64 glibc 115 on main
- See the binaries
- Compare on my branch


kirrg001:
Raised https://jsw.ibm.com/browse/INSTA-60323

PR #2092, opened 2025-10-24
fix(redis): resolved pool instrumentation to use instance instead of prototype

aryamohanan:
The instrumentation for pool was returning the prototype instead of the instance, which is wrong and will broke private-field access.

Redis v5.9.0 introduced [change](https://github.com/redis/node-redis/pull/3088/files), which modified how createClientPool initializes the client, causing private fields like #isOpen to become inaccessible when wrapped by our tracer.


packages/core/src/tracing/instrumentation/databases/redis.js
```diff
@@ -131,19 +131,17 @@ function instrument(redis) {
...
         const redisUrl = poolOptions?.url;
-        const redisPoolPrototype = Object.getPrototypeOf(redisPoolInstance);
 
-        shimAllCommands(redisPoolPrototype, redisUrl, false, redisCommandList);
+        shimAllCommands(redisPoolInstance, redisUrl, false, redisCommandList);
 
-        if (typeof redisPoolPrototype.multi === 'function') {
-          shimmer.wrap(redisPoolPrototype, 'multi', wrapMulti(redisUrl, false));
+        if (typeof redisPoolInstance.multi === 'function') {
+          shimmer.wrap(redisPoolInstance, 'multi', wrapMulti(redisUrl, false));
```
abhilash-sivan: Please correct me if I'm wrong. I understand we can either instrument the prototype or the instance itself. What exact advantage do we get by going with the instance? Are there any connection params passed in as private fields? Could you please explain?

aryamohanan: Yes, we could instrument either the prototype or the instance., but the prototype doesn’t have access to private fields. If our patched code doesn’t use any private fields, it will work,  but once the code that uses private fields, it throws an error.

The safest approach is to instrument the instance, which has full access to its private fields.

abhilash-sivan: Agree

kirrg001 (changes requested):
https://github.com/instana/nodejs/pull/2092/files#r2468858520

PR #2093, opened 2025-10-26
[Prod Dependency Bot] Bumped @opentelemetry/context-async-hooks from 1.25.0 to 2.2.0

instanacd:
Tada!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49224,7 +49224,7 @@
       "license": "MIT",
       "dependencies": {
         "@opentelemetry/api": "1.9.0",
-        "@opentelemetry/context-async-hooks": "1.25.0",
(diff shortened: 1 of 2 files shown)


kirrg001:
2.x.x only works with 18.19 because

> This new minimum allows the usage of 'module.register()' for better ESM instrumentation support.

We support > 18.

And @opentelemetry/context-async-hooks is used in our core otel integration, which adds a few core instrumentations to our tracer.

IMO its probably better to wait with this update. Is there any important update in 2.x.x, which puts pressure on this update?

aryamohanan:
Let’s reconsider this in the next major release when we drop support for Node.js v18. I’ll add a note to the release planning card.

PR #2096, opened 2025-10-27
fix: updated timestamp in internal logger from epoch time to ISO string

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-61665

### DIFF
--- a/packages/collector/src/logger.js
+++ b/packages/collector/src/logger.js
@@ -136,7 +136,6 @@ exports.init = function init(userConfig = {}) {
 
       // CASE: We copy any settings from the pino instance such as custom formats.
       if (symbols && Array.isArray(symbols)) {
-        // If there's a custom timestamp function from the userConfig logger  , we'll respect that
         const timeSym = symbols.find(sym => String(sym) === 'Symbol(pino.time)');
         // @ts-ignore
         const timestampFn = instance[timeSym];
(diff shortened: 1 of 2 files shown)


abhilash-sivan:
### Current default format:

Here the Instana logged time is in epoch format:
Screenshot 2025-12-17 at 12 02 26 AM

### After the changes:
1. The app uses no custom logger format
Screenshot 2025-12-17 at 12 07 14 AM



2. The app uses custom pino log anf time format, which is not affected by our change
Screenshot 2025-12-17 at 12 09 52 AM

 See our format: "time":"2025-12-16T18:39:22.006Z"

 See customer custom format for app logs: "time":"Tue, 16 Dec 2025 18:39:22 GMT"

packages/collector/src/logger.js
```diff
@@ -132,7 +135,8 @@ exports.init = function init(userConfig = {}) {
       const symbols = Object.getOwnPropertySymbols(instance);
 
       // CASE: We copy any settings from the pino instance such as custom formats.
-      if (symbols && Array.isArray(symbols)) {
+      if (userConfig.logger && symbols && Array.isArray(symbols)) {
```
kirrg001: Why was this change needed? 🤔

abhilash-sivan: Oh yes
`const instance = userConfig.logger || parentLogger;`

We fall back to our logger as instance. I have removed the condition now

packages/collector/src/logger.js
```diff
@@ -121,7 +123,8 @@ exports.init = function init(userConfig = {}) {
       const pinoOpts = {
         ...parentLogger.levels,
         level: parentLogger.level || 'info',
-        base: parentLogger.bindings()
+        base: parentLogger.bindings(),
+        timestamp: () => `,"time":"${new Date().toISOString()}"`
```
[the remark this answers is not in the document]
abhilash-sivan: There is a unit test for this:
https://github.com/instana/nodejs/pull/2096/changes#diff-401664c30893826ec114406103552839cdfb4396a0bc1fde7d13983e71608083R143

This verifies the format and fails if it does not match the ISO string.

PR #2105, opened 2025-10-28
chore: clarified pino implementation and extended test

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-23066

- see https://github.com/instana/nodejs/pull/2105#discussion_r2468756290

### DIFF
--- a/packages/collector/src/uninstrumentedLogger.js
+++ b/packages/collector/src/uninstrumentedLogger.js
@@ -18,6 +18,7 @@ const pinoWasRequiredBeforeUs = Object.keys(require.cache).some(key => key.inclu
 //       Furthermore, using "onModuleLoad" is much more clean because we don't have to rely on the precense of
 //       lib/tools.js.
 // SEE:  https://jsw.ibm.com/browse/INSTA-23066
+// SEE:  https://github.com/instana/nodejs/pull/2105/files#r2468756290
 
 // eslint-disable-next-line import/no-extraneous-dependencies, instana/no-unsafe-require
 const pino = require('pino').default;


packages/collector/src/uninstrumentedLogger.js
```diff
@@ -6,19 +6,22 @@
...
+//       BUT our pino require still loads pino into the internal node.js cache. Especially the files of the pino library
+//       such as lib/tools.js. These sub requires are not triggered anymore when pino is required the second time.
+//       That means if the customer is using the SAME npm version (our version), it won't get traced without our manual
+//       cache deletion below.
+//       If the customer has its own pino version, it will be traced correctly, because it will trigger "onFileLoad".
+// TODO: Consider migrating to "onModuleLoad". This will remove the need of having to manually delete the cache entry,
+//       because any later pino require will trigger the require hook again with "pino" as module name.
+//       Furthermore, using "onModuleLoad" is much more clean because we don't have to rely on the precense of
+//       lib/tools.js.
+// SEE:  https://jsw.ibm.com/browse/INSTA-23066
```
kirrg001: I have raised https://jsw.ibm.com/browse/INSTA-61459.

The new JIRA contains the new instrumentation for pino using `onModuleLoad` (see attachements). 

We can replace it whenever we want. 

Currently I don't see an urgent reason to replace the instrumentation. 
The tests proof that the current approach works fine, but with the downsides I have described in the comments:

> 
> Furthermore, using "onModuleLoad" is much more clean because we don't have to rely on the precense of lib/tools.js.

>  This will remove the need of having to manually delete the cache entry, because any later pino require will trigger the require hook again with "pino" as module name.

**There is a technical difference how the require cache works in Node.js.**

First require('pino') triggers Module._load and all sub requires within this library (such as require('lib/tools.js').

Second require('pino') triggers Module._load and NO sub requires because its already cached. But the general module require is triggered.

NOTE: Pino instrumentation is not active when we require pino in packages/collector/src/uninstrumentedLogger.js.

This explains the need to manually remove the cache entry completely for pino when using "onFileLoad".

kirrg001:
```suggestion
// SEE:  https://jsw.ibm.com/browse/INSTA-23066
// SEE:  https://github.com/instana/nodejs/pull/2105/files#r2468756290
```

aryamohanan: Thanks for the explanation. Make sense.

aryamohanan (approved):
LGTM.

abhilash-sivan (approved):

PR #2107, opened 2025-10-29
fix: added flag based check for restricting double instrumentation

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-61508


packages/core/src/tracing/instrumentation/databases/redis.js
```diff
@@ -46,6 +51,14 @@ function instrument(redis) {
   // NOTE: v4 no longer exposes the RedisClient. We need to wait till `createClient` get's called
   //       to get the instance of the redis client
   if (!redis.RedisClient) {
```
abhilash-sivan: we only need the restriction here for redis versions above 4, as v3 doesn't have this issue. so omitting the else case

PR #2112, opened 2025-10-31
chore: added ability to skip major production bumps

kirrg001:
We want to skip production bumps for major versions only.
Such as "Skip otel deps@2"

### DIFF
--- a/bin/dependencies/production/update-prod-dependencies.js
+++ b/bin/dependencies/production/update-prod-dependencies.js
@@ -5,10 +5,12 @@
 'use strict';
 
 const path = require('path');
+// eslint-disable-next-line import/no-extraneous-dependencies
+const semver = require('semver');
 const utils = require('../utils');
 
 const BRANCH = process.env.BRANCH;
-const SKIP_PUSH = process.env.SKIP_PUSH === 'true';
+const DRY_RUN = process.env.DRY_RUN === 'true';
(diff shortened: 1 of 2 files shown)


bin/dependencies/production/update-prod-dependencies.js
```diff
@@ -5,10 +5,12 @@
 'use strict';
 
 const path = require('path');
+// eslint-disable-next-line import/no-extraneous-dependencies
+const semver = require('semver');
 const utils = require('../utils');
 
 const BRANCH = process.env.BRANCH;
-const SKIP_PUSH = process.env.SKIP_PUSH === 'true';
+const DRY_RUN = process.env.DRY_RUN === 'true';
```
kirrg001: This is much simpler than using "SKIP_PUSH".

bin/dependencies/utils.js
```diff
@@ -253,12 +253,21 @@ exports.branchExists = (branchName, cwd) => {
  * @param {string} cwd - Current working directory
  * @param {boolean} isMainBranch - Whether the current branch is 'main'
  */
-exports.prepareGitEnvironment = (branchName, cwd, isMainBranch) => {
-  execSync('git checkout main', { cwd });
-  execSync('npm i --no-audit', { cwd });
+exports.prepareGitEnvironment = (branchName, cwd, isMainBranch, isDryRun) => {
```
kirrg001: Just added isDryRun in the target used fns.

PR #2115, opened 2025-11-03
Added ability to bump sub currency dependencies

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-61703

> node bin/dependencies/currency/bump-sub-currency-versions.js

Contains 2 commits:

1. The script.
2. The actual version bumps.


bin/clean-aws.js
```diff
@@ -8,7 +8,7 @@
 
 'use strict';
 
-const yargs = require('yargs/yargs');
```
kirrg001: After I have updated all the deps, there was TS warning.

> TS error for yargs - you can see that the lock file was updated caused by the version updates. Somehow yargs/yargs cannot be resolved anymore on TS level. Fixed by just using yargs. I could not find a hint why yargs/yargs was needed.

It still works fine

> $ node bin/clean-aws.js --help
Options:
      --help                 Show help                                 [boolean]
      --version              Show version number                       [boolean]
  -s, --service              s3, sqs, dynamodb, kinesis, sns
                                                        [string] [default: "s3"]
  -d, --dry                  Use --dry=true to only print the items that would
                             be deleted instead of actually deleting them, off
                             by default.              [boolean] [default: false]
      --skipDateCheck, --sd  The script will only delete data older than 1 day.
                             You can skip this condition.
                                                      [boolean] [default: false]
  -c, --criteria             Matching string for the item name.         [string]

PR #2117, opened 2025-11-04
test: removed testing against specific aws sdk v3 version

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-61703

- I could not find a reason in the GH history why we do that
- It was added here https://github.com/instana/nodejs/commit/30982f8db0ecaba5218150421bc05cb8b534186b
- I remember at this time we started with multi version testing. 
- But I don't see a reason to keep them.
- Most of the multi version testing where added because of instrumentation changes
- Plus: its good to test against multiple major versions
- But here we test with just another v3 version
- Maybe related to how AWS structured their AWS SDK? I dunno


packages/collector/test/tracing/cloud/aws-sdk/v3/sqs/receiver.js
```diff
@@ -15,7 +15,7 @@ const mock = require('@instana/core/test/test_util/mockRequire');
 
 /**
  * NOTE:
- * Link e.g. @aws-sdk/client-sqs-v3 to @aws-sdk/client-sqs
```
kirrg001: No need to remove the construct for multi version testing.

package.json
```diff
@@ -85,17 +85,15 @@
...
     "@aws-sdk/client-kms": "3.920.0",
     "@aws-sdk/client-lambda": "3.920.0",
     "@aws-sdk/client-s3": "3.920.0",
-    "@aws-sdk/client-s3-v3": "npm:@aws-sdk/client-s3@3.321.1",
     "@aws-sdk/client-sns": "3.920.0",
     "@aws-sdk/client-sqs": "3.920.0",
-    "@aws-sdk/client-sqs-v3": "npm:@aws-sdk/client-sqs@3.24.0",
     "@aws-sdk/client-ssm": "3.920.0",
     "@aws-sdk/lib-dynamodb": "3.920.0",
+    "@aws-sdk/middleware-retry": "^3.370.0",
```
kirrg001: The dependency is used in sns utils: packages/collector/test/tracing/cloud/aws-sdk/v3/sns/utils.js

And somehow these older aws depenedencies provided this dependency to us.
Now that we removed them, we have to manually install.

package.json
```diff
@@ -85,17 +85,15 @@
...
     "@aws-sdk/client-kms": "3.920.0",
     "@aws-sdk/client-lambda": "3.920.0",
     "@aws-sdk/client-s3": "3.920.0",
-    "@aws-sdk/client-s3-v3": "npm:@aws-sdk/client-s3@3.321.1",
     "@aws-sdk/client-sns": "3.920.0",
     "@aws-sdk/client-sqs": "3.920.0",
-    "@aws-sdk/client-sqs-v3": "npm:@aws-sdk/client-sqs@3.24.0",
     "@aws-sdk/client-ssm": "3.920.0",
     "@aws-sdk/lib-dynamodb": "3.920.0",
+    "@aws-sdk/middleware-retry": "3.370.0",
```
kirrg001: The dependency is used in sns utils: packages/collector/test/tracing/cloud/aws-sdk/v3/sns/utils.js

And somehow these older aws depenedencies provided this dependency to us.
Now that we removed them, we have to manually install.

.gitignore
```diff
@@ -10,6 +10,7 @@ coverage
 currency-report.md
 *.zip
 *instruction*.md
+eslint-report.json
```
kirrg001: Will extract directly to main before merge.

PR #2120, opened 2025-11-05
test: refactored test version checks for restify with clearer condition

aryamohanan:
- replaced combined supportedVersion check with direct semver comparison
- improved readability of the version check


packages/collector/test/tracing/opentelemetry/test.js
```diff
@@ -27,11 +27,6 @@ const globalAgent = require('../../globalAgent');
 const DELAY_TIMEOUT_IN_MS = 500;
 const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : describe.skip;
 
-const runTests =
```
aryamohanan: Moved restify-specific version check into its own test and cleaned up naming

kirrg001 (approved):

PR #2135, opened 2025-11-10
chore(prod-deps): corrected skip logic and improved version range handling

aryamohanan:
There are 2 issues in the prod-dep update logic

- Fixed skip pattern matching for major-version entries (e.g. "pino@10")
- Enhanced range handling in cleanVersionString:
  - ">=1.3.0 <1.10.0" → "1.9.0"
  - ">=1.3.0 <=1.10.0" → "1.10.0"

### DIFF
--- a/bin/dependencies/utils.js
+++ b/bin/dependencies/utils.js
@@ -393,6 +393,11 @@ exports.cleanVersionString = version => {
 
   // "<"  returns one minor version below the upper bound (e.g. "1.9.0")
   // "<=" returns the upper bound itself (e.g. "1.10.0")
+  // TODO: Fix version handling in installPackage().
+  // Currently, npm installs the latest version instead of respecting the defined upper bound.
+  // Example: "@opentelemetry/api": ">=1.3.0 <1.10.0"
+  // If version 1.10.0 exists, npm installs it instead of adhering to the intended limit.
+  // The range should instead be updated to "<1.11.0" to maintain correct behavior.
   const upperMatch = version.match(/<\s*=?\s*(\d+\.\d+\.\d+)/);
   if (upperMatch) {
     const upper = semver.coerce(upperMatch[1]).version;


bin/dependencies/production/update-prod-dependencies.js
```diff
@@ -75,7 +75,7 @@ Object.entries(dependencyMap).some(([dep, usageList]) => {
     // CASE: PROD_DEPS_SKIP="pino,@opentelemetry/instrumentation-tedious@2" (to skip all v2 updates)
     if (
       PROD_DEPS_SKIP.some(skipEntry => {
-        const matched = skipEntry.match(/^(.*)@(\d)$/);
+        const matched = skipEntry.match(/^(.*)@(\d+)$/);
```
aryamohanan: (\d) only single-digit majors like, @1

bin/dependencies/utils.js
```diff
@@ -369,10 +369,29 @@ exports.installPackage = options => {
...
 
 /**
- * Cleans a version string by removing non-numeric characters except dots
+ * Cleans and normalizes a version string into a valid semver value
  * @param {string} version - The version string to clean
  * @returns {string} Cleaned version string
  */
 exports.cleanVersionString = version => {
-  return version.replace(/[^0-9.]/g, '');
+  if (semver.valid(version)) return version;
```
aryamohanan: Recent updates to `@opentelemetry/api` introduced a version range (>=1.3.0 <1.10.0) that caused the dependency update script to fail. 

This change improves version parsing and normalization logic to correctly resolve the highest supported version without throwing errors.

kirrg001: I see. But the actual package.json update will be wrong too, right?

If e.g. 1.11 is out, it would do:

> (>=1.3.0 <1.10.0)
> 1.11.0

Not?

aryamohanan: Yes, this PR does not include a fix for that issue; it only ensures the production build pipeline passes.

kirrg001: Hm yeah I understand. The logic seems really wild. 

Can't we just do

> .match(/(\d+\.\d+\.\d+)/g)

aryamohanan: let me try this.

PR #2321, opened 2026-02-02
fix(aws-lambda): reduced cold start duration

aryamohanan:
This PR avoids lazy loading OpenTelemetry instrumentations in the AWS Lambda environment.

In AWS Lambda, lazy loading causes OpenTelemetry instrumentations to be initialized during the first invocation, which adds **1900–2100 ms** of latency to cold starts and increases billable execution time.

To address this, instrumentations are preloaded during the Lambda initialization step instead of being loaded lazily at runtime.

With this change, cold start overhead is significantly reduced, bringing initialization time down from approximately **1900 ms to 500 ms**, resulting in faster executions and lower costs in AWS Lambda environments.

Ref: https://ibm.ent.box.com/notes/2107188531951
https://jsw.ibm.com/browse/INSTA-74767


packages/aws-lambda/src/wrapper.js
```diff
@@ -32,6 +36,9 @@ coreConfig.init(logger);
 let config = coreConfig.normalize({}, lambdaConfigDefaults);
 let coldStart = true;
 
+// Preload OpenTelemetry instrumentations to avoid lazy loading overhead
+instanaCore.tracing.otelInstrumentations.preload();
```
kirrg001: Can't we solve this with a config? 🤔

kirrg001 (commented):
Commit suggestion:

> fix(aws-lambda): reduced cold start duration

kirrg001 (approved):
LGTM

Lets wait with the merge because of #2306  👍

PR #2331, opened 2026-02-09
feat(bunyan): added support for error.cause 

aryamohanan:
Enhanced bunyan instrumentation to support Node.js error.cause feature.

Same did for console, see https://github.com/instana/nodejs/pull/2286

ref: https://jsw.ibm.com/browse/INSTA-71403

### DIFF
--- a/packages/core/src/tracing/instrumentation/logging/bunyan.js
+++ b/packages/core/src/tracing/instrumentation/logging/bunyan.js
@@ -67,27 +67,17 @@ function instrumentedLog(ctx, originalLog, originalArgs, markAsError) {
 
     if (typeof fields === 'string') {
       message = fields;
-    } else if (fields && typeof fields.message === 'string' && typeof message === 'string') {
-      message = `${fields.message} -- ${message}`;
+    } else if (fields instanceof Error) {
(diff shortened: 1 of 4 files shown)


packages/core/src/tracing/instrumentation/logging/bunyan.js
```diff
@@ -67,23 +67,30 @@ function instrumentedLog(ctx, originalLog, originalArgs, markAsError) {
 
     if (typeof fields === 'string') {
       message = fields;
+    } else if (fields instanceof Error && typeof message === 'string') {
+      // CASE: logger.error(new Error('msg'), 'another msg')
+      const errorMessage = tracingUtil.extractErrorMessage(fields);
+      message = `${errorMessage} -- ${message}`;
+    } else if (fields instanceof Error) {
```
kirrg001: I think we can improve the readability here a lot. 

```js
const x = fields.err || fields.

+

if (fields instanceof Error) {
  const errorMessage = tracingUtil.extractErrorMessage(fields);
  
   if (typeof message === 'string')
   ....
}
```

abhilash-sivan (approved):

PR #2335, opened 2026-02-12
fix: handled non-promise return from instrumented pg call

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-70534

This PR is an extension of the following case: https://github.com/instana/nodejs/pull/2249

This is a split part of the following PR: https://github.com/instana/nodejs/pull/2283

### DIFF
--- a/packages/core/src/tracing/instrumentation/databases/pg.js
+++ b/packages/core/src/tracing/instrumentation/databases/pg.js
@@ -97,7 +97,7 @@ function instrumentedQuery(ctx, originalQuery, argsForOriginalQuery) {
           return error;
         });
     } else {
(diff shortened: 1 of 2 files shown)


packages/core/src/tracing/instrumentation/databases/pg.js
```diff
@@ -95,6 +97,9 @@ function instrumentedQuery(ctx, originalQuery, argsForOriginalQuery) {
           finishSpan(error, span);
           return error;
         });
+    } else {
+      tracingUtil.handleUnexpectedReturnValue(promise, 'pg', 'query');
```
kirrg001:
```suggestion
      tracingUtil.handleUnexpectedReturnValue(promise, exports.spanName, 'query');
```

kirrg001 (approved):

PR #2351, opened 2026-02-18
test: support package@all format to run all versions

aryamohanan:
Add support for running all versions of a package using the format:`package@all`

Example usage:
`runcollector express@all`

This eliminates the need to run each version separately.


abhilash-sivan (approved):
Nice 🥇

kirrg001 (approved):

PR #2356, opened 2026-02-20
test: ensured test folders regenerate before mocha validation

aryamohanan:
Implemented test folder regeneration before mocha validation with a reusable approach that's called from both `initEnv` and `run-test`


abhilash-sivan (approved):
tested, working 👍

abhilash-sivan (approved):

kirrg001 (approved):

PR #2373, opened 2026-02-24
test: cleaned up graphql subscription

kirrg001:
While working on #2306 I was splitting graphql from graphql subscriptions tests.
I noticed some left overs. Tidying this up as part of this PR.

Only moved the last tests over to graphql-subscriptions. No modifications.


aryamohanan (approved):
LGTM

PR #2422, opened 2026-03-24
test: updated prisma test dependencies to fetch the latest versions, aligned with v7

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-76669


kirrg001 (approved):
This is a first great step.
It seems we can finally update to the latest prisma client version.
This avoids a manual update of this file.

Next step: Investigation to get rid of the multiple pkg.json files.

aryamohanan (approved):

PR #2423, opened 2026-03-24
chore: fixed AWS resource cleanup script

aryamohanan:
- Migrate from deprecated AWS SDK v2 to v3 to eliminate deprecation warnings
- Change default behavior to clean ALL services instead of just S3
- Add debug logging to show sample items and matching statistics

image


kirrg001:
> total 181 s3 buckets and delete only resources that matches our criteria.

I think the output is confusing. 181 buckets company wide?

kirrg001 (approved):

PR #2436, opened 2026-03-27
refactor(core): normalized secret configuration

aryamohanan:
The secret config is scattered in index file so moved to a specific file


aryamohanan:
we decided not to do this now,

PR #2454, opened 2026-03-31
chore: switched to official s390x prebuild image

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-42185


aryamohanan (approved):
Preapproving

PR #2464, opened 2026-04-06
chore: corrected currency update logic to prevent mixing major and minor update

aryamohanan:
There are a couple of issues:

1. When a major update is applied, minor updates are also being included.
2. When multiple major updates occur, it additionally includes the major version updates from packages in the previously created PR, along with other minor updates.

see this PR https://github.com/instana/nodejs/pull/2461

### DIFF
--- a/bin/dependencies/currency/update-currencies.js
+++ b/bin/dependencies/currency/update-currencies.js
@@ -84,6 +84,7 @@ currencies.forEach(currency => {
 
     // 1. Reset currencies.json to discard changes from previous iterations of the loop.
     if (!DRY_RUN) {
+      execSync('git checkout main', { cwd });
       execSync('git checkout -- currencies.json', { cwd });
     }
 
--- a/bin/dependencies/currency/update-currencies.js
+++ b/bin/dependencies/currency/update-currencies.js
@@ -113,15 +113,15 @@ currencies.forEach(originalCurrency => {
   const newVersionObj =
     typeof installedVersionObj === 'string' ? latestVersion : { ...installedVersionObj, v: latestVersion };
 
-  if (isMajorUpdate) {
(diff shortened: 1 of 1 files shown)


abhilash-sivan (approved):

abhilash-sivan (approved):

bin/dependencies/currency/update-currencies.js
```diff
@@ -18,94 +18,109 @@ const cwd = path.join(__dirname, '..', '..', '..');
...
-    console.log(
-      `Skipping ${currency.name}. Installed version is ${installedVersion}. Latest version is ${latestVersion}`
-    );
+    console.log(`[UP-TO-DATE] ${currency.name} already up-to-date (${installedVersion})`);
     return;
   }
 
-  console.log(`Latest version: ${latestVersion}`);
-  console.log(`Installed version: ${installedVersion}`);
+  const isMajorUpdate = semver.major(latestVersion) !== semver.major(installedVersion);
```
aryamohanan: This condition was wrong earier

bin/dependencies/currency/update-currencies.js
```diff
@@ -18,94 +18,109 @@ const cwd = path.join(__dirname, '..', '..', '..');
...
 
-currencies.forEach(currency => {
-  console.log(`Checking currency update for ${currency.name}`);
+currencies.forEach(originalCurrency => {
+  let currency = originalCurrency;
+
+  if (MAJOR_UPDATES_MODE) {
+    utils.prepareGitEnvironment('main', cwd, true, DRY_RUN);
+
+    currencies = loadCurrencies();
```
aryamohanan: We load the currencies file fresh, else it will load from the altered one.

PR #2476, opened 2026-04-08
refactor(core): adapted optional object parameters in config normalizer functions

abhilash-sivan:
refactored fn params as objects


aryamohanan (approved):
❤️

aryamohanan:
commit suggestion: 
> refactor(core): used optional object parameters in config normalizer functions

something similiar.

kirrg001 (approved):

PR #2479, opened 2026-04-08
refactor(collector): moved resolve string config to core utility

aryamohanan:
moved string resolve config to core utility

image

### DIFF
--- a/packages/collector/src/types/collector.d.ts
+++ b/packages/collector/src/types/collector.d.ts
@@ -29,7 +29,7 @@ export interface CollectorConfig {
     stackTraceLength?: number;
     [key: string]: any;
   };
-  autoProfile?: boolean | string;
+  autoProfile?: boolean;
   reportUnhandledPromiseRejections?: boolean;
   logger?: GenericLogger;
   level?: string | number;
--- a/packages/collector/src/util/normalizeConfig.js
+++ b/packages/collector/src/util/normalizeConfig.js
@@ -80,11 +80,11 @@ function normalizeAgentRequestTimeout(userConfig, defaultConfig) {
 
 /**
  * @param {import('../types/collector').CollectorConfig} userConfig
- * @param {{ autoProfile: string | boolean }} defaultConfig
- * @returns {string | boolean}
+ * @param {{ autoProfile: boolean }} defaultConfig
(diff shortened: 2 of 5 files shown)


packages/collector/src/util/normalizeConfig.js
```diff
@@ -79,7 +84,12 @@ function normalizeAgentRequestTimeout(userConfig, defaultConfig) {
  * @returns {string | boolean}
  */
 function normalizeAutoProfile(userConfig, defaultConfig) {
-  return resolveConfig(process.env.INSTANA_AUTO_PROFILE, userConfig.autoProfile, defaultConfig.autoProfile);
+  return util.resolveConfig({
```
aryamohanan: This was missed refactoring. Moved the resolve method to core util

packages/core/src/config/util.js
```diff
@@ -183,3 +183,25 @@ exports.resolveBooleanConfigWithTruthyEnv = function resolveBooleanConfigWithTru
...
 };
+
+/**
+ * @param {Object} params
+ * @param {any} params.envVar
+ * @param {any} params.configValue
+ * @param {any} params.defaultValue
+ * @param {string} [params.configPath]
+ */
+exports.resolveConfig = function resolveConfig({ envVar, configValue, defaultValue, configPath }) {
```
kirrg001: How about we give it a more specific name?

kirrg001 (approved):
Pre-approving

packages/core/src/config/util.js
aryamohanan: renamed to `resolveStringConfig`

packages/collector/src/types/collector.d.ts
```diff
@@ -29,7 +29,7 @@ export interface CollectorConfig {
     stackTraceLength?: number;
     [key: string]: any;
   };
-  autoProfile?: boolean | string;
+  autoProfile?: boolean;
```
aryamohanan, line 32: `autoProfile` config expect true or fale and is always boolean. The type has been updated to boolean. The   environment value parsing has also been handled.

aryamohanan: TBD

PR #2486, opened 2026-04-13
chore(config): fixed INSTANA_TRACING_DISABLE env var handling

aryamohanan:
There were two issues with how `INSTANA_TRACING_DISABLE` was handled:

1. **ProcessControls behavior**
   `ProcessControls` was setting `INSTANA_TRACING_DISABLE=false` for all tests where `tracingEnabled=true` (which is the default). Based on the precedence rules in `disable.js`, setting `INSTANA_TRACING_DISABLE` to `'false'` overrides `INSTANA_TRACING_DISABLE_INSTRUMENTATIONS` and `INSTANA_TRACING_DISABLE_GROUPS`, thereby blocking more granular disable configurations.
   This has been corrected so that `ProcessControls` now sets `INSTANA_TRACING_DISABLE='true'` only when `tracingEnabled=false`, and leaves it unset when `tracingEnabled=true`.

2. **normalizeTracingEnabled fix**
   `normalizeTracingEnabled` has been updated to consider `INSTANA_TRACING_DISABLE` for the `tracing.enabled` configuration only when its value is `'true'` or `'false'`. If the variable contains instrumentation or group names (for example, `'logging'`), it is now handled separately by the disable normalizer.

### DIFF
--- a/packages/collector/test/test_util/ProcessControls.js
+++ b/packages/collector/test/test_util/ProcessControls.js
@@ -151,7 +151,6 @@ class ProcessControls {
         APP_CWD: this.cwd,
         INSTANA_AGENT_PORT: agentPort,
         INSTANA_LOG_LEVEL: 'warn',
-        INSTANA_TRACING_DISABLE: !this.tracingEnabled,
         INSTANA_FORCE_TRANSMISSION_STARTING_AT: '1',
         INSTANA_FULL_METRICS_INTERNAL_IN_S: 1,
         INSTANA_FIRE_MONITORING_EVENT_DURATION_IN_MS: 500,
@@ -163,6 +162,13 @@ class ProcessControls {
       },
       opts.env
     );
+
+    // Only set INSTANA_TRACING_DISABLE when tracing is actually disabled to avoid
+    // overriding other disable environment variables (INSTANA_TRACING_DISABLE_INSTRUMENTATIONS, etc.)
+    // See packages/core/src/config/configNormalizers/disable.js for precedence rules
+    if (!this.tracingEnabled) {
+      this.env.INSTANA_TRACING_DISABLE = 'true';
+    }
 
     if (this.usePreInit) {
       this.env.INSTANA_EARLY_INSTRUMENTATION = 'true';
--- a/packages/core/src/config/configNormalizers/disable.js
+++ b/packages/core/src/config/configNormalizers/disable.js
@@ -31,12 +31,10 @@ exports.normalize = function normalize(config) {
 
     if (envDisableConfig !== null) {
       if (envDisableConfig === true) {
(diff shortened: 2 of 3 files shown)


packages/collector/test/test_util/ProcessControls.js
```diff
@@ -151,7 +151,6 @@ class ProcessControls {
         APP_CWD: this.cwd,
         INSTANA_AGENT_PORT: agentPort,
         INSTANA_LOG_LEVEL: 'warn',
-        INSTANA_TRACING_DISABLE: !this.tracingEnabled,
```
aryamohanan, line 154: This was setting INSTANA_TRACING_DISABLE=false for all tests where tracingEnabled=true (the default). According to the precedence rules in disable.js, when INSTANA_TRACING_DISABLE is set to 'false', it take precedence over INSTANA_TRACING_DISABLE_INSTRUMENTATIONS and INSTANA_TRACING_DISABLE_GROUPS.

This caused logging integration tests to fail when trying to disable specific instrumentations using INSTANA_TRACING_DISABLE_INSTRUMENTATIONS
or INSTANA_TRACING_DISABLE_GROUPS.

aryamohanan: The case where `INSTANA_TRACING_DISABLE=false` was not handled in `disable.js` earlier; corrected the handling in this change: [https://github.com/instana/nodejs/pull/2478/changes#diff-2a559c2e74bbdf92a6ccc7a7dd679ea3eee0a586cd2490a7e2ef25bc3d31e51aR130](https://github.com/instana/nodejs/pull/2478/changes#diff-2a559c2e74bbdf92a6ccc7a7dd679ea3eee0a586cd2490a7e2ef25bc3d31e51aR130)

packages/core/src/config/configNormalizers/disable.js
```diff
@@ -31,12 +31,10 @@ exports.normalize = function normalize(config) {
 
     if (envDisableConfig !== null) {
       if (envDisableConfig === true) {
-        logger?.debug('[config] env:INSTANA_TRACING_DISABLE = true');
```
aryamohanan, line 34: we already logged this, removed duplicate logging

abhilash-sivan (approved):

packages/collector/test/test_util/ProcessControls.js
kirrg001: > when INSTANA_TRACING_DISABLE is set to 'false', it take precedence over INSTANA_TRACING_DISABLE_INSTRUMENTATIONS

Oh hm 🤔 
Shouldn't we just ignore it if its false? 🤔

kirrg001: Discussed: This is right as it is

kirrg001 (approved):

PR #2487, opened 2026-04-13
chore(core): added external config resolution

aryamohanan:
Added handling of external (agent-provided) configuration to ensure behavior is consistent with configuration precedence rules.

Precedence order:
env > in-code > agent-config > default

This change clarifies when external agent configuration is applied:

- External (agent) config is applied only when:
  - The current config value (from env or in-code) is not set, OR
  - The current value is still at its default state

- External config is NOT applied when:
  - The value is explicitly set via environment variables, OR
  - The value is defined in in-code configuration

This ensures that user-defined and environment-based settings always take priority,
while agent configuration only fills in missing or default values.

Tasks

- [ ] Add 100% coverage on config files
- [ ] Refactor Activate methods
- [ ] Refactor coreUtil.activate(extraConfig) #2489 
- [ ] Refactor tracingUtil.activate(extraConfig)
- [ ] Refactor  spanBuffer.activate(extraConfig)
- [ ] Refactor instrumentations
- [ ] Integration tests fix

---
image

### DIFF
--- a/packages/core/src/config/index.js
+++ b/packages/core/src/config/index.js
@@ -147,7 +147,6 @@ module.exports.init = _logger => {
 
 /**
  * @param {InstanaConfig} externalConfig
-
  */
 module.exports.activate = externalConfig => {
   if (externalConfig && typeof externalConfig === 'object' && Object.keys(externalConfig).length > 0) {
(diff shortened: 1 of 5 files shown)


packages/core/test/config/util_test.js
```diff
@@ -688,4 +688,178 @@ describe('config.util', () => {
       expect(result).to.equal(multilineValue);
     });
   });
+
+  describe('resolveExternalConfig', () => {
```
aryamohanan, line 692: coverage can be increased in another PR:
Current coverage:

Image

packages/core/src/tracing/index.js
```diff
@@ -263,6 +264,8 @@ function initInstanaInstrumentations(_config) {
 exports.activate = function activate(extraConfig = {}) {
   if (tracingEnabled && !tracingActivated) {
     tracingActivated = true;
+    coreConfig.activate(extraConfig);
```
aryamohanan, line 267: After the agent response is processed, `tracing.activate` is invoked from `agentready.js`. At this stage, the `coreConfig.activate` method normalizes the agent configuration by applying the correct precedence rules.

As a result, the subsequent `activate` methods in the following flow do not need to use the `extraConfig` configuration. They should rely solely on the already normalized config object. See example #2489.

packages/core/src/config/index.js
```diff
@@ -142,6 +145,17 @@ module.exports.init = _logger => {
   util.init(logger);
 };
 
+/**
+ * @param {import('@instana/collector/src/types/collector').AgentConfig} externalConfig
+ */
+module.exports.activate = externalConfig => {
```
aryamohanan, line 151: A wrapper activate function has been added to ensure consistent and standardized flow. Normally, the activate function is triggered once the agent has been resolved.

PR #2488, opened 2026-04-13
refactor(core): renamed configValue to inCodeValue for clarity

abhilash-sivan:
Renamed the `configValue` parameter to `inCodeValue` across all config
resolver functions to better reflect that it represents configuration
values set programmatically in code, as opposed to environment variables
or defaults. 

This aligns with the existing `incode:` prefix used in
debug logs throughout the configuration system.

Two options came in: `inCodeConfig` and `inCodeValue`. Currently I use `inCodeValue` because it seemed to go along with other objects in the whole config.

### DIFF
--- a/packages/collector/src/util/normalizeConfig.js
+++ b/packages/collector/src/util/normalizeConfig.js
@@ -44,7 +44,7 @@ module.exports = function normalizeConfig(userConfig = {}) {
 function normalizeAgentHost(userConfig, defaultConfig) {
   return util.resolveStringConfig({
     envVar: 'INSTANA_AGENT_HOST',
-    configValue: userConfig.agentHost,
+    inCodeValue: userConfig.agentHost,
     defaultValue: defaultConfig.agentHost,
     configPath: 'config.agentHost'
   });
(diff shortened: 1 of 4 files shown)


packages/collector/src/util/normalizeConfig.js
```diff
@@ -44,7 +44,7 @@ module.exports = function normalizeConfig(userConfig = {}) {
 function normalizeAgentHost(userConfig, defaultConfig) {
   return util.resolveStringConfig({
     envVar: 'INSTANA_AGENT_HOST',
-    configValue: userConfig.agentHost,
+    inCodeValue: userConfig.agentHost,
```
abhilash-sivan, line 47: Had to change here as well becuase it was failing due to type error; we can now use inCodeValue throughout the repo

abhilash-sivan: We use core util fn here also, so the type error makes sense :D

aryamohanan:
@abhilash-sivan Sorry, this PR is no longer required. The direction has changed, and it has already been addressed.

abhilash-sivan:
Okay closing now

PR #2489, opened 2026-04-13
refactor(core/util): use resolved config in activate instead of raw external config

aryamohanan:
This will be RFR after #2487


packages/core/test/util/disableInstrumentation_test.js
```diff
@@ -387,31 +387,6 @@ describe('util.disableInstrumentation', () => {
       expect(consoleResult).to.be.true;
     });
 
-    it('should accept service configuration and agent configuration when both are present', () => {
```
aryamohanan, line 390: This case was not correct, we should not merge both configs.

PR #2491, opened 2026-04-13
[Currency Bot] Bumped tsoa from 6.6.0 to 7.0.0-alpha.0

instanacd:
Tada!


abhilash-sivan:
not supporting alpha releases at the time being

PR #2492, opened 2026-04-14
fix: corrected the config precedence for agent config

aryamohanan:
TODO later

- [ ] Fix Disable config
- [ ] Fix Tests
- [ ] Add util.js test
- [ ] Add 100% test coverage
- [ ] Fix collector tests
- [ ] fix logging

### DIFF
--- a/packages/collector/src/announceCycle/agentready.js
+++ b/packages/collector/src/announceCycle/agentready.js
@@ -12,7 +12,7 @@ try {
   // Worker threads are not available, so we know that this is the main thread.
 }
 
-const { tracing } = require('@instana/core');
+const { tracing, coreConfig, util } = require('@instana/core');
 const agentConnection = require('../agentConnection');
 const agentOpts = require('../agent/opts');
 const initializedTooLate = require('../util/initializedTooLate');
@@ -130,6 +130,8 @@ function enter(_ctx) {
     }
   }
 
+  // here we update the config with the values from the agent
+  coreConfig.update(coreConfig, util.constants.CONFIG_SOURCES.AGENT);
   tracing.activate(agentOpts.config);
 
   if (agentOpts.autoProfile && autoprofile) {
--- a/packages/collector/src/announceCycle/agentready.js
+++ b/packages/collector/src/announceCycle/agentready.js
@@ -131,8 +131,8 @@ function enter(_ctx) {
   }
 
   // here we update the config with the values from the agent
-  const updatedConfig = coreConfig.update(agentOpts.config, util.constants.CONFIG_SOURCES.AGENT);
-  tracing.activate(updatedConfig);
+coreConfig.update(agentOpts.config, util.constants.CONFIG_SOURCES.AGENT);
+  tracing.activate();
 
   if (agentOpts.autoProfile && autoprofile) {
     profiler = autoprofile.start();
--- a/packages/collector/src/announceCycle/agentready.js
+++ b/packages/collector/src/announceCycle/agentready.js
@@ -131,7 +131,7 @@ function enter(_ctx) {
   }
 
   // here we update the config with the values from the agent
-  coreConfig.update(coreConfig, util.constants.CONFIG_SOURCES.AGENT);
+  coreConfig.update(agentOpts.config, util.constants.CONFIG_SOURCES.AGENT);
   tracing.activate(agentOpts.config);
 
   if (agentOpts.autoProfile && autoprofile) {
--- a/packages/collector/src/announceCycle/agentready.js
+++ b/packages/collector/src/announceCycle/agentready.js
@@ -131,8 +131,8 @@ function enter(_ctx) {
   }
 
   // here we update the config with the values from the agent
-coreConfig.update(agentOpts.config, util.constants.CONFIG_SOURCES.AGENT);
-  tracing.activate();
+  const updatedConfig = coreConfig.update(agentOpts.config, util.constants.CONFIG_SOURCES.AGENT);
+  tracing.activate(updatedConfig);
 
   if (agentOpts.autoProfile && autoprofile) {
     profiler = autoprofile.start();
--- a/packages/collector/src/announceCycle/agentready.js
+++ b/packages/collector/src/announceCycle/agentready.js
@@ -131,8 +131,8 @@ function enter(_ctx) {
   }
 
   // here we update the config with the values from the agent
-  coreConfig.update(agentOpts.config, util.constants.CONFIG_SOURCES.AGENT);
-  tracing.activate();
+  const updatedConfig = coreConfig.update(agentOpts.config, util.constants.CONFIG_SOURCES.AGENT);
+  tracing.activate(updatedConfig);
 
   if (agentOpts.autoProfile && autoprofile) {
     profiler = autoprofile.start();
--- a/packages/collector/src/announceCycle/agentready.js
+++ b/packages/collector/src/announceCycle/agentready.js
@@ -132,7 +132,7 @@ function enter(_ctx) {
 
   // here we update the config with the values from the agent
   coreConfig.update(agentOpts.config, util.constants.CONFIG_SOURCES.AGENT);
-  tracing.activate(agentOpts.config);
+  tracing.activate();
 
   if (agentOpts.autoProfile && autoprofile) {
     profiler = autoprofile.start();
--- a/packages/collector/src/announceCycle/agentready.js
+++ b/packages/collector/src/announceCycle/agentready.js
@@ -130,7 +130,6 @@ function enter(_ctx) {
     }
   }
 
-  // here we update the config with the values from the agent
   const updatedConfig = coreConfig.update(agentOpts.config, util.constants.CONFIG_SOURCES.AGENT);
   tracing.activate(updatedConfig);
 
--- a/packages/collector/src/util/normalizeConfig.js
+++ b/packages/collector/src/util/normalizeConfig.js
@@ -45,8 +45,8 @@ module.exports = function normalizeConfig(userConfig = {}) {
 function normalizeAgentHost(userConfig, defaultConfig) {
   const { value } = util.resolve(
     {
-      envVar: 'INSTANA_AGENT_HOST',
-      configValue: userConfig.agentHost,
+      envValue: 'INSTANA_AGENT_HOST',
+      inCodeValue: userConfig.agentHost,
       defaultValue: defaultConfig.agentHost,
       configPath: 'config.agentHost'
     },
@@ -63,8 +63,8 @@ function normalizeAgentHost(userConfig, defaultConfig) {
 function normalizeAgentPort(userConfig, defaultConfig) {
   const { value } = util.resolve(
     {
-      envVar: 'INSTANA_AGENT_PORT',
-      configValue: userConfig.agentPort,
+      envValue: 'INSTANA_AGENT_PORT',
+      inCodeValue: userConfig.agentPort,
       defaultValue: defaultConfig.agentPort,
       configPath: 'config.agentPort'
     },
@@ -81,8 +81,8 @@ function normalizeAgentPort(userConfig, defaultConfig) {
 function normalizeAgentRequestTimeout(userConfig, defaultConfig) {
   const { value } = util.resolve(
     {
-      envVar: 'INSTANA_AGENT_REQUEST_TIMEOUT',
-      configValue: userConfig.agentRequestTimeout,
+      envValue: 'INSTANA_AGENT_REQUEST_TIMEOUT',
+      inCodeValue: userConfig.agentRequestTimeout,
       defaultValue: defaultConfig.agentRequestTimeout,
       configPath: 'config.agentRequestTimeout'
     },
@@ -99,8 +99,8 @@ function normalizeAgentRequestTimeout(userConfig, defaultConfig) {
 function normalizeAutoProfile(userConfig, defaultConfig) {
   const { value } = util.resolve(
     {
-      envVar: 'INSTANA_AUTO_PROFILE',
-      configValue: userConfig.autoProfile,
+      envValue: 'INSTANA_AUTO_PROFILE',
+      inCodeValue: userConfig.autoProfile,
       defaultValue: defaultConfig.autoProfile,
       configPath: 'config.autoProfile'
     },
--- a/packages/collector/src/util/normalizeConfig.js
+++ b/packages/collector/src/util/normalizeConfig.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const util = require('@instana/core/src/config/util');
+const validate = require('@instana/core/src/config/validator');
 
 const defaults = {
   agentHost: '127.0.0.1',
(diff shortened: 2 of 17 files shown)


packages/core/src/tracing/instrumentation/messaging/kafkaJs.js
```diff
@@ -28,12 +28,8 @@ exports.updateConfig = function updateConfig(config) {
   traceCorrelationEnabled = config.tracing.kafka.traceCorrelation;
 };
 
-exports.activate = function activate(extraConfig) {
-  if (extraConfig && extraConfig.tracing && extraConfig.tracing.kafka) {
-    if (extraConfig.tracing.kafka.traceCorrelation != null) {
-      traceCorrelationEnabled = extraConfig.tracing.kafka.traceCorrelation;
-    }
-  }
+exports.activate = function activate(_config) {
```
kirrg001, line 32:
```suggestion
// TODO: We will remove _config as soon as the config object is a config instance (`config.get`)
exports.activate = function activate(_config) {
```

packages/core/src/config/index.js
```diff
@@ -175,32 +214,44 @@ module.exports.normalize = ({ userConfig = {}, finalConfigBase = {}, defaultsOve
...
-    configValue: userConfig.serviceName,
-    defaultValue: defaultConfig.serviceName,
-    configPath: 'config.serviceName'
-  });
+  const { value, source } = util.resolve(
+    {
+      envVar: 'INSTANA_SERVICE_NAME',
+      configValue: userConfig.serviceName,
+      defaultValue: defaultConfig.serviceName,
+      configPath: 'config.serviceName'
```
kirrg001: > configPath: 'config.serviceName'

This can be removed

packages/core/src/config/index.js
```diff
@@ -175,32 +214,44 @@ module.exports.normalize = ({ userConfig = {}, finalConfigBase = {}, defaultsOve
...
-  finalConfig.packageJsonPath = util.resolveStringConfig({
-    envVar: 'INSTANA_PACKAGE_JSON_PATH',
-    configValue: userConfig.packageJsonPath,
-    defaultValue: defaultConfig.packageJsonPath,
-    configPath: 'config.packageJsonPath'
-  });
+  const { value, source } = util.resolve(
+    {
+      envVar: 'INSTANA_PACKAGE_JSON_PATH',
+      configValue: userConfig.packageJsonPath,
```
kirrg001:
```suggestion
      inCodeValue: userConfig.packageJsonPath,
```

packages/core/src/config/index.js
```diff
@@ -175,32 +214,44 @@ module.exports.normalize = ({ userConfig = {}, finalConfigBase = {}, defaultsOve
...
 function normalizePackageJsonPath({ userConfig = {}, defaultConfig = {}, finalConfig = {} } = {}) {
-  finalConfig.packageJsonPath = util.resolveStringConfig({
-    envVar: 'INSTANA_PACKAGE_JSON_PATH',
-    configValue: userConfig.packageJsonPath,
-    defaultValue: defaultConfig.packageJsonPath,
-    configPath: 'config.packageJsonPath'
-  });
+  const { value, source } = util.resolve(
+    {
+      envVar: 'INSTANA_PACKAGE_JSON_PATH',
```
kirrg001:
```suggestion
      envValue: 'INSTANA_PACKAGE_JSON_PATH',
```

packages/core/src/util/spanFilter.js
```diff
@@ -19,26 +19,11 @@ function init(config) {
 }
 
 /**
- * @param {import('@instana/collector/src/types/collector').AgentConfig} extraConfig
+ * @param {import('../config').InstanaConfig} _config
```
kirrg001, line 22: >  * @param {import('@instana/collector/src/types/collector').AgentConfig} extraConfig

Why was that changed

packages/collector/src/announceCycle/agentready.js
```diff
@@ -130,7 +130,9 @@ function enter(_ctx) {
     }
   }
 
-  tracing.activate(agentOpts.config);
+  // here we update the config with the values from the agent
```
kirrg001:
```suggestion
```

IMO code is self explaining

packages/core/src/util/spanFilter.js
kirrg001: Ah ignore

packages/core/src/tracing/instrumentation/messaging/kafkaJs.js
```diff
@@ -28,12 +28,9 @@ exports.updateConfig = function updateConfig(config) {
...
 
-exports.activate = function activate(extraConfig) {
-  if (extraConfig && extraConfig.tracing && extraConfig.tracing.kafka) {
-    if (extraConfig.tracing.kafka.traceCorrelation != null) {
-      traceCorrelationEnabled = extraConfig.tracing.kafka.traceCorrelation;
-    }
-  }
+// TODO: We will remove _config as soon as the config object is a config instance (`config.get`)
+exports.activate = function activate(_config) {
+  traceCorrelationEnabled = _config.tracing?.kafka?.traceCorrelation;
```
kirrg001: Why do we need these chaining checks? 🤔

packages/core/src/config/index.js
```diff
@@ -8,9 +8,46 @@
...
+// @typedef {{ [x: string]: any }} configMeta
+/** @type {configMeta} */
+const configMeta = {};
+
+const configStore = {
+  /**
+   * @param {string} configPath
+   * @param {number} source
+   */
+  set(configPath, source) {
```
kirrg001: This must be 

> set(configPath, obj) {

Contains val and source

packages/core/src/config/configNormalizers/disable.js
```diff
@@ -5,6 +5,7 @@
 'use strict';
 
 const { DISABLABLE_INSTRUMENTATION_GROUPS } = require('../../tracing/constants');
+const { CONFIG_SOURCES } = require('../../util/constants');
```
kirrg001: TODO: Revert this file for this PR. Will do later

aryamohanan: Okie

packages/core/src/config/util.js
```diff
@@ -14,218 +16,67 @@ exports.init = _logger => {
...
 
-  if (envRaw != null) {
-    const envParsed = toValidNumber(envRaw);
-    if (envParsed !== undefined) {
-      logger.debug(`[config] env:${envVar} = ${envParsed}`);
-      return envParsed;
-    }
+  const inputs = {
+    env: envValue ? process.env[envValue] : undefined,
+    in_code: inCodeValue,
```
kirrg001: nitpik: We use camelcase everywhere. Here suddenly lowercase

packages/core/src/config/index.js
```diff
@@ -175,32 +214,42 @@ module.exports.normalize = ({ userConfig = {}, finalConfigBase = {}, defaultsOve
...
+  const { value, source } = util.resolve(
+    {
+      envValue: 'INSTANA_PACKAGE_JSON_PATH',
+      inCodeValue: userConfig.packageJsonPath,
+      defaultValue: defaultConfig.packageJsonPath
+    },
+    [validate.stringValidator]
+  );
+
+  configStore.set('config.packageJsonPath', { source });
```
kirrg001, line 244: val missing

packages/core/src/config/index.js
```diff
@@ -8,9 +8,46 @@
...
+
+  clear() {
+    Object.keys(configMeta).forEach(key => delete configMeta[key]);
+  }
+};
+
+/**
+ * @param {{ configPath: string, val?: any, source: number }} options
+ */
+function recordMeta({ configPath, val, source }) {
```
kirrg001: This is not rly meta or?
Why do we need this extra wrapper? 🤔 Looks redundant to me

aryamohanan: we can remove that

packages/core/src/config/index.js
```diff
@@ -258,54 +316,82 @@ function normalizeTracingEnabled({ userConfig = {}, defaultConfig = {}, finalCon
...
+    {
+      envValue: isBooleanValue ? 'INSTANA_TRACING_DISABLE' : undefined,
+      inCodeValue: userConfig.tracing.enabled,
+      defaultValue: defaultConfig.tracing.enabled
+    },
+    [validate.booleanValidator]
+  );
+
+  // The env var is TRACING_DISABLE, so we need to invert it when it comes from env
+  const finalValue = source === CONFIG_SOURCES.ENV ? !value : value;
```
kirrg001, line 322: ? Whats that?

aryamohanan: This is a special case with the `INSTANA_TRACING_DISABLE` environment variable. If it is set to `true`, then `finalConfig.tracing.enabled` should be set to `false` so the value needs to be inverted when assigning it.

packages/core/src/config/index.js
aryamohanan: Updated: We discussed this offline and decided not to proceed with adding this value for now.

packages/core/src/config/index.js
kirrg001: This is normalization. Lets add a todo to add normalizers to util.resolver later (no prio)

packages/core/src/tracing/index.js
```diff
@@ -260,12 +260,12 @@ function initInstanaInstrumentations(_config) {
   }
 }
 
-exports.activate = function activate(extraConfig = {}) {
+exports.activate = function activate(_config = config) {
```
kirrg001, line 263:
```suggestion
exports.activate = function activate(_config) {
```

packages/collector/src/util/normalizeConfig.js
kirrg001, line 112: Do we have to call util resolve here too?

packages/core/src/config/index.js
```diff
@@ -160,6 +190,8 @@ module.exports.normalize = ({ userConfig = {}, finalConfigBase = {}, defaultsOve
   } else {
     normalizedUserConfig = {};
   }
+  // Clear previous sources if any
+  configStore.clear();
```
kirrg001, line 203: Wondering if we have to call this only on init?
Instead of `// Clear previous sources if any`, could you please describe **why** we call this here?

packages/core/src/config/index.js
```diff
@@ -175,32 +207,42 @@ module.exports.normalize = ({ userConfig = {}, finalConfigBase = {}, defaultsOve
   normalizeTracingConfig({ userConfig: normalizedUserConfig, defaultConfig: defaults, finalConfig });
   normalizeSecrets({ userConfig: normalizedUserConfig, defaultConfig: defaults, finalConfig });
   normalizePreloadOpentelemetry({ userConfig: normalizedUserConfig, defaultConfig: defaults, finalConfig });
-
+  currentConfig = finalConfig;
```
kirrg001, line 219: Why do we need currentConfig, if we have to configStore? 🤔

packages/core/src/config/index.js
```diff
@@ -8,9 +8,39 @@
...
+    return configMeta[configPath];
+  },
+
+  clear() {
+    Object.keys(configMeta).forEach(key => delete configMeta[key]);
+  }
+};
+
+/** @type {InstanaConfig} */
+let currentConfig;
```
kirrg001, line 52: Comment needed as discussed ^

packages/core/src/config/index.js
aryamohanan: We don't clear configStore on init because init() is called only once during the application lifecycle, whereas normalize() can be called multiple times (e.g., when re-normalizing configuration). The configStore.clear() call in normalize() ensures we start fresh for each normalization.

packages/collector/src/util/normalizeConfig.js
aryamohanan: Yes. Updated

packages/core/src/config/index.js
aryamohanan: Discussed in offline.
Both have diff purpose.

currentConfig is simply a copy of the config object. Since obj are reference types, both  point to the same memory. Updating a value in currentConfig will also reflect in the original config object.

ConfigStore is used to maintain metadata about config values, now it track the source of each setting. This helps with decision-making and related processes. Not all configuration values are included here for example, collector-specific configurations are currently excluded.

The scope can be expanded in the future, and at present, the actual values  are not stored in it.

currentConfig can be removed in future when we implement config.get/ config.set. I will add a comment about this

packages/core/src/config/index.js
aryamohanan: added

kirrg001: We want a comment in the codebase here 👍

packages/core/src/config/index.js
kirrg001: normalize -> resets configStore
This is not clean, but we can revisit later

kirrg001: Lets add a clear comment in the codebase here 👍

packages/core/src/config/index.js
```diff
@@ -8,9 +8,48 @@
...
+
+/**
+ * @type {InstanaConfig}
+ *
+ * NOTE: currentConfig is a reference to the config object returned by normalize().
+ * This variable exists to allow dynamic config updates via the update() function without
+ * requiring the config object to be passed as a parameter.
+ *
+ * TODO: This can be removed in the future when we implement config.get()/config.set()
+ * methods that will provide a cleaner API for accessing and modifying configuration values.
```
kirrg001:
```suggestion
 * methods. The values will be kept in the configStore instance.
```

packages/core/src/config/index.js
```diff
@@ -258,54 +324,85 @@ function normalizeTracingEnabled({ userConfig = {}, defaultConfig = {}, finalCon
...
+    {
+      envValue: 'INSTANA_DISABLE_USE_OPENTELEMETRY',
+      inCodeValue: userConfig.tracing.useOpentelemetry,
+      defaultValue: defaultConfig.tracing.useOpentelemetry
+    },
+    [validate.booleanValidator]
+  );
+
+  // The env var is DISABLE_USE_OPENTELEMETRY, so we need to invert it when it comes from env
+  // TODO: Consider adding this normalization support to util.resolver
```
kirrg001:
```suggestion
  // TODO: add normalization helpers to util.resolve(...)
```

packages/core/src/config/index.js
```diff
@@ -258,54 +324,85 @@ function normalizeTracingEnabled({ userConfig = {}, defaultConfig = {}, finalCon
...
+    {
+      envValue: 'INSTANA_DISABLE_AUTO_INSTR',
+      inCodeValue: userConfig.tracing.automaticTracingEnabled,
+      defaultValue: defaultConfig.tracing.automaticTracingEnabled
+    },
+    [validate.booleanValidator]
+  );
+
+  // The env var is DISABLE_AUTO_INSTR, so we need to invert it when it comes from env
+  // TODO: Consider adding this normalization support to util.resolver
```
kirrg001:
```suggestion
  // TODO: add normalization helpers to util.resolve(...)
```

packages/core/src/config/index.js
```diff
@@ -160,6 +199,8 @@ module.exports.normalize = ({ userConfig = {}, finalConfigBase = {}, defaultsOve
   } else {
     normalizedUserConfig = {};
   }
+  // Clear previous sources if any
```
kirrg001:
```suggestion
  // TODO: This call needs to be reconsidered when we add the full config instance (`config.get(...)`).
```

kirrg001 (approved):

PR #2502, opened 2026-04-20
refactor(config): align disable normalizer with resolver return shape

aryamohanan:
The disable normalization is currently not returning the source info

see: https://github.com/instana/nodejs/pull/2492#discussion_r3099079615

### DIFF
--- a/packages/core/src/config/configNormalizers/disable.js
+++ b/packages/core/src/config/configNormalizers/disable.js
@@ -5,6 +5,7 @@
 'use strict';
 
 const { DISABLABLE_INSTRUMENTATION_GROUPS } = require('../../tracing/constants');
+const { CONFIG_SOURCES } = require('../../util/constants');
 /** @type {import('../../core').GenericLogger} */
 let logger;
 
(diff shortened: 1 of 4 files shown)


packages/core/src/config/configNormalizers/disable.js
```diff
@@ -31,11 +32,11 @@ exports.normalize = function normalize(config) {
 
     if (envDisableConfig !== null) {
       if (envDisableConfig === true) {
-        return true;
+        return { value: true, source: CONFIG_SOURCES.ENV };
       }
 
       if (envDisableConfig === false) {
-        return {};
+        return { value: {}, source: CONFIG_SOURCES.ENV };
```
kirrg001, line 39: Is this right?

aryamohanan: This is the case when INSTANA_TRACING_DISABLE=false.

kirrg001: 🤔 Why is the value {}

aryamohanan: we treat disable as object type internaly

aryamohanan: the external config type mixed with boolean and array obj, but internally we treat it as object.
INSTANA_TRACING_DISABLE can accept both boolean and obj array

kirrg001 (approved):

PR #2506, opened 2026-04-21
chore(config): preserved existing values when applying partial updates

aryamohanan:
External config updates now deep-merge nested plain objects into the existing normalized config instead of replacing them at the top level.

### DIFF
--- a/packages/core/src/config/index.js
+++ b/packages/core/src/config/index.js
@@ -974,9 +974,11 @@ exports.update = function update(externalConfig, source) {
       return currentConfig;
     }
 
-    /** @type {any} */ (currentConfig)[key] = externalConfig[key];
+    /** @type {Record<string, any>} */
+    const configAsRecord = currentConfig;
+    configAsRecord[key] = deepMerge(configAsRecord[key], externalConfig[key]);
+
     configStore.set(configPath, { source });
-    logger.debug(`[config] Updated ${key} from source ${source}: ${JSON.stringify(externalConfig[key])}`);
   });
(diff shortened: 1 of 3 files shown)


packages/core/src/config/index.js
```diff
@@ -974,9 +974,11 @@ exports.update = function update(externalConfig, source) {
       return currentConfig;
     }
 
-    /** @type {any} */ (currentConfig)[key] = externalConfig[key];
+    /** @type {Record<string, any>} */
+    const configAsRecord = currentConfig;
+    configAsRecord[key] = deepMerge(configAsRecord[key], externalConfig[key]);
+
     configStore.set(configPath, { source });
-    logger.debug(`[config] Updated ${key} from source ${source}: ${JSON.stringify(externalConfig[key])}`);
```
aryamohanan, line 979: This log not required

packages/core/src/config/index.js
```diff
@@ -974,9 +974,11 @@ exports.update = function update(externalConfig, source) {
       return currentConfig;
     }
 
-    /** @type {any} */ (currentConfig)[key] = externalConfig[key];
+    /** @type {Record<string, any>} */
+    const configAsRecord = currentConfig;
+    configAsRecord[key] = deepMerge(configAsRecord[key], externalConfig[key]);
```
aryamohanan, line 979: External config updates now deep-merge nested plain objects into the existing normalized config instead of replacing them at the top level.

kirrg001 (approved):

PR #2507, opened 2026-04-21
refactor(config): replaced agentConfig usage with updated config

aryamohanan:
In our codebase, agentConfig is used in several places such as spanBuffer and tracingUtil. This change replaces those usages. 

We will also need to update the instrumentations, which will be addressed in a separate PR

### DIFF
--- a/packages/core/src/tracing/tracingUtil.js
+++ b/packages/core/src/tracing/tracingUtil.js
@@ -37,8 +37,8 @@ exports.init = function (config) {
  * @param {import('../config').InstanaConfig} _config
  */
 exports.activate = function activate(_config) {
-  stackTraceLength = _config.tracing?.stackTraceLength;
-  stackTraceMode = _config.tracing?.stackTrace;
+  stackTraceLength = _config.tracing.stackTraceLength;
+  stackTraceMode = _config.tracing.stackTrace;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


packages/core/src/tracing/tracingUtil.js
```diff
@@ -34,26 +34,11 @@ exports.init = function (config) {
...
-  if (agentTraceConfig?.stackTrace && stackTraceMode === DEFAULT_STACK_TRACE_MODE) {
-    stackTraceMode = agentTraceConfig.stackTrace;
-  }
-
-  // stackTraceLength is valid when set to any number, including 0
-  if (agentTraceConfig?.stackTraceLength != null && stackTraceLength === DEFAULT_STACK_TRACE_LENGTH) {
-    stackTraceLength = agentTraceConfig.stackTraceLength;
-  }
+exports.activate = function activate(_config) {
+  stackTraceLength = _config.tracing?.stackTraceLength;
```
kirrg001: Hm I see different approaches in active method.

```
  if (_config.tracing.spanBatchingEnabled) {
    batchingEnabled = true;
```

````
stackTraceLength = _config.tracing?.stackTraceLength;
```

aryamohanan: updated

kirrg001 (approved):

PR #2508, opened 2026-04-21
chore: fixed failing config related tests

aryamohanan:


packages/core/src/config/util.js
```diff
@@ -66,10 +66,6 @@ exports.resolve = function resolve({ envValue, inCodeValue, agentValue, defaultV
       return true;
     }
 
-    if (rawValue !== undefined) {
-      logger?.warn(`[config] Validation failed for ${sourceKey}: ${JSON.stringify(rawValue)}`);
```
aryamohanan, line 70: This log is not required and logging in general will be fixed in another PR

packages/core/src/config/index.js
```diff
@@ -957,28 +957,40 @@ function normalizePreloadOpentelemetry({ userConfig = {}, defaultConfig = {}, fi
...
   Object.keys(externalConfig).forEach(key => {
-    const configPath = `config.${key}`;
-    const currentMeta = configStore.get(configPath);
+    const path = `${basePath}.${key}`;
+    const currentMeta = configStore.get(path);
 
     if (currentMeta && currentMeta.source < source) {
-      logger.debug(`[config] Skipping ${key}: current source ${currentMeta.source} > incoming ${source}`);
-      return currentConfig;
+      logger.debug(`[config] Skipping ${path}: current source ${currentMeta.source} > incoming ${source}`);
```
kirrg001, line 977: Do we really need this log?

kirrg001 (approved):

packages/core/src/config/index.js
aryamohanan: Not really. We can examine and cleanup logs in a separate PR.

kirrg001: Ok

PR #2511, opened 2026-04-22
fix: enforced max limit on metrics transmission config

abhilash-sivan:
- added max limit of 5000 
- max value is applied if config > 5000

### DIFF
--- a/packages/core/src/config/index.js
+++ b/packages/core/src/config/index.js
@@ -220,10 +220,16 @@ function normalizeMetricsConfig(config) {
     config.metrics.transmissionDelay,
     defaults.metrics.transmissionDelay,
     'config.metrics.transmissionDelay',
-    'INSTANA_METRICS_TRANSMISSION_DELAY',
-    null,
-    5000
+    'INSTANA_METRICS_TRANSMISSION_DELAY'
   );
+
+  // Validate max value of 5000 for transmissionDelay
+  if (config.metrics.transmissionDelay > 5000) {
+    logger.warn(
+      `The value of config.metrics.transmissionDelay (or INSTANA_METRICS_TRANSMISSION_DELAY) (${config.metrics.transmissionDelay}) exceeds the maximum allowed value of 5000. Assuming the default value ${defaults.metrics.transmissionDelay}.`
+    );
+    config.metrics.transmissionDelay = defaults.metrics.transmissionDelay;
+  }
 
   config.metrics.timeBetweenHealthcheckCalls =
     config.metrics.timeBetweenHealthcheckCalls || defaults.metrics.timeBetweenHealthcheckCalls;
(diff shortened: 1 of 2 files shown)


packages/core/src/config/index.js
```diff
@@ -718,9 +720,11 @@ function parseSecretsEnvVar(envVarValue) {
  * @param {*} defaultValue
  * @param {string} configPath
  * @param {string} envVarKey
+ * @param {number} [minValue]
+ * @param {number} [maxValue]
  * @returns {*}
  */
-function normalizeSingleValue(configValue, defaultValue, configPath, envVarKey) {
+function normalizeSingleValue(configValue, defaultValue, configPath, envVarKey, minValue, maxValue) {
```
abhilash-sivan: Since INSTANA_METRICS_TRANSMISSION_DELAY normalization uses the shared `normalizeSingleValue` function, the min/max validation logic has been added.

Even though the minimum value isn’t currently enforced or used, it’s included for consistency with the common normalisation approach.

packages/core/src/config/index.js
```diff
@@ -220,7 +220,9 @@ function normalizeMetricsConfig(config) {
     config.metrics.transmissionDelay,
     defaults.metrics.transmissionDelay,
     'config.metrics.transmissionDelay',
-    'INSTANA_METRICS_TRANSMISSION_DELAY'
+    'INSTANA_METRICS_TRANSMISSION_DELAY',
+    null,
```
aryamohanan: qs: why null passed as min value? what if customer configure 500ms?  It still works?

packages/core/src/config/index.js
kirrg001: Please revert this and handle the check only in the normalizeMetricsConfig fn

packages/core/src/config/index.js
abhilash-sivan: yes

packages/core/src/config/index.js
abhilash-sivan: done

packages/core/src/config/index.js
```diff
@@ -223,6 +223,14 @@ function normalizeMetricsConfig(config) {
     'INSTANA_METRICS_TRANSMISSION_DELAY'
   );
 
+  // Validate max value of 5000 for transmissionDelay
+  if (config.metrics.transmissionDelay > 5000) {
+    logger.warn(
+      `The value of config.metrics.transmissionDelay (or INSTANA_METRICS_TRANSMISSION_DELAY) (${config.metrics.transmissionDelay}) exceeds the maximum allowed value of 5000. Assuming the default value ${defaults.metrics.transmissionDelay}.`
+    );
+    config.metrics.transmissionDelay = defaults.metrics.transmissionDelay;
```
kirrg001: This should be 5000, right?

abhilash-sivan: The default is still 1000 as per the discussions

kirrg001: But if a customer sets already >5s, we should set the MAX limit?

aryamohanan: In this case, it should be 5000ms, that is the max allowed limit.

abhilash-sivan: ah that makes sense

kirrg001 (approved):
Please merge as fix

PR #2517, opened 2026-04-24
fix: disabled collector init process events

aryamohanan:
Sending 'instana.collector.initialized' event is disabled by default.
     * This prevents interference with application-level communication channels.
     * The event is primarily used for testing purposes now.
     
     
  ref https://jsw.ibm.com/browse/INSTA-85329


packages/collector/src/announceCycle/agentready.js
```diff
@@ -156,8 +156,12 @@ function enter(_ctx) {
...
     const { parentPort } = require('worker_threads');
 
-    if (parentPort) {
-      // CASE: This is for the worker thread if available.
+    /**
+     * Sending 'instana.collector.initialized' event is disabled by default.
+     * This prevents interference with application-level communication channels.
+     * The event is primarily used for testing purposes now.
+     */
+    if (parentPort && !agentOpts.config.disableCollectorInitEvent) {
```
kirrg001:
```suggestion
    if (parentPort && agentOpts.config.disableCollectorInitEvent === false) {
```

kirrg001 (approved):

kirrg001 (approved):

PR #2528, opened 2026-05-03
[Prod Dependency Bot] Bumped @opentelemetry/api from 1.9.0 to 1.9.1

instanacd:
Tada!


aryamohanan:
As we keep range for @opentelemetry/api  in package.json, this update is not required now.

PR #2530, opened 2026-05-03
feat: added support for sqs-consumer v15

instanacd:
ref https://jsw.ibm.com/browse/INSTA-88562

https://github.com/bbc/sqs-consumer/releases/tag/v15.0.0

Tada!


currencies.json
```diff
@@ -1059,7 +1059,7 @@
     "core": false,
     "versions": [
       {
-        "v": "14.2.8"
+        "v": "15.0.1"
```
aryamohanan: Breaking change:
Support for Node.js now requires version 22 or later.
https://github.com/bbc/sqs-consumer/blob/main/package.json#L103

https://github.com/bbc/sqs-consumer/pull/785/changes

abhilash-sivan (approved):

PR #2531, opened 2026-05-04
chore: added node constraint for mongoose v9

aryamohanan:
Mongoose v9 requires v20.19 and above.

See: https://github.com/Automattic/mongoose/issues/16267#issuecomment-4366936565

Currently our v18 pipeline is failing for this issue.


currencies.json
```diff
@@ -779,7 +779,8 @@
     "core": false,
     "versions": [
       {
-        "v": "9.6.1"
+        "v": "9.6.1",
+        "node": ">=20.19"
```
aryamohanan, line 783: https://github.com/Automattic/mongoose/blob/master/package.json#L120

abhilash-sivan (approved):

PR #2432, opened 2026-03-26
Config Precedence Changes for v6 (Refactor + Breaking Changes)

aryamohanan:
This PR includes all configuration-related changes, including breaking changes and refactoring.
It is intended for the upcoming v6 branch, which will be created later, and will be merged into v6.

Breaking changes 

1. Remove deprecated config INSTANA_DEV_MIN_DELAY_BEFORE_SENDING_SPANS
2. Precedence order change env>in-code > agent config > default

Tasks

- [x]  Create a util file for shared config helpers and move resolveNumericConfig to the util file #2430
- [x]  Remove deprecated config INSTANA_DEV_MIN_DELAY_BEFORE_SENDING_SPANS #2431 
- [x]  Replace all inline numeric parsing (parseInt, isNaN) with resolveNumericConfig
- [x]  Extract common env parsing logic into util #2433 
- [x] Remove duplicated parsing/validation logic
- [x]  Fix and standardize logging messages and consistent format #2446
- [x]  Normalize boolean handling across config
- [x]  Minor other cleanups https://github.com/instana/nodejs/pull/2448
- [x] Todo comments #2438
- [x] Improve tests #2445
- [x] Add more tests or missing tests https://github.com/instana/nodejs/pull/2447
- [x] 100% coveragehttps://github.com/instana/nodejs/pull/2451
- [x] PR to use the options pattern in config/index.js functions @abhilash-sivan 
- [x] Cleanup config/index.js config-> use in-code  instead of configvalue @abhilash-sivan 
- [x] Add infos to public docs for order config + upcoming changes on config page @kirrg001 
- [x] Unskip tests
- [x] Update order env> in-code #2478 
- [x] Collector config refactor #2457 
- [x] collector test coverage https://github.com/instana/nodejs/pull/2456
- [x] Fix testcases in collector #2486
- [x]  Add new method in core ->config->index to sync agent config #2492
- [x]  Refactor activate fns in core #2507 
- [x]  Fix Disable config #2502 
- [x] Add util.js test
- [x] Add 100% test coverage
- [x]  Fix collector tests #2506  #2508
- [x] fix logging #2514
- [x] Fix instrumentations #2509
- [x] incorporate the main config changes
- [x]  Integration tests #2515
- [ ]  More tests (manual + e2e)
- [ ] public docs any updates

TBD

- [x] In debug mode, the configured variables will be logged twice. https://jsw.ibm.com/browse/INSTA-89582 

Testing tasks

[...]

PR #2547, opened 2026-05-12
chore: respected config precedence for agent-provided secrets

aryamohanan:
See: https://github.com/instana/nodejs/pull/2432#discussion_r3224949007

Previously, agent-provided secrets configuration was directly applied via secrets.setMatcher(), bypassing the configuration precedence system. This meant agent config would override user-provided config (ENV vars
or in-code config), violating the intended precedence order: ENV > INCODE > AGENT > DEFAULT

The fix follows the established activate pattern used by other modules

### DIFF
--- a/packages/core/src/tracing/index.js
+++ b/packages/core/src/tracing/index.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const sdk = require('./sdk');
+const secrets = require('../secrets');
 const constants = require('./constants');
 const tracingMetrics = require('./metrics');
 const opentracing = require('./opentracing');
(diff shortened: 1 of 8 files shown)


packages/core/src/tracing/index.js
```diff
@@ -266,6 +267,7 @@ exports.activate = function activate(_config = config) {
     coreUtil.activate(_config);
     tracingUtil.activate(_config);
     spanBuffer.activate(_config);
+      secrets.activate(_config);
```
kirrg001:
```suggestion
    secrets.activate(_config);
```

kirrg001 (approved):

PR #2551, opened 2026-05-15
chore: fixed mariadb url capture for esm

aryamohanan:
ESM tests were failing on main due to prisma mariadb adapter not capturing url for ESM apps.

### DIFF
--- a/packages/core/src/tracing/instrumentation/databases/prisma.js
+++ b/packages/core/src/tracing/instrumentation/databases/prisma.js
@@ -69,7 +69,9 @@ function instrumentMariaDbAdapter(mariadbAdapterModule) {
   if (isGetterExport) {
     return {
       ...mariadbAdapterModule,
-      PrismaMariaDb: InstanaPrismaMariaDb
+      get PrismaMariaDb() {
+        return InstanaPrismaMariaDb;
+      }
     };
   }
   // ESM: PrismaMariaDb is a direct export
--- a/packages/core/src/tracing/instrumentation/databases/prisma.js
+++ b/packages/core/src/tracing/instrumentation/databases/prisma.js
@@ -30,7 +30,7 @@ const mariadbAdapterConfigMap = new WeakMap();
 exports.init = function init(config) {
   logger = config.logger;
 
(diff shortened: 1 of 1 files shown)


packages/core/src/tracing/instrumentation/databases/prisma.js
```diff
@@ -64,10 +64,17 @@ function instrumentMariaDbAdapter(mariadbAdapterModule) {
     }
   }
 
-  return {
-    ...mariadbAdapterModule,
-    PrismaMariaDb: InstanaPrismaMariaDb
-  };
+  const isGetterExport = !!Object.getOwnPropertyDescriptor(mariadbAdapterModule, 'PrismaMariaDb')?.get;
+  // CJS: PrismaMariaDb is a getter property
+  if (isGetterExport) {
```
kirrg001, line 69: I think something is wrong here.

CJS e.g.

Original vs Modified

{ PrismaMariaDb: [Getter] }
{ PrismaMariaDb: [class InstanaPrismaMariaDb extends PrismaMariaDbAdapterFactory] }

I expect:

{ InstanaPrismaMariaDb: [Getter] }

packages/core/src/tracing/instrumentation/databases/prisma.js
```diff
@@ -64,10 +64,17 @@ function instrumentMariaDbAdapter(mariadbAdapterModule) {
...
+    return {
+      ...mariadbAdapterModule,
+      PrismaMariaDb: InstanaPrismaMariaDb
+    };
+  }
+  // ESM: PrismaMariaDb is a direct export
+  mariadbAdapterModule.PrismaMariaDb = InstanaPrismaMariaDb;
```
kirrg001, line 78: ESM is correct

kirrg001 (changes requested):

packages/core/src/tracing/instrumentation/databases/prisma.js
aryamohanan: Updated to preserve the original getter semantics

kirrg001 (approved):

PR #2559, opened 2026-05-21
chore: make agent data port configurable independently from the agent connection port

abhilash-sivan:
This PR prepares the codebase for independently configurable agent ports, allowing a separate port to be used for sending telemetry data such as traces and metrics.

This addresses the dual-port setup use case where the agent connection and announce cycle continue to use the default agent port, while telemetry data is sent to a separately configured OTLP endpoint.

### DIFF
--- a/packages/collector/src/agentConnection.js
+++ b/packages/collector/src/agentConnection.js
@@ -133,6 +133,7 @@ exports.announceNodeCollector = function announceNodeCollector(callback) {
 
   logger.debug(`Announcing the Node.js collector to the Instana host agent at ${agentOpts.host}:${agentOpts.port}`);
 
+  // Agent announcement uses agentOpts.port (agent connection port)
   const req = http.request(
     {
       host: agentOpts.host,
@@ -235,6 +236,8 @@ exports.sendLogToAgent = function sendLogToAgent(logLevel, message, stackTrace) 
 
   const payload = Buffer.from(JSON.stringify(payloadObject), 'utf8');
 
+  // Log forwarding uses agentOpts.port (agent connection port) for now
+  // this can be customized if we use OTLP transformation for logs in the future
   const req = http.request(
     {
       host: agentOpts.host,
@@ -271,6 +274,7 @@ exports.sendLogToAgent = function sendLogToAgent(logLevel, message, stackTrace) 
 function checkWhetherResponseForPathIsOkay(path, cb) {
   cb = util.atMostOnce('callback for checkWhetherResponseForPathIsOkay', cb);
 
+  // Agent readiness check uses agentOpts.port (agent connection port)
   const req = http.request(
     {
       host: agentOpts.host,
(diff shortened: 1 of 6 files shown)


packages/collector/src/agentConnection.js
```diff
@@ -452,10 +456,12 @@ function sendData(path, data, cb, ignore404 = false) {
     return setImmediate(cb.bind(null, error));
   }
 
+  // Use dataPort for sending all telemetry data (metrics, spans, profiles, events, etc.)
+  // dataPort defaults to agentPort but can be configured separately for future use cases (e.g., OTel format)
   const req = http.request(
     {
       host: agentOpts.host,
-      port: agentOpts.port,
+      port: agentOpts.dataPort,
```
kirrg001, line 464: If the customer sets AGENT_PORT, we will ignore the port or?

abhilash-sivan: We will continue using the customer-configured agent port for operations such as announcements and agent readiness checks etc.

For sending telemetry data, we will use a separate data port, which will be selected based on whether OTLP is enabled via configuration (yet to implement).

If the customer configures `AGENT_PORT` and OTLP is not enabled, then the same port will also be used for sending data.

kirrg001: Oki makes sense

packages/collector/src/util/normalizeConfig.js
```diff
@@ -73,6 +74,24 @@ function normalizeAgentPort(userConfig, defaultConfig) {
   return value;
 }
 
+/**
+ * @param {import('../types/collector').CollectorConfig} userConfig
+ * @param {{ agentPort: number }} defaultConfig
+ * @returns {number}
+ */
+function normalizeAgentDataPort(userConfig, defaultConfig) {
```
aryamohanan, line 82: This method is same as `normalizeAgentPort`, is this intentional or is there something missing?

packages/collector/src/agentConnection.js
aryamohanan: I have a question: if the customer has configured an agent port and enabled OTLP, what will be the data port?

abhilash-sivan: In this case, we will use the OTLP port as the data port because we will start communicating using OTLP, and currently only the dedicated OTLP ports can listen to and understand the OTLP format. So, we will use port 4318.

packages/collector/src/util/normalizeConfig.js
abhilash-sivan: Yes, this is currenlty same logic, but in future we will add logic for port switching here

>   // Future logic for OTLP enabled check can be added here to determine which port to use

aryamohanan: Could we simplify this by delegating to `normalizeAgentPort` when OTLP is not enabled?

```
if (userConfig.otlpEnabled) {
  return defaultConfig.otlpPort; // 4317 or 4318
}

return normalizeAgentPort(userConfig, defaults);
```

abhilash-sivan: Yeah, this seems a clean approach.
But currently we don't have the logic

> if (userConfig.otlpEnabled) {
>   return defaultConfig.otlpPort; // 4317 or 4318
> }

so this will just return the normalizeAgentPort()

Also what if we allow the customers to configure data ports separately (in future). we can discuss this tomorrow

kirrg001 (approved):

aryamohanan (approved):

PR #2564, opened 2026-05-24
feat: added support for typeorm v1

instanacd:
https://jsw.ibm.com/browse/INSTA-91638

https://typeorm.io/docs/releases/1.0/release-notes/


abhilash-sivan (commented):
Since we don't have dedicated TypeORM instrumentation, TypeORM database operations should still be traced through the underlying database driver instrumentations (PostgreSQL, MySQL, etc.).

So the changes in "TypeORM" for MySQL and PG will not affect us. Please confirm!

kirrg001 (approved):

PR #2566, opened 2026-05-25
test: enabled Prisma >v4.x tests for Node.js v26

aryamohanan:
As part of v26 support, we didabled all prisma versions test against v26.

Prisma v4.x tests fail on CI with Node.js v26 with the following error:
```
Error: Unknown binaryTarget debian-openssl-3.5.x and no custom engine files were provided
```

Node.js v26 ships with OpenSSL 3.5.x. Prisma v4.x  does not include precompiled binaries for the `debian-openssl-3.5.x` platform target. The issue only affects Debian-based Linux, Linux requires exact OpenSSL version matches in binary target names (e.g., `debian-openssl-3.0.x`, `debian-openssl-3.5.x`).

ref https://jsw.ibm.com/browse/INSTA-89970

### DIFF
--- a/currencies.json
+++ b/currencies.json
@@ -1165,22 +1165,19 @@
     "versions": [
       {
         "v": "7.8.0",
-        "node": ">=20.19.0 <23 || >=24 <26",
+        "node": ">=20.19.0 <23 || >=24",
         "skipValidation": true
       },
       {
         "v": "6.19.0",
-        "node": "<26",
         "skipValidation": true
       },
       {
         "v": "4.16.2",
-        "node": "<26",
         "skipValidation": true
       },
       {
         "v": "4.5.0",
-        "node": "<26",
         "skipValidation": true
       }
     ]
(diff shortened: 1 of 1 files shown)


currencies.json
```diff
@@ -1165,12 +1165,11 @@
     "versions": [
       {
         "v": "7.8.0",
-        "node": ">=20.19.0 <23 || >=24 <26",
+        "node": ">=20.19.0 <23 || >=24",
```
abhilash-sivan, line 1168: So v23 is still having issues ?

aryamohanan: I'm only testing against v26, and I created this PR as part of the post-release tasks for v26.

Regarding your question, this defines the officially supported Node.js versions for prisma:
https://github.com/prisma/prisma/blob/main/package.json#L7C1-L7C42

abhilash-sivan: Hmm I think they only support LTS versions, I got it 👍

abhilash-sivan (approved):

PR #2578, opened 2026-06-02
feat: added support for redis v6

instanacd:
Added support for Redis v6. 

 https://github.com/redis/node-redis/releases/tag/redis%406.0.0

#### Changes

Based on the Redis 6.0 release notes, Redis v6 introduces the following changes that affect our instrumentation:

* **Node.js 20** is now the minimum supported runtime.
* The package now publicly exposes **`RedisClient`**, **`RedisCluster`**, **`RedisSentinel`**, and pool-related classes.

#### Instrumentation Changes

Previously, the instrumentation used the presence of `RedisClient` to determine whether to apply legacy prototype-based instrumentation. This assumption is no longer valid because:

| Version | RedisClient | Multi |
| ------- | ----------- | ----- |
| v3      | ✅           | ✅     |
| v4      | ❌           | ❌     |
| v5      | ❌           | ❌     |
| v6      | ✅           | ❌     |

As Redis v6 reintroduces the `RedisClient` export, checking for `RedisClient` alone incorrectly classifies v6 as a legacy client. Detect the legacy implementation via the combined presence of:

- RedisClient.prototype
- Multi

This uniquely identifies the Redis v3 API shape while ensuring Redis v4–v6 continue to use the modern factory-based instrumentation path.

#### Reference

* Tracking issue: [INSTA-92521](https://jsw.ibm.com/browse/INSTA-92521)


packages/core/src/tracing/instrumentation/databases/redis.js
```diff
@@ -48,9 +48,15 @@ function captureCommands(file) {
...
-  // NOTE: v4 no longer exposes the RedisClient. We need to wait till `createClient` get's called
-  //       to get the instance of the redis client
-  if (!redis.RedisClient) {
+  // NOTE: Redis versions differ in export shape:
+  // v4/v5: are strictly factory-based (createClient/createCluster) and do not expose RedisClient.
+  // v6: expose RedisClient again, but this does NOT imply legacy/class-based behavior.
+  // RedisClient exists in both v3 and v6. Detect the legacy v3 API via
+  // the combined presence of RedisClient and Multi rather than RedisClient alone.
+  const isLegacyRedisAPI =
+    redis.RedisClient?.prototype && typeof redis.Multi?.prototype?.exec_transaction === 'function';
```
aryamohanan: RedisClient alone cannot be used to distinguish v3 from newer v6 because it is present in both v3 and v6.

Replaced the RedisClient-only detection with a check for the legacy class-based API shape.
Detect the legacy implementation via the combined presence of:

- RedisClient
- Multi

Continue using prototype instrumentation only for v3. Use the createClient-based instrumentation for v4+.

abhilash-sivan (approved):
LGTM 🚀 

I couldn't find any breaking changes that appear to impact tracing overall.

https://github.com/redis/node-redis/releases/tag/redis%406.0.0

PR #2600, opened 2026-06-19
chore: updated yargs to v17.7.3 for v26 compatibility

aryamohanan:
Node.js v25.7 and v26 introduced a module-loading change that breaks yargs v16 and v17, which impacts test execution through transitive dependencies such as mocha.

We cannot migrate to yargs v18 because it requires newer Node.js versions and does not support our Node.js  v18. As a temporary workaround, the CI pipeline installs yargs v18 when running on Node.js v26, but local Node.js v26 test runs remained affected.

Now that yargs [17.7.3](https://github.com/yargs/yargs/releases/tag/v17.7.3) has been released with the compatibility fix, add it as a direct dependency and override it across the dependency tree to ensure a consistent version, including transitive dependencies.
ref: https://jsw.ibm.com/browse/INSTA-89969

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     "underscore": "1.13.1",
     "uuid": "9.0.1",
     "ws": "8.16.0",
-    "yargs": "17.5.1"
+    "yargs": "17.7.3"
   },
   "overrides": {
     "prebuildify": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -178,7 +178,8 @@
     },
     "gcstats.js": {
       "nan": "2.27.0"
-    }
+    },
+    "yargs": "17.7.3"
```
aryamohanan: To avoid relying on transitive dependencies such as mocha for yargs resolution, added yargs on overrides too.

abhilash-sivan (approved):
LGTM

PR #2615, opened 2026-06-24
chore: updated mapping

abhilash-sivan:
grpc
aws s3
azure storage
aws lambda
aws lambda invoke

revisitted messaging libs for gap filling

- [ ] add rpc test

### DIFF
--- a/packages/core/src/otlpExporter/common/semconv/base/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/base/mappings.js
@@ -36,7 +36,6 @@ const MAPPINGS = {
     ROUTE: 'http.route',
     STATUS_TEXT: 'http.status_text',
     URL_TEMPLATE: 'http.url.template',
-    NETWORK_PROTOCOL: 'network.protocol.name',
     REQUEST_HEADER: 'http.request.header',
     RESPONSE_HEADER: 'http.response.header'
   },
@@ -46,11 +45,8 @@ const MAPPINGS = {
     BATCH_MESSAGE_COUNT: 'messaging.batch.message_count',
     MESSAGE_ID: 'messaging.message.id',
     MESSAGE_BODY_SIZE: 'messaging.message.body.size',
-    DESTINATION_NAME: 'messaging.destination.name',
-    DESTINATION_TEMPLATE: 'messaging.destination.template',
     DESTINATION_PARTITION_ID: 'messaging.destination.partition.id',
     kafka: {
-
       MESSAGE_KEY: 'messaging.kafka.message.key'
     },
     rabbitmq: {
@@ -60,8 +56,9 @@ const MAPPINGS = {
 
   database: {
     COLLECTION_NAME: 'db.collection.name',
-    TABLE: 'db.sql.table',
-    AWS_DYNAMODB_TABLE_NAMES: 'aws.dynamodb.table_names'
+    dynamodb: {
+      COLLECTION_NAME: 'aws.dynamodb.table_names'
+    }
   },
 
   rpc: {
--- a/packages/core/src/otlpExporter/common/semconv/base/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/base/mappings.js
@@ -63,7 +63,10 @@ const MAPPINGS = {
     USER: 'db.user',
     COLLECTION_NAME: 'db.collection.name',
     TABLE: 'db.sql.table',
-    AWS_DYNAMODB_TABLE_NAMES: 'aws.dynamodb.table_names'
+    AWS_DYNAMODB_TABLE_NAMES: 'aws.dynamodb.table_names',
+    mongodb: {
+      COLLECTION: 'db.mongodb.collection'
+    }
   },
 
   rpc: {
--- a/packages/core/src/otlpExporter/common/semconv/base/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/base/mappings.js
@@ -103,12 +103,7 @@ const MAPPINGS = {
     },
     aws: {
       S3_BUCKET: 'aws.s3.bucket',
-      S3_KEY: 'aws.s3.key',
-      KINESIS_STREAM: 'aws.kinesis.stream_name',
-      KINESIS_SHARD: 'aws.kinesis.shard_id',
-      KINESIS_SHARD_ITERATOR_TYPE: 'aws.kinesis.shard_iterator_type',
-      KINESIS_STARTING_SEQUENCE_NUMBER: 'aws.kinesis.starting_sequence_number',
-      KINESIS_EXPLICIT_HASH_KEY: 'aws.kinesis.explicit_hash_key'
+      S3_KEY: 'aws.s3.key'
     },
     azure: {
       STORAGE_ACCOUNT: 'az.storage.account.name',
--- a/packages/core/src/otlpExporter/common/semconv/base/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/base/mappings.js
@@ -62,12 +62,10 @@ const MAPPINGS = {
   },
 
   database: {
-    OPERATION: 'db.operation.name',
-    NAMESPACE: 'db.namespace',
-    NAME: 'db.name',
     USER: 'db.user',
     COLLECTION: 'db.collection.name',
-    TABLE: 'db.sql.table'
+    TABLE: 'db.sql.table',
+    AWS_DYNAMODB_TABLE_NAMES: 'aws.dynamodb.table_names'
   },
 
   rpc: {
--- a/packages/core/src/otlpExporter/common/semconv/base/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/base/mappings.js
@@ -43,7 +43,6 @@ const MAPPINGS = {
 
   messaging: {
     SYSTEM: 'messaging.system',
-    CONSUMER_GROUP: 'messaging.consumer.group.name',
     BATCH_MESSAGE_COUNT: 'messaging.batch.message_count',
     MESSAGE_ID: 'messaging.message.id',
     MESSAGE_BODY_SIZE: 'messaging.message.body.size',
@@ -51,7 +50,7 @@ const MAPPINGS = {
     DESTINATION_TEMPLATE: 'messaging.destination.template',
     DESTINATION_PARTITION_ID: 'messaging.destination.partition.id',
     kafka: {
-      OFFSET: 'messaging.kafka.message.offset',
+
       MESSAGE_KEY: 'messaging.kafka.message.key'
     },
     rabbitmq: {
@@ -60,13 +59,9 @@ const MAPPINGS = {
   },
 
   database: {
-    USER: 'db.user',
     COLLECTION_NAME: 'db.collection.name',
     TABLE: 'db.sql.table',
-    AWS_DYNAMODB_TABLE_NAMES: 'aws.dynamodb.table_names',
-    mongodb: {
-      COLLECTION: 'db.mongodb.collection'
-    }
+    AWS_DYNAMODB_TABLE_NAMES: 'aws.dynamodb.table_names'
   },
 
   rpc: {
--- a/packages/core/src/otlpExporter/common/semconv/base/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/base/mappings.js
@@ -56,19 +56,17 @@ const MAPPINGS = {
     },
     rabbitmq: {
       ROUTING_KEY: 'messaging.rabbitmq.destination.routing_key'
-    },
-    gcp: { PROJECT_ID: 'gcp.project_id' }
+    }
   },
 
   database: {
     USER: 'db.user',
-    COLLECTION: 'db.collection.name',
+    COLLECTION_NAME: 'db.collection.name',
     TABLE: 'db.sql.table',
     AWS_DYNAMODB_TABLE_NAMES: 'aws.dynamodb.table_names'
   },
 
   rpc: {
-    SYSTEM_NAME: 'rpc.system.name',
     METHOD: 'rpc.method',
     METHOD_ORIGINAL: 'rpc.method_original',
     GRPC_STATUS: 'rpc.grpc.status_code',
@@ -93,22 +91,11 @@ const MAPPINGS = {
     ACCOUNT_ID: 'cloud.account.id',
     RESOURCE_ID: 'cloud.resource_id',
     gcp: {
-      PROJECT_ID: 'gcp.project_id',
-      STORAGE_BUCKET: 'gcp.storage.bucket',
-      STORAGE_OBJECT: 'gcp.storage.object',
-      STORAGE_SOURCE_BUCKET: 'gcp.storage.source.bucket',
-      STORAGE_DESTINATION_BUCKET: 'gcp.storage.destination.bucket',
-      STORAGE_SOURCE_OBJECT: 'gcp.storage.source.object',
-      STORAGE_DESTINATION_OBJECT: 'gcp.storage.destination.object'
+      PROJECT_ID: 'gcp.project_id'
     },
     aws: {
       S3_BUCKET: 'aws.s3.bucket',
       S3_KEY: 'aws.s3.key'
-    },
-    azure: {
-      STORAGE_ACCOUNT: 'az.storage.account.name',
-      CONTAINER: 'az.storage.container.name',
-      BLOB: 'az.storage.blob.name'
     }
   },
 
--- a/packages/core/src/otlpExporter/common/semconv/base/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/base/mappings.js
@@ -46,6 +46,7 @@ const MAPPINGS = {
     OPERATION_TYPE: 'messaging.operation.type',
     OPERATION_NAME: 'messaging.operation.name',
     CONSUMER_GROUP: 'messaging.consumer.group.name',
+    BATCH_MESSAGE_COUNT: 'messaging.batch.message_count',
     MESSAGE_ID: 'messaging.message.id',
     MESSAGE_BODY_SIZE: 'messaging.message.body.size',
     DESTINATION_NAME: 'messaging.destination.name',
@@ -74,11 +75,9 @@ const MAPPINGS = {
   },
 
   rpc: {
-    SYSTEM: 'rpc.system',
     SYSTEM_NAME: 'rpc.system.name',
     METHOD: 'rpc.method',
     METHOD_ORIGINAL: 'rpc.method_original',
-    SERVICE: 'rpc.service',
     GRPC_STATUS: 'rpc.grpc.status_code',
     GRPC_ERROR: 'rpc.grpc.status_message'
   },
@@ -97,7 +96,9 @@ const MAPPINGS = {
   cloud: {
     REGION: 'cloud.region',
     PROVIDER: 'cloud.provider',
+    PLATFORM: 'cloud.platform',
     ACCOUNT_ID: 'cloud.account.id',
+    RESOURCE_ID: 'cloud.resource_id',
     gcp: {
       PROJECT_ID: 'gcp.project_id',
       STORAGE_BUCKET: 'gcp.storage.bucket',
@@ -125,8 +126,15 @@ const MAPPINGS = {
 
   faas: {
     NAME: 'faas.name',
+    VERSION: 'faas.version',
+    INVOCATION_ID: 'faas.invocation_id',
     INVOCATION_TYPE: 'faas.invocation_type',
-    TRIGGER: 'faas.trigger'
+    TRIGGER: 'faas.trigger',
+    COLDSTART: 'faas.coldstart'
+  },
+
+  process: {
+    RUNTIME_NAME: 'process.runtime.name'
   },
 
   exception: {
--- a/packages/core/src/otlpExporter/common/semconv/base/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/base/mappings.js
@@ -55,8 +55,7 @@ const MAPPINGS = {
       MESSAGE_KEY: 'messaging.kafka.message.key'
     },
     rabbitmq: {
-      ROUTING_KEY: 'messaging.rabbitmq.destination.routing_key',
-      MESSAGE_ROUTING_KEY: 'messaging.rabbitmq.destination.routing_key'
+      ROUTING_KEY: 'messaging.rabbitmq.destination.routing_key'
     },
     gcp: { PROJECT_ID: 'gcp.project_id' }
   },
--- a/packages/core/src/otlpExporter/common/semconv/base/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/base/mappings.js
@@ -43,8 +43,6 @@ const MAPPINGS = {
 
   messaging: {
     SYSTEM: 'messaging.system',
-    OPERATION_TYPE: 'messaging.operation.type',
-    OPERATION_NAME: 'messaging.operation.name',
     CONSUMER_GROUP: 'messaging.consumer.group.name',
     BATCH_MESSAGE_COUNT: 'messaging.batch.message_count',
     MESSAGE_ID: 'messaging.message.id',
@@ -58,7 +56,7 @@ const MAPPINGS = {
     },
     rabbitmq: {
       ROUTING_KEY: 'messaging.rabbitmq.destination.routing_key',
-      MESSAGE_ROUTING_KEY: 'messaging.rabbitmq.message.routing_key'
+      MESSAGE_ROUTING_KEY: 'messaging.rabbitmq.destination.routing_key'
     },
     gcp: { PROJECT_ID: 'gcp.project_id' }
   },
@@ -66,12 +64,10 @@ const MAPPINGS = {
   database: {
     OPERATION: 'db.operation.name',
     NAMESPACE: 'db.namespace',
-    QUERY_TEXT: 'db.query.text',
     NAME: 'db.name',
     USER: 'db.user',
     COLLECTION: 'db.collection.name',
-    TABLE: 'db.sql.table',
-    CONNECTION: 'db.connection'
+    TABLE: 'db.sql.table'
   },
 
   rpc: {
--- a/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
@@ -26,7 +26,7 @@ const MAPPINGS = {
     CONNECTION_STRING: 'db.connection_string',
     MONGO_DB_COLLECTION: 'db.mongodb.collection',
     NAME: 'db.name',
-    OPERATION: 'db.operation',
+    OPERATION_NAME: 'db.operation',
     PEER_NAME: 'net.peer.name',
     PEER_PORT: 'net.peer.port',
     QUERY_STATEMENT: 'db.statement',
--- a/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
@@ -18,9 +18,6 @@ const MAPPINGS = {
     DESTINATION_NAME: 'messaging.destination',
     kafka: {
       PARTITION: 'messaging.kafka.partition'
-    },
-    sqs: {
-      SYSTEM: 'aws.sqs'
     }
   },
 
--- a/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
@@ -36,6 +36,10 @@ const MAPPINGS = {
   network: {
     PEER_NAME: 'net.peer.name',
     PEER_PORT: 'net.peer.port'
+  },
+
+  rpc: {
+    SYSTEM_NAME: 'rpc.system'
   }
 };
 
--- a/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
@@ -29,7 +29,7 @@ const MAPPINGS = {
     OPERATION: 'db.operation',
     PEER_NAME: 'net.peer.name',
     PEER_PORT: 'net.peer.port',
-    STATEMENT: 'db.statement',
+    QUERY_STATEMENT: 'db.statement',
     SYSTEM: 'db.system'
   },
 
--- a/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
@@ -16,15 +16,18 @@ const MAPPINGS = {
 
   messaging: {
     DESTINATION_NAME: 'messaging.destination',
+    OPERATION: 'messaging.operation',
     kafka: {
       PARTITION: 'messaging.kafka.partition'
     }
   },
 
   database: {
-    SYSTEM: 'db.system',
+    CONNECTION_STRING: 'db.connection_string',
     PEER_NAME: 'net.peer.name',
-    PEER_PORT: 'net.peer.port'
+    PEER_PORT: 'net.peer.port',
+    STATEMENT: 'db.statement',
+    SYSTEM: 'db.system'
   },
 
   network: {
--- a/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
@@ -16,7 +16,7 @@ const MAPPINGS = {
 
   messaging: {
     DESTINATION_NAME: 'messaging.destination',
-    OPERATION: 'messaging.operation',
+    OPERATION_NAME: 'messaging.operation',
     kafka: {
       PARTITION: 'messaging.kafka.partition'
     }
--- a/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
@@ -24,6 +24,9 @@ const MAPPINGS = {
 
   database: {
     CONNECTION_STRING: 'db.connection_string',
+    MONGO_DB_COLLECTION: 'db.mongodb.collection',
+    NAME: 'db.name',
+    OPERATION: 'db.operation',
     PEER_NAME: 'net.peer.name',
     PEER_PORT: 'net.peer.port',
     STATEMENT: 'db.statement',
--- a/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
+++ b/packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
@@ -18,19 +18,24 @@ const MAPPINGS = {
     DESTINATION_NAME: 'messaging.destination',
     OPERATION_NAME: 'messaging.operation',
     kafka: {
-      PARTITION: 'messaging.kafka.partition'
+      PARTITION: 'messaging.kafka.partition',
+      CONSUMER_GROUP: 'messaging.kafka.consumer.group',
+      OFFSET: 'messaging.kafka.message.offset'
     }
   },
 
   database: {
     CONNECTION_STRING: 'db.connection_string',
-    MONGO_DB_COLLECTION: 'db.mongodb.collection',
-    NAME: 'db.name',
+    NAMESPACE: 'db.name',
     OPERATION_NAME: 'db.operation',
     PEER_NAME: 'net.peer.name',
     PEER_PORT: 'net.peer.port',
-    QUERY_STATEMENT: 'db.statement',
-    SYSTEM: 'db.system'
+    QUERY_TEXT: 'db.statement',
+    SYSTEM_NAME: 'db.system',
+    USER: 'db.user',
+    mongodb: {
+      COLLECTION: 'db.mongodb.collection'
+    }
   },
 
   network: {
--- a/packages/core/src/otlpExporter/traces/mappers/constants.js
+++ b/packages/core/src/otlpExporter/traces/mappers/constants.js
@@ -35,7 +35,27 @@ exports.INSTRUMENTATION_TYPES = {
   S3: 's3',
   KINESIS: 'kinesis',
   AZSTORAGE: 'azstorage',
-  AWS_LAMBDA_INVOKE: 'aws.lambda.invoke'
+  AWS_LAMBDA_INVOKE: 'aws.lambda.invoke',
+  AWS_LAMBDA_ENTRY: 'lambda'
+};
+
+/**
+ * Maps Instana lambda.trigger values to OTel faas.trigger values.
+ * @type {Record<string, string>}
+ */
+exports.LAMBDA_TRIGGER_MAP = {
+  'aws:api.gateway': 'http',
+  'aws:api.gateway.noproxy': 'http',
+  'aws:application.load.balancer': 'http',
+  'aws:lambda.function.url': 'http',
+  'aws:s3': 'datasource',
+  'aws:dynamodb': 'datasource',
+  'aws:kinesis': 'datasource',
+  'aws:kinesis.firehouse': 'datasource',
+  'aws:sqs': 'pubsub',
+  'aws:sns': 'pubsub',
+  'aws:cloudwatch.events': 'timer',
+  'aws:lambda.invoke': 'other'
 };
 
 exports.OTLP_STATUS_CODES = {
--- a/packages/core/src/otlpExporter/traces/mappers/constants.js
+++ b/packages/core/src/otlpExporter/traces/mappers/constants.js
@@ -35,7 +35,7 @@ exports.INSTRUMENTATION_TYPES = {
   S3: 's3',
   KINESIS: 'kinesis',
   AZSTORAGE: 'azstorage',
-  AWS_LAMBDA_INVOKE: 'lambda',
+  AWS_LAMBDA_INVOKE: 'aws.lambda.invoke',
   AWS_LAMBDA_ENTRY: 'lambda'
 };
 
@@ -55,7 +55,8 @@ exports.LAMBDA_TRIGGER_MAP = {
   'aws:sqs': 'pubsub',
   'aws:sns': 'pubsub',
   'aws:cloudwatch.events': 'timer',
-  'aws:lambda.invoke': 'other'
+  'aws:lambda.invoke': 'other',
+  'aws:cloudwatch.logs': 'other'
 };
 
 exports.OTLP_STATUS_CODES = {
--- a/packages/core/src/otlpExporter/traces/mappers/constants.js
+++ b/packages/core/src/otlpExporter/traces/mappers/constants.js
@@ -51,7 +51,6 @@ exports.LAMBDA_TRIGGER_MAP = {
   'aws:s3': 'datasource',
   'aws:dynamodb': 'datasource',
   'aws:kinesis': 'datasource',
-  'aws:kinesis.firehouse': 'datasource',
   'aws:sqs': 'pubsub',
   'aws:sns': 'pubsub',
   'aws:cloudwatch.events': 'timer',
--- a/packages/core/src/otlpExporter/traces/mappers/constants.js
+++ b/packages/core/src/otlpExporter/traces/mappers/constants.js
@@ -35,7 +35,7 @@ exports.INSTRUMENTATION_TYPES = {
   S3: 's3',
   KINESIS: 'kinesis',
   AZSTORAGE: 'azstorage',
-  AWS_LAMBDA_INVOKE: 'aws.lambda.invoke',
+  AWS_LAMBDA_INVOKE: 'lambda',
   AWS_LAMBDA_ENTRY: 'lambda'
 };
 
--- a/packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
+++ b/packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
@@ -147,9 +147,7 @@ const instrumentationMappings = {
     spanName: data => `${data.sort} ${data.queue}`,
     spanAttributes: [
       { otlp: OTLP.messaging.SYSTEM, value: 'aws_sqs' },
-      { otlp: OTLP.faas.TRIGGER, value: 'pubsub' },
       { otlp: OTLP.messaging.OPERATION_NAME, instana: 'type' },
-      { otlp: OTLP.messaging.OPERATION_TYPE, value: 'process' },
       { otlp: OTLP.messaging.DESTINATION_NAME, instana: 'queue' },
       { otlp: OTLP.messaging.BATCH_MESSAGE_COUNT, instana: 'size' },
       { otlp: OTLP.messaging.MESSAGE_ID, instana: 'messageId' },
@@ -200,7 +198,7 @@ const instrumentationMappings = {
   [INSTRUMENTATION_TYPES.MYSQL]: {
     spanName: data => data.stmt,
     spanAttributes: [
-      { otlp: OTLP.database.SYSTEM, value: 'mysql' },
+      { otlp: OTLP.database.SYSTEM, value: INSTRUMENTATION_TYPES.MYSQL },
       { otlp: OTLP.database.QUERY_STATEMENT, instana: 'stmt' },
       { otlp: OTLP.database.USER, instana: 'user' },
       { otlp: OTLP.database.NAME, instana: 'db' },
@@ -213,7 +211,7 @@ const instrumentationMappings = {
   [INSTRUMENTATION_TYPES.MSSQL]: {
     spanName: data => data.stmt,
     spanAttributes: [
-      { otlp: OTLP.database.SYSTEM, value: 'mssql' },
+      { otlp: OTLP.database.SYSTEM, value: INSTRUMENTATION_TYPES.MSSQL },
       { otlp: OTLP.database.QUERY_STATEMENT, instana: 'stmt' },
       { otlp: OTLP.database.USER, instana: 'user' },
       { otlp: OTLP.database.NAME, instana: 'db' },
@@ -262,7 +260,7 @@ const instrumentationMappings = {
   [INSTRUMENTATION_TYPES.COUCHBASE]: {
     spanName: data => data.bucket,
     spanAttributes: [
-      { otlp: OTLP.database.SYSTEM, value: 'couchbase' },
+      { otlp: OTLP.database.SYSTEM, value: [INSTRUMENTATION_TYPES.COUCHBASE] },
       { otlp: OTLP.database.NAMESPACE, instana: 'bucket' },
       { otlp: OTLP.database.NAME, instana: 'bucket' },
       { otlp: OTLP.database.QUERY_STATEMENT, instana: 'sql' },
@@ -299,7 +297,7 @@ const instrumentationMappings = {
   [INSTRUMENTATION_TYPES.MEMCACHED]: {
     spanName: data => data.operation,
     spanAttributes: [
-      { otlp: OTLP.database.SYSTEM, value: 'memcached' },
+      { otlp: OTLP.database.SYSTEM, value: INSTRUMENTATION_TYPES.MEMCACHED },
       { otlp: OTLP.database.CONNECTION_STRING, instana: 'connection' },
       { otlp: OTLP.database.OPERATION_NAME, instana: 'operation' },
       { otlp: OTLP.server.ADDRESS, instana: 'connection', transform: extractHost },
--- a/packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
+++ b/packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
@@ -12,7 +12,10 @@ const {
   extractHost,
   extractPort,
   getRPCMethod,
-  getRPCMethodOriginal
+  getRPCMethodOriginal,
+  toBoolean,
+  getRegionFromArn,
+  getAccountIdFromArn
 } = require('./util');
 
 const ctx = require('../../common/context');
@@ -240,7 +243,7 @@ const instrumentationMappings = {
     spanAttributes: [
       { otlp: OTLP.database.SYSTEM, value: 'mongodb' },
       { otlp: OTLP.database.NAMESPACE, instana: 'namespace' },
-      { otlp: OTLP.database.mongodb.COLLECTION_NAME, instana: 'collection' },
+      { otlp: OTLP.database.mongodb.COLLECTION, instana: 'collection' },
       { otlp: OTLP.database.COLLECTION_NAME, instana: 'collection' },
       { otlp: OTLP.database.OPERATION, instana: 'command', transform: toUpperCase },
       { otlp: OTLP.database.QUERY_TEXT, instana: ['json', 'filter'], transform: firstDefined },
@@ -436,30 +439,19 @@ const instrumentationMappings = {
       {
         otlp: OTLP.cloud.REGION,
         instana: 'arn',
-        transform: arn => {
-          // arn:aws:lambda:{region}:{account-id}:function:{name}:{version}
-          const parts = typeof arn === 'string' ? arn.split(':') : [];
-          return parts[3] || undefined;
-        }
+        transform: arn => getRegionFromArn(arn)
       },
       {
         otlp: OTLP.cloud.ACCOUNT_ID,
         instana: 'arn',
-        transform: arn => {
-          const parts = typeof arn === 'string' ? arn.split(':') : [];
-          return parts[4] || undefined;
-        }
+        transform: arn => getAccountIdFromArn(arn)
       },
       { otlp: OTLP.cloud.RESOURCE_ID, instana: 'arn' },
       { otlp: OTLP.faas.INVOCATION_ID, instana: 'reqId' },
       {
         otlp: OTLP.faas.COLDSTART,
         instana: 'coldStart',
-        transform: value => {
-          if (typeof value === 'boolean') return value;
-          if (typeof value === 'string') return value === 'true';
-          return undefined;
-        }
+        transform: value => toBoolean(value)
       },
       { otlp: OTLP.process.RUNTIME_NAME, instana: 'runtime' },
       { otlp: OTLP.exception.MESSAGE, instana: 'error' }
--- a/packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
+++ b/packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
@@ -84,7 +84,7 @@ const instrumentationMappings = {
   [INSTRUMENTATION_TYPES.KAFKA]: {
     spanName: data => `${data.operation} ${data.endpoints}`,
     spanAttributes: [
-      { otlp: OTLP.messaging.SYSTEM, value: 'kafka' },
+      { otlp: OTLP.messaging.SYSTEM, value: INSTRUMENTATION_TYPES.KAFKA},
       { otlp: OTLP.messaging.DESTINATION_NAME, instana: 'endpoints' },
       { otlp: OTLP.messaging.OPERATION_NAME, instana: 'operation' },
       {
@@ -96,12 +96,12 @@ const instrumentationMappings = {
   },
 
   [INSTRUMENTATION_TYPES.RABBITMQ]: {
-    spanName: data => `${data.sort || 'process'} ${data.exchange || data.key || data.queue || 'unknown'}`,
+    spanName: data => `${data.sort} ${data.exchange} || data.key}`
     spanAttributes: [
-      { otlp: OTLP.messaging.SYSTEM, value: 'rabbitmq' },
+      { otlp: OTLP.messaging.SYSTEM, value: INSTRUMENTATION_TYPES.RABBITMQ },
       { otlp: OTLP.messaging.OPERATION_NAME, instana: 'sort' },
       { otlp: OTLP.messaging.DESTINATION_NAME, instana: ['exchange', 'key', 'queue'] },
-      { otlp: OTLP.messaging.rabbitmq.ROUTING_KEY, instana: 'exchange' },
+      { otlp: OTLP.messaging.rabbitmq.ROUTING_KEY, instana: 'key' },
       { otlp: OTLP.messaging.rabbitmq.MESSAGE_ROUTING_KEY, instana: 'key' },
       { otlp: OTLP.messaging.MESSAGE_BODY_SIZE, instana: 'size' },
       { otlp: OTLP.server.ADDRESS, instana: 'address', transform: extractHost },
--- a/packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
+++ b/packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
@@ -226,7 +226,7 @@ const instrumentationMappings = {
   },
 
   [INSTRUMENTATION_TYPES.MONGO]: {
-    spanName: data => `mongo.${data.command}`,
+    spanName: data => data.command,
     spanAttributes: [
       { otlp: OTLP.database.SYSTEM, value: 'mongodb' },
       { otlp: OTLP.database.NAMESPACE, instana: 'namespace' },
@@ -242,7 +242,7 @@ const instrumentationMappings = {
   },
 
   [INSTRUMENTATION_TYPES.REDIS]: {
-    spanName: data => `redis.${data.operation}`,
+    spanName: data => data.operation,
     spanAttributes: [
       { otlp: OTLP.database.SYSTEM, value: INSTRUMENTATION_TYPES.REDIS },
       { otlp: OTLP.database.OPERATION, instana: 'operation' },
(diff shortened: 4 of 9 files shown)


packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -76,11 +84,23 @@ const instrumentationMappings = {
   },
 
   [INSTRUMENTATION_TYPES.KAFKA]: {
-    spanName: data => `${data.operation} ${data.endpoints}`,
+    // In KafkaNode, we collect service directly
+    spanName: data => `${data.operation || data.access} ${data.service || data.endpoints}`,
```
aryamohanan: I don't think this change is correct. Kafka doesn't have an `access` field in the internal format, we already normalize it to `operation`, same for  `service` and `endpoint`. 

```suggestion
    spanName: data => `${data.operation} ${data.endpoints}`,
```

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -76,11 +84,23 @@ const instrumentationMappings = {
...
     spanAttributes: [
       { otlp: OTLP.messaging.SYSTEM, value: 'kafka' },
-      { otlp: OTLP.messaging.DESTINATION_NAME, instana: 'endpoints' },
-      { otlp: OTLP.messaging.OPERATION_NAME, instana: 'operation' },
+      { otlp: OTLP.messaging.DESTINATION_NAME, instana: ['service', 'endpoints'], transform: firstDefined },
```
aryamohanan: here too

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -76,11 +84,23 @@ const instrumentationMappings = {
...
+      { otlp: OTLP.messaging.OPERATION_NAME, instana: ['access', 'operation'], transform: firstDefined },
+      {
+        otlp: OTLP.messaging.OPERATION_TYPE,
+        instana: ['access', 'operation'],
+        transform: values => {
+          const op = firstDefined(values);
+          if (op === 'send') return 'send';
+          if (op === 'consume') return 'receive';
```
aryamohanan: Why are we updating the value? IMO, we shouldn't modify it, we should send it as-is. These are package-specific operation names, and they can differ across packages or even between language implementations.

TBD

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -76,11 +84,23 @@ const instrumentationMappings = {
...
+          return op || null;
+        }
+      },
+      { otlp: OTLP.messaging.BATCH_MESSAGE_COUNT, spanLevel: 'b.s' },
```
aryamohanan: I don't think this logic is correct. b.s represents the span batch size and shouldn't be coupled with the messaging batch size.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -112,10 +133,19 @@ const instrumentationMappings = {
...
+      return `${op} ${data.queue || 'unknown'}`;
+    },
     spanAttributes: [
       { otlp: OTLP.messaging.SYSTEM, value: 'bull' },
-      { otlp: OTLP.messaging.OPERATION_NAME, instana: 'sort' },
+      {
+        // bull.sort 'exit' → 'send', 'entry' → 'process'
+        otlp: OTLP.messaging.OPERATION_NAME,
+        instana: 'sort',
+        transform: sort => BULL_SORT_OP[String(sort)] || sort
```
aryamohanan: Here too

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -125,11 +155,12 @@ const instrumentationMappings = {
...
     spanName: data => `${data.type || data.sort || 'process'} ${data.queue || 'unknown'}`,
     spanAttributes: [
-      { otlp: OTLP.messaging.SYSTEM, value: OTLP.messaging.sqs?.SYSTEM || 'aws.sqs' },
+      { otlp: OTLP.messaging.SYSTEM, value: 'aws_sqs' },
       { otlp: OTLP.messaging.OPERATION_NAME, instana: 'type' },
       { otlp: OTLP.messaging.CONSUMER_GROUP, instana: 'group' },
       { otlp: OTLP.messaging.DESTINATION_NAME, instana: 'queue' },
-      { otlp: OTLP.messaging.MESSAGE_BODY_SIZE, instana: 'size' },
+      // sqs.size = number of messages in a batch send (SendMessageBatch), not body bytes
+      { otlp: OTLP.messaging.BATCH_MESSAGE_COUNT, instana: 'size' },
```
aryamohanan, line 148: This too

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: In KafkaNode, we collect access and service directly.
see: https://github.com/instana/nodejs/blob/e9408f7df8f506cb6933b073d740259384134464/packages/core/src/tracing/instrumentation/messaging/kafkaNode.js#L127

Seems like this is a valid case 🤔

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: Yeah, we already discussed this. I took the mapping from RFD, but as you said, modifying the value doesn't seem correct. I will revert this change

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: Agree

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan: I think this is something we ignored earlier. I think we need a refactoring PR for kafka node to use internal format.

aryamohanan: its not a good idea to handle internal and be formats

abhilash-sivan: https://github.com/instana/nodejs/pull/2621

packages/core/src/otlpExporter/traces/mappers/constants.js
```diff
@@ -35,7 +35,27 @@ exports.INSTRUMENTATION_TYPES = {
...
+  AWS_LAMBDA_INVOKE: 'aws.lambda.invoke',
+  AWS_LAMBDA_ENTRY: 'lambda'
+};
+
+/**
+ * Maps Instana lambda.trigger values to OTel faas.trigger values.
+ * @type {Record<string, string>}
+ */
+exports.LAMBDA_TRIGGER_MAP = {
+  'aws:api.gateway': 'http',
```
aryamohanan, line 47: Why this?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -296,14 +308,48 @@ const instrumentationMappings = {
...
+      },
+      { otlp: OTLP.rpc.SYSTEM_NAME, instana: 'flavor' },
+      { otlp: OTLP.server.ADDRESS, instana: 'host' },
+      { otlp: OTLP.server.PORT, instana: 'port' },
       { otlp: OTLP.rpc.GRPC_ERROR, instana: 'error' }
     ]
   },
 
+  // Note: Graphql is not specified in RFD
+  // Will revisit if RFD is updated
```
aryamohanan: Is this available in the OTel specification? IMO, the RFD may not include all of our instrumentation mappings, so the OTel specification should be treated as the source of truth.

packages/core/src/otlpExporter/traces/mappers/constants.js
abhilash-sivan: https://github.ibm.com/instana/technical-documentation/pull/389/files#diff-8c000a237a5b02307d5be328873fa852d6b88797e1774a06511664f7e3f2aa05R1720

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: https://opentelemetry.io/docs/specs/semconv/graphql/graphql-spans/

It is available in the OTEL spec; I will create a separate PR for the missing items later. This needs to be added to the RFD as well, but no need to do it right now; I will update RFD once we merge the changes

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan:
```suggestion
      { otlp: OTLP.messaging.SYSTEM, value: INSTRUMENTATION_TYPES.RABBITMQ },
```

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: I just want to let you know that the current changes are inherited from looking at the Otel semconv,` but it is only specified for server spans; we adapted it for both entry and exit spans in our case

packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
```diff
@@ -18,9 +18,6 @@ const MAPPINGS = {
     DESTINATION_NAME: 'messaging.destination',
     kafka: {
       PARTITION: 'messaging.kafka.partition'
```
aryamohanan: I miss messaging.operation in v1.23 and messaging.operation.name in v1.41

packages/core/src/otlpExporter/common/semconv/base/mappings.js
```diff
@@ -46,6 +46,7 @@ const MAPPINGS = {
     OPERATION_TYPE: 'messaging.operation.type',
```
aryamohanan: This should be in v1.41

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -81,6 +87,10 @@ const instrumentationMappings = {
       { otlp: OTLP.messaging.SYSTEM, value: 'kafka' },
```
aryamohanan:
```suggestion
      { otlp: OTLP.messaging.SYSTEM, value: INSTRUMENTATION_TYPES.KAFKA},
```

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -93,6 +103,7 @@ const instrumentationMappings = {
       { otlp: OTLP.messaging.DESTINATION_NAME, instana: ['exchange', 'key', 'queue'] },
```
aryamohanan: qs: Are we collecting `rabbitmq.queue` from our instrumentation?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -93,6 +103,7 @@ const instrumentationMappings = {
...
       { otlp: OTLP.messaging.rabbitmq.ROUTING_KEY, instana: 'exchange' },
```
aryamohanan: issue:
```suggestion
      { otlp: OTLP.messaging.rabbitmq.ROUTING_KEY, instana: 'key' },
```

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -93,6 +103,7 @@ const instrumentationMappings = {
...
       { otlp: OTLP.messaging.rabbitmq.MESSAGE_ROUTING_KEY, instana: 'key' },
```
aryamohanan: Where did you get this mapping key `messaging.rabbitmq.message.routing_key`

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan:
```suggestion
    spanName: data => `${data.sort} ${data.exchange} || data.key}`
```
According to instr, `queue` will never be present, so there's no benefit in including it and sort will be always there in rabbitmq instr.

aryamohanan:
```suggestion
      { otlp: OTLP.messaging.SYSTEM, value: INSTRUMENTATION_TYPES.NATS },
```

aryamohanan:
```suggestion
    spanName: data => `${data.sort } ${data.subject '}`,
```
`sort` and `subject` always have a value, please double confirm

aryamohanan:
```suggestion
      { otlp: OTLP.messaging.SYSTEM, value: INSTRUMENTATION_TYPES.BULL },
```

aryamohanan:
```suggestion
    spanName: data => `${data.sort} ${data.queue}`,
```

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: This is a manual error; the actual key is `messaging.rabbitmq.destination.routing_key`

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: No, we are not. The purpose of adding some of these fields is that I thought even if we do not collect the data now, we can add this later because some of this was marked as mandatory in the RFD, so even if we keep this mapping here right now, the functionality won't be affected because we will clean up the undefined data.

from rfd:


rabbitmq.exchange + rabbitmq.key + rabbitmq.queue | messaging.destination.name
-- | --




That said, with the more recent changes in mapping, we went with stricter checking of what our instrumentation produces and retained only the relevant mappings. Given that, this can be removed. Thanks!

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -125,11 +136,12 @@ const instrumentationMappings = {
   [INSTRUMENTATION_TYPES.SQS]: {
     spanName: data => `${data.type || data.sort || 'process'} ${data.queue || 'unknown'}`,
     spanAttributes: [
-      { otlp: OTLP.messaging.SYSTEM, value: OTLP.messaging.sqs?.SYSTEM || 'aws.sqs' },
+      { otlp: OTLP.messaging.SYSTEM, value: 'aws_sqs' },
       { otlp: OTLP.messaging.OPERATION_NAME, instana: 'type' },
```
aryamohanan, line 146: Note: Type seems mandatory field but we only collect that in exit spans? I checked v3/sqs instrumentation. Please research and confirm.

The values we collected are "single.sync" or "batch.sync".

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -125,11 +136,12 @@ const instrumentationMappings = {
   [INSTRUMENTATION_TYPES.SQS]: {
     spanName: data => `${data.type || data.sort || 'process'} ${data.queue || 'unknown'}`,
```
aryamohanan:
```suggestion
    spanName: data => `${data.sort} ${data.queue}`,
```

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -125,11 +136,12 @@ const instrumentationMappings = {
...
     spanAttributes: [
-      { otlp: OTLP.messaging.SYSTEM, value: OTLP.messaging.sqs?.SYSTEM || 'aws.sqs' },
+      { otlp: OTLP.messaging.SYSTEM, value: 'aws_sqs' },
       { otlp: OTLP.messaging.OPERATION_NAME, instana: 'type' },
       { otlp: OTLP.messaging.CONSUMER_GROUP, instana: 'group' },
       { otlp: OTLP.messaging.DESTINATION_NAME, instana: 'queue' },
-      { otlp: OTLP.messaging.MESSAGE_BODY_SIZE, instana: 'size' },
+      // TBD sqs.size = number of messages in a batch send (SendMessageBatch), not body bytes
+      { otlp: OTLP.messaging.BATCH_MESSAGE_COUNT, instana: 'size' },
```
aryamohanan, line 148: I agree, may be we can remove the comment

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: In v2, also, we collect type

>  return cls.ns.runAndReturn(() => {
>     const span = cls.startSpan({
>       spanName: 'sqs',
>       kind: EXIT
>     });
>     span.ts = Date.now();
>     span.stack = tracingUtil.getStackTrace(instrumentedSendMessage);
>     span.data.sqs = {
>       sort: sortTypes.EXIT,
>       type: messageData.Entries ? callTypes.SEND_MESSAGE_BATCH : callTypes.SEND_MESSAGE,
>       group: messageData.MessageGroupId,
>       queue: originalArgs[0].QueueUrl || ''
>     };

here also we only collect "type" for exit spans.

But I don't think that impacts this mapping. We map everything here—if the field is present, we convert it; otherwise, we clean it up. Also, there is no type field in the resulting OpenTelemetry span. correct ?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan: > The purpose of adding some of these fields is that I thought even if we do not collect the data now, we can add this later because some of this was marked as mandatory in the RFD,

Please don't rely solely on the RFD. As you know, it isn't thoroughly reviewed or guaranteed to match every tracer, at least for us. It's better to validate the requirements against our instrumentation rather than assuming everything in the RFD is accurate.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan: I'm bit confused: now https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs-message

- [faas.trigger](https://opentelemetry.io/docs/specs/semconv/faas/faas-spans/) MUST be set to pubsub.
- [messaging.operation.type](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/) MUST be set to process.
- [messaging.system](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/) MUST be set to aws_sqs.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan, line 160: I think we should add a note here that we don't have the actual operation name. For now, we're setting it to "send" based on the RFD.
```suggestion
     // We don't have the actual operation name. Using "send" based on the RFD.
      { otlp: OTLP.messaging.OPERATION_NAME, value: 'send' },
```

aryamohanan:
```suggestion
    spanName: data => `publish ${data.topic }`,
```

aryamohanan, line 158: We have topic and target, its good to double check the mapping

aryamohanan:
```suggestion
```

aryamohanan: qs: Where does this logic come from?

packages/core/src/otlpExporter/common/semconv/base/mappings.js
aryamohanan: db.operation.name not in v1.23

aryamohanan: db.query.text not in v1.23

aryamohanan: Is db.name a valid OTLP attribute? I couldn't find it in the OpenTelemetry specification. Could you please double-check? I might have missed something.

aryamohanan: `db.connection` a valid OTLP attribute? ot its `db.connection_string`?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan: same qs: Where does this logic come from?

aryamohanan:
```suggestion
      { otlp: OTLP.database.SYSTEM, value: INSTRUMENTATION_TYPES.DB2 },
```

aryamohanan: We may have missed capturing the connection string. For DB2, we capture the `dsn` as the connection string. Could you please check?

aryamohanan: Its not valid

aryamohanan: Please double check

aryamohanan: Do we have collection in span properties?

aryamohanan: qs: Do we need transform to uppercase?

aryamohanan:
```suggestion
      { otlp: OTLP.database.SYSTEM, value: INSTRUMENTATION_TYPES.REDIS},
```

aryamohanan:
```suggestion
    spanName: data => `redis.${data.operation }`,
```
Also need to convert to uppercase. Please check

aryamohanan: What about server address and server port? Thats missing

aryamohanan: This should be "couchdb", see: https://opentelemetry.io/docs/specs/semconv/db/couchdb/#spans

aryamohanan, line 261: We collected `hostname`, that can be used as connectonstring mapping please check

aryamohanan: Please validate whether its db.namespace. I think its db.namespace, see: https://opentelemetry.io/docs/specs/semconv/db/couchdb/#spans

aryamohanan: operation not valid in case of couchdb, we are not collecting it.

aryamohanan:
```suggestion
      { otlp: OTLP.database.SYSTEM, value: INSTRUMENTATION_TYPES.ELASTICSEARCH},
```

aryamohanan, line 272: Are we added db.statement in semconv v1.23? I think query_text is latest semconv.

aryamohanan: I think its namespace

aryamohanan: request?

aryamohanan: request?

aryamohanan:
```suggestion
      { otlp: OTLP.database.SYSTEM, value: INSTRUMENTATION_TYPES.DYNAMODB},
```

aryamohanan: I missed `db.collection.name` in this https://opentelemetry.io/docs/specs/semconv/db/dynamodb/

Its [aws.dynamodb.table_names](https://opentelemetry.io/docs/specs/semconv/registry/attributes/aws/)

aryamohanan: command is not valid

aryamohanan: server address and port is missing, we can get the value from connection

aryamohanan, line 314: We can also find server address and port from url

aryamohanan:
```suggestion
    spanName: data => `prisma.${data.action`,
```

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -296,23 +308,60 @@ const instrumentationMappings = {
   [INSTRUMENTATION_TYPES.RPC]: {
     spanName: data => data.call || 'rpc.call',
```
aryamohanan: 'rpc.call'?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -296,23 +308,60 @@ const instrumentationMappings = {
...
     spanAttributes: [
-      { otlp: OTLP.rpc.METHOD, instana: 'call' },
-      { otlp: OTLP.rpc.SYSTEM, instana: 'flavor' },
-      { otlp: OTLP.network.PEER_NAME, instana: 'host' },
-      { otlp: OTLP.network.PEER_PORT, instana: 'port' },
+      {
+        otlp: OTLP.rpc.METHOD,
+        instana: 'call',
+        transform: (value, spanData) => {
```
aryamohanan: complex logic move to util.js

aryamohanan (changes requested):

packages/core/src/otlpExporter/common/semconv/base/mappings.js
abhilash-sivan: https://opentelemetry.opendocs.io/docs/specs/semconv/database/database-spans/#call-level-attributes

This is all I could find on v1.23. This seems like an archive for v123,

abhilash-sivan: moved to v1.41 and added db.statement for v1.23
refs 
v1.23: https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/database/database-spans.md

v1.42: https://opentelemetry.io/docs/specs/semconv/db/database-spans/

abhilash-sivan: official doc: https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/database/database-spans.md

abhilash-sivan: yes it is connection_string and there is no replacement found in v 1.43, updated

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: seems like so

abhilash-sivan: There is no spec for connection string in RFD and also could not find the the official spec for DB2,

abhilash-sivan: This si present in v1.23 https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/database/database-spans.md

abhilash-sivan: host, port, user info are not currently collected. It is now removed
However, if we decide to keep all possible data for otel, we can add this later

abhilash-sivan: It seems like we do not collect collections directly even though it is required by the OTEL spec. We can remove this and add later if required

abhilash-sivan: I think no need to use case conversion

Span name SHOULD follow the general [database span name convention](https://opentelemetry.io/docs/specs/semconv/db/database-spans/#name) except that db.namespace SHOULD NOT be used in the span name since it is a numeric value that ends up looking confusing.

abhilash-sivan: Thanks, this was a mistake in th RFD

abhilash-sivan: This is not attribbute, it was a fallback of string "operation" but not required, as bucket will always be there

abhilash-sivan: db.name is available in v1.23, whereas db.namespace was introduced in v1.41.

Although db.name and db.namespace seem similar, they are not equivalent. In most of our instrumentations, we only collect the database name; MongoDB is one of the exception where we specifically collect the namespace.

Because of that, I'm not sure we can simply populate both attributes in the mapper. Doing so would mean treating db.name as db.namespace, which isn't always semantically correct. As a result, v1.23 and v1.41 may not always contain equivalent data.

abhilash-sivan: this is only a fallback string; I plan to remove all the fallback strings
spanName population need to be revisited, as this was not specifically mentioned in the RFD. I did not modify much

abhilash-sivan: corrected now

packages/core/src/otlpExporter/common/semconv/base/mappings.js
```diff
@@ -57,28 +56,22 @@ const MAPPINGS = {
     },
     rabbitmq: {
       ROUTING_KEY: 'messaging.rabbitmq.destination.routing_key',
-      MESSAGE_ROUTING_KEY: 'messaging.rabbitmq.message.routing_key'
+      MESSAGE_ROUTING_KEY: 'messaging.rabbitmq.destination.routing_key'
```
aryamohanan: Its a duplication of ROUTING_KEY

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -33,6 +38,7 @@ const OTLP = /** @type {any} */ (ctx.semConv);
  * @typedef {Object} AttributeMapping
  * @property {string} otlp
  * @property {string | string[]} [instana]
+ * @property {string} [spanLevel]
```
aryamohanan: Remove unused spanLevel

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan: The mapping should only contains available values from Instana.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: I am confused because we have added a logic in aws.lambda.entry span and trigger mapping is happeining there. Maybe we need this here too. TBD

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: Added host, port and connection_string attributes

abhilash-sivan: name attribute is for v1.23, but I am not sure if 'name' == `namespace`

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: done

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: removed

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: discussed: use statement

abhilash-sivan: added

abhilash-sivan: done

packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
```diff
@@ -16,18 +16,21 @@ const MAPPINGS = {
...
-    },
-    sqs: {
-      SYSTEM: 'aws.sqs'
     }
   },
 
   database: {
-    SYSTEM: 'db.system',
+    CONNECTION_STRING: 'db.connection_string',
+    MONGO_DB_COLLECTION: 'db.mongodb.collection',
```
aryamohanan: Can we use database.mongodb.collection instead of MONGO_DB_COLLECTION?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -202,119 +226,143 @@ const instrumentationMappings = {
...
   },
 
   [INSTRUMENTATION_TYPES.MONGO]: {
-    spanName: data => `mongo.${data.command}`,
+    spanName: data => data.command,
     spanAttributes: [
       { otlp: OTLP.database.SYSTEM, value: 'mongodb' },
       { otlp: OTLP.database.NAMESPACE, instana: 'namespace' },
+      { otlp: OTLP.database.MONGO_DB_COLLECTION, instana: 'collection' },
       { otlp: OTLP.database.COLLECTION, instana: 'collection' },
```
aryamohanan: nitpick: rename all references COLLECTION => COLLECTION_NAME

packages/core/src/otlpExporter/common/semconv/base/mappings.js
```diff
@@ -97,7 +89,9 @@ const MAPPINGS = {
   cloud: {
     REGION: 'cloud.region',
     PROVIDER: 'cloud.provider',
+    PLATFORM: 'cloud.platform',
     ACCOUNT_ID: 'cloud.account.id',
+    RESOURCE_ID: 'cloud.resource_id',
     gcp: {
       PROJECT_ID: 'gcp.project_id',
       STORAGE_BUCKET: 'gcp.storage.bucket',
```
aryamohanan: These are not valid OTLP attributes, so I'd prefer not to add them here. For now, we can map GCS to the existing database or cloud properties and refrain from introducing new semantic conventions.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -202,119 +226,143 @@ const instrumentationMappings = {
...
+    spanName: data => `${data.operationType}`,
     spanAttributes: [
+      { otlp: OTLP.graphql.OPERATION_TYPE, instana: 'operationType' },
       { otlp: OTLP.graphql.OPERATION_NAME, instana: 'operationName' },
-      { otlp: OTLP.graphql.OPERATION_TYPE, instana: 'operationType' }
+      { otlp: OTLP.error.TYPE, instana: 'errors' }
     ]
   },
 
+  // Note: GCS is not specified in RFD
```
aryamohanan: The official OpenTelemetry specification also doesn't include GCS. I think we should add a better comment and map it to the existing database or cloud properties instead of introducing our own properties. This is a known gap in the current specification, and support hasn't been added yet.

aryamohanan:
```
  // Note: Google Cloud Storage (GCS) is not included in the official OpenTelemetry semantic conventions
  // specification. This mapping uses standard cloud.* and database.* attributes to represent GCS operations.
  [INSTRUMENTATION_TYPES.GCS]: {
    spanName: data => `gcs.${data.op}`,
    spanAttributes: [
      { otlp: OTLP.cloud.PROVIDER, value: 'gcp' },
      { otlp: OTLP.database.OPERATION, instana: 'op' },
      { otlp: OTLP.cloud.gcp.PROJECT_ID, instana: 'projectId' },
      { otlp: OTLP.database.NAMESPACE, instana: 'bucket' },
      { otlp: OTLP.database.COLLECTION_NAME, instana: 'object' },
      { otlp: OTLP.error.TYPE, instana: 'error' }
    ]
  },
```

WDYT?

packages/core/src/otlpExporter/common/semconv/base/mappings.js
```diff
@@ -56,29 +55,22 @@ const MAPPINGS = {
...
-    NAME: 'db.name',
     USER: 'db.user',
     COLLECTION: 'db.collection.name',
     TABLE: 'db.sql.table',
-    CONNECTION: 'db.connection'
+    AWS_DYNAMODB_TABLE_NAMES: 'aws.dynamodb.table_names'
   },
 
   rpc: {
-    SYSTEM: 'rpc.system',
```
aryamohanan, line 77: 'rpc.system' should be added in v1.23

aryamohanan: IMO, it should be SYSTEM_NAME and v1.23 value should be 'rpc.system'  and v1.41 its 'rpc.system.name'

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: Yeah we can use existing working attributes

packages/core/src/otlpExporter/common/semconv/base/mappings.js
abhilash-sivan: Yeah we can use same key SYSTEM_NAME and different values in version specific files

packages/core/src/otlpExporter/common/semconv/base/mappings.js
```diff
@@ -109,12 +103,7 @@ const MAPPINGS = {
...
-      S3_KEY: 'aws.s3.key',
-      KINESIS_STREAM: 'aws.kinesis.stream_name',
-      KINESIS_SHARD: 'aws.kinesis.shard_id',
-      KINESIS_SHARD_ITERATOR_TYPE: 'aws.kinesis.shard_iterator_type',
-      KINESIS_STARTING_SEQUENCE_NUMBER: 'aws.kinesis.starting_sequence_number',
-      KINESIS_EXPLICIT_HASH_KEY: 'aws.kinesis.explicit_hash_key'
+      S3_KEY: 'aws.s3.key'
     },
     azure: {
       STORAGE_ACCOUNT: 'az.storage.account.name',
```
aryamohanan: These are not valid OTLP attributes, so I'd prefer not to add them here. For now, we can map azure to the existing database or cloud properties and refrain from introducing new semantic conventions.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan: same as gcp

packages/core/src/otlpExporter/common/semconv/base/mappings.js
abhilash-sivan: Yeah, removed.

packages/core/src/otlpExporter/common/semconv/base/mappings.js
aryamohanan: yes.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: sure

packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
abhilash-sivan: did you mean `DATABASE_MONGODB_COLLECTION: 'db.mongodb.collection',` or ?

packages/core/src/otlpExporter/traces/mappers/constants.js
```diff
@@ -35,7 +35,27 @@ exports.INSTRUMENTATION_TYPES = {
   S3: 's3',
   KINESIS: 'kinesis',
   AZSTORAGE: 'azstorage',
-  AWS_LAMBDA_INVOKE: 'aws.lambda.invoke'
+  AWS_LAMBDA_INVOKE: 'lambda',
+  AWS_LAMBDA_ENTRY: 'lambda'
```
aryamohanan, line 39: issue: Both `AWS_LAMBDA_INVOKE` and `AWS_LAMBDA_ENTRY` were incorrectly mapped to the same value 'lambda', which would cause conflicts since they represent different span types:

```
  AWS_LAMBDA_INVOKE: 'aws.lambda.invoke',
  AWS_LAMBDA_ENTRY: 'aws.lambda.entry'
```

packages/core/src/otlpExporter/common/semconv/base/mappings.js
abhilash-sivan: azure STORAGE_ACCOUNT, right? Yeah, this was also open for dicussion. removed now

packages/core/src/otlpExporter/traces/mappers/constants.js
abhilash-sivan: invoke is `aws.lambda.invoke`

got away from my sight, but I am wondering, should we add 'invoke' separately? We should discuss this

packages/core/src/otlpExporter/traces/mappers/constants.js
```diff
@@ -35,7 +35,27 @@ exports.INSTRUMENTATION_TYPES = {
...
+ */
+exports.LAMBDA_TRIGGER_MAP = {
+  'aws:api.gateway': 'http',
+  'aws:api.gateway.noproxy': 'http',
+  'aws:application.load.balancer': 'http',
+  'aws:lambda.function.url': 'http',
+  'aws:s3': 'datasource',
+  'aws:dynamodb': 'datasource',
+  'aws:kinesis': 'datasource',
+  'aws:kinesis.firehouse': 'datasource',
```
aryamohanan:
```suggestion
```

packages/core/src/otlpExporter/traces/mappers/constants.js
```diff
@@ -35,7 +35,27 @@ exports.INSTRUMENTATION_TYPES = {
...
+  'aws:sqs': 'pubsub',
+  'aws:sns': 'pubsub',
+  'aws:cloudwatch.events': 'timer',
+  'aws:lambda.invoke': 'other'
```
aryamohanan: I miss `aws:cloudwatch.logs`,`aws:lambda.function.url` in the mapping

aryamohanan: It can be tagged as other for now

abhilash-sivan: okay, will add both as other

abhilash-sivan: 'aws:lambda.function.url': 'http',

this is already present, I will add `aws:cloudwatch.logs` as other

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -366,15 +415,63 @@ const instrumentationMappings = {
...
+      { otlp: OTLP.faas.NAME, instana: 'functionName' },
+      { otlp: OTLP.faas.VERSION, instana: 'functionVersion' },
+      { otlp: OTLP.cloud.PROVIDER, value: 'aws' },
+      { otlp: OTLP.cloud.PLATFORM, value: 'aws_lambda' },
+      {
+        otlp: OTLP.cloud.REGION,
+        instana: 'arn',
+        transform: arn => {
+          // arn:aws:lambda:{region}:{account-id}:function:{name}:{version}
+          const parts = typeof arn === 'string' ? arn.split(':') : [];
```
aryamohanan: move to util -> getRegionFrom Arn or similar method

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -366,15 +415,63 @@ const instrumentationMappings = {
...
+          return parts[3] || undefined;
+        }
+      },
+      {
+        otlp: OTLP.cloud.ACCOUNT_ID,
+        instana: 'arn',
+        transform: arn => {
+          const parts = typeof arn === 'string' ? arn.split(':') : [];
```
aryamohanan: same here

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -366,15 +415,63 @@ const instrumentationMappings = {
...
+        }
+      },
+      { otlp: OTLP.cloud.RESOURCE_ID, instana: 'arn' },
+      { otlp: OTLP.faas.INVOCATION_ID, instana: 'reqId' },
+      {
+        otlp: OTLP.faas.COLDSTART,
+        instana: 'coldStart',
+        transform: value => {
+          if (typeof value === 'boolean') return value;
+          if (typeof value === 'string') return value === 'true';
```
aryamohanan: util method toBoolean

packages/core/src/otlpExporter/traces/mappers/constants.js
aryamohanan: I think it should be

```
  AWS_LAMBDA_INVOKE: 'aws.lambda.invoke',
  AWS_LAMBDA_ENTRY: 'lambda'
```

Please double check

aryamohanan: AWS_LAMBDA_ENTRY mapping
`→ spanData = span.data['lambda']`
AWS_LAMBDA_INVOKE mapping
`→ spanData = span.data['aws.lambda.invoke']`

abhilash-sivan: Yeah this is correct:

>   AWS_LAMBDA_INVOKE: 'aws.lambda.invoke',
>   AWS_LAMBDA_ENTRY: 'lambda'

aryamohanan: Okie, please correct it.

packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
aryamohanan: I meant something like the following: 

```
  database: {
    CONNECTION_STRING: 'db.connection_string',
     ...  
    mongodb: {
      COLLECTION: 'db.mongodb.collection'
    }
  },
```

aryamohanan: So we can call      ` { otlp: OTLP.database.mongodb.COLLECTION, instana: 'collection' },`

abhilash-sivan: yeah got it

packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
```diff
@@ -16,23 +16,30 @@ const MAPPINGS = {
 
   messaging: {
     DESTINATION_NAME: 'messaging.destination',
+    OPERATION: 'messaging.operation',
```
kirrg001: bug:

v1.23 uses OPERATION
But we reference messaging.OPERATION_NAME

So we will add undefined keys?

kirrg001: I think there is some more undefined keys. Was that intentional?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -202,179 +229,235 @@ const instrumentationMappings = {
...
     ]
   },
 
   [INSTRUMENTATION_TYPES.S3]: {
-    spanName: data => `s3.${data.op || 'operation'}`,
+    spanName: data => `s3.${data.op}`,
     spanAttributes: [
-      { otlp: OTLP.database.OPERATION, instana: 'op' },
+      { otlp: OTLP.rpc.SYSTEM_NAME, value: 'aws-api' },
+      { otlp: OTLP.rpc.METHOD, instana: 'op', transform: value => (value ? `S3.${value}` : value) },
```
kirrg001: > The name of the operation corresponding to the request, as returned by the AWS SDK

https://opentelemetry.io/docs/specs/semconv/object-stores/s3/

I think this must be only {value} or?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -202,179 +229,235 @@ const instrumentationMappings = {
...
       { otlp: OTLP.error.TYPE, instana: 'error' }
     ]
   },
 
   [INSTRUMENTATION_TYPES.COUCHBASE]: {
-    spanName: data => `couchbase.${data.bucket || 'operation'}`,
+    spanName: data => data.bucket,
     spanAttributes: [
-      { otlp: OTLP.database.SYSTEM, value: 'other_nosql' },
+      { otlp: OTLP.database.SYSTEM, value: 'couchdb' },
```
kirrg001: couchbase or?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -202,179 +229,235 @@ const instrumentationMappings = {
...
+  // https://opentelemetry.io/docs/specs/semconv/graphql/graphql-spans/
   [INSTRUMENTATION_TYPES.GRAPHQL]: {
-    spanName: data =>
-      data.operationName ? `${data.operationType || 'query'} ${data.operationName}` : data.operationType || 'graphql',
+    spanName: data => `${data.operationType}`,
     spanAttributes: [
+      { otlp: OTLP.graphql.OPERATION_TYPE, instana: 'operationType' },
       { otlp: OTLP.graphql.OPERATION_NAME, instana: 'operationName' },
-      { otlp: OTLP.graphql.OPERATION_TYPE, instana: 'operationType' }
+      { otlp: OTLP.error.TYPE, instana: 'errors' }
```
kirrg001, line 345: https://opentelemetry.io/docs/specs/semconv/graphql/graphql-spans/

I can't find error.type?

packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
abhilash-sivan: This is not intentional. we can unify the keys if possible.
This seems possible and missed. I will rename the OPERATION in v1.23 to OPERATION_NAME

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: see: 

db.system.name MUST be set to "couchdb" and SHOULD be provided at span creation time.

https://opentelemetry.io/docs/specs/semconv/db/couchdb/

packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
kirrg001: We have to use the same keys in each version, right?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: There is a generic description for error 

see: https://opentelemetry.io/docs/specs/semconv/general/recording-errors/#recording-errors-on-spans

packages/core/src/otlpExporter/common/semconv/v1.23/mappings.js
abhilash-sivan: yes, then it will override with the version

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -123,50 +146,52 @@ const instrumentationMappings = {
   },
 
   [INSTRUMENTATION_TYPES.SQS]: {
-    spanName: data => `${data.type || data.sort || 'process'} ${data.queue || 'unknown'}`,
+    spanName: data => `${data.sort} ${data.queue}`,
     spanAttributes: [
-      { otlp: OTLP.messaging.SYSTEM, value: OTLP.messaging.sqs?.SYSTEM || 'aws.sqs' },
+      { otlp: OTLP.messaging.SYSTEM, value: 'aws_sqs' },
+      { otlp: OTLP.faas.TRIGGER, value: 'pubsub' },
```
kirrg001: Why is this field here? 🤔 I think this is not right or?

abhilash-sivan: saw this https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs-message

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
kirrg001: I looked here

https://opentelemetry.io/docs/specs/semconv/db/database-spans/

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -202,179 +228,235 @@ const instrumentationMappings = {
...
       { otlp: OTLP.error.TYPE, instana: 'error' }
     ]
   },
 
   [INSTRUMENTATION_TYPES.DYNAMODB]: {
-    spanName: data => `dynamodb.${data.operation || 'request'}`,
+    spanName: data => data.operation,
     spanAttributes: [
-      { otlp: OTLP.database.SYSTEM, value: 'dynamodb' },
+      { otlp: OTLP.database.SYSTEM, value: INSTRUMENTATION_TYPES.DYNAMODB },
```
kirrg001: S3 has rpc fields. Dynamodb not. Why ist that?
https://opentelemetry.io/docs/specs/semconv/db/dynamodb/

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
kirrg001: But this only belongs to the AWS Lambda Entry span and not to the sqs span data or?
This is the faas trigger (who triggered the aws lambda fn)

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: Oh, CouchDB is not same as Couchbase.

Then it is a mistake and it belongs to `other_nosql` as per RFD.

https://github.ibm.com/instana/technical-documentation/pull/389/files#diff-8c000a237a5b02307d5be328873fa852d6b88797e1774a06511664f7e3f2aa05R664

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -202,179 +228,235 @@ const instrumentationMappings = {
...
-      { otlp: OTLP.cloud.aws.KINESIS_EXPLICIT_HASH_KEY, instana: 'record' },
-      { otlp: OTLP.cloud.aws.KINESIS_SHARD_ITERATOR_TYPE, instana: 'shardType' },
-      { otlp: OTLP.cloud.aws.KINESIS_STARTING_SEQUENCE_NUMBER, instana: 'startSequenceNumber' },
-      { otlp: OTLP.cloud.aws.KINESIS_SHARD, instana: 'shard' },
       { otlp: OTLP.error.TYPE, instana: 'error' }
     ]
   },
 
+  // Note: Azure Storage is not included in the official OpenTelemetry semantic conventions
+  // specification. This mapping uses standard cloud.* and database.* attributes to represent azure operations.
```
kirrg001, line 389: Okay that means Otel spans do not contain the blob name information?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: We started a discussion here https://github.com/instana/nodejs/pull/2615/changes#r3491396080 


> But this only belongs to the AWS Lambda Entry span and not to the sqs span data or?
> This is the faas trigger (who triggered the aws lambda fn)

Yeah, I had the same question! So this is only applicable in the AWS Lambda context, not in the exit span, right?

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -202,179 +228,235 @@ const instrumentationMappings = {
...
-      { otlp: OTLP.cloud.gcp.STORAGE_DESTINATION_OBJECT, instana: 'destinationObject' },
+      { otlp: OTLP.database.NAMESPACE, instana: 'bucket' },
+      { otlp: OTLP.database.COLLECTION_NAME, instana: 'object' },
       { otlp: OTLP.error.TYPE, instana: 'error' }
     ]
   },
 
   [INSTRUMENTATION_TYPES.S3]: {
-    spanName: data => `s3.${data.op || 'operation'}`,
+    spanName: data => `s3.${data.op}`,
```
kirrg001: S3 uppercase

https://opentelemetry.io/docs/specs/semconv/cloud-providers/aws-sdk/

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -123,50 +145,52 @@ const instrumentationMappings = {
...
-      { otlp: OTLP.messaging.gcp.PROJECT_ID, instana: 'projid' },
+      { otlp: OTLP.cloud.gcp.PROJECT_ID, instana: 'projid' },
       { otlp: OTLP.messaging.MESSAGE_ID, instana: 'messageId' },
       { otlp: OTLP.error.TYPE, instana: 'error' }
     ]
   },
 
   [INSTRUMENTATION_TYPES.PG]: {
-    spanName: data => data.stmt?.split(/\s+/)[0]?.toUpperCase() || 'POSTGRESQL',
+    spanName: data => data.stmt,
```
kirrg001: Could you please explain this change? 🤔

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -176,10 +200,11 @@ const instrumentationMappings = {
   },
 
   [INSTRUMENTATION_TYPES.MYSQL]: {
-    spanName: data => data.stmt?.split(/\s+/)[0]?.toUpperCase() || 'MYSQL',
+    spanName: data => data.stmt,
     spanAttributes: [
       { otlp: OTLP.database.SYSTEM, value: 'mysql' },
       { otlp: OTLP.database.QUERY_TEXT, instana: 'stmt' },
+      { otlp: OTLP.database.STATEMENT, instana: 'stmt' },
```
kirrg001: This is a normalisation problem I guess.
Both same key, but different values.

QUERY_TEXT, STATEMENT

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -202,179 +228,235 @@ const instrumentationMappings = {
...
       { otlp: OTLP.database.OPERATION, instana: 'op' },
-      { otlp: OTLP.cloud.gcp.STORAGE_BUCKET, instana: 'bucket' },
-      { otlp: OTLP.cloud.gcp.STORAGE_OBJECT, instana: 'object' },
       { otlp: OTLP.cloud.gcp.PROJECT_ID, instana: 'projectId' },
-      { otlp: OTLP.cloud.gcp.STORAGE_SOURCE_BUCKET, instana: 'sourceBucket' },
-      { otlp: OTLP.cloud.gcp.STORAGE_SOURCE_OBJECT, instana: 'sourceObject' },
-      { otlp: OTLP.cloud.gcp.STORAGE_DESTINATION_BUCKET, instana: 'destinationBucket' },
-      { otlp: OTLP.cloud.gcp.STORAGE_DESTINATION_OBJECT, instana: 'destinationObject' },
+      { otlp: OTLP.database.NAMESPACE, instana: 'bucket' },
+      { otlp: OTLP.database.COLLECTION_NAME, instana: 'object' },
```
kirrg001, line 358: Same problem I think with normalization. NAMEPSACE & COLLECTION_NAME

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan: Yes. This resolved.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan: There is a mismatch between the RFD and the OTEL specification. We will likely submit a PR to correct the RFD and updated our mapping accordingly.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan: Yes, this is a known gap. We currently map the values to the available database and cloud attributes. Once the OTEL specification is officially released, we can accommodate the new attributes accordingly.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: removed the trigger attribute

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: Added the rpc.system/rpc.system.name field now

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan: `NAMESPACE` and `COLLECTION_NAME` represent different concepts:

* `db.collection.name` – The name of a collection, table, or container within a database.
* `db.namespace` – The name of the database.

`NAMESPACE` and `NAME` can be normalized, but `NAMESPACE` and `COLLECTION_NAME` should not be treated as equivalent.

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
abhilash-sivan: Initially I did map the spanName with the query name, but then we decided to change it to 'statement' so that the UI will show what Instana has been already showing.

abhilash-sivan: > The span name SHOULD be {db.query.summary} if a summary is available.

https://opentelemetry.io/docs/specs/semconv/db/database-spans/#name

So this query also in line with Otel, AFAIK.

kirrg001 (approved):

aryamohanan (approved):

PR #2534, opened 2026-05-04
chore: added experimental OTLP trace exporter support

abhilash-sivan:
## Summary - Phase 1

Introduces the Instana-to-OTLP exporter.

* Adds support for exporting traces via OTLP (`4318/v1/traces`) with agent.
* Supports OpenTelemetry Semantic Conventions (SemConv) v1.23 to v1.41.1.
* Protocol: http/json
* Infrastructure correlation

## Enabling OTLP Export

OTLP export is disabled by default and can be enabled in one of the following ways.

1. **Environment Variable**
`INSTANA_TRACING_OTLP_ENABLED=true`

2. **Collector Configuration**
```
instanaCollector({
  tracing: {
    otlp: {
      enabled: true
    }
  }
});
```
3. **Agent Configuration**
```
com.instana:
  tracing:
    otlp:
      enabled: true
```

### Out of Scope

* Metrics export
* Logs export
* Other protocols like http/protobuf or grpc
* Extended configuration options
* Serverless OTLP endpoint
* Events (exception)

### Core Implementation

* [x] Handle agent port usage (data port vs. connection port) (#2559)
* [x] Support multiple span data keys (e.g., `span.data.peer`, `span.data.mongo`)
* [x] Finalize overall design
* [x] Implement SemConv lookup handling
* [x] Implement span name mapping
* [x] Complete metadata attribute mapping:
* [x] Direct mappings
* [x] Computed mappings
* [x] Research span status mapping
* [x] Research span name mapping (@abhilash-sivan)
* [x] Implement OpenTelemetry instrumentation-specific handling
* [x] Implement log-specific handling
* [x] Implement resource mapping
* [x] Implement required metrics mappings
* [x] Implement agent connection logic
* [x] Basic Error handling @abhilash-sivan 
* [x] Resource-service name mapping
* [x] Dual transformation flow in span data  @abhilash-sivan 
* [x] Implement configuration option for enabling OTLP export 
* [x] agent data port issue
* [x] normalize HTTP client span attributes #2617
* [x] span status + http error 400>
* [x] general cleanup coding
* [x] Fix type definitions
* [x] Pipeline green
* [x] remove example apps

### Span Attribute Mapping @abhilash-sivan 

* [x] HTTP
* [x] Kafka
* [x] aws sqs
* [x] nats 
* [x] nats streaming
* [x] bull
* [x] ampq (rabbitmq)
* [x] SNS
* [x] GCPS (Google Cloud Pub/Sub)
[...]

PR #2620, opened 2026-06-26
refactor: optimized span transformation in spanBuffer

aryamohanan:
- Move span transformation from addSpan to transmission points
- Transform spans in batch during transmitSpans and getAndResetSpans

This is a prior PR for OTLP exporter #2534

### DIFF
--- a/packages/core/src/tracing/spanBuffer.js
+++ b/packages/core/src/tracing/spanBuffer.js
@@ -6,7 +6,7 @@
 'use strict';
 
 const tracingMetrics = require('./metrics');
-const { transform } = require('./backend_mappers');
+const instanaBackendMapper = require('./backend_mappers');
 
 /** @type {import('../core').GenericLogger} */
 let logger;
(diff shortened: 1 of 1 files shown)


packages/core/src/tracing/spanBuffer.js
```diff
@@ -185,9 +185,6 @@ exports.addSpan = function (span) {
     return;
   }
 
-  // Transform internal span data format into external (backend) readable format.
-  span = applySpanTransformation(span);
```
aryamohanan, line 189: Move span transformation from addSpan to transmitSpans/getAndResetSpans to avoid dual transformation. 

OTLP Exporter performs batch transformation.

packages/core/src/tracing/spanBuffer.js
```diff
@@ -477,7 +475,7 @@ exports.getAndResetSpans = function getAndResetSpans() {
   const spansToSend = spans;
   spans = [];
   batchingBuckets.clear();
-  return spansToSend;
+  return applySpanTransformation(spansToSend);
```
aryamohanan, line 478: When the transformation was moved from addSpan() to transmitSpans(), the normal Collector → Agent flow continued to work, but the serverless/AWS Lambda path use getAndResetSpans() over transmitSpans().As a result, it now receives internal spans instead of transformed backend spans.

Added the transformation here too. Only lambda, calling this function to get the span

kirrg001 (approved):

abhilash-sivan (approved):

PR #2621, opened 2026-06-26
refactor: renamed kafka-node span attributes to use internal format

abhilash-sivan:
This refactor is needed for the OTLP mapper so that the mapper can rely on internal format

We already have the backend mapper in place, so the end-user functionality is not affected


aryamohanan (approved):
LGTM

PR #2638, opened 2026-07-08
chore: added baggage example for agent otel app

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-79730

Proof that baggage information appear in UI under tags ✅

baggageotelattrs
baggageoteltags


aryamohanan (approved):
wow

PR #2642, opened 2026-07-08
feat: added support for graphql v17

instanacd:
refs https://jsw.ibm.com/browse/INSTA-97806

release note: https://graphql.org/blog/2026-06-15-introducing-graphql-js-v17/


abhilash-sivan:
https://graphql.org/blog/2026-06-15-introducing-graphql-js-v17/

abhilash-sivan:
https://www.graphql-js.org/upgrade-guides/v16-v17/

kirrg001:
https://www.graphql-js.org/upgrade-guides/v16-v17/
Did you check the new features? It seems there is different public API notations etc. We might need to add some.

We can temporarily skip apollo server - the PR to support v17 is already open though.

abhilash-sivan:
> Did you check the new features? It seems there is different public API notations etc. We might need to add some.

Added support for the harness API, which is the new default for execution commands. We can add more in the future if there are no critical items. wdyt ?

kirrg001:
Could you please run ESM tests on this branch?

> Added support for the harness API, which is the new default for execution commands. We can add more in the future if there are no critical items. wdyt ?

Could you please list anything which affects our instrumentation or is not covered? Ty

packages/core/src/tracing/instrumentation/protocols/graphql.js
```diff
@@ -24,7 +24,20 @@ const subscriptionUpdate = 'subscription-update';
 exports.init = function init(config) {
   logger = config.logger;
 
+  // For graphql < v17: hook into the execution file as it is loaded (CJS).
   hook.onFileLoad(/\/graphql\/execution\/execute.js/, instrumentExecute);
+
+  // For graphql >= v17: ships as module-sync ESM (index.mjs) on Node 22+.
+  // require('graphql') resolves to index.mjs
+  // graphqlExports.defaultHarness is the live harness.mjs object that graphql() calls directly
+  hook.onModuleLoad('graphql', graphqlExports => {
```
abhilash-sivan: https://github.com/graphql/graphql-js/blob/17.x.x/src/index.ts#L48

kirrg001:
> We can add more in the future if there are no critical items. wdyt ?

We can add any features later, yes. We just want to create a follow-up ticket in case there is new functions or features we have to cover.

abhilash-sivan:
> Could you please run ESM tests on this branch?
> 
> > Added support for the harness API, which is the new default for execution commands. We can add more in the future if there are no critical items. wdyt ?
> 
> Could you please list anything which affects our instrumentation or is not covered? Ty

Previously also we were only shimming the execute method only.
Right now they splitted it to:

> Last but certainly not least, v17 includes long-awaited experimental support for incremental delivery with @defer and @stream. GraphQL.js v17 keeps ordinary execute() as a single-result executor and exposes experimentalExecuteIncrementally() for the current incremental response shape. 

so we can consider adding `experimentalExecuteIncrementally` if this is not already coverd with current instrumentation changes

packages/core/src/tracing/instrumentation/protocols/graphql.js
```diff
@@ -24,7 +24,20 @@ const subscriptionUpdate = 'subscription-update';
 exports.init = function init(config) {
   logger = config.logger;
 
+  // For graphql < v17: hook into the execution file as it is loaded (CJS).
   hook.onFileLoad(/\/graphql\/execution\/execute.js/, instrumentExecute);
+
+  // For graphql >= v17: ships as module-sync ESM (index.mjs) on Node 22+.
```
kirrg001: Is that comment right? 🤔

packages/collector/test/integration/currencies/protocols/graphql/rawGraphQLServer.mjs
```diff
@@ -0,0 +1,103 @@
...
+const require = createRequire(import.meta.url);
+
+import bodyParser from 'body-parser';
+import express from 'express';
+import morgan from 'morgan';
+import amqp from 'amqplib';
+
+const schemaFactory = require('./schema.js');
+const { schema } = schemaFactory();
+const graphQL = require('graphql');
```
kirrg001: Why is this a require? Huh?

abhilash-sivan: > SyntaxError: The requested module 'graphql' does not provide an export named 'default'

Got an issue and updated it with createRequire

> 
> import { createRequire } from 'module';
> const require = createRequire(import.meta.url);

abhilash-sivan: I think I can also use named import using a different structure, I will check and update

packages/collector/test/integration/currencies/protocols/graphql/package.json.template.v16
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "@apollo/server": "latest",
```
kirrg001, line 3: Lets also add that to the follow up graphql JIRA 👍 
We can re-test when the release for v17 support is out

packages/collector/test/integration/currencies/protocols/graphql/test_base.js
```diff
@@ -322,7 +328,7 @@ module.exports = function (name, version, isLatest, mode) {
       before(async () => {
         serverControls = new ProcessControls({
           dirname: __dirname,
-          appName: 'apolloServer',
+          appName: mode === 'raw' ? 'rawGraphQLServer' : 'apolloServer',
```
kirrg001, line 331: Was that a bug? We changed the logic here

abhilash-sivan: This isn't a bug.

I removed the Apollo Server test from the GraphQL v17 setup. If that file is loaded, it causes issues because Apollo Server currently depends on GraphQL v16. As a result, both GraphQL v16 and v17 end up being loaded, which triggers the following error:

> Duplicate "graphql" modules cannot be used at the same time since different versions may have different capabilities and behavior.

With the Apollo Server test removed, the setup is now clean and the correct mode loads the correct files.

packages/core/src/tracing/instrumentation/protocols/graphql.js
```diff
@@ -24,11 +24,27 @@ const subscriptionUpdate = 'subscription-update';
 exports.init = function init(config) {
   logger = config.logger;
 
+  // For graphql < v17: hook into the execution file as it is loaded (CJS).
   hook.onFileLoad(/\/graphql\/execution\/execute.js/, instrumentExecute);
+
+  // For graphql >= v17: it is shipped as ESM first
```
kirrg001:
```suggestion
  // For graphql >= v17: They shipped ESM support including a larger refactoring loading the execute file from an internal harness module. CJS apps load the index.mjs entrypoint because of the `module-sync` feature which was added in Node v22 (https://nodejs.org/api/packages.html#conditional-exports). Thats why CJS apps with GraphQL v17 do not trigger `onFileLoad` for `execution/execute`.
```

packages/core/src/tracing/instrumentation/protocols/graphql.js
```diff
@@ -24,11 +24,27 @@ const subscriptionUpdate = 'subscription-update';
 exports.init = function init(config) {
   logger = config.logger;
 
+  // For graphql < v17: hook into the execution file as it is loaded (CJS).
```
kirrg001:
```suggestion
  // For graphql < v17: require('graphql') triggers the internal file load 'execution/execute'
```

packages/core/src/tracing/instrumentation/protocols/graphql.js
```diff
@@ -24,11 +24,27 @@ const subscriptionUpdate = 'subscription-update';
...
+
+  // For graphql >= v17: it is shipped as ESM first
+  hook.onModuleLoad('graphql', instrumentGraphQL, { nativeEsm: true });
+
   hook.onFileLoad(/\/@apollo\/gateway\/dist\/executeQueryPlan.js/, instrumentApolloGatewayExecuteQueryPlan);
   hook.onModuleLoad('@apollo/federation', logDeprecatedWarning);
 };
 
+function instrumentGraphQL(graphqlExports) {
+  // require('graphql') resolves to index.mjs via CJS; nativeEsm covers the case where
```
kirrg001: IMO this does not explain why. I'd remove this comment - see suggestion

abhilash-sivan: Okay, sure. I'll apply the changes once the current pipeline passes, before merging.

kirrg001 (approved):
Pre-approving 👍

packages/collector/test/integration/currencies/protocols/graphql/test_base.js
aryamohanan: I can't follow this change, is this cause a change in v16 test flow?

kirrg001: Yeah me neither :D

abhilash-sivan: Yeah, it will change the app itself. I would say previously it was wrong :D 

It did not create any issues earlier because this was `suppressed` test, so it doesn't matter which app it actually started

aryamohanan (approved):
Great work 🎉

PR #2648, opened 2026-07-13
test: renamed internal agent stub environment variables

aryamohanan:
The environment variable names were confusing and did not clearly indicate that they are specific to the agent configuration used by the internal agent stub in the test setup.

Renamed all agent stub environment variables to make their scope clearer.


abhilash-sivan (approved):

kirrg001 (approved):
I love u

PR #2649, opened 2026-07-13
fix(winston): avoid wrapping error logger method twice

aryamohanan:
Removed the duplicate `error` shim registration from the Winston instrumentation to avoid wrapping the same method twice.


packages/core/src/tracing/instrumentation/logging/winston.js
```diff
@@ -50,7 +50,6 @@ function instrumentWinston3(createLogger) {
     shimLevelMethod(derivedLogger, 'emerg', true);
     shimLevelMethod(derivedLogger, 'alert', true);
     shimLevelMethod(derivedLogger, 'crit', true);
-    shimLevelMethod(derivedLogger, 'error', true);
```
aryamohanan, line 53: error is already wrapped at line 46. The second registration wraps the same method again, so it can be removed.

kirrg001 (approved):

abhilash-sivan (approved):

PR #2653, opened 2026-07-14
fix: preserved trace headers in aws4-axios package triggered calls

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-100764

headers in the aws4-axios package triggered http calls:

>   'user-agent': 'axios/1.18.1',
>   'x-amz-date': '20260714T084817Z',

### DIFF
--- a/packages/collector/test/integration/currencies/protocols/http/client/test_base.js
+++ b/packages/collector/test/integration/currencies/protocols/http/client/test_base.js
@@ -1418,50 +1418,55 @@ module.exports = function (name, version, isLatest) {
           )
         ));
 
-    if (!appUsesHttps) {
-      // Verifies that Instana trace headers are propagated through requests signed by aws4-axios.
-      // aws4-axios sets `Authorization: AWS4-HMAC-SHA256` + `User-Agent: axios/1.18.1`.
-      it('must propagate X-INSTANA and traceparent headers through an aws4-axios signed request', () =>
-        clientControls
-          .sendRequest({
-            method: 'GET',
-            path: '/aws4-axios-signed-request'
-          })
-          .then(body => {
-            expect(body['x-instana-t'], 'X-INSTANA-T must reach the downstream service').to.be.a('string');
-            expect(body['x-instana-s'], 'X-INSTANA-S must reach the downstream service').to.be.a('string');
-            expect(body.traceparent, 'traceparent must reach the downstream service').to.be.a('string');
+    it('must propagate X-INSTANA and traceparent headers through an external package signed request', () =>
+      clientControls
+        .sendRequest({
+          method: 'GET',
+          path: '/aws4-axios-signed-request'
+        })
+        .then(body => {
+          expect(body['x-instana-t'], 'X-INSTANA-T must reach the downstream service').to.be.a('string');
+          expect(body['x-instana-s'], 'X-INSTANA-S must reach the downstream service').to.be.a('string');
+          expect(body.traceparent, 'traceparent must reach the downstream service').to.be.a('string');
 
-            return retry(() =>
-              globalAgent.instance.getSpans().then(spans => {
-                const exitSpan = expectExactlyOneMatching(spans, [
-                  span => expect(span.n).to.equal('node.http.client'),
-                  span => expect(span.k).to.equal(constants.EXIT),
-                  span => expect(span.data.http.url).to.match(/\/aws4-axios-target/)
-                ]);
+          return retry(() =>
+            globalAgent.instance.getSpans().then(spans => {
+              const exitSpan = expectExactlyOneMatching(spans, [
+                span => expect(span.n).to.equal('node.http.client'),
+                span => expect(span.k).to.equal(constants.EXIT),
+                span => expect(span.data.http.url).to.match(/\/aws4-axios-target/)
+              ]);
 
-                expect(body['x-instana-t']).to.equal(exitSpan.t);
-                expect(body['x-instana-s']).to.equal(exitSpan.s);
-                expect(body.traceparent).to.include(exitSpan.t);
-              })
-            );
-          }));
+              expect(body['x-instana-t']).to.equal(exitSpan.t);
+              expect(body['x-instana-s']).to.equal(exitSpan.s);
+              expect(body.traceparent).to.include(exitSpan.t);
 
-      it('must NOT create http client EXIT span for an aws-sdk signed request', () =>
-        clientControls
-          .sendRequest({
-            method: 'GET',
-            path: '/aws-sdk-signed-request'
-          })
-          .then(() => {
-            return delay(500).then(() =>
-              globalAgent.instance.getSpans().then(spans => {
-                const exitSpans = spans.filter(span => span.n === 'node.http.client' && span.k === constants.EXIT);
-                expect(exitSpans, 'must not create a node.http.client exit span').to.have.length(0);
-              })
-            );
-          }));
-    }
+              // 1 x http server
+              // 1 x http client
+              // 1 x http server
+              expect(spans).to.have.lengthOf(3);
+            })
+          );
+        }));
+
+    it('must NOT create http client EXIT span for an aws-sdk signed request', () =>
+      clientControls
+        .sendRequest({
+          method: 'GET',
+          path: '/aws-sdk-signed-request'
+        })
+        .then(() => {
+          return delay(500).then(() =>
+            globalAgent.instance.getSpans().then(spans => {
+              const exitSpans = spans.filter(span => span.n === 'node.http.client' && span.k === constants.EXIT);
+              expect(exitSpans, 'must not create a node.http.client exit span').to.have.length(0);
+
+              // 1 x http server
+              // 1 x http server
+              expect(spans).to.have.lengthOf(2);
+            })
+          );
+        }));
   }
 
   function registerConnectionRefusalTest(appUsesHttps) {
--- a/packages/collector/test/integration/currencies/protocols/http/client/test_base.js
+++ b/packages/collector/test/integration/currencies/protocols/http/client/test_base.js
@@ -1428,8 +1428,6 @@ module.exports = function (name, version, isLatest) {
             path: '/aws4-axios-signed-request'
           })
           .then(body => {
-            // The downstream server echoes back whatever trace headers it received.
-            // All three must be present — none should have been swallowed by the AWS-signed-request guard.
             expect(body['x-instana-t'], 'X-INSTANA-T must reach the downstream service').to.be.a('string');
             expect(body['x-instana-s'], 'X-INSTANA-S must reach the downstream service').to.be.a('string');
             expect(body.traceparent, 'traceparent must reach the downstream service').to.be.a('string');
@@ -1442,52 +1440,24 @@ module.exports = function (name, version, isLatest) {
                   span => expect(span.data.http.url).to.match(/\/aws4-axios-target/)
                 ]);
 
-                // The trace ID the server received must be the same one the exit span carries,
-                // proving end-to-end trace context continuity.
                 expect(body['x-instana-t']).to.equal(exitSpan.t);
-
-                // The span ID the server received as X-INSTANA-S must equal the exit span's own span ID.
                 expect(body['x-instana-s']).to.equal(exitSpan.s);
-
-                // traceparent must encode the same trace ID (last 16 hex chars of the 32-char trace ID field).
                 expect(body.traceparent).to.include(exitSpan.t);
               })
             );
           }));
 
-      // Verifies that Instana trace headers are NOT injected into requests from the aws-sdk.
-      // aws-sdk sets `Authorization: AWS4-HMAC-SHA256` + `User-Agent: aws-sdk-js/...`.
-      it('must NOT inject X-INSTANA or traceparent headers into an aws-sdk signed request', () =>
+      it('must NOT create http client EXIT span for an aws-sdk signed request', () =>
         clientControls
           .sendRequest({
             method: 'GET',
             path: '/aws-sdk-signed-request'
           })
-          .then(body => {
-            return retry(() =>
+          .then(() => {
+            return delay(500).then(() =>
               globalAgent.instance.getSpans().then(spans => {
-                // A node.http.client exit span is still created — the guard only suppresses header
-                // injection, not span creation.
-                const exitSpan = expectExactlyOneMatching(spans, [
-                  span => expect(span.n).to.equal('node.http.client'),
-                  span => expect(span.k).to.equal(constants.EXIT),
-                  span => expect(span.data.http.url).to.match(/\/aws4-axios-target/)
-                ]);
-
-                // The downstream service must NOT have received the exit span's trace ID as X-INSTANA-T.
-                // body['x-instana-t'] is either null (not sent) or a parent span's value — either way
-                // it must differ from the exit span's own trace ID.
-                expect(body['x-instana-t']).to.not.equal(exitSpan.t);
-
-                // Same for X-INSTANA-S: the exit span's span ID must not have been forwarded.
-                expect(body['x-instana-s']).to.not.equal(exitSpan.s);
-
-                // traceparent must not carry the exit span's trace ID.
-                // When null (not sent at all), that already satisfies the requirement.
-                // When present (e.g. a parent span's traceparent), it must not reference this exit span.
-                if (body.traceparent !== null) {
-                  expect(body.traceparent).to.not.include(exitSpan.t);
-                }
+                const exitSpans = spans.filter(span => span.n === 'node.http.client' && span.k === constants.EXIT);
+                expect(exitSpans, 'must not create a node.http.client exit span').to.have.length(0);
               })
             );
           }));
--- a/packages/collector/test/integration/currencies/protocols/http/client/test_base.js
+++ b/packages/collector/test/integration/currencies/protocols/http/client/test_base.js
@@ -1417,6 +1417,81 @@ module.exports = function (name, version, isLatest) {
             })
           )
         ));
+
+    if (!appUsesHttps) {
+      // Verifies that Instana trace headers are propagated through requests signed by aws4-axios.
+      // aws4-axios sets `Authorization: AWS4-HMAC-SHA256` + `User-Agent: axios/1.18.1`.
+      it('must propagate X-INSTANA and traceparent headers through an aws4-axios signed request', () =>
+        clientControls
+          .sendRequest({
+            method: 'GET',
+            path: '/aws4-axios-signed-request'
+          })
+          .then(body => {
+            // The downstream server echoes back whatever trace headers it received.
+            // All three must be present — none should have been swallowed by the AWS-signed-request guard.
+            expect(body['x-instana-t'], 'X-INSTANA-T must reach the downstream service').to.be.a('string');
+            expect(body['x-instana-s'], 'X-INSTANA-S must reach the downstream service').to.be.a('string');
+            expect(body.traceparent, 'traceparent must reach the downstream service').to.be.a('string');
+
+            return retry(() =>
+              globalAgent.instance.getSpans().then(spans => {
+                const exitSpan = expectExactlyOneMatching(spans, [
+                  span => expect(span.n).to.equal('node.http.client'),
+                  span => expect(span.k).to.equal(constants.EXIT),
+                  span => expect(span.data.http.url).to.match(/\/aws4-axios-target/)
+                ]);
+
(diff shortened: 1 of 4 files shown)


packages/core/src/tracing/instrumentation/protocols/httpClient.js
kirrg001: I was not able to reproduce this ^ with the AWS SDK's v2/v3.
**Only with a script and only for v2.** It seems this issue is an edge case - maybe it got fixed by AWS.

We should make that clear here in the comments.
I can push the scripts tmrw and we can mention that here in the comment.

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -495,16 +495,46 @@ function isItSafeToModifiyHeadersInOptions(options) {
       // instrumented Lambda function behind an API gateway and the user is using
       // https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/APIGateway.html to invoke this Gateway/Lambda
       // combination, which _would_ benefit from tracing headers.)
-      return false;
+
+      // Exception: when the request comes from axios (e.g. via aws4-axios), the signing library re-signs
+      // the request fresh on every call, so adding trace headers is safe. The receiving end is also likely
+      // an instrumented service (e.g. a Lambda behind API Gateway) that benefits from receiving trace headers.
+      if (!isAxiosRequest(options.headers)) {
```
kirrg001: The fix is not right IMO.

See https://github.com/instana/nodejs/commit/cc3a99b738c9058e2abf713cb33622f4cb23d1d2

We only should not add the headers if the request is coming from the SDK!

>if (
      'authorization' === key.toLowerCase() &&
      typeof options.headers[key] === 'string' &&
      options.headers[key].indexOf('AWS') === 0
    ) {

This ^ is not enough.

Please remove `isItSafeToModifiyHeadersInOptions`.
And add:

```js
    const isAWSSDKRequest = isAWSSDKRequest(...)

    // CASE 1: if parent is exit, we skip anyway (e.g. http request from aws sdk)
    // CASE 2: Some special cases like sqs entry etc.
    if (skipTracingResult.skip || shouldBeBypassed(skipTracingResult.parentSpan, options, isAWSSDKRequest)) {

       if (isAWSSDKRequest) {
          return originalRequest.apply(coreModule, arguments);
       }
   
```

kirrg001 (changes requested):
See comments

packages/core/src/tracing/instrumentation/protocols/httpClient.js
kirrg001: See https://github.com/instana/nodejs/commit/2481c1ee4c325b8d18acccbc663865e7e505fc64

aws sdk v2 fails
aws sdk v3 works

packages/core/src/tracing/instrumentation/protocols/httpClient.js
aryamohanan: I agree with detecting AWS SDK requests rather than looking at the `Authorization` header. The issue is specific to AWS SDK requests, not to signed requests in general, so `Authorization` isn't a reliable discriminator. Libraries like `aws4-axios` also generate SigV4 signatures but don't exhibit the same retry behaviour.

~~One question though: based on the reproduction scripts, this issue appears to be specific to **AWS SDK v2**. In v3, each retry uses a fresh request/options object, so the injected `X-Instana-*` headers don't leak into `SignedHeaders`~~

packages/core/src/tracing/instrumentation/protocols/httpClient.js
aryamohanan: Those scripts are 🔥

packages/core/src/tracing/instrumentation/protocols/httpClient.js
abhilash-sivan: I think yes. Seems like it was introduced to cover the retry/signing issue. Maybe we only need to address for v2

packages/core/src/tracing/instrumentation/protocols/httpClient.js
abhilash-sivan: ~The script suggests that we have issue in `main` also for v2~

packages/core/src/tracing/instrumentation/protocols/httpClient.js
aryamohanan: I was wrong, so it was intentionally skipped header propagation for the full aws sdk , see reason https://github.com/instana/nodejs/blob/cc3a99b738c9058e2abf713cb33622f4cb23d1d2/packages/core/src/tracing/instrumentation/protocols/httpClient.js#L320

abhilash-sivan: Makes sense. thanks

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -98,14 +98,34 @@ function evaluateHeaderValue(headerValue, validator) {
...
   const userAgent = (headers && headers['User-Agent']) || (headers && headers['user-agent']);
+  return isAWSNodeJSHeader(userAgent);
+}
 
-  const isAWSNodeJSHeader = evaluateHeaderValue(
+/**
+ * @param {string | string[] | undefined} userAgent
+ * @returns {boolean}
+ */
+function isAWSNodeJSHeader(userAgent) {
```
kirrg001: `isAWSNodeJSHeader`? 🤔 Do you mean `isAWSSDKHeader`?

Could you please explain the need for both `isAWSNodeJSHeader` and `isAwsSdkRequest`?

abhilash-sivan: isAWSNodeJSHeader is the old internal name from before the refactor!

There is no need to keep both; it can be combined. Also I can update the name to isAWSSDKHeader to be more generic

abhilash-sivan: renamed to `isAWSdkHeader`

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -98,14 +98,26 @@ function evaluateHeaderValue(headerValue, validator) {
...
     userAgent,
     header => header.toLowerCase().indexOf('aws-sdk-nodejs') > -1 || header.toLowerCase().indexOf('aws-sdk-js') > -1
   );
+}
+
+function shouldBeBypassed(parentSpan, options) {
+  const headers = options && options.headers;
+  const userAgent = (headers && headers['User-Agent']) || (headers && headers['user-agent']);
+
+  const awsSdkHeader = isAWSdkHeader(options);
```
aryamohanan: Can we pass the `awsSdkRequest` value to this function? We can avoid dual computation.

`function shouldBeBypassed(parentSpan, options, awsSdkRequest) {`

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -388,7 +407,19 @@ function tryToAddHeadersToOpts(options, span, w3cTraceContext) {
...
+    // "SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided.
+    // Check your key and signing method."
+    // See https://docs.aws.amazon.com/general/latest/gr/signing_aws_api_requests.html
+    //
+    // Additionally, adding our headers to this request would not have any benefit — the receiving end will
+    // be an AWS service like S3 and those are not instrumented. (There is a very small chance that the
+    // receiving end is an instrumented Lambda function behind an API gateway and the user is using
+    // https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/APIGateway.html to invoke this
+    // Gateway/Lambda combination, which _would_ benefit from tracing headers.)
+    if (awsSdkRequest) {
```
kirrg001: Why?

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -388,7 +407,19 @@ function tryToAddHeadersToOpts(options, span, w3cTraceContext) {
   // (see setHeadersOnRequest).
 
   if (hasHeadersOption(options)) {
-    if (!isItSafeToModifiyHeadersInOptions(options)) {
+    // This is a signed AWS API request (from the aws-sdk package, identified by its User-Agent).
```
kirrg001: We discussed that we will update the comment.

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -443,8 +469,11 @@ function removeInstanaHeadersFromOpts(options) {
   delete options.headers[constants.w3cTraceState];
 }
 
-function setHeadersOnRequest(clientRequest, span, w3cTraceContext) {
-  if (!isItSafeToModifiyHeadersForRequest(clientRequest)) {
+function setHeadersOnRequest(clientRequest, span, w3cTraceContext, awsSdkRequest) {
+  // This is a signed AWS API request (from the aws-sdk package, identified by its User-Agent).
+  // Adding our headers to this request would trigger a SignatureDoesNotMatch error in case the request
+  // will be retried. See the comment in tryToAddHeadersToOpts for the full explanation.
+  if (awsSdkRequest) {
```
kirrg001: Why?

packages/core/src/tracing/instrumentation/protocols/httpClient.js
abhilash-sivan: done

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -178,7 +188,16 @@ function instrument(coreModule, forceHttps) {
 
     const parentSpan = skipTracingResult.parentSpan;
 
-    if (skipTracingResult.skip || shouldBeBypassed(skipTracingResult.parentSpan, options)) {
+    const awsSdkRequest = isAWSdkHeader(options);
+
+    // aws-sdk requests that are being skipped/bypassed must be returned completely untouched —
+    // no suppression or W3C headers either, as nothing must disturb the signed request because
+    // adding headers would cause SignatureDoesNotMatch on retries.
```
kirrg001: This comment is confusing

packages/core/src/tracing/instrumentation/protocols/httpClient.js
abhilash-sivan: Instead of checking the auth headers, we decided to check if the request is coming from the AWS SDK, right? 
But I think I get the logical issue here; we still want to consider other requests without authHeader. Now we only skip for awsSdk.

> 
> const authHeader = clientRequest.getHeader('Authorization');
>   return !authHeader

packages/core/src/tracing/instrumentation/protocols/httpClient.js
abhilash-sivan: I thought if the request is coming from the AWS SDK, we do not want to modify the headers. that was the intention behind removal of `isItSafeToModifiyHeadersInOptions`

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -98,14 +98,24 @@ function evaluateHeaderValue(headerValue, validator) {
  * https://nodejs.org/dist/latest-v14.x/docs/api/http.html#http_http_request_options_callback
  * @return {boolean} true, if the HTTP request that is about to happen should _not_ create a span.
  */
-function shouldBeBypassed(parentSpan, options) {
+/**
+ * Returns true if the request options originate from the aws-sdk (v2 or v3), identified by the
+ * User-Agent header. Used to skip Instana header injection for SigV4-signed requests.
```
kirrg001: > Used to skip Instana header injection for SigV4-signed requests.

Why do we mention that everywhere? That is not right.

packages/core/src/tracing/instrumentation/protocols/httpClient.js
kirrg001: But this code is not reached or?
Could you please explain the use case? Is there a test?

packages/collector/test/integration/currencies/protocols/http/client/test_base.js
```diff
@@ -1417,6 +1417,81 @@ module.exports = function (name, version, isLatest) {
...
+
+            return retry(() =>
+              globalAgent.instance.getSpans().then(spans => {
+                const exitSpan = expectExactlyOneMatching(spans, [
+                  span => expect(span.n).to.equal('node.http.client'),
+                  span => expect(span.k).to.equal(constants.EXIT),
+                  span => expect(span.data.http.url).to.match(/\/aws4-axios-target/)
+                ]);
+
+                // The trace ID the server received must be the same one the exit span carries,
```
kirrg001: Please lets not add these useless AI comments.

packages/core/src/tracing/instrumentation/protocols/httpClient.js
abhilash-sivan: No test, let me try to add. Maybe there is no issue.

packages/core/src/tracing/instrumentation/protocols/httpClient.js
kirrg001: Every use case needs to be understood & covered & explained right

abhilash-sivan: if (!isItSafeToModifiyHeadersInOptions(options)) {

this check is for only identifying the aws sdk requests, it previously checked whether 

> options.headers[key].indexOf('AWS') === 0

So replacing it with  `if (awsSdkRequest)` seems reasonable. 

Sorry, am I missing any other use cases?

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -178,7 +186,13 @@ function instrument(coreModule, forceHttps) {
 
     const parentSpan = skipTracingResult.parentSpan;
 
-    if (skipTracingResult.skip || shouldBeBypassed(skipTracingResult.parentSpan, options)) {
+    const awsSdkRequest = isAWSdkHeader(options);
```
aryamohanan:
```suggestion
    const awsSdkRequest = isAwsSdkHeader(options);
```
Typo in fn name

aryamohanan: isAwsSdkRequest is more correct

abhilash-sivan: Yeah, I prefer `isAwsSdkRequest`. Added

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -194,6 +204,11 @@ function instrument(coreModule, forceHttps) {
       return clientRequest;
     }
 
+    // "Note: There is no need to create http EXIT spans for AWS SDK created events"
```
abhilash-sivan:
```suggestion
    // Note: There is no need to create http EXIT spans for AWS SDK created events
```

packages/collector/test/integration/currencies/protocols/http/client/test_base.js
```diff
@@ -1417,6 +1417,51 @@ module.exports = function (name, version, isLatest) {
             })
           )
         ));
+
+    if (!appUsesHttps) {
+      // Verifies that Instana trace headers are propagated through requests signed by aws4-axios.
+      // aws4-axios sets `Authorization: AWS4-HMAC-SHA256` + `User-Agent: axios/1.18.1`.
+      it('must propagate X-INSTANA and traceparent headers through an aws4-axios signed request', () =>
```
kirrg001: I think we can make this test general or?
This is no longer related to aws4-axios. It was related to all http exists with AWS headers, right?

packages/collector/test/integration/currencies/protocols/http/client/test_base.js
```diff
@@ -1417,6 +1417,51 @@ module.exports = function (name, version, isLatest) {
...
+            method: 'GET',
+            path: '/aws4-axios-signed-request'
+          })
+          .then(body => {
+            expect(body['x-instana-t'], 'X-INSTANA-T must reach the downstream service').to.be.a('string');
+            expect(body['x-instana-s'], 'X-INSTANA-S must reach the downstream service').to.be.a('string');
+            expect(body.traceparent, 'traceparent must reach the downstream service').to.be.a('string');
+
+            return retry(() =>
+              globalAgent.instance.getSpans().then(spans => {
```
kirrg001: span length check missing.

packages/collector/test/integration/currencies/protocols/http/client/test_base.js
abhilash-sivan: Yeah, it is related to all HTTP exits with AWS headers. Can make it general

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -194,6 +204,11 @@ function instrument(coreModule, forceHttps) {
       return clientRequest;
     }
 
+    // Note: There is no need to create http EXIT spans for AWS SDK created events
+    if (awsSdkRequest) {
+      return originalRequest.apply(coreModule, arguments);
```
kirrg001: Why here and not earlier? 🤔

abhilash-sivan: It can be kept above shouldBypass function. Then there is a cleanup required in the shouldBypass function.

There is an SQS polling case handled in this function, which will not be required after the current instrumentation restriction. But I think it can be done separately, as we need this fix to go ASAP.

Removal of SQS polling logic might need some manual verifications

abhilash-sivan: + I wanted to see if all the tests will pass in the current logic!!

abhilash-sivan: Since you also pointed it out, I will remove this SQS case and move the instrumentation skipping logic to top.
If the test passes, we can agree it works

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -102,23 +114,6 @@ function shouldBeBypassed(parentSpan, options) {
...
-
-  const hostInfo = (headers && headers.Host) || (headers && headers.host);
-
-  // Same regex used by AWS SDK at /lib/services/sqs.js
-  const hostMatchesSQS = evaluateHeaderValue(hostInfo, header => header.match(/^sqs\.(?:.+?)\./) !== null);
-
-  // 'user-agent': 'aws-sdk-js/3.329.0 os/darwin/22.5.0 lang/js md/nodejs/18.14.2 api/sqs/3.329.0'
-  const agentMatchesSQS = evaluateHeaderValue(userAgent, header => header.toLowerCase().indexOf('api/sqs') > -1);
-
-  if (parentSpan && parentSpan.n === 'sqs' && isAWSNodeJSHeader && (hostMatchesSQS || agentMatchesSQS)) {
```
abhilash-sivan, line 118: After deciding to not instrument the SDK spans, this case is redundant

kirrg001 (approved):
Pre-approving

Requirements:
1. We add span lengths to main branch for sqs test + rebase this PR
2. We fix sonar report

aryamohanan (approved):
🎉

packages/collector/test/integration/currencies/protocols/http/client/test_base.js
```diff
@@ -1417,6 +1417,56 @@ module.exports = function (name, version, isLatest) {
             })
           )
         ));
+
+    it('must propagate X-INSTANA and traceparent headers through an external package signed request', () =>
+      clientControls
+        .sendRequest({
+          method: 'GET',
+          path: '/aws4-axios-signed-request'
```
kirrg001, line 1425: We can also rename this path. Not important though, just for consistency

packages/core/src/tracing/instrumentation/protocols/httpClient.js
```diff
@@ -178,6 +172,11 @@ function instrument(coreModule, forceHttps) {
 
     const parentSpan = skipTracingResult.parentSpan;
 
+    // Note: There is no need to create http EXIT spans for AWS SDK created events
```
kirrg001:
```suggestion
    // Note: There is no need to create http EXIT spans for AWS SDK operations. They have their own AWS instrumentations and spans.
    // Note: There was a case in the past where customers hit SignatureDoesNotMatch from AWS, because we have attached Instana headers to outgoing AWS SDK requests. We could only reproduce these in a script - see https://github.com/instana/nodejs/tree/main/packages/collector/test/bin/reproduce-scripts
```

PR #2657, opened 2026-07-15
chore: added support for INSTANA_LOG_LEVEL_CAPTURE env variable

aryamohanan:
introduced a new configuration option to control the minimum log level for log span capture.

Previously, log spans were created only for logs at **WARN** level and above. With this change, users can configure the minimum log level that should generate log spans.

## Configuration

- Environment variable: `INSTANA_LOG_LEVEL_CAPTURE`
- In-code configuration: `tracing.logLevelCapture`

Supported values:

- `INFO`
- `WARN` (default)
- `ERROR`
- `OFF`

## Log capture behavior

| Configuration | Captured log levels |
|---------------|---------------------|
| `INFO` | INFO, WARN, ERROR, FATAL |
| `WARN` (default) | WARN, ERROR, FATAL |
| `ERROR` | ERROR, FATAL |
| `OFF` | No log spans are captured |

### DIFF
--- a/packages/core/src/config/index.js
+++ b/packages/core/src/config/index.js
@@ -7,7 +7,12 @@
 
 const normalizers = require('./normalizers');
 const deepMerge = require('../util/deepMerge');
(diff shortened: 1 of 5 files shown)


kirrg001 (approved):
Great 👹

PR #2651, opened 2026-07-14
feat(bunyan): added option to capture info logs

aryamohanan:
## Summary

This PR adds support for capturing **info**-level logs for Bunyan. By default, log capture starts at **WARN** and above. When configured, **INFO** logs can also be captured.

Reference: https://www.npmjs.com/package/bunyan#levels

## Configuration

The log capture level can be configured using either:

- Environment variable: `INSTANA_TRACING_CAPTURE_LOG_LEVEL`
- In-code configuration: `tracing.captureLogLevel`

Supported values:

- `INFO`
- `WARN` (default)
- `ERROR`
- `OFF`

## Log capture behavior

| Configuration | Captured log levels        |
| ------------- | -------------------------- |
| `INFO`        | INFO, WARN, ERROR, FATAL   |
| `WARN`        | WARN, ERROR, FATAL         |
| `ERROR`       | ERROR, FATAL               |
| `OFF`         | No log spans are captured. |

## TODO

- [x] config support  https://github.com/instana/nodejs/pull/2657
- [x] pino  #2661
- [ ] winston #2662
- [ ] log4js #2659
- [x] console https://github.com/instana/nodejs/pull/2658
- [ ] bunyan #2651
- [ ] public docs

ref: https://jsw.ibm.com/browse/INSTA-101942
tech spec: https://github.ibm.com/instana/technical-documentation/blob/master/tracing/specification/README.md#log-spans

PR #2658, opened 2026-07-15
feat(console): added option to capture info logs

aryamohanan:
Previously, console log spans were created only for logs at WARN level and above. With this change, users can configure the minimum log level that should generate log spans.

### Configuration

**Environment variable**: INSTANA_LOG_LEVEL_CAPTURE
**In-code configuration**: tracing.logLevelCapture

#### Supported values:

- INFO
- WARN (default)
- ERROR
- OFF

### DIFF
--- a/packages/core/src/tracing/tracingUtil.js
+++ b/packages/core/src/tracing/tracingUtil.js
@@ -10,7 +10,7 @@ const path = require('path');
 const StringDecoder = require('string_decoder').StringDecoder;
 
 const stackTrace = require('../util/stackTrace');
-const { STACK_TRACE_MODES, LOG_LEVEL_PRIORITY, DEFAULT_LOG_LEVEL, LOG_LEVEL } = require('../util/constants');
+const { STACK_TRACE_MODES, LOG_LEVEL_PRIORITY, LOG_LEVEL } = require('../util/constants');
 
 /** @type {import('../core').GenericLogger} */
 let logger;
@@ -463,10 +463,6 @@ exports.shouldCaptureLogSpan = function shouldCaptureLogSpan(level) {
     return false;
   }
   const minLevel = LOG_LEVEL_PRIORITY[logLevelConfig];
-
-  if (typeof level === 'number') {
-    return level >= minLevel;
-  }
 
   const numericLevel = LOG_LEVEL_PRIORITY[level?.toLowerCase()];
 
--- a/packages/core/src/tracing/tracingUtil.js
+++ b/packages/core/src/tracing/tracingUtil.js
@@ -52,7 +52,6 @@ exports.activate = function activate(_config) {
   stackTraceLength = _config.tracing.stackTraceLength;
   stackTraceMode = _config.tracing.stackTrace;
   httpExitConfig = _config.tracing.http.exit;
-  logLevelConfig = _config.tracing.logLevelCapture;
 };
 
 /**
--- a/packages/core/src/tracing/tracingUtil.js
+++ b/packages/core/src/tracing/tracingUtil.js
@@ -464,7 +464,7 @@ exports.shouldCaptureLogSpan = function shouldCaptureLogSpan(level) {
   }
   const minLevel = LOG_LEVEL_PRIORITY[logLevelConfig];
 
-  const numericLevel = LOG_LEVEL_PRIORITY[level?.toLowerCase()];
+  const numericLevel = LOG_LEVEL_PRIORITY[level.toLowerCase()];
 
   return numericLevel !== undefined && numericLevel >= minLevel;
 };
--- a/packages/core/src/tracing/tracingUtil.js
+++ b/packages/core/src/tracing/tracingUtil.js
@@ -10,7 +10,7 @@ const path = require('path');
 const StringDecoder = require('string_decoder').StringDecoder;
 
 const stackTrace = require('../util/stackTrace');
-const { STACK_TRACE_MODES, LOG_LEVEL_PRIORITY, DEFAULT_LOG_LEVEL } = require('../util/constants');
+const { STACK_TRACE_MODES, LOG_LEVEL_PRIORITY, DEFAULT_LOG_LEVEL, LOG_LEVEL } = require('../util/constants');
 
 /** @type {import('../core').GenericLogger} */
 let logger;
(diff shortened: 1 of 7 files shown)


packages/core/src/tracing/tracingUtil.js
```diff
@@ -447,3 +454,22 @@ exports.handleUnexpectedReturnValue = function handleUnexpectedReturnValue(retur
 
   return true;
 };
+
+/**
+ * @param {string} level
+ * @returns {boolean}
+ */
+exports.shouldCaptureLogSpan = function shouldCaptureLogSpan(level) {
+  if (logLevelConfig === 'off') {
```
kirrg001: Do you want to use the constant?

packages/core/src/tracing/tracingUtil.js
```diff
@@ -447,3 +454,22 @@ exports.handleUnexpectedReturnValue = function handleUnexpectedReturnValue(retur
...
+    return false;
+  }
+  const minLevel = LOG_LEVEL_PRIORITY[logLevelConfig];
+
+  if (typeof level === 'number') {
+    return level >= minLevel;
```
kirrg001: I thought that the log providers use different levels 🤔 
Some use 10-60, some different ranges.

packages/core/src/tracing/tracingUtil.js
```diff
@@ -447,3 +454,22 @@ exports.handleUnexpectedReturnValue = function handleUnexpectedReturnValue(retur
...
+
+/**
+ * @param {string} level
+ * @returns {boolean}
+ */
+exports.shouldCaptureLogSpan = function shouldCaptureLogSpan(level) {
+  if (logLevelConfig === 'off') {
+    return false;
+  }
+  const minLevel = LOG_LEVEL_PRIORITY[logLevelConfig];
```
kirrg001, line 465: This can go into 

> if (typeof level === 'number') {

Right?

packages/core/src/tracing/tracingUtil.js
aryamohanan: We'll map this in the corresponding instrumentations to ensure it matches. For console, Pino etc, the mapping will be the same.

kirrg001: I guess I need to see another PR in action. 

Not sure about

> LOG_LEVEL_PRIORITY[logLevelConfig];

Because the log level comes from the library IMO

> exports.shouldCaptureLogSpan = function shouldCaptureLogSpan(level, **levels**) {

packages/core/src/tracing/tracingUtil.js
aryamohanan: It's used in both cases, so keeping it outside avoids duplicating LOG_LEVEL_PRIORITY[logLevelConfig]. Since logLevelConfig is always a string(default -warn), computing it once is ok?

packages/core/src/tracing/tracingUtil.js
aryamohanan: I missed this. I added support for `off` later and forgot to update on the PR. 😄 Thanks for catching it!

packages/core/src/tracing/tracingUtil.js
aryamohanan: Here you go: https://github.com/instana/nodejs/pull/2659/changes

aryamohanan: you can also see winston mapping here https://github.com/instana/nodejs/pull/2651/changes#diff-759e7c503a0a1298914ab9a5a2bcfa3e58cc258bb076f0458e8baf89d0a3e5b9

kirrg001: Already discussed all topics offline.

packages/core/test/tracing/tracingUtil_test.js
```diff
@@ -1504,4 +1505,183 @@ describe('tracing/tracingUtil', () => {
...
+      });
+
+      it('should return true for numeric fatal level', () => {
+        expect(shouldCaptureLogSpan(LOG_LEVEL_PRIORITY[LOG_LEVEL.FATAL])).to.equal(true);
+      });
+    });
+
+    describe('when using a custom level map', () => {
+      const LEVEL_MAP = {
+        5000: LOG_LEVEL.TRACE,
```
kirrg001: This needs to be 5000: Number not reference.
Otherwise you can test against a false positive if the level in constants is wrong

packages/core/src/tracing/tracingUtil.js
```diff
@@ -447,3 +453,28 @@ exports.handleUnexpectedReturnValue = function handleUnexpectedReturnValue(retur
...
+/**
+ * @param {string | number } level
+ * @param {Object.<number, string>} [levelMap]
+ * @returns {boolean}
+ */
+exports.shouldCaptureLogSpan = function shouldCaptureLogSpan(level, levelMap) {
+  if (logLevelConfig === LOG_LEVEL.OFF) {
+    return false;
+  }
+  const minLevel = LOG_LEVEL_PRIORITY[logLevelConfig];
```
kirrg001, line 465: I thought we will do:

> localLevelPriority || GLOBAL_LOG_LEVEL_PRIORITY

aryamohanan: We have the std priority order, the order is same for all

packages/core/src/tracing/tracingUtil.js
```diff
@@ -29,6 +29,11 @@ let stackTraceMode;
  * @type {import("../config").HTTPExitTracingOptions | undefined}
  */
 let httpExitConfig;
+
+/**
+ * @type {string}
+ */
+let logLevelConfig = DEFAULT_LOG_LEVEL;
```
kirrg001: 🤔 Thats the default which we should set in core config, right

aryamohanan:
```suggestion
let logLevelConfig;
```

packages/core/src/tracing/tracingUtil.js
```diff
@@ -447,3 +453,18 @@ exports.handleUnexpectedReturnValue = function handleUnexpectedReturnValue(retur
...
+ * @param {string} level
+ * @returns {boolean}
+ */
+exports.shouldCaptureLogSpan = function shouldCaptureLogSpan(level) {
+  if (logLevelConfig === LOG_LEVEL.OFF) {
+    return false;
+  }
+  const minLevel = LOG_LEVEL_PRIORITY[logLevelConfig];
+
+  const numericLevel = LOG_LEVEL_PRIORITY[level?.toLowerCase()];
```
kirrg001: Do we need a protection for level being null?

kirrg001 (approved):
Pre-approving

PR #2659, opened 2026-07-15
feat(log4js): added option to capture info logs

aryamohanan:
## Summary

This PR adds support for capturing **info**-level logs for log4js. By default, log capture starts at **WARN** and above. When configured, **INFO** logs can also be captured.

Reference: https://github.com/log4js-node/log4js-node/blob/master/lib/levels.js#L96

https://log4js-node.github.io/log4js-node/api.html

## Configuration

- Environment variable: `INSTANA_TRACING_CAPTURE_LOG_LEVEL`
- In-code configuration: `tracing.captureLogLevel`

Supported values:

- `INFO`
- `WARN` (default)
- `ERROR`
- `OFF`

## Log capture behavior

### DIFF
--- a/packages/core/src/tracing/instrumentation/logging/log4js.js
+++ b/packages/core/src/tracing/instrumentation/logging/log4js.js
@@ -8,12 +8,22 @@
 const util = require('util');
 const shimmer = require('../../shimmer');
 
+const { LOG_LEVEL } = require('../../../util/constants');
 const hook = require('../../../util/hook');
 const tracingUtil = require('../../tracingUtil');
 const constants = require('../../constants');
 const cls = require('../../cls');
 
 let isActive = false;
+
+const LEVEL_MAP = {
+  5000: LOG_LEVEL.TRACE,
+  10000: LOG_LEVEL.DEBUG,
+  20000: LOG_LEVEL.INFO,
+  30000: LOG_LEVEL.WARN,
+  40000: LOG_LEVEL.ERROR,
+  50000: LOG_LEVEL.FATAL
+};
 
 exports.init = function init() {
   hook.onFileLoad(/\/log4js\/lib\/logger\.js/, instrumentLog4jsLogger);
@@ -34,15 +44,20 @@ function shimLog(originalLog) {
       return originalLog.apply(this, arguments);
     }
 
-    if (level == null || typeof level.level !== 'number' || level.level < 30000) {
+    if (level == null || typeof level.level !== 'number') {
       return originalLog.apply(this, arguments);
     }
 
-    return instrumentedLog(this, data, originalLog, level.level >= 40000, level);
+    const resolvedLogLevel = resolveLogLevel(level.level);
+    if (!resolvedLogLevel || !tracingUtil.shouldCaptureLogSpan(resolvedLogLevel)) {
+      return originalLog.apply(this, arguments);
+    }
+
+    return instrumentedLog(this, data, originalLog, resolvedLogLevel);
   };
 }
 
-function instrumentedLog(ctx, data, originalLog, markAsError, level) {
+function instrumentedLog(ctx, data, originalLog, level) {
   return cls.ns.runAndReturn(() => {
     let message;
 
@@ -61,7 +76,7 @@ function instrumentedLog(ctx, data, originalLog, markAsError, level) {
     span.data.log = {
       message
     };
-    if (markAsError) {
+    if (tracingUtil.isLogLevelAnError(level)) {
       span.ec = 1;
     }
 
@@ -72,6 +87,10 @@ function instrumentedLog(ctx, data, originalLog, markAsError, level) {
       span.transmit();
     }
   });
+}
+
+function resolveLogLevel(level) {
+  return LEVEL_MAP[level];
 }
 
 exports.activate = function activate() {
--- a/packages/core/src/tracing/instrumentation/logging/pino.js
+++ b/packages/core/src/tracing/instrumentation/logging/pino.js
@@ -10,7 +10,7 @@
 const { inspect } = require('util');
 const shimmer = require('../../shimmer');
 
-const { LOG_LEVEL_PRIORITY, LOG_LEVEL } = require('../../../util/constants');
+const { LOG_LEVEL_PRIORITY } = require('../../../util/constants');
 const hook = require('../../../util/hook');
 const tracingUtil = require('../../tracingUtil');
 const constants = require('../../constants');
@@ -92,7 +92,7 @@ function shimGenLog(originalGenLog) {
             message
           };
 
-          if (levelName === LOG_LEVEL.ERROR || levelName === LOG_LEVEL.FATAL) {
+          if (tracingUtil.isLogLevelAnError(levelName)) {
             span.ec = 1;
           }
 
--- a/packages/collector/test/integration/currencies/logging/log4js/test_base.js
+++ b/packages/collector/test/integration/currencies/logging/log4js/test_base.js
@@ -67,7 +67,7 @@ module.exports = function (name, version, isLatest) {
       }
 
       it(`must not trace info ${suffix}`, async () => {
-        await trigger({ level: 'info', message: 'Info message - must not be traced.', useLogMethod });
+        await trigger({ level: 'info', message: 'Info message - must not be traced by default.', useLogMethod });
         return retry(async () => {
           const spans = await agentControls.getSpans();
           const entrySpan = expectAtLeastOneMatching(spans, [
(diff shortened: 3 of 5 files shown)


packages/core/src/tracing/instrumentation/logging/log4js.js
```diff
@@ -8,13 +8,23 @@
...
+const { LOG_LEVEL } = require('../../../util/constants');
 const hook = require('../../../util/hook');
 const tracingUtil = require('../../tracingUtil');
 const constants = require('../../constants');
 const cls = require('../../cls');
 
 let isActive = false;
 
+const LEVEL_MAP = {
+  5000: LOG_LEVEL.TRACE,
```
aryamohanan, line 20: https://github.com/log4js-node/log4js-node/blob/master/lib/levels.js#L96

packages/core/src/tracing/instrumentation/logging/log4js.js
```diff
@@ -44,7 +54,9 @@ function shimLog(originalLog) {
     }
 
     const actualLevel = levels && levels.getLevel(level, levels.INFO);
-    if (actualLevel == null || typeof actualLevel.level !== 'number' || actualLevel.level < 30000) {
+
+    const levelName = resolveLogLevel(actualLevel && actualLevel.level);
```
kirrg001: I would return early if actualLevel is null. Hard to read
Because we have to do

> resolveLogLevel(actualLevel && actualLevel.level);

With potential null

And another !levelName check.

packages/core/src/tracing/instrumentation/logging/log4js.js
```diff
@@ -44,7 +54,9 @@ function shimLog(originalLog) {
     }
 
     const actualLevel = levels && levels.getLevel(level, levels.INFO);
```
kirrg001: Whats this exactly? Can you pls give an example?

aryamohanan: This was already there before. 

This is my understanding: `getLevel()` is used to resolve the provided log4js level into a standard `Level` object before applying our existing mapping. This handles different custom inputs supported by log4js.

Examples:

```js
logger.log('ERROR', 'Something failed');
````

resolves to:

```js
Level { level: 40000, levelStr: 'ERROR', colour: 'red' }
```

If an custom numeric value is provided:

```js
logger.log(3, 'Something wrong');
```

`getLevel()` cannot resolve the value, so it falls back to `levels.INFO`:

```js
Level { level: 20000, levelStr: 'INFO', colour: 'green' }
```

This ensures we always get a valid log4js `Level` object before applying our Instana log level mapping.

kirrg001: Seems redundant to me or?

aryamohanan: I don't think they're doing the same thing. getLevel() first converts the input into a log4js Level object (for example, from "ERROR" or a Level instance). Our LEVEL_MAP only maps the numeric value from that object (for example, 40000 -> LOG_LEVEL.ERROR).

We could potentially remove getLevel(), but then we'd need to handle all the input formats that log4js accepts ourselves. Since this code was already there, I assumed it was intentionally relying on log4js to normalize the input before applying our mapping.

kirrg001: I am confused.

If I print these:

```js
LEVEL Level { level: 50000, levelStr: 'FATAL', colour: 'magenta' }
[class Level] {
  levels: [
    Level { level: 5e-324, levelStr: 'ALL', colour: 'grey' },
    Level { level: 5000, levelStr: 'TRACE', colour: 'blue' },
    Level { level: 10000, levelStr: 'DEBUG', colour: 'cyan' },
    Level { level: 20000, levelStr: 'INFO', colour: 'green' },
    Level { level: 30000, levelStr: 'WARN', colour: 'yellow' },
    Level { level: 40000, levelStr: 'ERROR', colour: 'red' },
    Level { level: 50000, levelStr: 'FATAL', colour: 'magenta' },
    Level { level: 9007199254740992, levelStr: 'MARK', colour: 'grey' },
    Level {
      level: 1.7976931348623157e+308,
      levelStr: 'OFF',
      colour: 'grey'
    }
  ],
  ALL: Level { level: 5e-324, levelStr: 'ALL', colour: 'grey' },
  TRACE: Level { level: 5000, levelStr: 'TRACE', colour: 'blue' },
  DEBUG: Level { level: 10000, levelStr: 'DEBUG', colour: 'cyan' },
  INFO: Level { level: 20000, levelStr: 'INFO', colour: 'green' },
  WARN: Level { level: 30000, levelStr: 'WARN', colour: 'yellow' },
  ERROR: Level { level: 40000, levelStr: 'ERROR', colour: 'red' },
  FATAL: Level { level: 50000, levelStr: 'FATAL', colour: 'magenta' },
  MARK: Level { level: 9007199254740992, levelStr: 'MARK', colour: 'grey' },
  OFF: Level {
    level: 1.7976931348623157e+308,
    levelStr: 'OFF',
    colour: 'grey'
  }
}
ACTUAL LEVEL Level { level: 50000, levelStr: 'FATAL', colour: 'magenta' }
```

aryamohanan: You were right that getLevel() was redundant — but only because we were shimming the wrong method. The real bug is that we patched `log` instead of `_log`. log receives raw user input (strings, numbers, whatever), so getLevel() was there as a normalization workaround. _log is the internal method log4js always calls after resolving the level, so it always receives a proper Level instance - no normalization needed.

When shimming log:

logger.error('msg') → calls this.log(level, ...args) where level is already a Level object →  works
logger.log('error', 'msg') → level is the string 'error' → typeof 'error'.level !== 'number' →  never traced (without getLevel())

see: https://github.com/log4js-node/log4js-node/blob/master/lib/logger.js#L224

I can raise a PR against main to fix the bug

kirrg001: I am wondering if older versions need this logic? 🤔

kirrg001: >logger.log('error', 'msg') → level is the string 'error' → typeof 'error'.level !== 'number' → never traced (without getLevel())

Ah yeah this ^ 
IMO we just need a use case comment in the codebase 👍

packages/core/src/tracing/instrumentation/logging/log4js.js
```diff
@@ -90,6 +97,13 @@ function instrumentedLog(ctx, originalArgs, originalLog, markAsError) {
   });
 }
 
+function resolveLogLevel(level) {
+  return LEVEL_MAP[level];
+}
+
+function shouldMarkAsError(level) {
```
kirrg001: https://github.com/instana/nodejs/pull/2651#discussion_r3628979737

aryamohanan: done

packages/core/src/tracing/instrumentation/logging/log4js.js
aryamohanan: resolved this case

aryamohanan:
I will merge this as 2 commits

> - chore: added helper to determine error log levels
> - feat(log4js): added option to capture info logs

packages/core/src/tracing/instrumentation/logging/log4js.js
```diff
@@ -34,15 +44,20 @@ function shimLog(originalLog) {
...
       return originalLog.apply(this, arguments);
     }
 
-    return instrumentedLog(this, data, originalLog, level.level >= 40000, level);
+    const levelName = resolveLogLevel(level.level);
+    if (!levelName || !tracingUtil.shouldCaptureLogSpan(levelName)) {
+      return originalLog.apply(this, arguments);
+    }
+
+    return instrumentedLog(this, data, originalLog, level);
```
aryamohanan:
```suggestion
    return instrumentedLog(this, data, originalLog, levelName);
```

packages/core/src/tracing/instrumentation/logging/pino.js
```diff
@@ -92,7 +92,7 @@ function shimGenLog(originalGenLog) {
             message
           };
 
-          if (levelName === LOG_LEVEL.ERROR || levelName === LOG_LEVEL.FATAL) {
+          if (tracingUtil.shouldMarkErrorForLogLevel(level)) {
```
kirrg001: suggestion: The name `shouldMarkErrorForLogLevel` feels a bit clunky to read. A couple of cleaner alternatives could be `shouldMarkSpanAsError` or `isLogLevelAnError` (preferred).

kirrg001 (approved):
Pre-approving

PR #2660, opened 2026-07-16
fix(core): resolved potential bootstrap problems during preinit

aryamohanan:
… available


kirrg001 (approved):
We know that this is required for upcoming logging PR's, because otherwise tracingUtil won't have the config object available during preinit.

Potentially this fixes other unknown bugs on bootstrap during preinit.

PR #2661, opened 2026-07-16
feat(pino): added option to capture info logs

aryamohanan:
## Summary

This PR adds support for capturing **info**-level logs for Pino. By default, log capture starts at **WARN** and above. When configured, **INFO** logs can also be captured.

Reference: https://github.com/pinojs/pino/blob/HEAD/docs/api.md#opt-customlevels

## Configuration

- Environment variable: `INSTANA_LOG_LEVEL_CAPTURE`
- In-code configuration: `tracing.logLevelCapture`

Supported values:

- `INFO`
- `WARN` (default)
- `ERROR`
- `OFF`

## Log capture behavior

### DIFF
--- a/packages/core/src/tracing/instrumentation/logging/pino.js
+++ b/packages/core/src/tracing/instrumentation/logging/pino.js
@@ -10,13 +10,17 @@
 const { inspect } = require('util');
 const shimmer = require('../../shimmer');
 
-const { LOG_LEVEL } = require('../../../util/constants');
+const { LOG_LEVEL_PRIORITY } = require('../../../util/constants');
 const hook = require('../../../util/hook');
 const tracingUtil = require('../../tracingUtil');
 const constants = require('../../constants');
 const cls = require('../../cls');
 
 let isActive = false;
+
+// Reverse the log level mapping once for numeric-to-name lookups.
+// Pino internally represents log levels as numbers (for example, 30 -> "info").
+const LEVEL_NAMES = Object.fromEntries(Object.entries(LOG_LEVEL_PRIORITY).map(([name, value]) => [value, name]));
 
 exports.init = function init() {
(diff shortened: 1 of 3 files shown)


packages/core/src/tracing/instrumentation/logging/pino.js
```diff
@@ -103,6 +104,30 @@ function shimGenLog(originalGenLog) {
   };
 }
 
+function resolveLogLevel(level) {
+  if (typeof level === 'string') {
+    return level.toLowerCase();
+  }
+
+  if (level >= 60) {
```
aryamohanan: Pino accepts custom log levels as numeric values, so instead of using a one-to-one mapping, we check the value against a range.

https://github.com/pinojs/pino/blob/HEAD/docs/api.md#opt-customlevels

packages/core/src/tracing/instrumentation/logging/pino.js
```diff
@@ -103,6 +104,30 @@ function shimGenLog(originalGenLog) {
   };
 }
 
+function resolveLogLevel(level) {
```
kirrg001, line 111: suggestion:

```
// reverse once (or define twice in constants)
const levels = Object.fromEntries(
  Object.entries(constants.LOG_LEVEL_PRIORITY).map(([key, value]) => [value, key])
);

function shimGenLog(originalGenLog) {
  ....
  // NOTE: The `logLevelCapture` only accepts official log level names. Custom pino levels (e.g. 55 is "vahalla") are rounded down to nearest ten.
  const levelName = levels[Math.floor(level / 10) * 10]
}
```

aryamohanan: updated

kirrg001 (approved):

abhilash-sivan (approved):

PR #2662, opened 2026-07-17
feat(winston): added option to capture info logs

aryamohanan:
## Summary

This PR adds support for capturing **info**-level logs for Winston. By default, log capture starts at **WARN** and above. When configured, **INFO** logs can also be captured.

Reference: https://github.com/winstonjs/winston#logging-levels

## Configuration

- Environment variable: `INSTANA_TRACING_CAPTURE_LOG_LEVEL`
- In-code configuration: `tracing.captureLogLevel`

Supported values:

- `INFO`
- `WARN` (default)
- `ERROR`
- `OFF`

## Log capture behavior

### DIFF
--- a/packages/core/src/tracing/instrumentation/logging/winston.js
+++ b/packages/core/src/tracing/instrumentation/logging/winston.js
@@ -61,31 +61,31 @@ function instrumentWinston3(createLogger) {
     const derivedLogger = createLogger.apply(this, arguments);
 
     // npm levels
-    shimLevelMethod(derivedLogger, 'error', true);
-    shimLevelMethod(derivedLogger, 'warn', false);
-    shimLevelMethod(derivedLogger, 'info', false);
+    shimLevelMethod(derivedLogger, 'error');
+    shimLevelMethod(derivedLogger, 'warn');
+    shimLevelMethod(derivedLogger, 'info');
 
     // syslog levels (RFC5424)
-    shimLevelMethod(derivedLogger, 'emerg', true);
-    shimLevelMethod(derivedLogger, 'alert', true);
-    shimLevelMethod(derivedLogger, 'crit', true);
-    shimLevelMethod(derivedLogger, 'warning', false);
-    shimLevelMethod(derivedLogger, 'notice', false);
+    shimLevelMethod(derivedLogger, 'emerg');
+    shimLevelMethod(derivedLogger, 'alert');
+    shimLevelMethod(derivedLogger, 'crit');
+    shimLevelMethod(derivedLogger, 'warning');
+    shimLevelMethod(derivedLogger, 'notice');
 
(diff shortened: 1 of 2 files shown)


packages/core/src/tracing/instrumentation/logging/winston.js
```diff
@@ -9,9 +9,27 @@ const hook = require('../../../util/hook');
...
+  // npm levels
+  error: LOG_LEVEL.ERROR,
+  warn: LOG_LEVEL.WARN,
+  info: LOG_LEVEL.INFO,
+  debug: LOG_LEVEL.DEBUG,
+  verbose: LOG_LEVEL.TRACE,
+  silly: LOG_LEVEL.TRACE,
+
+  // RFC 5424 syslog levels
+  emerg: LOG_LEVEL.FATAL,
```
aryamohanan, line 26: reference: https://github.com/winstonjs/winston#logging-levels

abhilash-sivan (approved):

packages/core/src/tracing/instrumentation/logging/winston.js
```diff
@@ -78,6 +98,10 @@ function instrumentedLevelMethod(originalMethod, markAsError) {
       return originalMethod.apply(this, arguments);
     }
 
+    if (!tracingUtil.shouldCaptureLogSpan(level)) {
```
kirrg001: resolveLogLevel is missing or

aryamohanan:
```suggestion
    if (!tracingUtil.shouldCaptureLogSpan(resolveLogLevel(level))) {
```

aryamohanan: Good catch. shouldCaptureLogSpan currently expects a resolved log level, so resolveLogLevel should be called before passing the value. I'll update it.

packages/core/src/tracing/instrumentation/logging/winston.js
```diff
@@ -153,12 +177,13 @@ function instrumentedLog(originalMethod) {
...
 
-function levelIsTraced(level) {
-  return levelIsError(level) || level === 'warn' || level === 'warning';
+function resolveLogLevel(level) {
+  return LEVEL_MAP[level];
 }
 
 function levelIsError(level) {
-  return level === 'error' || level === 'emerg' || level === 'alert' || level === 'crit';
+  const resolvedLevel = resolveLogLevel(level);
```
kirrg001: Currently we call this fn in three places and worst case we call it 3x.

We can remember it:

```js
const normalizedLogLevel = resolveLogLevel(level):

// use it normalizedLogLevel

kirrg001 (approved):
Pre-approving

PR #2665, opened 2026-07-20
fix: renamed INSTANA_DISABLE_W3C_TRACE_CORRELATION to INSTANA_TRACING_DISABLE_W3C_CORRELATION

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-93111

### DIFF
--- a/packages/core/src/config/index.js
+++ b/packages/core/src/config/index.js
@@ -992,7 +992,12 @@ function normalizeDisableW3cTraceCorrelation({ userConfig = {}, defaultConfig = 
 
   // Pass the env var name to util.resolve (which calls process.env[envValue] internally).
   // New env var takes precedence; fall back to deprecated if new is not set.
-  const envValue = isNewEnvSet ? newEnvName : isDeprecatedEnvSet ? deprecatedEnvName : undefined;
+  let envValue;
+  if (isNewEnvSet) {
+    envValue = newEnvName;
+  } else if (isDeprecatedEnvSet) {
+    envValue = deprecatedEnvName;
+  }
 
   const deprecatedInCode = userConfig.tracing.disableW3cTraceCorrelation;
   const newInCode = userConfig.tracing.disableW3cCorrelation;
--- a/packages/core/src/config/index.js
+++ b/packages/core/src/config/index.js
@@ -990,7 +990,6 @@ function normalizeDisableW3cTraceCorrelation({ userConfig = {}, defaultConfig = 
     );
   }
 
-  // Pass the env var name to util.resolve (which calls process.env[envValue] internally).
   // New env var takes precedence; fall back to deprecated if new is not set.
   let envValue;
   if (isNewEnvSet) {
--- a/packages/core/src/config/index.js
+++ b/packages/core/src/config/index.js
@@ -980,22 +980,19 @@ function normalizeDisableW3cTraceCorrelation({ userConfig = {}, defaultConfig = 
   const newEnvName = 'INSTANA_TRACING_DISABLE_W3C_CORRELATION';
   const deprecatedEnvName = 'INSTANA_DISABLE_W3C_TRACE_CORRELATION';
 
-  const newEnvVar = process.env[newEnvName];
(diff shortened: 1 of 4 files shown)


abhilash-sivan (approved):

packages/core/src/tracing/tracingHeaders.js
```diff
@@ -14,7 +14,7 @@ const w3c = require('./w3c_trace_context');
  */
 let logger;
 
-let disableW3cTraceCorrelation = false;
```
aryamohanan, line 17: In `tracingHeaders_test`, we're still using `disableW3cTraceCorrelation: false` on line 38. We should update that as well.

packages/core/src/config/index.js
```diff
@@ -975,22 +977,62 @@ function normalizeDisableW3c({ finalConfig = {} } = {}) {
  * @param {{ userConfig?: InstanaConfig|null, defaultConfig?: InstanaConfig, finalConfig?: InstanaConfig }} [options]
  */
 function normalizeDisableW3cTraceCorrelation({ userConfig = {}, defaultConfig = {}, finalConfig = {} } = {}) {
+  const newEnvName = 'INSTANA_TRACING_DISABLE_W3C_CORRELATION';
```
aryamohanan, line 980: Nit: I'd rename newEnvName to envName. The "new" qualifier is only relevant during the deprecation window, whereas envName will remain accurate even after the deprecated variable is removed.

packages/core/src/config/index.js
```diff
@@ -975,22 +977,62 @@ function normalizeDisableW3c({ finalConfig = {} } = {}) {
...
+  const isDeprecatedEnvSet = process.env[deprecatedEnvName] !== undefined;
+
+  if (isDeprecatedEnvSet) {
+    logger.warn(
+      // eslint-disable-next-line max-len
+      `[Deprecation Warning] The environment variable ${deprecatedEnvName} is deprecated and will be removed in a future release. Please use ${newEnvName} instead.`
+    );
+  }
+
+  // Pass the env var name to util.resolve (which calls process.env[envValue] internally).
```
aryamohanan:
```suggestion
```
Is this comment intentional?

packages/core/src/config/index.js
```diff
@@ -975,22 +977,62 @@ function normalizeDisableW3c({ finalConfig = {} } = {}) {
...
+    envValue = newEnvName;
+  } else if (isDeprecatedEnvSet) {
+    envValue = deprecatedEnvName;
+  }
+
+  const deprecatedInCode = userConfig.tracing.disableW3cTraceCorrelation;
+  const newInCode = userConfig.tracing.disableW3cCorrelation;
+  let inCodeValue;
+
+  if (deprecatedInCode !== undefined) {
```
aryamohanan, line 1005: Nit: Consider checking `disableW3cCorrelation` first since it takes precedence over the deprecated option. This makes the precedence more explicit and avoids relying on the later assignment to override the earlier one.

packages/core/src/tracing/tracingHeaders.js
kirrg001: 👍

packages/core/src/config/index.js
kirrg001: I think its fine as it is 👍

packages/core/src/config/index.js
kirrg001: See 

> 
  if (newInCode !== undefined) {
    inCodeValue = newInCode;
  }

We override

aryamohanan: Yes, I understand the override gives newInCode higher precedence. My suggestion was only about making the precedence more explicit rather than relying on the later assignment. For example:

```
if (deprecatedInCode !== undefined) {
  logger.warn(...);
}

const inCodeValue =
  newInCode !== undefined ? newInCode : deprecatedInCode;
```

I'm okay with the current implementation as well, it was just a readability nit.

aryamohanan (approved):

PR #2672, opened 2026-07-23
chore(log4js): shim _log instead of log 

aryamohanan:
The old shim wrapped `log()`, which received the raw value passed by the user (for example, `"ERROR"` or `3`). So we needed `getLevel()` to convert it into a `Level` object.

Now we're wrapping `_log()`, which log4js calls **after** it has already resolved the level. That means `level` is always a `Level` object, so `getLevel()` is no longer needed.

This also lets us remove the extra mapping logic (`LEVEL_MAP`, `LOG_LEVEL`, `shouldCaptureLogSpan`, and `resolveLogLevel`), since we can use the resolved `Level` object directly.

### DIFF
--- a/packages/core/src/tracing/instrumentation/logging/log4js.js
+++ b/packages/core/src/tracing/instrumentation/logging/log4js.js
@@ -15,25 +15,16 @@ const cls = require('../../cls');
 
 let isActive = false;
 
-let levels;
-
 exports.init = function init() {
-  hook.onFileLoad(/\/log4js\/lib\/levels\.js/, saveLevelsRef);
   hook.onFileLoad(/\/log4js\/lib\/logger\.js/, instrumentLog4jsLogger);
 };
 
-function saveLevelsRef(levelsModule) {
-  if (typeof levelsModule.getLevel === 'function' && levelsModule.INFO != null) {
-    levels = levelsModule;
-  }
-}
-
 function instrumentLog4jsLogger(loggerModule) {
-  shimmer.wrap(loggerModule.prototype, 'log', shimLog);
+  shimmer.wrap(loggerModule.prototype, '_log', shimLog);
 }
 
 function shimLog(originalLog) {
-  return function (level) {
+  return function (level, data) {
     // The __instana attribute identifies the Instana logger, so prevent these logs from being traced.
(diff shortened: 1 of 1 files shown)


packages/core/src/tracing/instrumentation/logging/log4js.js
```diff
@@ -15,25 +15,16 @@ const cls = require('../../cls');
...
 
-function saveLevelsRef(levelsModule) {
-  if (typeof levelsModule.getLevel === 'function' && levelsModule.INFO != null) {
-    levels = levelsModule;
-  }
-}
-
 function instrumentLog4jsLogger(loggerModule) {
-  shimmer.wrap(loggerModule.prototype, 'log', shimLog);
+  shimmer.wrap(loggerModule.prototype, '_log', shimLog);
```
aryamohanan, line 23: see: https://github.com/log4js-node/log4js-node/blob/master/lib/logger.js#L224

packages/core/src/tracing/instrumentation/logging/log4js.js
```diff
@@ -43,31 +34,23 @@ function shimLog(originalLog) {
...
-      originalArgs[i] = arguments[i];
-    }
-
-    return instrumentedLog(this, originalArgs, originalLog, actualLevel.level >= 40000);
+    return instrumentedLog(this, data, originalLog, level.level >= 40000, level);
   };
 }
 
-function instrumentedLog(ctx, originalArgs, originalLog, markAsError) {
+function instrumentedLog(ctx, data, originalLog, markAsError, level) {
```
kirrg001, line 45: nitpik: order of arguments could be optimized

kirrg001 (approved):
Great!

packages/core/src/tracing/instrumentation/logging/log4js.js
aryamohanan: This is refactored in #2659

aryamohanan: It will be resolved.

kirrg001: Okii

PR #2678, opened 2026-07-27
fix: captured missing log level in log spans 

aryamohanan:
For some instrumentations, we're only sending the log message and not the log level. In the backend, if no log level is provided and ec != 0, the log is treated as WARN by default.

This means that if an INFO log span is sent without a level, the backend incorrectly classifies it as a WARN log, which doesn't seem right.

see: https://github.ibm.com/instana/backend/blob/develop/forge/src/main/java/com/instana/forge/connection/LogSpan.java#L116

Since we're already capturing the log level in our instrumentation, I think we should send it as part of the log span as well. The console instrumentation already includes the log level in span.data, so we could follow the same approach here.

UI: https://instana.io/s/RfyXRzHfRlGabC1AYVTaKg

### DIFF
--- a/packages/collector/test/integration/currencies/logging/bunyan/test_base.js
+++ b/packages/collector/test/integration/currencies/logging/bunyan/test_base.js
@@ -69,20 +69,21 @@ module.exports = function (name, version, isLatest) {
         )
       ));
 
-    it('must trace warn', () => runTest('warn', false, 'Warn message - should be traced.', controls));
+    it('must trace warn', () => runTest('warn', false, 'Warn message - should be traced.', 'warn', controls));
 
-    it('must trace error', () => runTest('error', true, 'Error message - should be traced.', controls));
+    it('must trace error', () => runTest('error', true, 'Error message - should be traced.', 'error', controls));
 
-    it('must trace fatal', () => runTest('fatal', true, 'Fatal message - should be traced.', controls));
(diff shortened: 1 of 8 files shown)


packages/collector/test/integration/currencies/logging/bunyan/test_base.js
```diff
@@ -69,78 +69,137 @@ module.exports = function (name, version, isLatest) {
         )
       ));
 
-    it('must trace warn', () => runTest('warn', false, 'Warn message - should be traced.', controls));
-
-    it('must trace error', () => runTest('error', true, 'Error message - should be traced.', controls));
-
-    it('must trace fatal', () => runTest('fatal', true, 'Fatal message - should be traced.', controls));
+    it('must trace warn', () =>
```
aryamohanan, line 72: I used options object and add level in it, no other changes

packages/collector/test/integration/currencies/logging/winston/test_base.js
```diff
@@ -245,6 +251,9 @@ module.exports = function (name, version, isLatest, mode) {
     expect(span.data).to.exist;
     expect(span.data.log).to.exist;
     expect(span.data.log.message).to.equal(message);
+    if (level) {
```
kirrg001: If the log level is always there, why do we need this condition 🤔

aryamohanan: yeah, this is not needed.

kirrg001 (approved):

PR #2679, opened 2026-07-27
feat: added support for latest opentelemetry esm instrumentations

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-97288

### Summary

1. Updated @opentelemetry/instrumentation-tedious from 0.28.0 to 0.40.0.
2. Updated import-in-the-middle from 2.0.5 to ^3.3.3 to maintain compatibility with the newer OpenTelemetry instrumentation.
3. Updated @instana/instrumentation-confluent-kafka-javascript 

### Reason

@opentelemetry/instrumentation-tedious@0.40.0 depends on a newer @opentelemetry/instrumentation version that requires import-in-the-middle@^3. Keeping import-in-the-middle@2.x caused npm to install multiple IITM versions, which broke ESM module interception and prevented Tedious instrumentation from patching correctly.

### Fix

Pinned import-in-the-middle to ^3.3.3 ensuring all instrumentation uses a single shared IITM instance, restoring ESM tracing while remaining compatible with the updated OpenTelemetry instrumentation.

### PR body
- bumped @opentelemetry/instrumentation-tedious from 0.28.0 to 0.40.0
- bumped import-in-the-middle from 2.0.5 to 3.3.3
- bumped @instana/instrumentation-confluent-kafka-javascript from 1.0.1 to 1.1.0


packages/collector/test/integration/currencies/databases/tedious/test_base.js
```diff
@@ -89,11 +89,10 @@ module.exports = function (name, version, isLatest) {
                         expect(span.data.tags.name).to.eql(`${queryType} azure-nodejs-test`);
 
                         expect(span.data.operation).to.equal('tedious');
-                        expect(span.data.tags['db.system']).to.eql('mssql');
-                        expect(span.data.tags['db.name']).to.eql('azure-nodejs-test');
-                        expect(span.data.tags['db.user']).to.eql('admin@instana@nodejs-team-db-server');
-                        expect(span.data.tags['db.statement']).to.eql(expectedStatement);
-                        expect(span.data.tags['net.peer.name']).to.eql('nodejs-team-db-server.database.windows.net');
+                        expect(span.data.tags['db.system.name']).to.eql('microsoft.sql_server');
```
abhilash-sivan, line 92: changed assertions to adapt new semconv

The stable OpenTelemetry semantic conventions (v1.33.0+) are supported starting with @opentelemetry/instrumentation-tedious v0.39.0. See the [Semantic Conventions](https://www.npmjs.com/package/@opentelemetry/instrumentation-tedious) section

kirrg001:
The problem is that Otel with ESM requires users to manually register the IITM hook.

See:
https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md

That means, our `esm-register.mjs` file automatically takes care of that requirement. We need to a proper explanation to our integration + our esm register.

If we just update IITM in core with v3, we can still run into issues with Otel instrumentations who use an older instrumentations dependency. But the Otel ecosystem suffers from the same problem.

**As soon as you have one Otel instrumentation which is still on the old instrumentation version, it won't work for the customer OR the instrumentation dependency got already deduped to the root (which is for example not the case for our setup - its still on 207).**

Right now with the Tedious 0.40 update, all our Otel dependencies use the newer instrumentation dependency already.
So we will have problems with Customer codebases and custom Otel installations and if we allow custom otel dependencies.

As soon the customer has any slightly different installation tree and IITM v2 is on the root, the Otel instrumentations would no longer work because the Otel instrumentation dependency loads IITM as a dependency and would load v2 from the root. But we instantiate v3.

Similar to our setup: as soon as we have an Otel instrumentation which needs IITM v2, it does not work anymore.

Can you please add proper explanations & tests which break the solution? The test can just be a simple reproduce script. For now we can only update to v3 + ^.

Long-term: we may need to figure out if multiple IITM versions are being used (similar to the Otel API fix)

abhilash-sivan:
> The problem is that Otel with ESM requires users to manually register the IITM hook.
> 
> See: https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md
> 
> That means, our `esm-register.mjs` file automatically takes care of that requirement. We need to a proper explanation to our integration + our esm register.
> 
> If we just update IITM in core with v3, we can still run into issues with Otel instrumentations who use an older instrumentations dependency. But the Otel ecosystem suffers from the same problem.
> 
> **As soon as you have one Otel instrumentation which is still on the old instrumentation version, it won't work for the customer OR the instrumentation dependency got already deduped to the root (which is for example not the case for our setup - its still on 207).**
> 
> Right now with the Tedious 0.40 update, all our Otel dependencies use the newer instrumentation dependency already. So we will have problems with Customer codebases and custom Otel installations and if we allow custom otel dependencies.
> 
> As soon the customer has any slightly different installation tree and IITM v2 is on the root, the Otel instrumentations would no longer work because the Otel instrumentation dependency loads IITM as a dependency and would load v2 from the root. But we instantiate v3.
> 
> Similar to our setup: as soon as we have an Otel instrumentation which needs IITM v2, it does not work anymore.
> 
> Can you please add proper explanations & tests which break the solution? The test can just be a simple reproduce script. For now we can only update to v3 + ^.
> 
> Long-term: we may need to figure out if multiple IITM versions are being used (similar to the Otel API fix)


I added a warning comment to IITM loader
https://github.com/instana/nodejs/pull/2679/changes#diff-af165377f863b6d4ad356ba4927e9b67ada30e8c4c531b4aee8c6db5c18c8e91R44

Can be deferred to debug or maybe remove entirely.

I also added a supporting test where multiple instantiations logs this warning

kirrg001:
Ty. Will check out asap

I think we have to close #2690 and merge the update both deps updates together right?

packages/collector/esm-register.mjs
```diff
@@ -16,6 +16,22 @@
  * node --import @instana/collector/esm-register.mjs server.js
  */
 
+/**
+ * Version conflict risk
```
kirrg001:
```suggestion
 * OpenTelemetry Integration correlation
```

packages/collector/esm-register.mjs
```diff
@@ -16,6 +16,22 @@
...
+ * ---------------------
```
kirrg001:
```suggestion
 * 
 * Our OpenTelemetry integration works because we load the IITM hook for ESM here. See https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md.
 * 
```

packages/collector/esm-register.mjs
```diff
@@ -16,6 +16,22 @@
...
+ * Instana pins IITM v3 (via @instana/core). Some older OTel instrumentation packages still depend on
```
kirrg001:
```suggestion
 * Instana pins IITM v3 (via @instana/core). In general, some OTel instrumentation packages still depend on
```

packages/collector/esm-register.mjs
```diff
@@ -16,6 +16,22 @@
...
+ * IITM v2. When npm deduplication places an IITM v2 copy at the project root, both versions end up
+ * loaded simultaneously: OTel loads v2 from the root, while Instana loads its own v3. Because each
+ * IITM instance keeps its own independent hook registry, hooks registered through one instance are
+ * invisible to the other. As a result, OTel instrumentations stop working entirely.
+ *
+ * Instana detects this condition at startup and emits a warning when multiple IITM instances are
+ * found in the module cache. See packages/core/src/util/iitmHook.js for the detection logic.
+ *
+ * Reference: https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md#additional-notes-on-experimental-loaders
+ * Note: A long-term fix would be to negotiate a single shared hook registry across all IITM versions.
```
kirrg001:
```suggestion
 * TODO: Address further cases as described as part of [INSTA-107020](https://jsw.ibm.com/browse/INSTA-107020)
```

Could we track this issue as a JIRA and link it here? I am not sure what the solution will be yet. I think we might need to load both long-term. But we need to learn from this first.

packages/collector/test/integration/currencies/databases/tedious/test_base.js
kirrg001: And our BE supports this version already, yeah?

packages/collector/esm-register.mjs
abhilash-sivan: Yeah I will create a card and track in JIRA

packages/collector/esm-register.mjs
```diff
@@ -16,6 +16,24 @@
  * node --import @instana/collector/esm-register.mjs server.js
  */
 
+/**
+ * OpenTelemetry Integration correlation
+ * 
+ * Our OpenTelemetry integration works because we load the IITM hook for ESM here. See https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md.
```
kirrg001, line 22:
```suggestion
 * Our OpenTelemetry integration works because we load the IITM hook for ESM here. See https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md.
 * > register('@instana/core/iitm-loader.mjs', import.meta.url);
 * This line adds support for Instana instrumentations and our Otel integration instrumentations working.
```

packages/collector/test/integration/currencies/databases/tedious/tedious-app.js
```diff
@@ -28,6 +28,14 @@ if (!tediousPath.includes(expectedLocalPath)) {
   );
 }
 
+const localNodeModules = tediousPath.substring(0, tediousPath.lastIndexOf('node_modules') + 'node_modules'.length);
+const tediousVersion = require(path.join(localNodeModules, 'tedious', 'package.json')).version;
+const otelTediousVersion = require(path.join(localNodeModules, '@opentelemetry', 'instrumentation-tedious', 'package.json')).version;
+
+console.log('[tedious-app] tedious version:', tediousVersion);
+console.log('[tedious-app] tedious resolved path:', tediousPath);
+console.log('[tedious-app] @opentelemetry/instrumentation-tedious version:', otelTediousVersion);
```
kirrg001: Do you want to remove them?

abhilash-sivan: yeah I cleaned up now :D

packages/core/src/util/iitmHook.js
```diff
@@ -20,9 +34,25 @@ exports.init = function init(config) {
...
+  if (iitmInstances.length > 1) {
+    logger.warn(
+      '[Instana] Multiple import-in-the-middle (IITM) instances detected in the module cache. ' +
+        'This typically happens when different packages depend on incompatible IITM versions ' +
+        'and npm deduplication places both on disk. ' +
+        'Each instance maintains its own hook registry, so hooks registered through one instance ' +
+        'will not fire when another instance dispatches module load events — ' +
+        'OpenTelemetry instrumentations are likely broken. ' +
+        'To fix this, align all IITM dependencies to the same version. ' +
+        `Detected instances:\n${iitmInstances.map(p => `  - ${p}`).join('\n')}`
```
aryamohanan: Do you have a sample log output showing how this warning appears? I'd like to see what it looks like in customer logs.

packages/core/test/util/iitmHook_test.js
```diff
@@ -0,0 +1,93 @@
...
+
+  const registrySuffix = path.join('import-in-the-middle', 'lib', 'register.js');
+
+  // Fake keys used to simulate additional IITM instances on disk.
+  const fakeIitmKey1 = path.join('/project/node_modules', registrySuffix);
+  const fakeIitmKey2 = path.join('/project/packages/core/node_modules', registrySuffix);
+
+  // Real lib/register.js entries that may be in require.cache when the full suite runs.
+  // We snapshot them before each test so we have full control over what the detector sees.
+  let snapshotKeys;
```
aryamohanan: Could you please remove the unnecessary comments from the test file and keep only the ones that are essential for understanding the test logic?

packages/collector/test/integration/currencies/databases/tedious/test_base.js
abhilash-sivan: Yeah, BE is updated to 1.46 already. Also these are tags, shouldn't be an issue right ?

packages/core/package.json
```diff
@@ -57,18 +57,18 @@
...
     "@opentelemetry/instrumentation-fs": "0.39.0",
     "@opentelemetry/instrumentation-oracledb": "0.46.0",
     "@opentelemetry/instrumentation-restify": "0.63.0",
     "@opentelemetry/instrumentation-socket.io": "0.67.0",
-    "@opentelemetry/instrumentation-tedious": "0.28.0",
+    "@opentelemetry/instrumentation-tedious": "0.40.0",
     "@opentelemetry/sdk-trace-base": "2.9.0",
     "cls-bluebird": "^2.1.0",
-    "import-in-the-middle": "2.0.5",
+    "import-in-the-middle": "^3.0.0",
```
aryamohanan: qs: why can't we update to the latest v3.3.3?

packages/collector/test/integration/currencies/databases/tedious/test_base.js
kirrg001: Oki Yeah I think so

packages/core/package.json
abhilash-sivan: discussed: pinning 3.3.3

packages/core/test/util/iitmHook_test.js
abhilash-sivan: sure

packages/core/src/util/iitmHook.js
abhilash-sivan: abhi ~/instana/nodejs [fix-tedious-esm] $ node packages/core/src/util/demo-iitm-conflict.js
{"level":30,"time":"2026-08-04T11:42:37.003Z","threadId":0,"name":"@instana/collector","msg":"PID Store starting with pid 80447"}
[WARN]  [Instana] Multiple import-in-the-middle (IITM) instances detected in the module cache. This typically happens when different packages depend on incompatible IITM versions and npm deduplication places both on disk. Each instance maintains its own hook registry, so hooks registered through one instance will not fire when another instance dispatches module load events — OpenTelemetry instrumentations are likely broken. To fix this, align all IITM dependencies to the same version. Detected instances:
  - /Users/abhi/instana/nodejs/packages/core/node_modules/import-in-the-middle/lib/register.js
  - /Users/abhi/instana/nodejs/node_modules/import-in-the-middle/lib/register.js




{"level":30,"time":"2026-08-04T11:42:37.172Z","threadId":0,"name":"@instana/collector","msg":"@instana/collector module version: 6.5.0"}
{"level":30,"time":"2026-08-04T11:42:37.172Z","threadId":0,"name":"@instana/collector","msg":"Transitioning from <init> to agentHostLookup"}
{"level":20,"time":"2026-08-04T11:42:37.175Z","threadId":0,"name":"@instana/collector","msg":"Attempt to load native add-on gcstats.js directly has been successful."}
{"level":20,"time":"2026-08-04T11:42:37.176Z","threadId":0,"name":"@instana/collector","msg":"Attempt to load native add-on event-loop-stats directly has been successful."}
{"level":30,"time":"2026-08-04T11:42:37.195Z","threadId":0,"name":"@instana/collector","msg":"Found an agent on 127.0.0.1:42699, proceeding to announce request."}
{"level":30,"time":"2026-08-04T11:42:37.195Z","threadId":0,"name":"@instana/collector","msg":"Transitioning from agentHostLookup to unannounced"}
{"level":20,"time":"2026-08-04T11:42:37.195Z","threadId":0,"name":"@instana/collector","msg":"Announcing the Node.js collector to the Instana host agent at 127.0.0.1:42699"}

abhilash-sivan: It's not necessary that we add a warn log, ether we can use a debug log or just not log anything. I am okay with majority decision

aryamohanan: Is this warning logged only once during startup, or could it be emitted more than once? I agree that we probably don't need to log this as a warning. While it can help diagnose instrumentation issues, it may be confusing or noisy for customers, especially since there's often little they can do about it directly. I'd prefer logging it at the debug level instead. 

If we do keep it as a warning, I think the message should be shorter.

abhilash-sivan: Yeah, debug is better.
I think iitmhook register is only happening once during startup.

kirrg001: I am not sure about this warning rn. We only want to see a warning when there is ITTM on the root or?

packages/core/package.json
```diff
@@ -57,18 +57,18 @@
...
     "@opentelemetry/instrumentation-fs": "0.39.0",
     "@opentelemetry/instrumentation-oracledb": "0.46.0",
     "@opentelemetry/instrumentation-restify": "0.63.0",
     "@opentelemetry/instrumentation-socket.io": "0.67.0",
-    "@opentelemetry/instrumentation-tedious": "0.28.0",
+    "@opentelemetry/instrumentation-tedious": "0.40.0",
     "@opentelemetry/sdk-trace-base": "2.9.0",
     "cls-bluebird": "^2.1.0",
-    "import-in-the-middle": "2.0.5",
+    "import-in-the-middle": "^3.3.3",
```
kirrg001, line 71: We can switch to caret agree.

packages/core/src/util/iitmHook.js
abhilash-sivan: We check `require.cache` for multiple occurrences of `import-in-the-middle`'s `lib/register.js`.

I only added this to prove the test, but I don't think we need this logic in the tracer. At best, it would run and emit debug logs, but it still wouldn't pinpoint the exact reason why an OpenTelemetry instrumentation isn't working for a customer. In practice, we also inspect the customer's `package.json` and dependency tree as part of the investigation.

So, I'm planning to remove both the test and this logic. I think we only need the dependency version bumps and the explanation.

kirrg001 (approved):

PR #2680, opened 2026-07-28
chore(http): captured request and response headers separately

aryamohanan:
This PR adds separate request and response HTTP header fields to HTTP spans while preserving the existing combined header field for backwards compatibility.

These fields are internal and are used only for OTLP field mapping. They are removed during transformation and are not forwarded to the backend.

ref https://jsw.ibm.com/browse/INSTA-102440


aryamohanan:
This is currently blocked pending community feedback on the HTTP header mapping approach. Since this is essentially a bug fix rather than a feature, it is currently considered low priority and can be implemented in a later phase.

PR #2684, opened 2026-07-29
chore(otlp-exporter): added missing resource mapping

aryamohanan:
Some resource mapping were not added earlier, added those.
**Whats not included**

Faas resource mapping can be handlied in a different PR

Future:

There are some rc candidates(not added in the tracer spec), can be extended later

1. [process.runtime.name](https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/)/
2. [process.runtime.version](https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/)
---
Tracer spec: https://github.ibm.com/instana/technical-documentation/blob/master/tracing/specification/otlp-exporter.md#resource-attribute-mapping

otel spec https://opentelemetry.io/docs/specs/semconv/resource/

ref https://jsw.ibm.com/browse/INSTA-100980

### DIFF
--- a/packages/core/src/otlpExporter/common/context.js
+++ b/packages/core/src/otlpExporter/common/context.js
@@ -15,8 +15,6 @@ class OtlpConfigContext {
     /** @type {any} */
     this._compiledSemConv = null;
     /** @type {string | null} */
-    this._hostId = null;
-    /** @type {string | null} */
     this._pid = null;
     /** @type {string | null} */
     this._serviceName = null;
--- a/packages/core/src/otlpExporter/common/context.js
+++ b/packages/core/src/otlpExporter/common/context.js
@@ -20,6 +20,8 @@ class OtlpConfigContext {
     this._pid = null;
     /** @type {string | null} */
     this._serviceName = null;
+    /** @type {string | null} */
+    this._serviceVersion = null;
   }
 
   /**
@@ -54,6 +56,20 @@ class OtlpConfigContext {
 
   get serviceName() {
     return this._serviceName;
+  }
+
+  /**
+   * @param {string} serviceVersion
+   */
+  setServiceVersion(serviceVersion) {
+    if (!serviceVersion || this._serviceVersion === serviceVersion) {
+      return;
+    }
+    this._serviceVersion = serviceVersion;
+  }
+
+  get serviceVersion() {
+    return this._serviceVersion;
   }
 }
 
--- a/packages/core/src/otlpExporter/common/context.js
+++ b/packages/core/src/otlpExporter/common/context.js
@@ -18,8 +18,6 @@ class OtlpConfigContext {
     this._pid = null;
     /** @type {string | null} */
     this._serviceName = null;
-    /** @type {string | null} */
-    this._serviceVersion = null;
   }
 
   /**
@@ -54,20 +52,6 @@ class OtlpConfigContext {
 
   get serviceName() {
     return this._serviceName;
-  }
-
-  /**
-   * @param {string} serviceVersion
-   */
-  setServiceVersion(serviceVersion) {
-    if (!serviceVersion || this._serviceVersion === serviceVersion) {
-      return;
-    }
-    this._serviceVersion = serviceVersion;
-  }
-
-  get serviceVersion() {
-    return this._serviceVersion;
   }
 }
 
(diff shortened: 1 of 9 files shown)


packages/core/src/otlpExporter/common/context.js
```diff
@@ -15,11 +15,11 @@ class OtlpConfigContext {
     /** @type {any} */
     this._compiledSemConv = null;
     /** @type {string | null} */
-    this._hostId = null;
```
aryamohanan, line 18: _hostId is not set from anywhere to the context so remove d from here

packages/core/src/otlpExporter/metrics/converter.js
```diff
@@ -23,10 +23,13 @@ function init(config) {
 /**
  * @param {any} metrics
  */
-function resolveServiceName(metrics) {
+function resolveServiceIdentity(metrics) {
```
aryamohanan: From the metrcis, we get the servicename nad service version.

packages/core/src/otlpExporter/metrics/util.js
```diff
@@ -4,15 +4,6 @@
 
 'use strict';
 
-/**
- * @param {Record<string, any>} from
- * @returns {string}
- */
-function getResourceKey(from) {
```
aryamohanan, line 11: This was used only on test file, removed

packages/core/src/otlpExporter/common/transformers/resource.js
```diff
@@ -38,7 +45,25 @@ const resourceMapper = {
...
+    return resource[RESOURCE.SERVICE_VERSION] || ctx.serviceVersion || undefined;
+  },
+
+  /**
+   * @param {RawPayload} rawPayload
+   * @returns {string | undefined}
+   */
+  serviceInstanceId(rawPayload) {
+    const resource = rawPayload.data?.resource || rawPayload.resource || {};
+    return resource[RESOURCE.SERVICE_INSTANCE_ID] || generatedServiceInstanceId;
```
aryamohanan: In otel, serviceInstanceId is generated from randomUUID, see https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-resources/src/detectors/platform/node/ServiceInstanceIdDetector.ts#L1C1-L27C74

packages/core/src/otlpExporter/common/transformers/resource.js
```diff
@@ -182,6 +201,20 @@ function extractResourceAttributes(rawPayload) {
   return { attributes };
 }
 
+/**
+ * @param {string} nodePlatform
+ */
+function normalizeOsType(nodePlatform) {
```
aryamohanan, line 207: Reference https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-resources/src/detectors/platform/node/OSDetector.ts#L20

packages/core/src/otlpExporter/common/transformers/resource.js
```diff
@@ -114,9 +117,20 @@ const resourceMapper = {
...
+  },
+  /**
+   * @param {RawPayload} rawPayload
+   * @returns {string | undefined}
+   */
+  osType(rawPayload) {
+    const resource = rawPayload.data?.resource || rawPayload.resource || {};
+
+    if (resource[RESOURCE.OS_TYPE]) {
+      return String(resource[RESOURCE.OS_TYPE]);
```
kirrg001: Are you doing String conversation because of null values or why?

packages/core/src/otlpExporter/common/transformers/resource.js
kirrg001, line 207: Are these normalized across tracers btw?

kirrg001 (approved):
Great

packages/core/src/otlpExporter/common/transformers/resource.js
aryamohanan: This is  the otel standard. In otlp mapping we follow that. 

See: https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-resources/src/detectors/platform/node/utils.ts#L20

aryamohanan: These are the value in the spec : windows; linux; darwin, it normalized

kirrg001: Nice

packages/core/src/otlpExporter/common/transformers/resource.js
aryamohanan: This conversion not required

PR #2688, opened 2026-07-31
chore(otlp-exporter): added faas.name resource mapping

aryamohanan:
At the moment, we only map `faas.name`, as it is the only Required FaaS resource attribute in the OpenTelemetry specification.

Otel spec: https://opentelemetry.io/docs/specs/semconv/resource/faas/

ref: https://jsw.ibm.com/browse/INSTA-100980

### DIFF
--- a/packages/core/src/otlpExporter/common/transformers/resource.js
+++ b/packages/core/src/otlpExporter/common/transformers/resource.js
@@ -34,6 +34,19 @@ const INSTRUMENTATION_SCOPE = {
  * @property {Record<string, any>} [f]
  */
 
+/**
+ * Extracts faas.name from provider-specific span data fields.
+ *
+ * @param {RawPayload} rawPayload
+ * @returns {string | undefined}
+ */
+function getFaasNameFromSpanData(rawPayload) {
+  return (
+    // AWS Lambda
+    rawPayload.data?.lambda?.functionName
+  );
+}
+
 const resourceMapper = {
   /**
    * @param {RawPayload} rawPayload
@@ -119,6 +132,16 @@ const resourceMapper = {
 
     return resource[RESOURCE.HOST_ID] || metadata.h;
   },
+
+  /**
+   * @param {RawPayload} rawPayload
+   * @returns {string | undefined}
+   */
+  faasName(rawPayload) {
+    const resource = rawPayload.data?.resource || rawPayload.resource || {};
+    return resource[RESOURCE.FAAS_NAME] || getFaasNameFromSpanData(rawPayload) || undefined;
+  },
+
   /**
    * @param {RawPayload} rawPayload
    * @returns {string | undefined}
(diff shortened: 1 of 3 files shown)


packages/core/src/otlpExporter/common/transformers/resource.js
```diff
@@ -34,6 +34,19 @@ const INSTRUMENTATION_SCOPE = {
...
+/**
+ * Extracts faas.name from provider-specific span data fields.
+ *
+ * @param {RawPayload} rawPayload
+ * @returns {string | undefined}
+ */
+function getFaasNameFromSpanData(rawPayload) {
+  return (
+    // AWS Lambda
+    rawPayload.data?.lambda?.functionName
```
aryamohanan, line 46: Currently only AWS Lambda is supported. When support for additional FaaS providers (e.g. Azure Functions, GCP Cloud Functions) is added, extend this function with the corresponding span data path.

abhilash-sivan (approved):

packages/core/src/otlpExporter/common/transformers/resource.js
kirrg001: Do we know whats the plan is for serverless in general?
Will there be more mappings? I guess so right?
Wondering if this mapping needs to happen somewhere else because its specific instrumentation knowledge (similar to our meta mapper).

aryamohanan: > Do we know whats the plan is for serverless in general?

Serverless endpoint support is already planned as a Phase 2 item, so we can revisit this as part of that work. We already have the mappings for aws, azure etc

> Will there be more mappings? I guess so right?
> Wondering if this mapping needs to happen somewhere else because its specific instrumentation knowledge (similar to our meta mapper)

For now, `faas.name` is the only required FaaS resource attribute, and we only support AWS Lambda. I agree as we add more FaaS providers or resource attributes, we can introduce a `resourceAttributes` mapping similar to `spanAttributes`. At the moment, with a single provider and a single required attribute, keeping it in `getFaasNameFromSpanData` keeps the implementation simple. what doyou think?

kirrg001: As soon as we will add Cloud mapping, we do not want to have multiple places for the data extraction. Checking the data lambda structure here is IMO not clean. Thats why I think we have to do it similar to [metrics mapping](https://github.com/instana/nodejs/blob/main/packages/core/src/otlpExporter/traces/transformers/spanMetaData.js#L117-L118)

kirrg001: We can merge as it is for now and we can find a solution when we will work on the cloud mapping. Up to you 👍

aryamohanan: Yeah sure.

kirrg001 (approved):

PR #2697, opened 2026-08-04
chore: removed unused env variable

aryamohanan:
I couldn’t find any real usage or references for these environment variable. It looks like they no longer be used, as I couldn’t find any active references in the codebase or documentation.


kirrg001 (approved):
Nice cleanup

abhilash-sivan (approved):

PR #2707, opened 2026-08-10
ci: improved credentials in tekton containers

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-108160


.tekton/tasks/test-groups/autoprofile-task.yaml
```diff
@@ -161,6 +128,30 @@ spec:
...
+          done
+          NEEDED_CREDS=$(echo "$NEEDED_CREDS" | tr ' ' '\n' | sort -u | tr '\n' ' ')
+          if [ -n "$NEEDED_CREDS" ]; then
+            ibmcloud plugin install secrets-manager -f -r 'IBM Cloud'
+            ibmcloud sm init --service-url "$SM_SERVICE_URL"
+            for CRED in $NEEDED_CREDS; do
+              export $CRED=$(ibmcloud sm secret-value-get --name "$CRED" --service-name "$SM_SERVICE_NAME" --output json | jq -r '.resources[0].secret_data.payload')
+            done
+          fi
+          unset API_KEY
```
kirrg001: This is the most important thing.

.tekton/tasks/test-groups/aws-lambda-split-5-task.yaml
```diff
@@ -161,6 +128,30 @@ spec:
           # Write mapping to shared file (printed by tasks-results)
           SHORT_NAMES=$(echo "$CLAIMED_TESTS" | tr ' ' '\n' | sed 's|.*/currencies/||; s|.*/misc/||; s|.*/unit/||; s|/test\.js$||' | tr '\n' ', ' | sed 's/,$//')
           echo "aws-lambda-split-5: $SHORT_NAMES" >> "$ARTIFACTS_PATH/test-mapping.txt"
+          NEEDED_CREDS=""
```
aryamohanan, line 131: Could we create a common script for this and call it from the respective tasks instead of repeating the same code in multiple places? Would that be possible?

kirrg001: The code is in the template.

.tekton/tasks/test-groups/collector-1-redis-task.yaml
```diff
@@ -168,6 +133,8 @@ spec:
         echo "CIRCLE_COMPARE_URL: $CIRCLE_COMPARE_URL"
         # echo "MODIFIED_FILES: $MODIFIED_FILES"
         echo "GIT_COMMIT_DESC: $GIT_COMMIT_DESC"
+        SM_SERVICE_URL="https://2502e0f8-4e08-47d1-bc55-fb1713f22522.private.eu-de.secrets-manager.appdomain.cloud"
```
aryamohanan, line 136: Could we define these common environment variables in one place and reuse them across the tasks instead of repeating them?

.tekton/tasks/test-groups/aws-lambda-split-5-task.yaml
kirrg001: The files are generated

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-s3/.credentials
```diff
@@ -0,0 +1,2 @@
+AWS_ACCESS_KEY_ID
```
aryamohanan, line 1: I have a question: the `execute-tools` task requires AWS credentials. Will those credentials still be available  fotr the task after this change?

kirrg001: No. Its a separate container 👍

aryamohanan (approved):
The idea looks cool. Let’s try it out.

.tekton/tasks/test-groups/autoprofile-task.yaml
abhilash-sivan, line 120: In the non-collector tests, including serverless, AWS Lambda, etc., the logic for cleaning up `API_KEY` is currently inside dead code that never executes:

```bash
if [ "false" == "true" ]; then
```

We can simply unset `API_KEY` just after the offline cache fetch step.

Also, this logic is not required in these files and can be removed, right?

abhilash-sivan: Hmm... since this is template-generated, we can maybe use the end of this block to unset API_KEY, so that it will be applied to all tasks irrespective of the SM fetch step 🤔

kirrg001: Maybe add the comment on the template?

.tekton/templates/test-task.yaml.template
```diff
@@ -181,6 +149,31 @@ spec:
...
+          NEEDED_CREDS=$(echo "$NEEDED_CREDS" | tr ' ' '\n' | sort -u | tr '\n' ' ')
+          if [ -n "$NEEDED_CREDS" ]; then
+            ibmcloud plugin install secrets-manager -f -r 'IBM Cloud'
+            ibmcloud sm init --service-url "$SM_SERVICE_URL"
+            for CRED in $NEEDED_CREDS; do
+              export $CRED=$(ibmcloud sm secret-value-get --name "$CRED" --service-name "$SM_SERVICE_NAME" --output json | jq -r '.resources[0].secret_data.payload')
+            done
+          fi
+          unset API_KEY
         fi
```
abhilash-sivan, line 176:
```suggestion
          fi
        fi

unset API_KEY
```

abhilash-sivan: The unset API_KEY goes below the `if [ "{{claimTests}}" == "true" ]; then` block, so it will get applied on all cases

kirrg001: Yeah not needed,but also does not hurt or?

abhilash-sivan: Yeah, no probs, but would be nice to remove the API_KEY from non collector tasks too

abhilash-sivan (approved):
🦸‍♀️

PR #1649, opened 2025-03-25
chore: added test comments for gcp

abhilash-sivan:

PR #2708, opened 2026-08-11
test: switched aws sdk tests to use localstack

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-108647

- added localstack support for aws tests
- to run against real AWS instead of LocalStack, log in and unset set RUN_REAL_AWS

test result: https://ibm.box.com/s/c4kttm6qsxcb2sizhev1lkz28giw4w6j

to verify:

### run in localstack

- runcollector-nw @aws-sdk
- runcollector-nw aws-sdk
- runaws-lambda

### run in aws
export RUN_REAL_AWS=true

- runcollector-nw @aws-sdk
- runcollector-nw aws-sdk
- runaws-lambda


packages/collector/test/integration/currencies/cloud/aws-sdk/lambda/suite.js
```diff
@@ -14,14 +14,15 @@ const { retry, stringifyItems, delay } = require('@_local/core/test/test_util');
...
+const { createFunction, removeFunction } = require('./util');
 
 const SPAN_NAME = 'aws.lambda.invoke';
 const functionName = 'nodejs-tracer-lambda';
 
 const withErrorOptions = [false, true];
 const availableCtx = [null, '{"Custom": {"awesome_company": "Instana"}}', '{"Custom": "Something"}'];
 const requestMethods = ['Callback', 'Promise'];
-const availableOperations = ['invoke', 'invokeAsync'];
+const availableOperations = ['invoke'];
```
abhilash-sivan: self note: check invokeAsync

abhilash-sivan: https://docs.localstack.cloud/aws/services/lambda/#api-coverage

invokeAsync is not supported by localstack

aryamohanan: We removed invokeAsync completely? This is really concerning.

packages/aws-lambda/src/ssm.js
```diff
@@ -72,7 +72,18 @@ module.exports.init = (config, _coldStart) => {
     };
 
     logger.debug(`INSTANA_SSM_PARAM_NAME is ${process.env.INSTANA_SSM_PARAM_NAME}.`);
-    const client = new SSMClient({ region: process.env.AWS_REGION });
+    let ssmEndpoint = process.env.INSTANA_CONNECT_LOCALSTACK_AWS;
```
aryamohanan: Why was this change made? This is actual production code

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-dynamodb/app.js
```diff
@@ -15,29 +15,24 @@ require('@instana/collector')();
 
 const express = require('express');
 
-const awsRegion = 'us-east-2';
```
aryamohanan, line 18: I’m wondering if there’s a way we can switch back to using the actual AWS services. I thought we had a flag that allowed us to switch between LocalStack and real AWS.

packages/aws-lambda/src/ssm.js
abhilash-sivan: sorry my bad. I will check the ssm area again

packages/aws-lambda/test/integration_test/test_definition.js
```diff
@@ -495,7 +522,7 @@ function registerTests(handlerDefinitionPath, reduced) {
       });
     });
 
-    describe('[with decryption] succeeds', () => {
+    (getLocalstackEndpoint() ? describeOrSkipIfReduced(reduced) : describe.skip)('[with decryption] succeeds', () => {
```
abhilash-sivan: self note: allow

.tekton/templates/test-task.yaml.template
```diff
@@ -118,6 +118,9 @@ spec:
         fi
 
         export CI=true
+        if echo "$AVAILABLE_SIDECARS" | grep -q 'localstack'; then
+          export INSTANA_CONNECT_LOCALSTACK_AWS=http://localhost:4566
```
kirrg001: This is already defined in our host config json. Why do we need this here?
We just need a toggle/mode in which the tests are running.

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-dynamodb/app.js
abhilash-sivan: Yes, I agree. Initially I thought we can use the Localstack env variable, but it seems better to add a new one. I added a new RUN_AWS env variable which will run the actual AWS

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-sns/test_base.js
```diff
@@ -27,13 +27,28 @@ const inVersionDir =
...
+  ? `${awsEndpoint}/${accountId}/`
+  : `https://sqs.us-east-2.amazonaws.com/${accountId}/`;
+
 let topicArn;
 const topicName = `nodejs-team-v3-${semver.major(process.versions.node)}`;
 const availableStyles = ['default', 'callback', 'v2'];
 const availableCommands = {
   PublishCommand: {
     Message: 'STRING_VALUE',
-    TargetArn: 'STRING_VALUE',
```
abhilash-sivan, line 36: in availableCommands, PublishCommand includes both TargetArn and TopicArn (the TopicArn is added in the loop at line 157). AWS SNS won't accept both simultaneously.

abhilash-sivan: It was working previously because the actual test never ran against AWS, only LS

.tekton/tasks/test-groups/aws-lambda-split-5-task.yaml
```diff
@@ -44,8 +51,8 @@ spec:
           memory: "8Gi"
       script: |
         #!/bin/bash
-        AVAILABLE_SIDECARS=""
-        SIDECAR_COUNTS=""
+        AVAILABLE_SIDECARS="localstack"
+        SIDECAR_COUNTS="localstack=1"
```
kirrg001: Did you manually edit these?

abhilash-sivan: Yes, do we have a script to add sidecars? I though no

kirrg001: 🤔 But these are test groups files. These are generated not?

>.tekton/tasks/test-groups/aws-lambda-split-2-task.yaml

packages/collector/test/integration/currencies/cloud/aws-sdk/lambda/suite.js
abhilash-sivan: Added back for actual AWS test

.tekton/tasks/test-groups/aws-lambda-split-5-task.yaml
abhilash-sivan: The 'generate test files' script does not make any changes. So the addition is correct right?

kirrg001: The sidecars are automatically added to the generated test files.
No manual edits in the generated files

abhilash-sivan: Oky. I was on the wrong branch. The 'generate test files' does make changes, i will update

.tekton/generate-test-files.js
```diff
@@ -37,7 +37,7 @@ const packages = {
 
   'test:ci:autoprofile': {},
   'test:ci:aws-fargate': {},
-  'test:ci:aws-lambda': { split: 5 },
+  'test:ci:aws-lambda': { split: 5, sidecars: ['localstack'] },
```
kirrg001, line 40: I thought aws lambda pkg does not run against AWS? Why do we have to change that

abhilash-sivan: We have 2 tests SSM and KMS client tests, running against AWS in this package

packages/collector/test/integration/currencies/cloud/aws-sdk/lambda/app.mjs
```diff
@@ -14,8 +14,25 @@ process.on('SIGTERM', () => {
 import AWS from 'aws-sdk';
 import express from 'express';
 const logPrefix = `AWS SDK v2 Lambda (${process.pid}):\t`;
+
+function getClientConfig() {
```
aryamohanan: Can we call the getConfig from util.js?

packages/collector/test/integration/currencies/cloud/aws-sdk/lambda/suite.js
```diff
@@ -14,14 +14,15 @@ const { retry, stringifyItems, delay } = require('@_local/core/test/test_util');
...
+const { createFunction, removeFunction } = require('./util');
 
 const SPAN_NAME = 'aws.lambda.invoke';
 const functionName = 'nodejs-tracer-lambda';
 
 const withErrorOptions = [false, true];
 const availableCtx = [null, '{"Custom": {"awesome_company": "Instana"}}', '{"Custom": "Something"}'];
 const requestMethods = ['Callback', 'Promise'];
-const availableOperations = ['invoke', 'invokeAsync'];
+const availableOperations = process.env.RUN_AWS ? ['invoke', 'invokeAsync'] : ['invoke'];
```
aryamohanan: I think we should add a proper comment here explaining that `invokeAsync` is currently not supported by LS. We could also consider requesting this feature from the LocalStack team, as they may add support for it in the future.

packages/collector/test/integration/currencies/cloud/aws-sdk/sqs/sqsUtil.js
```diff
@@ -18,8 +18,32 @@ AWS.HttpRequest.prototype.appendToUserAgent = function () {
   this.headers[this.getUserAgentHeaderName()] = [this._userAgent];
 };
 
+function getLocalstackEndpoint() {
```
aryamohanan: We’re repeating the same functions in multiple places. Could we move them into a generic test utility file and reuse them across the tests instead? This would help avoid duplication and make them easier to maintain.

abhilash-sivan: I already tried it. I ran into an issue while copying the files to the test directory from a common location.

Also, there are already utility files in multiple AWS tests, so I think we can keep the functions for now. It also gives us the flexibility to configure each test differently if needed.

I don't think it's worth spending more time optimising this right now

packages/collector/test/integration/currencies/cloud/aws-sdk/lambda/app.mjs
abhilash-sivan: Yes

.tekton/generate-test-files.js
kirrg001: https://github.com/instana/nodejs/pull/1649

Its all packages except aws lambda. I remembered wrong.

packages/collector/test/integration/currencies/cloud/aws-sdk/sqs/sqsUtil.js
aryamohanan: We define getLocalstackEndpoint() in 15 different places. I think it would be better to move this, along with getClientConfig(), into a common test utility file and reuse them across the tests. We already have a common utilities file for this, so we can avoid duplicating these functions.

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-s3/util.js
```diff
@@ -6,7 +6,27 @@
 'use strict';
 
 const awsSdk3 = require('@aws-sdk/client-s3');
-const s3 = new awsSdk3.S3({ region: 'us-east-2' });
+const { getLocalstackEndpoint } = require('@_local/collector/test/integration/currencies/cloud/@aws-sdk/aws-utils');
+
+exports.getLocalstackEndpoint = getLocalstackEndpoint;
+
+exports.getClientConfig = function () {
```
aryamohanan, line 16: Why can’t we use the shared function here? `If forcePathStyle: true` is the reason, couldn’t we pass it as an arg to the shared function that accepts options param?

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-s3/test_base.js
```diff
@@ -42,6 +42,7 @@ let libraryEnv;
 function start() {
   const mochaSuiteFn = supportedVersion(process.versions.node) ? describe : describe.skip;
 
+  // NOTE: Set RUN_AWS=true to run against real AWS instead of LocalStack.
```
aryamohanan: With no LocalStack and no real AWS, this test is still passing, which suggests something is wrong with the test itself. It may not be related to your PR.

```text
getting result with error: yes
✓ should instrument createBucket, putObject, listObjects, listObjectsV2, headObject, getObject, deleteObject, deleteBucket with error (246ms)
```

aryamohanan:
Please fix the code styling

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-s3/test_base.js
abhilash-sivan: Yeah, this is not related. But we can check later. I guess this only checks an exit span exists or something, no proper exact assertions, because some errors were thrown manually to reproduce this case

aryamohanan: This is the same case for Kinesis as well. Maybe we can create a card for it and plan to address it later in Q4.

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-kinesis/test_base.js
```diff
@@ -58,9 +58,9 @@ function start() {
     mochaSuiteFn = describe;
   }
 
-  const { cleanup } = require('./util');
-  const { checkStreamExistence } = require('./util');
+  const { cleanup, checkStreamExistence } = require('./util');
 
+  // NOTE: Set RUN_AWS=true to run against real AWS instead of LocalStack.
```
aryamohanan: Nit: I noticed this comment in several places. We could add it to the common util instead.

aryamohanan (approved):
Pre-approving 🎉

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-s3/util.js
abhilash-sivan: done

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-s3/test_base.js
abhilash-sivan: okay

packages/collector/test/integration/currencies/cloud/@aws-sdk/client-kinesis/test_base.js
abhilash-sivan: updated only 1 per package

PR #2709, opened 2026-08-11
chore: fixed TEST_NAME_FILTER as mocha grep when no named file matches

abhilash-sivan:
When no name match is found for `TEST_NAME_FILTER`, we now use the `--grep` pattern to run the matching tests.

For example:

```text
runcollector-nw aws-sdk s3
runcollector-nw aws-sdk dynamodb
```

Previously, these commands were running the entire `aws-sdk` test suite because the test file was shared across multiple services and wasn’t named by category.

The script has now been updated to handle these special cases as well, ensuring that only the relevant tests are executed.

### DIFF
--- a/bin/run-tests.sh
+++ b/bin/run-tests.sh
@@ -239,7 +239,6 @@ if [ -n "$PACKAGE" ]; then
     fi
     args="-- --grep \"$COMBINED_GREP\" $args"
     echo "Running tests for $ACTUAL_PACKAGE${VERSION:+ version $VERSION} (grep pattern: $COMBINED_GREP)"
-    echo "Warning: Could not identify specific test files, running full suite with filter."
   fi
 
 else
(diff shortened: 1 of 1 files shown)


bin/run-tests.sh
```diff
@@ -201,16 +222,23 @@ if [ -n "$PACKAGE" ]; then
...
     # If we couldn't determine specific files, fall back to old behavior but warn
-    args="-- --grep \"$GREP_PATTERN\" $args"
-    echo "Running tests for $ACTUAL_PACKAGE${VERSION:+ version $VERSION} (grep pattern: $GREP_PATTERN)"
+    COMBINED_GREP="$GREP_PATTERN"
+    if [ -n "$TEST_NAME_FILTER" ]; then
+      COMBINED_GREP="$GREP_PATTERN.*$TEST_NAME_FILTER"
+    fi
+    args="-- --grep \"$COMBINED_GREP\" $args"
+    echo "Running tests for $ACTUAL_PACKAGE${VERSION:+ version $VERSION} (grep pattern: $COMBINED_GREP)"
     echo "Warning: Could not identify specific test files, running full suite with filter."
```
abhilash-sivan:
```suggestion
```

aryamohanan (approved):

kirrg001 (approved):

PR #2711, opened 2026-08-12
chore: removed unsused deps

aryamohanan:
Evaluate production and development dependencies and remove those that are no longer required.

Found few unused deps, there are few more that we can replace with native node modules.


package.json
```diff
@@ -82,14 +82,12 @@
...
     "@aws-sdk/lib-dynamodb": "3.985.0",
     "@aws-sdk/middleware-retry": "3.370.0",
-    "@aws-sdk/util-retry": "3.370.0",
     "@azure/storage-blob": "12.30.0",
     "@commitlint/cli": "14.1.0",
     "@commitlint/config-conventional": "14.1.0",
     "@elastic/ecs-pino-format": "^1.5.0",
     "@ibm/tekton-lint": "1.1.0",
     "@types/chai": "4.3.0",
-    "@types/detect-libc": "1.0.0",
```
kirrg001, line 92: These got moved into the library?

aryamohanan: detect-libc provides its own type definitions, so you do not need this installed.

kirrg001 (approved):

PR #2713, opened 2026-08-18
chore: allowed metrics interval up to 60s

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-102260

before: 5s
after: upto 60s

Allowed poll rate values: 1s, 5s, 10s, 20s, 30s, 60s, ~120s, 180s, 240s, 300s, 360s, 420s, 480s, 540s, 600s~
ref from https://instana.slack.com/archives/C09S45LHBSQ/p1785315424877589?thread_ts=1784806406.182009&cid=C09S45LHBSQ

EDIT:

updated the config upto 60s, will add support for 60 - 600 once BE is capable

PR #2716, opened 2026-08-31
feat(core): added support for poll_rate agent config

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-102260


packages/core/src/config/index.js
```diff
@@ -156,7 +154,7 @@ let defaults = {
   packageJsonPath: null,
 
   metrics: {
-    transmissionDelay: 1000,
+    transmissionDelay: 30000,
```
kirrg001, line 157: ?

packages/core/src/util/constants.js
```diff
@@ -45,3 +45,7 @@ exports.LOG_LEVEL_PRIORITY = {
 };
 
 exports.DEFAULT_LOG_LEVEL = exports.LOG_LEVEL.WARN;
+
+// Allowed poll rate values in milliseconds, matching the agent's supported poll_rate seconds:
+// [1, 5, 10, 20, 30, 60]
```
kirrg001, line 50: Do we have a card to update it to 600s?

packages/core/src/util/constants.js
```diff
@@ -45,3 +45,7 @@ exports.LOG_LEVEL_PRIORITY = {
...
+exports.allowedTransmissionDelayValues = [1000, 5000, 10000, 20000, 30000, 60000];
```
kirrg001, line 51: 🤔 I thought customer can set any value between 1 and 60s

kirrg001 (changes requested):
See comments.
Also: The commit is a feature not a fix.

packages/core/src/util/constants.js
kirrg001: Found it https://jsw.ibm.com/browse/INSTA-111840

packages/core/src/util/constants.js
abhilash-sivan: Don't think so.

Code from sensor:

>  private static final int[] VALID_POLL_RATES = {1, 5, 10, 20, 30, 60, 120, 180, 240, 300, 360, 420, 480, 540, 600};
> 
>   /**
>    * Normalizes a poll rate configuration value to the nearest valid poll rate.
>    * <p>
>    * Valid poll rates are: 1, 5, 10, 20, 30, 60, 120, 180, 240, 300, 360, 420, 480, 540, 600 seconds.
>    * </p>
>    *
>    * @param input the poll rate configuration value (can be any Number subclass or String)
>    * @return the normalized poll rate in seconds (always a valid poll rate value)
>    */
>   public static int normalizePollRate(Object input) {
>     return normalizePollRate(input, 1);
>   }


There is already a common normaliser which normalized the value to the allowed list, out allowed list is a subset of this

packages/core/src/config/index.js
abhilash-sivan: Ah, this was a mistake; I fixed the changes in https://github.com/instana/nodejs/pull/2713
I will merge 2713 first and rebase this; I will keep it as 1000

packages/core/src/config/index.js
```diff
@@ -133,11 +133,6 @@ let currentConfig;
 /** @type {String[]} */
 const allowedSecretMatchers = ['equals', 'equals-ignore-case', 'contains', 'contains-ignore-case', 'regex', 'none'];
 
-// Allowed poll rate values in milliseconds
-// (1s, 5s, 10s, 20s, 30s, 60s)
-// Will add support for 120s, 180s, 240s, 300s, 360s, 420s, 480s, 540s, 600s once BE is capable
-const allowedTransmissionDelayValues = [1000, 5000, 10000, 20000, 30000, 60000];
```
abhilash-sivan, line 139: moving to constants

packages/core/src/config/index.js
```diff
@@ -322,22 +317,7 @@ function normalizeMetricsConfig({ userConfig = {}, defaultConfig = {}, finalConf
...
 
-  // Validate that transmissionDelay is one of the allowed values (in ms)
-  if (!allowedTransmissionDelayValues.includes(transmissionDelay)) {
-    const nearest = allowedTransmissionDelayValues.reduce(
-      (prev, curr) => (Math.abs(curr - transmissionDelay) < Math.abs(prev - transmissionDelay) ? curr : prev),
-      allowedTransmissionDelayValues[0]
-    );
-    logger.warn(
-      // eslint-disable-next-line max-len
-      `The value of config.metrics.transmissionDelay (or INSTANA_METRICS_TRANSMISSION_DELAY) (${transmissionDelay}) is not one of the allowed values (${allowedTransmissionDelayValues.join(
```
abhilash-sivan, line 333: logic moved to validators

packages/core/src/util/constants.js
kirrg001: Yeah they cannot, but its technically possible to send the interval we want. The agreement was a product-wide decision to only send these defined intervals for now.

kirrg001 (approved):

PR #2741, opened 2026-09-07
ci: verify pr

kirrg001:


kirrg001:
Some tests did not pass and are stuck. This is known - see #2742

PR #2765, opened 2026-09-11
ci: cleanup pipline.yaml sps

aryamohanan:


kirrg001:
Done via https://github.com/instana/nodejs/pull/2766

PR #2775, opened 2026-09-16
ci: use npm ci --ignore-scripts in sps 

aryamohanan:
Replace `npm install --foreground-scripts` with `npm ci --ignore-scripts` for dependency installs in the Tekton pipeline.

- `npm ci` installs strictly from package-lock.json, preventing a  compromised newer version from silently resolving in
- `--ignore-scripts` blocks preinstall/postinstall/install lifecycle hooks from running in CI, closing the exact execution vector used  by the CHAINDROP/Shai-Hulud npm supply chain worm (Aug 2026)

`--foreground-scripts` added as part of https://github.com/instana/nodejs/commit/f558acc11d71cc8a376cecc193b079545cd1ec7f

Since we ignore lifecycle scripts during installation, this option is no longer needed, so I removed it.

ref : https://nodejs.org/learn/getting-started/security-best-practices#supply-chain-attacks

https://www.elastic.co/security-labs/shai-hulud-chaindrop-npm-supply-chain#recommendations-how-to-respond-to-the-shai-hulud-compromise


.sps/pipeline-config.yaml
```diff
@@ -21,7 +21,7 @@ tasks:
           echo "Node.js: $(node --version)"
 
           cd "$WORKSPACE/$(load_repo app-repo path)"
-          npm install --loglevel warn --foreground-scripts
+          npm ci --loglevel warn --ignore-scripts
```
kirrg001, line 24: Using npm ci has no effect in our test env right now.

aryamohanan: True, but I think keeping `npm ci` is good practice here. It ensures deterministic builds by adhering to `package-lock.json` and is recommended for CI.

kirrg001: Yeah but you did not switch to npm ci for the other parts, because we do run that locally too.
So its a bit inconsistent right now.
You could use `isCI` helper.

aryamohanan: Good point, updated. The generated test wrappers in create-version-test-folders.js now use the isCI() helper npm ci when running in CI, npm install locally.

bin/create-version-test-folders.js
```diff
@@ -246,15 +247,17 @@ ${
 `
     : ''
 }      log('[INFO] Running npm install for ${suiteName}@${displayVersion}...');
-      const npmCmd = process.env.CI ?
+      const npmCmd = isCI() ?
         'npm install --cache ${rootDir}/.npm-offline-cache --prefer-offline ' +
```
kirrg001: This should be npm ci or

PR #2780, opened 2026-09-17
ci: added overrides support in currencies.json and fix sequelize vulnerabilities

aryamohanan:
### What this PR adds:
- currencies.json — "overrides": { "uuid": "^11.1.1" } on the sequelize entry (CVE fix)
- bin/generate-lock-files.js — reads currency.overrides and passes them into the generated package.json so the committed .template lock files resolve the overridden version
- bin/create-version-test-folders.js — reads currency.overrides and writes them into the package.json inside each _v*/ test folder so npm install at test time also respects the override
- package-lock.json.v6.37.8.template (both databases/ and frameworks/) — regenerated; uuid now resolves to 11.1.1
- .cra/.fileignore — removed the sequelize _v6.37.8 exclusion since the lock is now clean

this is the cve: https://github.com/advisories/GHSA-w5hq-g745-h8pq

Sequalize was using "uuid": "^8.3.2"


bin/generate-lock-files.js
```diff
@@ -90,6 +91,28 @@ function generateLockFile(currencyName, version, testDir, instanaVersion, baseLo
       name: `lock-gen-${safeName}-v${version}`,
       dependencies
     };
+
+    const templates = [
```
aryamohanan: Previously, `bin/generate-lock-files`.js only wrote `{ [currencyName]: version }` into the temporary package.json, ignoring any `package.json.template*` files. With this change, any overrides or extra settings defined in the templates are respected when generating the package-lock.json.v<version> files.

bin/generate-lock-files.js
```diff
@@ -149,10 +172,15 @@ function generateLockFileFromTemplate(templatePath, instanaVersion) {
...
-    );
+    const pkgJson = {
+      name: `lock-gen-tpl-${safeName}`,
+      dependencies
+    };
+    if (tpl.overrides) {
+      pkgJson.overrides = tpl.overrides;
+    }
+
+    fs.writeFileSync(path.join(tmpDir, 'package.json'), `${JSON.stringify(pkgJson, null, 2)}\n`);
```
aryamohanan: The only addition is merging the package.json.template* files into the temporary package.json so that template settings (such as overrides or extra dependencies) are respected when generating the lockfiles.

packages/collector/test/integration/currencies/databases/sequelize/package.json.template
```diff
@@ -4,5 +4,8 @@
   },
   "optionalDependencies": {
     "pg-native": "3.8.0"
+  },
+  "overrides": {
```
aryamohanan: this is the cve fix: https://github.com/advisories/GHSA-w5hq-g745-h8pq

Sequalize was using  `"uuid": "^8.3.2"`

kirrg001 (approved):

kirrg001 (approved):

PR #2755, opened 2026-09-10
feat: added W3C baggage support

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-103622

- [Docs](https://github.ibm.com/instana/technical-documentation/tree/master/tracing/specification#baggage-support)
- Kafka does not support W3C at all right now ([Ticket](https://jsw.ibm.com/browse/INSTA-116166))
- grpc does not support W3C at all right now ([Ticket](https://jsw.ibm.com/browse/INSTA-116166))

# Tests

## Propagate only

```
[Downstream Server] Received headers from Main Server:
  - baggage:     userId=alice,region=eu-central,tier=premium;metadata=test
  - traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-fdfb138e4e34a0f5-01
```

## Capture is ON

w3cb1

w3cb2

## Disable is ON

```
[Downstream Server] Received headers from Main Server:
  - baggage:     <none>
```

PR #2782, opened 2026-09-17
feat(otlp): added support for mapping Instana spans

kirrg001:
refs https://jsw.ibm.com/browse/INSTA-115765


packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -537,15 +551,21 @@ module.exports = {
...
+      if (!handler || !spanData) {
         continue;
       }
 
-      for (let j = 0; j < handler.length; j++) {
-        const attribute = applyMapping(handler[j], spanData);
-
-        if (attribute) {
-          attributes.push(attribute);
+      if (typeof handler === 'function') {
```
kirrg001, line 565: handler is either 1 to 1 mapping or is a fn which returns dynamic attrs.
Tags -> Customer defined.

aryamohanan (commented):
Good catch 👏

packages/core/test/otlpExporter/traces/mappers/instanaInstrumentationMappings_test.js
```diff
@@ -708,4 +741,200 @@ describe('otlpExporter/traces/mappers/instanaInstrumentationMappings', () => {
...
+                }
+              }
+            }
+          }
+        };
+
+        const result = spanStatus(span);
+        expect(result).to.deep.equal({
+          code: OTLP_STATUS_CODES.ERROR,
+          message: 'sdk failed'
```
kirrg001, line 955: This is the error fallback (ec 1, but no error msg).

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
aryamohanan, line 478: sdk should be excluded here?

packages/core/test/otlpExporter/traces/mappers/instanaInstrumentationMappings_test.js
aryamohanan: why? 🤔 

This test expects` message: 'sdk failed' `, but  SDK span with ec=1 and `custom.tags.message = 'something went wrong'`.

kirrg001: This test does not test tags.message :)

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
kirrg001: Why?

aryamohanan: I think it would be safer to exclude `sdk` here as well, similar to how it’s handled in `spanHandle.js`. Since `sdk` can coexist with the actual span type, relying on insertion order could cause us to pick the wrong span type. Just a suggestion to keep the behavior consistent.

packages/core/test/otlpExporter/traces/mappers/instanaInstrumentationMappings_test.js
aryamohanan: we can include that too or?

aryamohanan (dismissed):
Pre-approving. Great work, as always! 🚀

packages/core/test/otlpExporter/traces/mappers/instanaInstrumentationMappings_test.js
kirrg001: Not needed because its already covered in prev tests

packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
kirrg001: But why do we want to remove the sdk type? 🤔 
We need it anyway for #2755

kirrg001: Ah ok only for the span type. Okay I can add for **multi** data keys use case

aryamohanan (approved):

PR #2784, opened 2026-09-18
fix(otlp-exporter): included query parameters in http url.full attribute

aryamohanan:
Combine `endpoints` and `params` into the full URL representation for `OTLP.http.URL_FULL` . This was not correct before


packages/core/src/otlpExporter/traces/mappers/instanaInstrumentationMappings.js
```diff
@@ -77,7 +78,7 @@ const instrumentationMappings = {
     },
     spanAttributes: [
       { otlp: OTLP.http.REQUEST_METHOD, instana: 'operation', transform: toUpperCase },
-      { otlp: OTLP.http.URL_FULL, instana: 'endpoints' },
+      { otlp: OTLP.http.URL_FULL, instana: ['endpoints', 'params'], transform: buildFullUrl },
```
kirrg001, line 81: Such a great design 👍

kirrg001 (approved):

PR #2786, opened 2026-09-18

instanacd:
Tada!


packages/collector/test/integration/currencies/cloud/@aws-sdk/client-lambda/package-lock.json.v3.1131.0.template
```diff
@@ -0,0 +1,1282 @@
+{
```
abhilash-sivan, line 1: should have been replaced with `@aws-sdk/client-lambda/package-lock.json.v3.1128.0.template`. need to check the package lock generation logic

aryamohanan: This is actually replace. Could you double check

abhilash-sivan: Yes, it seems like the files are correctly generated, but the UI shows it wrong. No worries then

abhilash-sivan (approved):

PR #2788, opened 2026-09-22
feat: added configuration for bind variables collection

abhilash-sivan:
refs https://jsw.ibm.com/browse/INSTA-117026


packages/collector/src/announceCycle/unannounced.js
```diff
@@ -417,6 +419,22 @@ function applyHttpExitConfiguration(agentResponse) {
   }
 }
 
+/**
+ * Applies the global db-bind-variables configuration from the agent's `tracing.global` block.
+ *
+ * @param {AgentAnnounceResponse} agentResponse
+ */
+function applyDbBindVariablesConfiguration(agentResponse) {
```
aryamohanan: Missing unit tests for these changes

packages/core/src/config/index.js
```diff
@@ -1230,6 +1239,60 @@ function parseSecretsEnvVar(envVarValue) {
     keywords: keywordsArray
   };
 }
+
+/**
+ * @param {{ userConfig?: InstanaConfig|null, defaultConfig?: InstanaConfig, finalConfig?: InstanaConfig }} [options]
+ */
+function normalizeDbBindVariables({ userConfig = {}, defaultConfig = {}, finalConfig = {} } = {}) {
+  // Priority 1: environment variables (highest priority)
+  const disableEnv = process.env.INSTANA_TRACING_DB_BIND_VARIABLES_DISABLE;
```
aryamohanan: Could we follow the existing config normalization pattern here instead of introducing a separate validation flow? It would keep the priority, validation, fallback, and logging behavior consistent with the other config options.

packages/core/src/config/validator.js
```diff
@@ -219,3 +219,38 @@ exports.validateTransmissionDelay = function validateTransmissionDelay(value) {
...
+
+/**
+ * Validates a db-bind-variables `disable` value.
+ * Accepts booleans only; strings such as 'true'/'false' are intentionally rejected here
+ * because env-var coercion is handled by the normalizer.
+ *
+ * @param {any} value
+ * @returns {{ isValid: boolean, error: string | null }}
+ */
+exports.validateDbBindVariablesDisable = function validateDbBindVariablesDisable(value) {
```
aryamohanan: why can't we use existing validator "booleanValidator"?

packages/core/src/config/validator.js
```diff
@@ -219,3 +219,38 @@ exports.validateTransmissionDelay = function validateTransmissionDelay(value) {
   );
   return nearest;
 };
+
```
aryamohanan: missing corresponding unit test for this validator module

packages/core/src/config/index.js
```diff
@@ -1230,6 +1239,60 @@ function parseSecretsEnvVar(envVarValue) {
     keywords: keywordsArray
   };
 }
+
+/**
+ * @param {{ userConfig?: InstanaConfig|null, defaultConfig?: InstanaConfig, finalConfig?: InstanaConfig }} [options]
+ */
+function normalizeDbBindVariables({ userConfig = {}, defaultConfig = {}, finalConfig = {} } = {}) {
```
aryamohanan: add unit tests for this

aryamohanan (commented):
Added some initial comments.

aryamohanan:
IMO, this PR only adds the configuration option, but it isn’t consumed anywhere yet. I’d consider this a chore rather than a feature until the configuration is actually applied.

